PR #13424, opened 2020-02-17
[ci skip] Fixed non-working examples in Manual: Exception Handling

sigmapie8:
Tried to fix some examples in Exception Handling section of doc/manual.rst which were not working when copy-pasted.

### DIFF
--- a/doc/manual.rst
+++ b/doc/manual.rst
@@ -4221,7 +4221,7 @@ Raise statement
 
 Example:
 
-.. code-block:: nim/
+.. code-block:: nim
   raise newException(IOError, "IO failed")
 
 Apart from built-in operations like array indexing, memory allocation, etc.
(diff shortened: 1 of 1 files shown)


doc/manual.rst
```diff
@@ -4219,8 +4221,8 @@ Raise statement
 
 Example:
 
-.. code-block:: nim
-  raise newEOS("operating system failed")
+.. code-block:: nim/
```
Varriount: Why the forward slash?

Araq:
The CIs do check the documentation so don't use the ``[ci skip]`` for it please.

doc/manual.rst
Araq: Remove the slash please.

sigmapie8: Sorry that was a typing error. I'll change it.

sigmapie8: > The CIs do check the documentation so don't use the `[ci skip]` for it please.
Okay.

PR #13429, opened 2020-02-18
Add `sequtils.unzip` to complement `sequtils.zip`

kaushalmodi:


lib/pure/collections/sequtils.nim
```diff
@@ -281,6 +281,19 @@ when (NimMajor, NimMinor) <= (1, 0):
...
+proc unzip*[S, T](s: openArray[(S, T)]): (seq[S], seq[T]) =
+  ## Returns a tuple of two sequences split out from a sequence of 2-field tuples.
+  runnableExamples:
+    let
+      zipped = @[(1, 'a'), (2, 'b'), (3, 'c')]
+      unzipped1 = @[1, 2, 3]
+      unzipped2 = @['a', 'b', 'c']
+    assert zipped.unzip() == (unzipped1, unzipped2)
+    assert zip(unzipped1, unzipped2).unzip() == (unzipped1, unzipped2)
+  for elem in s:
```
Vindaar, line 295: We should make sure to initialize the `result` sequences with the correct length, since it's the same as the length of `s`.
```suggestion
  result[0] = newSeqOfCap[S](s.len)
  result[1] = newSeqOfCap[T](s.len)
  for elem in s:
```

def-:
I would suggest `newSeq(result[0], s.len)` and then `result[0][i] = s[i][0]` etc. in order to only allocate memory once.

kaushalmodi:
@def- Thanks! ~@Vindaar's suggested commit in https://github.com/nim-lang/Nim/pull/13429/commits/980c0ea042945c40eb8e0b7c1a35c0ab7261ed60 should take care of that.~

Actually https://github.com/nim-lang/Nim/pull/13429/commits/2ff1e41412be2b89a7018f2b72760293a7a8f6d3 should take care of that.

Vindaar:
> @def- Thanks! ~@Vindaar's suggested commit in [980c0ea](https://github.com/nim-lang/Nim/commit/980c0ea042945c40eb8e0b7c1a35c0ab7261ed60) should take care of that.~
> 
> Actually [2ff1e41](https://github.com/nim-lang/Nim/commit/2ff1e41412be2b89a7018f2b72760293a7a8f6d3) should take care of that.

Just FYI these two things should do essentially the same thing. `newSeqOfCap` simply allows one to use `add` on the sequence instead of having to access the `i`-th element, but it already allocates the space given as the argument.

lib/pure/collections/sequtils.nim
```diff
@@ -281,6 +281,21 @@ when (NimMajor, NimMinor) <= (1, 0):
 else:
   zipImpl(s1, s2, seq[(S, T)])
 
+proc unzip*[S, T](s: openArray[(S, T)]): (seq[S], seq[T]) =
```
Araq: Needs a ``.since`` annotation.

kaushalmodi:
@Vindaar Thanks I wasn't aware of that. I thought that the `newSeqOfCap` might be doing just a length check internally.

lib/pure/collections/sequtils.nim
kaushalmodi: Done

kaushalmodi:
> 
> 
> @Vindaar Thanks I wasn't aware of that. I thought that the `newSeqOfCap` might be doing just a length check internally.

I have now reverted back to using `newSeqOfCap` as I like the `.add` syntax better.

timotheecour:
> Just FYI these two things should do essentially the same thing. newSeqOfCap simply allows one to use add on the sequence instead of having to access the i-th element, but it already allocates the space given as the argument.

[EDIT] not true, see https://github.com/nim-lang/Nim/pull/13448 (using `[]=` can be 7X faster than `add`)

Vindaar:
> > Just FYI these two things should do essentially the same thing. newSeqOfCap simply allows one to use add on the sequence instead of having to access the i-th element, but it already allocates the space given as the argument.
> 
> @Vindaar this is not true: as shown in #13448 (which fixes performance issue with `unzip`) `newSeqOfCap(n)` + `add` is 5-15X slower than `newSeq(n)` + `[]=`

Wait, what? 😨

PR #13453, opened 2020-02-21
Make gdb script work on mac and linux; add windows gdb script

jyapayne:
I could never get the built in `nim-gdb` script to work with Mac or my linux box. This PR fixes it for both and adds support for windows in the form of a batch script. This is useful for getting debugging working in VSCode on all platforms.

### DIFF
--- a/bin/nim-gdb
+++ b/bin/nim-gdb
@@ -5,17 +5,25 @@ set -e
 
 which nim > /dev/null || (echo "nim not in PATH"; exit 1)
 which gdb > /dev/null || (echo "gdb not in PATH"; exit 1)
+which readlink > /dev/null || which greadlink > /dev/null || (echo "readlink not in PATH. Please install coreutils from brew if on Mac."; exit 1)
+
+nreadlink () {
+    (which greadlink > /dev/null && greadlink "$@") || (which readlink > /dev/null && readlink "$@") || echo "Readlink could not be found"
+}
 
 # Find out where the pretty printer Python module is
-NIM_SYSROOT=$(dirname $(dirname $(readlink -e $(which nim))))
+NIM_SYSROOT=$(dirname $(dirname $(nreadlink -e $(which nim))))
 GDB_PYTHON_MODULE_PATH="$NIM_SYSROOT/tools/nim-gdb.py"
 
 # Run GDB with the additional arguments that load the pretty printers
 # Set the environment variable `NIM_GDB` to overwrite the call to a
 # different/specific command (defaults to `gdb`).
 NIM_GDB="${NIM_GDB:-gdb}"
+
+# This is needed for some reason. I can't get -eval-command to work ever
+echo "source $GDB_PYTHON_MODULE_PATH" > gdbcommands.txt
 # exec replaces the new process of bash with gdb. It is always good to
 # have fewer processes.
-exec ${NIM_GDB} \
-  -eval-command "source $GDB_PYTHON_MODULE_PATH" \
-  "$@"
+exec ${NIM_GDB} --command="gdbcommands.txt" "$@"
+
+rm gdbcommands.txt
--- a/bin/nim-gdb.bat
+++ b/bin/nim-gdb.bat
@@ -1,0 +1,16 @@
+@echo off
+for %%i in (nim.exe) do (set NIM_BIN=%%~dp$PATH:i)
+
+for %%i in ("%NIM_BIN%\..\") do (set NIM_ROOT=%%~fi)
+
+set @GDB_PYTHON_MODULE_PATH=%NIM_ROOT%\tools\nim-gdb.py
+set @NIM_GDB=gdb.exe
+
+@echo source %@GDB_PYTHON_MODULE_PATH%> wingdbcommand.txt
+
+%@NIM_GDB% --command="wingdbcommand.txt" %*
+
+del wingdbcommand.txt /f /q
+
+EXIT /B %ERRORLEVEL%
+@echo on


bin/nim-gdb
```diff
@@ -5,17 +5,25 @@ set -e
...
+NIM_SYSROOT=$(dirname $(dirname $(nreadlink -e $(which nim))))
 GDB_PYTHON_MODULE_PATH="$NIM_SYSROOT/tools/nim-gdb.py"
 
 # Run GDB with the additional arguments that load the pretty printers
 # Set the environment variable `NIM_GDB` to overwrite the call to a
 # different/specific command (defaults to `gdb`).
 NIM_GDB="${NIM_GDB:-gdb}"
+
+# This is needed for some reason. I can't get -eval-command to work ever
+echo "source $GDB_PYTHON_MODULE_PATH" > gdbcommands.txt
```
krux02, line 24: can you elaborate on this? Why is this needed? What version of gdb are you using (``gdb --version``)? Is it related to an outdated version of ``gdb``?

bin/nim-gdb
```diff
@@ -5,17 +5,25 @@ set -e
...
 # exec replaces the new process of bash with gdb. It is always good to
 # have fewer processes.
-exec ${NIM_GDB} \
-  -eval-command "source $GDB_PYTHON_MODULE_PATH" \
-  "$@"
+exec ${NIM_GDB} --command="gdbcommands.txt" "$@"
+
+rm gdbcommands.txt
```
krux02, line 29: this line will never be executed. Nothing after ``exec`` will be executed because like I explained in the comment above, ``exec`` replaces the process of bash with gdb. There is no ``bash`` precess running anymore after ``exec``.

bin/nim-gdb
```diff
@@ -5,17 +5,25 @@ set -e
 
 which nim > /dev/null || (echo "nim not in PATH"; exit 1)
 which gdb > /dev/null || (echo "gdb not in PATH"; exit 1)
+which readlink > /dev/null || which greadlink > /dev/null || (echo "readlink not in PATH. Please install coreutils from brew if on Mac."; exit 1)
+
+nreadlink () {
+    (which greadlink > /dev/null && greadlink "$@") || (which readlink > /dev/null && readlink "$@") || echo "Readlink could not be found"
+}
```
krux02, line 12: Please make a variable with the ``readlink`` command. Something like ``READLINK``. Also please ensure that you are on ``osx`` before printing any troubleshooting guidelines about fixing problems on ``osx``. This you can do with ``uname -s``

krux02 (commented):
@Araq Thanks for including me in the review process for this.

bin/nim-gdb
jyapayne: Yep, you're right. Don't know why I thought this worked. I'll open a PR and fix the issues you mentioned here and add you as a reviewer

bin/nim-gdb
jyapayne: It could have been. I believe I was using 8.3 at the time, but I could be wrong. Just tried with gdb 8.3.1 and it works without this hack.

bin/nim-gdb
jyapayne: Not sure if I'm doing it wrong, but a variable doesn't work, only a function will because in NIM_SYSROOT below, the variable expansion leads to bash thinking the expansion is not a command with arguments.

For example,

```bash
READLINK=$((which greadlink > /dev/null && greadlink "$@") || (which readlink > /dev/null && readlink "$@") || echo "Readlink could not be found")

NIM_SYSROOT=$(dirname $(dirname $(READLINK -e $(which nim))))
```

If I run the above script on my Mac, I get:

```
greadlink: missing operand
Try 'greadlink --help' for more information.
READLINK: illegal option -- e
usage: READLINK [-FlLnqrsx] [-f format] [-t timefmt] [file ...]
usage: dirname path
usage: dirname path
```

I will take the suggestion about detecting OS and put it in the followup PR

bin/nim-gdb
jyapayne: Hmm, it looks like my issue was with VSCode. Using your original code makes VSCode error out with:

```
Loading Nim Runtime support.
('set Nim pretty printers for ', '/Users/joey/Projects/nim-nanovg/build/demosdl2/demo_sdl2_debug.dSYM/Contents/Resources/DWARF/demo_sdl2_debug')
undefinedBFD: /usr/lib/dyld(i386:x86-64): unknown load command 0x34
BFD: /usr/lib/dyld(i386:x86-64): unknown load command 0x34
BFD: /usr/lib/dyld(i386:x86-64): unknown load command 0x34
No source file named /Users/joey/Projects/nim-nanovg/examples/demo_sdl2_gpu.nim.
Running executable
[New Thread 0xf03 of process 2992]
```

I'm not sure what to do with that :/

jyapayne: Hmm, it looks like it works if I rename `nim-gdb` to `nim-gdb.bash`. Would that be acceptable to do, and then have a symbolic link from `nim-gdb` to `nim-gdb.bash`?

jyapayne:
Comments addressed in https://github.com/nim-lang/Nim/pull/13658

bin/nim-gdb
krux02: I was thinking about this:
```sh
if [[ $(uname -s) = Darwin]]; then  
  READLINK=greadlink 
  which greadlink > /dev/null || echo "greadlink not in PATH. Please install coreutils from brew."
else
  READLINK=readlink
fi
```
And then you call readlink with ``$READLINK`` instead of ``readlink``.

jyapayne: Ah, yep. Looks like that works. I updated the PR. Modified your code a bit to fit in with the code I had written.

PR #13459, opened 2020-02-21
Update pegs.nim to work at compiletime. No range errors.

solo989:


krux02:
test case is missing. The best way to test this is to transform an already existing pegs test into a test that runs both at run time as well as at compile time.

solo989:
Added a test case.

lib/pure/pegs.nim
```diff
@@ -2040,6 +2058,8 @@ proc escapePeg*(s: string): string =
 when isMainModule:
   assert escapePeg("abc''def'") == r"'abc'\x27\x27'def'\x27"
   assert match("(a b c)", peg"'(' @ ')'")
+  static:
+    assert match("(a b c)", peg"'(' @ ')'")
```
krux02, line 2061: Here is an example how to lift a test to be executed at compile time as well as runtime:
https://github.com/nim-lang/Nim/blob/devel/tests/stdlib/tjsonmacro.nim#L638

krux02 (changes requested):

solo989:
Lifted all tests.

krux02 (approved):

PR #13460, opened 2020-02-21
Consider proc as a pointer type in options

metagn:


Araq:
This is a breaking change, it needs a changelog entry and a ``.since`` workaround.

Issue #12790, opened 2019-12-02
Proposal: endinanness-aware I/O in stdlib streams


johnnovak:
I've been working on an endianness-aware buffered file I/O module for quite a while now and have just realised that 1) I could actually use `streams` instead, 2) this functionality is generic enough to be added to the `streams` standard module.

Not sure what the general view is on extending the standard library at this point, so I thought I'd ask here first to avoid wasting my time.

The changes I'm proposing would not change anything existing; I'd just add the following functions to support endianness-aware I/O. The `BE` and `LE` postfixes indicate the endinanness of the data in the stream. For example, `readInt16BE` would read 2 bytes from the stream and swap them on a little-endian platform, but perform no conversion on a big-endian platform.

The single-value read & write functions would look like this:

```nim
proc readInt16BE(s: Stream): int16
proc readInt16LE(s: Stream): int16
...
proc readFloat64BE(s: Stream): float64
proc readFloat64LE(s: Stream): float64

proc peekInt16BE(s: Stream): int16
proc peekInt16LE(s: Stream): int16
...
proc peekFloat64BE(s: Stream): float64
proc peekFloat64LE(s: Stream): float64

proc writeInt16BE(s: Stream): int16
proc writeInt16LE(s: Stream): int16
...
proc writeFloat64BE(s: Stream): float64
proc writeFloat64LE(s: Stream): float64
```

And here's the interface for the chunked versions (`LE` variants omitted for brevity):

```nim
proc writeDataBE*(s: Stream, buffer: openArray[int16|uint16], numItems: Natural)
proc writeDataBE*(s: Stream, buffer: openArray[int32|uint32|float32], numItems: Natural)
proc writeDataBE*(s: Stream, buffer: openArray[int64|uint64|float64], numItems: Natural)

proc writeData16BE*(s: Stream, buffer: pointer, bufLen: Natural)
proc writeData32BE*(s: Stream, buffer: pointer, bufLen: Natural)
proc writeData64BE*(s: Stream, buffer: pointer, bufLen: Natural)

proc readData16BE*(s: Stream, buffer: var openArray[int16|uint16], numItems: Natural)
proc readData32BE*(s: Stream, buffer: var openArray[int32|uint32|float32], numItems: Natural)
proc readData64BE*(s: Stream, buffer: var openArray[int64|uint64|float64], numItems: Natural)

proc readData16BE*(s: Stream, buffer: pointer, bufLen: Natural)
[...]

Issue #9206, opened 2018-10-05
endians.nim API alignment-unsafe


arnetheduck:
The current implementation of `endians.nim` is not safe against alignment issues and will cause crashes on certain platforms, and (silent) slowdowns on others.

https://github.com/nim-lang/Nim/blob/269b957e2968345a4c9a5037ab97ab726135a921/lib/pure/endians.nim#L48

The issue is that there's no guarantee that the incoming pointer is aligned - it is common that it is not, when dealing with network protocols and other binary data, precisely where endian conversions are common.

There's a few ways to fix it:

1. document it and hope the developer will remember these things (haha)
2. avoid using pointers in the API - go with values only
3. assume unaligned buffers and copy data to local, aligned buffers - this copy will likely be elided by good backend optimisers if they can determine that original buffer is aligned
4. add proper alignment support to the language

thoughts? 2 or 3 seem reasonable, 4 is some ways off and would also require changing the API so compatibility-wise is the same.

Araq:
I think (3) is best.

arnetheduck:
on occasion, I've missed a value-based api a little when using these - it would be convenient for several reasons - the hard part being what the returned value should be.. ie something like 
```
enum Endian = endLittle, endBig
proc to[T](t: T, endian: Endian): array[sizeof(T), byte]

echo 42.to(endLittle)
```
kind of works, though it doesn't tag the return value with any endianess - int et al are known to be native endian, but if the type system could be exploited to provide "correct" endianess for these, that would be cool - this would allow the creation of nice serialization/networking API that declare the endianess expectations up-front.

arnetheduck:
also, the current approach with byte size in name makes it hard to write generic code - there's always a block needed to convert from integer size to byte-name - the value/type-based api solves this neatly.

krux02:
looks like this issue has been solved

PR #13463, opened 2020-02-21
better endians

johnnovak:
As the title says :) This is a sort of a prequel to my upcoming `std/binstreams` module, as discussed in https://github.com/nim-lang/Nim/issues/12790.

Heavy inspired by https://github.com/status-im/nim-stew/blob/1c4293b3e754b5ea68a188b60b192801162cd44e/stew/endians2.nim (suggested by @mratsim ).

I didn't bother fixing up the rest of the tests and the documentation until I get a general approval from you guys.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -104,6 +104,7 @@
   serve no purpose whatsoever.
 - `httpclient.newHttpClient` and `httpclient.newAsyncHttpClient` added `headers`
   argument to set initial HTTP Headers, instead of a hardcoded empty `newHttpHeader()`.
+- `endians` introduces a new and improved API and deprecates the existing API.
 
 
 ## Language additions
--- a/lib/pure/endians.nim
+++ b/lib/pure/endians.nim
@@ -17,9 +17,7 @@
 ## intrinsics.
 ##
 ## Most of the API is compatible with NimVM, except the procs that take
-## `pointer` arguments. Also note that `float32` is currently not supported on
-## NimVM, so calling functions that operate on `float32` arguments would
-## trigger compiler crashes on that backend.
+## `pointer` arguments.
 ##
 
 include "system/inclrtl"
@@ -477,7 +475,7 @@ func fromBytesLE*(T: typedesc[SomeNumber], buf: openArray[byte],
     fromBytesLE(T, buf[startIndex].unsafeAddr)
 
 
-when defined(testing) and isMainModule:
+when isMainModule:
   const
     i8 = 0xf2'i8
     u8 = 0xf2'u8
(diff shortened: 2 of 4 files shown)


Araq:
Requires a (little) RFC outlining the benefits and incompatibilities (if there are any). And a changelog entry and ``.since`` support.

johnnovak:
@Araq Okay, so here's the requested RFC. I'll do the documentation, changelog and the `.since` stuff as the last step once we have agreed on the API. I'll also need to change a few other stdlib modules that rely on this, but that's simple.

I've done a force push because the openarrays variants were not quite right and added tests for them too.

----------------------------

The proposed new `endians` module is a superset of the existing one. It provides the same functionality but with a better API and some quality-of-life enhancements.

# Existing API

The problem with the existing API is that the functions only accept pointers to memory locations and perform the endianness conversion in-place. Moreover, the function names are a little confusing and there's no generic support.

```nim
proc swapEndian64*(outp, inp: pointer)
proc swapEndian32*(outp, inp: pointer)
proc swapEndian16*(outp, inp: pointer)
proc littleEndian64*(outp, inp: pointer)
proc littleEndian32*(outp, inp: pointer)
proc littleEndian16*(outp, inp: pointer)
proc bigEndian64*(outp, inp: pointer)
proc bigEndian32*(outp, inp: pointer)
proc bigEndian16*(outp, inp: pointer)
```


# Proposed new API

The new API supports generics and makes code dealing with endianness conversion more readable thanks to the `toXX/fromXX` naming scheme. It uses the same compiler intrinsincs under the hood if available just like the existing module.

All conversion functions have pure function, openarray and pointer variants.

```nim
func toBE*[T: SomeNumber](value: T): T
func toLE*[T: SomeNumber](value: T): T
func fromBE*[T: SomeNumber](value: T): T
func fromLE*[T: SomeNumber](value: T): T

proc toBytesBE*[T: SomeNumber](value: T, buf: pointer)
proc toBytesBE*[T: SomeNumber](value: T, buf: openArray[byte],
                               index: Natural = 0)

(...similarly for `toBytesLE`, `fromBytesBE` & `fromBytesLE`...)

func swapEndian*[T: SomeNumber](value: T): T
proc swapEndian*[T: SomeNumber](buf: pointer)
proc swapEndian*[T: SomeNumber](buf: var openArray[T], index: Natural = 0)
```

# Backwards compatibility

The new library is capable of expressing the same transforms (and more) as the old one. However, existing code relying on `endians` needs to be changed to use the new API. I think this is fine because the existing module is marked as unstable and it's not used that often anyway.

Araq:
> I think this is fine because the existing module is marked as unstable and it's not used that often anyway.

The old procs should remain but get a ``.deprecated`` annotation.

johnnovak:
Okay, the final version is up.

changelog.md
```diff
@@ -1,4 +1,4 @@
-# 1.2 - xxxx-xx-xx
+# 1.1 - xxxx-xx-xx
```
narimiran: Don't change this. The next stable version will indeed be 1.2.

lib/pure/endians.nim
```diff
@@ -98,16 +123,438 @@ else:
...
+                 buf: pointer) {.inline, mySince: (1,1).} =
+  ## Swaps the byte order of a number at a memory location in-place. The type
+  ## of the number has to be passed in as an argument.
+  runnableExamples:
+    var i = 0xdeadbeef'i32
+    swapEndian(int32, i.addr)
+    doAssert i == 0xefbeadde'i32
+
+  var value = cast[ptr T](buf)[]
+  cast[ptr T](buf)[] = swapEndian(value)
```
mratsim, line 234: AFAIK this doesn't work in NimVM

mratsim (commented):
The main reason we wrote `endians2` at Status was to allow compile-time endianness manipulation.

(https://github.com/status-im/nim-stew/blob/master/stew/endians2.nim)

Compile-time endianness is particularly useful for big integer (such as [Stint](https://github.com/status-im/nim-stint)) or cryptographic libraries (such as [Constantine](https://github.com/mratsim/constantine)) where it is desirable to parse big-endian hex constant parameters at compile-time and transform them into the internal library representation, for example [cryptographic elliptic curve moduli](https://github.com/mratsim/constantine/blob/80f822c2275744a4fd3c1029d91d32cd34595f2b/constantine/config/curves.nim#L42-L57):

```Nim
declareCurves:
  curve BN254:
    bitsize: 254
    modulus: "0x30644e72e131a029b85045b68181585d97816a916871ca8d3c208c16d87cfd47"
    # Equation: Y^2 = X^3 + 3
  curve BLS12_381:
    bitsize: 381
    modulus: "0x1a0111ea397fe69a4b1ba7b6434bacd764774b84f38512bf6730d2a0f6b0f6241eabfffeb153ffffb9feffffffffaaab"
    # Equation: y^2 = x^3 + 4
  curve P256: # secp256r1 / NIST P-256
    bitsize: 256
    modulus: "0xffffffff00000001000000000000000000000000ffffffffffffffffffffffff"
```

Or in Stint, a significant amount of endian-aware code is special-cased for Nim VM:
- `initFromBytesBE` https://github.com/status-im/nim-stint/blob/9e49b00148884a01d61478ae5d2c69b543b93ceb/stint/io.nim#L421-L466
- `toByteArrayBE` https://github.com/status-im/nim-stint/blob/9e49b00148884a01d61478ae5d2c69b543b93ceb/stint/io.nim#L492-L514

Ideally the update module has compile-time tests and the same interface in NimVM and at runtime. The implementation can change and leverage compiler builtins and pointer casting where relevant.

changelog.md
johnnovak: Okay, didn't know about that. So this means all `{.since: (1, 1).}` pragmas in the codebase are wrong too and should be updated?

lib/pure/endians.nim
johnnovak: Cool, I'll make it work with NimVM then too somehow. I'm not using compile-time evaluation much myself, but I see that it can useful plus at least I'll learn a bit more about NimVM in the process. Back to work... 😄

changelog.md
nc-x: > So this means all {.since: (1, 1).} pragmas in the codebase are wrong too and should be updated?
  
No, v1.1 is the current devel. And if you change the since pragma to 1.2, then those procs won't be usable in devel.

narimiran: > So this means all {.since: (1, 1).} pragmas in the codebase are wrong too and should be updated?

No, as @nc-x said, they are correct and should be left as they are.

Araq:
> ````proc toBytesBE*[T: SomeNumber](value: T, buf: openArray[byte], index: Natural = 0)```

Why the index here? Nim's openArray does support slicing nowadays.

changelog.md
johnnovak: Okay fine, I'll do as instructed but I'm not quite getting this new versioning scheme... Is there a summary somewhere about it so I can understand it?

narimiran: Odd *patch* versions were traditionally devel versions, e.g. 0.18.1, 0.19.3, 1.0.5, etc.

We have adopted similar versioning for post-1.0 *minor* versions, e.g. 1.1.1 is current devel version.
The reason behind such decision is exactly stuff like this where we can do `since (x, y)` and it can be already tested with devel (version `x.y`) and will be included in the next stable release (version `x.y+1`).

johnnovak: Thanks for the explanation, I suspected this after thinking a bit about it and it makes sense. It's similar to the NodeJS versioning scheme.

mratsim:
> > ````proc toBytesBE*[T: SomeNumber](value: T, buf: openArray[byte], index: Natural = 0)```
> 
> Why the index here? Nim's openArray does support slicing nowadays.

I assume you are talking about this: https://github.com/nim-lang/Nim/pull/13463/files#diff-edbfa477abacea63b65f8bbf454c7727R296-R301

#### `var openarray`

Note that in this case the input is a `var openarray` and none of the openarray slicing methods return a `var openarray`.

#### Verbose slicing

Now even if we were not in a var context, openarray slicing is very verbose. Case in point:

Slicing with array syntax, allocates an array for `y[2 ..< 2+N]`
```Nim
let index = 2
type T = uint64
var x[0 ..< sizeof(T)] = y[index ..< index+sizeof(T)]
```
Slicing with openarray
```Nim
var x[0 ..< sizeof(T)] = y.toOpenArray(index, index+sizeof(T)-1)
```

#### In the end

So with the toBytesBE proc:

```Nim
value.toBytesBE(buf, index)
```
vs
```Nim
value.toBytesBE(buf.toOpenArray(index, index+sizeof(value)-1))
```

Araq:
Well the "verbose slicing" problem should be fixed differently, ``toOpenArray`` was the first iteration and now we can figure out a better syntax for it.

johnnovak:
> Well the "verbose slicing" problem should be fixed differently, toOpenArray was the first iteration and now we can figure out a better syntax for it.

We can update this API once we have less verbose slicing :) As @mratsim pointed out, this syntax is hard to beat:

```
value.toBytesBE(buf, index)
```

Araq:
Well then rename ``index`` to ``startIndex`` or similar.

johnnovak:
@mratsim Changed a few things here and there to make it more NimVM friendly plus added static tests. It's still not the "ideal" single API for both NimVM and runtime that you envisioned, but I'd prefer to keep the pointer variants because they're so much easier to use than faking openarrays. Also, endianness conversion comes up more often with raw buffers, so I think they're worth keeping.

I think this will fully satisfy your NimVM use cases. There's one small problem, namely that casting `int32/uint32` to `float32` is broken under NimVM. I couldn't find a workaround for this, so I've just raised a bug (https://github.com/nim-lang/Nim/issues/13479) and commented those tests out.

Otherwise, I think I'm done. I might add a few more runnable examples, but that's it.

Oh, and if someone could run the tests on a big endian hardware, that would be great!

johnnovak:
> Well then rename index to startIndex or similar.

That's not how it works, it just writes to a single index in the target array (it has to be an openarray of type we're coverting):

```nim
var i16arr: array[2, int16]
toBytesLE(1234'i16, i16arr, 1)
```

But actually I just realised this is a bit silly as you could just do `i16arr[1] = toBE(1234'i16)` just as easily. My bad...

I think I should rewrite the `toBytesXX/fromBytesXX` functions so they operate on `openArray[byte]` and the the `startIndex` would make sense (just like here: https://github.com/status-im/nim-stew/blob/1c4293b3e754b5ea68a188b60b192801162cd44e/stew/endians2.nim)

The pointer variants still makes sense though.

Thoughts?

mratsim:
> Oh, and if someone could run the tests on a big endian hardware, that would be great!

I think your best bet would be to add `s390x` or `ppc64` to Travis CI (https://github.com/travis-ci/travis-ci/issues/8869#issuecomment-515875324) but that probably should be done in a separate PR.

johnnovak:
> I think your best bet would be to add `s390x` or `ppc64` to Travis CI ([travis-ci/travis-ci#8869 (comment)](https://github.com/travis-ci/travis-ci/issues/8869#issuecomment-515875324)) but that probably should be done in a separate PR.

I have no experience with either Travis CI or Nim's build/testing process and I'd rather spend my time focusing on my binstream module, for which this PR is a prerequisite. Maybe you could help out, @mratsim ?

johnnovak:
Rewrote the openArray stuff with proper NimVM support and updated the tests. It was actually quite interesting, you can do lots of cool stuff with templates and typedescs :)

Documentation update and some more runnable examples will come tomorrow, but I think the code is pretty much final now.

lib/pure/endians.nim
```diff
@@ -98,16 +135,853 @@ else:
...
+    else:       cast[T](    swap16(cast[uint16](value)))
+  elif T is SomeNumber32:
+    when nimvm: cast[T](slowSwap32(cast[uint32](value)))
+    else:       cast[T](    swap32(cast[uint32](value)))
+  elif T is SomeNumber64:
+    when nimvm: cast[T](slowSwap64(cast[uint64](value)))
+    else:       cast[T](    swap64(cast[uint64](value)))
+
+
+proc swapEndian*(T: typedesc[SomeNumber],
```
Clyybber: Can these be funcs too?

lib/pure/endians.nim
```diff
@@ -98,16 +135,853 @@ else:
...
+  template swap32(a: uint32): uint32 = slowSwap32(a)
+  template swap64(a: uint64): uint64 = slowSwap64(a)
+
+
+func swapEndian*[T: SomeNumber](value: T): T {.inline, mySince: (1,1).} =
+  ## Swaps the byte order of a number.
+  runnableExamples:
+    doAssert swapEndian(0xdeadbeef'i32) == 0xefbeadde'i32
+
+  when T is SomeInteger8:
```
Clyybber: Maybe its better to just do `when sizeof(T) == ...:` ?

johnnovak: I kinda like it this way.

lib/pure/endians.nim
johnnovak: They have side effects.

Araq: Wait what?

Araq:
Please remove your SomeNumber32 et al additions. No new symbols for system.nim, it's already so bloated that it's the biggest nim module in the Nim compiler.

lib/pure/endians.nim
johnnovak: Well, some of these procs mutate memory locations.

johnnovak:
> Please remove your SomeNumber32 et al additions.

That's ok, I was on the fence about these actually. Maybe I'll just use `sizeof` as suggested earlier.

lib/pure/endians.nim
johnnovak: Changed my mind, I'll use `sizeof`.

johnnovak:
@Araq Updated again, removed the SomeNumberXX stuff and update the docs. I also had to remove the `{.since.}` from a function to make the CI build pass (see the last commit). Not sure why that was necessary...

lib/pure/endians.nim
johnnovak: Okay, just realised that when you guys say "side effects", you don't actually mean the [standard definition](https://en.wikipedia.org/wiki/Side_effect_(computer_science)) :) I do pure-FP style coding as my day job in Scala, so this surprised me a bit. But okay... in the sense how Nim views side effects, these are side effect free (they don't touch any globals). Made them all funcs.

lib/pure/endians.nim
```diff
@@ -98,16 +126,919 @@ else:
...
+    doAssert fromBytesLE(int32, buf, 1) == 0xdeadbeef'i32
+
+  when nimvm: slowFromBytes(T, littleEndian, buf, startIndex)
+  else:
+    assert startIndex < buf.len
+    assert startIndex + sizeof(T) <= buf.len
+    fromBytesLE(T, buf[startIndex].unsafeAddr)
+
+
+when defined(testing) and isMainModule:
```
krux02: Why do you check on `defined(testing)`? As far as I am aware a `when isMainModule` is enough.
Apart from that. The pattern that when I want to test something on the VM and at runtime, I run the exact same code at compile time and at runtime:

```nim
when isMainModule:
  proc localTest() = [...] # can be a template, too
  localTest()
  static:
    localTest()
```

lib/pure/endians.nim
```diff
@@ -98,16 +126,919 @@ else:
...
+    toBytesLE(i64, buf, 3); assert fromBytesLE(  int64, buf, 3) == i64
+    toBytesLE(u64, buf, 3); assert fromBytesLE( uint64, buf, 3) == u64
+    toBytesLE(f32, buf, 3); assert fromBytesLE(float32, buf, 3) == f32
+    toBytesLE(f64, buf, 3); assert fromBytesLE(float64, buf, 3) == f64
+
+
+  # NimVM tests
+  # -----------
+  # Most float32 tests are commented out because they crash the VM.
+  # This is a known limitation, the VM only supports float64 currently.
```
krux02: This should be fixed by now.

mratsim:
> I have no experience with either Travis CI or Nim's build/testing process and I'd rather spend my time focusing on my binstream module, for which this PR is a prerequisite. Maybe you could help out, @mratsim ?

When I have one weekend or 2 to lose ;), I don't want to fight against CI for the next 6 months.

johnnovak:
> I don't want to fight against CI for the next 6 months.

Exactly how I feel about this too... :)

lib/pure/endians.nim
johnnovak: Indeed, all tests are passing now with the head version.

lib/pure/endians.nim
johnnovak: Good points, I've done these changes.

lib/pure/endians.nim
```diff
@@ -98,16 +124,734 @@ else:
...
+  elif sizeof(T) == 4:
+    when nimvm: cast[T](slowSwap32(cast[uint32](value)))
+    else:       cast[T](    swap32(cast[uint32](value)))
+  elif sizeof(T) == 8:
+    when nimvm: cast[T](slowSwap64(cast[uint64](value)))
+    else:       cast[T](    swap64(cast[uint64](value)))
+
+
+func swapEndian*(T: typedesc[SomeNumber],
+                 buf: pointer) {.inline, since: (1,1).} =
```
Araq, line 225: Er... what? How about ``swapEndian*[T: SomeNumber](buf: var T)``

lib/pure/endians.nim
```diff
@@ -98,16 +124,734 @@ else:
...
-  proc littleEndian16*(outp, inp: pointer){.inline.} = copyMem(outp, inp, 2)
-  proc bigEndian64*(outp, inp: pointer) {.inline.} = swapEndian64(outp, inp)
-  proc bigEndian32*(outp, inp: pointer) {.inline.} = swapEndian32(outp, inp)
-  proc bigEndian16*(outp, inp: pointer) {.inline.} = swapEndian16(outp, inp)
+  template swap16(a: uint16): uint16 = slowSwap16(a)
+  template swap32(a: uint32): uint32 = slowSwap32(a)
+  template swap64(a: uint64): uint64 = slowSwap64(a)
+
+
+# TODO had to remove {.since: (1,1).} because it broke `koch runCI`
```
Araq, line 203: Well, .since is not negotiable here. How come "koch runCI" failed with this? We use it successfully everywhere else.

lib/pure/endians.nim
```diff
@@ -98,16 +124,734 @@ else:
...
+
+  var value = cast[ptr T](buf)[]
+  cast[ptr T](buf)[] = swapEndian(value)
+
+
+func swapEndian*(T: typedesc[SomeNumber], buf: var openArray[byte],
+                 startIndex: Natural) {.inline, since: (1,1).} =
+  ## Swaps the byte order of a number stored as a byte sequence in `buf`
+  ## starting at `startIndex`. The type of the number must be passed in as an
+  ## argument.
```
Araq, line 241: Huh? What about ``var openArray[T]``? Same for all the other funcs that use that weird type signature.

lib/pure/endians.nim
```diff
@@ -98,16 +124,734 @@ else:
...
+  ## endiannes unaware I/O to native byte order, so it can be used for
+  ## arithmetic calculations in a program. This is equivalent to `toLE`,
+  ## it's only included for readability.
+  ##
+  ## Consider using `fromBytesLE` instead to make the intent clearer.
+  toLE(value)
+
+func toBytesBE*[T: SomeNumber](value: T, buf: pointer)
+    {.inline, since: (1,1).} =
+  ## Writes a number to a memory location as a big-endian byte sequence.
```
Araq, line 315: ``pointer`` is unnecessarily unsafe and low level.

Araq:
Too many procs. We need to reduce the API surface. My guess is that the openArray versions are rather pointless in practice as one has to deal with arrays of structs with fields of 2, 4, 8 bytes in size, so it cannot be used and you end up with ``fieldPairs`` over custom objects.

lib/pure/endians.nim
Araq: Ah, I misunderstand this proc. But it's not necessary, remove it.

Araq:
Having thought about this more, neither the new nor the old API is ideal. The old is a little too bare-bones, while the new is bloated by a factor of 3 ("pure function, openarray and pointer variants").

In reality there is only ``swap`` for the sizes 2, 4, 8. That's 3 different procs. The size can be derived from the type. Assuming that you have a type like

```nim

type
  Foo = object
    length: uint16
    data: ptr UncheckedArray[uint64]
```

and it arrives in Network order but you need it in little endian (why though? Shouldn't it be "native endian"), the code looks like:

```nim
var f: Foo
when system.cpuEndian == littleEndian:
  swapEndian(f.length)
  for i in 0..< f.length: swapEndian(f.data[i])
```

Suggesting we should have ``proc swapEndian[T: SomeNumber](x: var T)``. And nothing else is required.

johnnovak:
@Araq Thanks for the review. The funny thing is that I'm in the process of implementing `binstreams` which actually only uses `swapEndian` internally, so my initial reaction was to agree with you. But then I thought about it some more during the day and I'm not so sure anymore.

First, one man's bloated API is another's well specified :) So there's definitely a taste aspect here. I for one have been using the Java and Scala standard libraries for more than a decade now and I really like the fact that they are exhaustive and well-specified. Also, for the record, I used [endians2 ](https://github.com/status-im/nim-stew/blob/1c4293b3e754b5ea68a188b60b192801162cd44e/stew/endians2.nim) as a starting point which I thought was already pretty nice, so I guess I'm not the only one who likes more well-specified APIs.

Only keeping `proc swapEndian[T: SomeNumber](x: var T)`, a single function... well it doesn't get more barebones than that, does it? :) Of course you can express everything else in terms of this single function, but isn't it the point of a standard library to make common programming tasks easier and convenient, so people won't need to come up with their own helpers all the time? That's why we have so many functions in the strutils module, everybody could write their own, but what's the point of that? And you don't need all of them all the time, but it's handy that they're already there when you do.

Reflecting on your example, writing `when system.cpuEndian == littleEndian: ...` gets old really fast. It's lots of noise, the first thing I'd do is to write a `toBE/LE` & `fromBE/LE` template myself in my program. So why not provide these in the module then? Even Nim's own nativesockets module contains the `ntohl` and `ntohs` functions. The core Apple SDK contains [26 functions](https://developer.apple.com/search/?q=endians) and macros related just for such endianness conversion. I mean, if we have an endianness *module*, it should help with common operations, otherwise what's the point of having a whole module dedicated to this in the first place? We might as well just chuck the single generic swap function into system and be done with it.

Also, I think a general purpose endianness API should be able to handle byte buffers. See Java's [ByteBuffer](https://docs.oracle.com/javase/10/docs/api/java/nio/ByteBuffer.html) implementation as an example. In my experience these can be very handy, for example at one point I had to send requests to an SMS sending service, where I had to encode UTF-8 code points with big-endian byte-order, then take the whole byte array and URI encode it and THEN send it in a HTTP request as text. Byte buffers (`openArray[byte]`) make these situations easier to deal with in conjunction with `to/fromBytesBE/LE` and result in more readable code that makes the intent clear.

There also lots of bizarre mixed-endian binary formats out there, it's much easier and more readable to chuck the data into a byte buffer and use `to/fromBytesBE/LE` on the data, as opposed to litter the code with lots of `when system.cpuEndian == xxxEndian` stuff. Some examples:

https://github.com/haskell-hvr/uuid/issues/48
https://stackoverflow.com/questions/8841857/is-there-a-performance-rationale-for-mixing-a-binary-files-endianness

Finally, I don't see why should we optimise for the least number of API functions. Especially when with Nim the unused functions are not even compiled into the executable because of static linking, right? 

To sum it up, this is how I see it:

- `swapEndian` is a good basic function to have, but I think I'd prefer the non-var variant that returns a value instead of mutating it
- `toBE/toLE/fromBE/fromLE` are useful, many other standard libraries define them as macros as well because they result in more readable code
- the `openArray[byte]` variants of `fromBytesBE/fromBytesLE/toBytesLE/toBytesBE` are very useful for more complicated stuff when you need to prepare some binary data packet to be sent to a socket or whatever (maybe they would be handy with various CPU emulators too, but that's not my area of expertise)
- if we're worried about bloat, I think we can safely drop the `pointer` variants of all functions

@mratsim I think it would be valuable if you shared your perspective on the matter too.

Araq:
Well I want the minimum, you want the maximum so as a compromise we'll find a nice middle ground. But we need to look at your upcoming binstreams.nim etc to see what it really needs, rather than providing 1000 lines of code "just in case" for what could be 90 lines of code. And yeah, strutils is so bloated that I consider to move the stuff that's really useful to ``strbasics``.

johnnovak:
> But we need to look at your upcoming binstreams.nim etc to see what it really needs

It's almost done and it only needs `swapEndian`, as I said. So if that's the deciding factor, we'll end up with a single function module :P

johnnovak:
What about just keeping `swapEndian` and `toBE/toLE/fromBE/fromLE` in the `endians` module, that's just 5 functions in total, and moving everything else into a new `endianutils` module? That would keep everybody happy I guess.

mratsim:
> Too many procs. We need to reduce the API surface. My guess is that the openArray versions are rather pointless in practice as one has to deal with arrays of structs with fields of 2, 4, 8 bytes in size, so it cannot be used and you end up with `fieldPairs` over custom objects.
>
> Having thought about this more, neither the new nor the old API is ideal. The old is a little too bare-bones, while the new is bloated by a factor of 3 ("pure function, openarray and pointer variants").
>
> In reality there is only swap for the sizes 2, 4, 8. That's 3 different procs. The size can be derived from the type. Assuming that you have a type like

1. We only need either "pure functions+openarray" or "pure functions+pointers". We can easily cast pointer+len to openarray or convert openarrays to pointer.
I think openarray are better than pointer because we would have the normal Nim checks, and they would also work at compile-time and point users to https://github.com/nim-lang/Nim/blob/version-1-0/lib/system.nim#L4542

2. The proposed swapEndian signature is not enough, often you have a preallocated buffer that is your internal representation and want to convert from an immutable stream/raw data to that internal representation. So besides

```Nim
proc swapEndian[T: SomeNumber](x: var T)
```

We also need

```Nim
proc swapEndian[T: SomeNumber](dst: var T, src: T)
```

3. We need to be able to convert whole buffers from one representation to the others. In practice we need:
- nativeToBig
- nativeToLittle
- bigToNative
- littleToNative
If for some reasons big<->little is needed it's easy to use `when cpuEndian == littleEndian` with one of the previous primitives or no-op.

## Use-cases:

### Big Integers

The canonical/serialized representation of big integers is an `openarray[byte]` with the integers stores as big-endian. This is called `octet-string` in cryptographic spec.
  In actual Nim code, the size to convert is **32 bytes** to **48 bytes**, repos:
  - https://github.com/status-im/nim-stint
  - https://github.com/status-im/nimbus/
  - https://github.com/status-im/nim-bncurve
  - https://github.com/status-im/nim-blscurve
  - https://github.com/mratsim/constantine

Internally we need to deal with:
- word-endianness: often native-endian
- limb-endianness: often little-endian so that myBigInt.limbs[0] is the less significant word
- byte-endianness for serialization often big-Endian
Though Stint and Nimbus are fully native-endian (i.e. if there was native uint256, the representation could be casted to Stint types)

### Cryptography

At the interface of cryptographic and big-integers there are often conversions going on as cryptography standards use bigEndian integers:
- RSA is usually between 256 bytes to 512 bytes (2048-bit to 4096-bit)

Cryptographic secret keys, public keys and signatures are in the 32 to 96 bytes range and require native to bigEndian conversion on serialization and bigEndian to native on parsing.

### Networking

We only have 2 to 8 bytes integer there so `swapEndians` is enough. But networking often requires cryptography.

### Example endian aware code

This is endian-aware code that we have to write in our libraries:

- https://github.com/status-im/nim-stew/blob/50562b51/stew/endians2.nim
- https://github.com/status-im/nim-stint/blob/9e49b001/stint/io.nim#L382-L514
- https://github.com/status-im/nimbus/blob/15c9fa54/nimbus/vm/precompiles.nim#L153-L169
- https://github.com/status-im/nim-bncurve/blob/9b8aac56/bncurve/fp.nim#L117-L158 (This is used in Nimbus codebase and conversion Stint <-> BNcurve is inefficient as it requires an intermediate bigEndian conversion step)
- https://github.com/mratsim/constantine/blob/ca817fcb/constantine/io/io_bigints.nim#L162-L170

The last one in particular is the most useful to export from complex internal representation to canonical little or big-endian representation including VM compatibility:
```Nim
template blobFrom(dst: var openArray[byte], src: SomeUnsignedInt, startIdx: int, endian: static Endianness) =
  ## Write an integer into a raw binary blob
  ## Swapping endianness if needed
  when endian == cpuEndian:
    for i in 0 ..< sizeof(src):
      dst[startIdx+i] = toByte((src shr (i * 8)))
  else:
    for i in 0 ..< sizeof(src):
      dst[startIdx+sizeof(src)-1-i] = toByte((src shr (i * 8)))
```
It can be improved for runtime by using the swap compiler intrinsics.

arnetheduck:
Generally, I'd avoid pointer and openArray as far as possible.

* value semantics allow the compiler to perform better optimizations and more closely models the hardware instructions
* passing `var openArray` is problematic: there's now the possibility that exceptions are raised (indexerror) and the additional pointer indirection prevents certain optimizations (alignment issues, etc) - for example in wasm, passing by pointer results in more verbose assembly code
* working with arrays instead avoids all these problems - the caller can copy the resulting array into their openArray if they want to - in practice, in the assembly code, this is what happens anyway.
* `swapEndian` taking int as a type is problematic: either the input or the output _is not an integer_

the set of functions that I'd consider critical is `toXEndian(uintX): array[X: bytes]` and `fromXEndian(array[X, byte]): uintX` - this because you often need to have a specific endianness (in binary protocols) and you want the code to work generically for all platforms. again, we're using value semantics because that results in more efficient code which is reasonable for low-level code like this.

the other function would be `swapBytes(array[N, byte): array[N, byte]` or some variation on that - `swapEndians` seems like the kind of function that allows bytes that are not in the right endianess for the platform to propagate, so if you want to call it, you should explicitly convert to bytes first.

returning an `int` in `toBE(intX): intX` I consider a broken API: after byte-swapping, the result is no longer an integer that you can reasonably perform arithmetic on - I believe all such functions should be removed.

finally, during the review of `endians2`, the issue came up that if you have 3 bytes in your buffer and you want to convert to an `uint32` - there exists a reasonable semantic to just use the 3 bytes and treat the missing byte as zero-fill (which side depends on which endian) - I consider this such a minor use case that it should not be addressed at this low level.

arnetheduck:
the suggested functions form a nucleus of total, exception and defect-free functions that can be used safely from everywhere - this is what the hardware typically looks like - on top of that, one can imagine a few helpers, for example what @mratsim suggests, that make it convenient to call when working with other kinds of byte buffers, but then you diverge from the hardware and have to pay for it with semantic and performance-based inefficiencies like range checks, exceptions, indirections, nulls etc etc

johnnovak:
> value semantics allow the compiler to perform better optimizations and more closely models the hardware instructions

We have the `fromBE/fromLE/toBE/toLE/swapEndian` pure functions so you can write code like this if you are trying to micro-optimise things:

```nim
var in: array[uint32, 1024]
for i in 0..in.high:
  out[i] = toBE(in[i])
```

> passing var openArray is problematic: there's now the possibility that exceptions are raised (indexerror) and the additional pointer indirection prevents certain optimizations (alignment issues, etc)

Is this all true though? I am under the impression that passing in an array as an openArray argument incurs no performance penalty compared to plain arrays. Arrays would be treated as arrays and seqs as seqs, you just don't need to write two functions. If this is not the case, maybe there's a more general problem here that should be fixed at the language level because then it kind of beats the purpose of having openArrays in the first place.

I'm not exactly sure what you mean by alignment issues either, I just ran this quick test (on Intel hardware, a Macbook Pro):


```nim
var s8 = @[1'u8, 2'u8]
var s16 = @[1'u16, 2'u16]
var s32 = @[1'u32, 2'u32]
var s64 = @[1'u64, 2'u64]

echo "uint8"
echo cast[ByteAddress](s8[0].addr) mod 8
echo cast[ByteAddress](s8[1].addr) mod 8

echo "\nuint16"
echo cast[ByteAddress](s16[0].addr) mod 8
echo cast[ByteAddress](s16[1].addr) mod 8

echo "\nuint32"
echo cast[ByteAddress](s32[0].addr) mod 8
echo cast[ByteAddress](s32[1].addr) mod 8

echo "\nuint64"
echo cast[ByteAddress](s64[0].addr) mod 8
echo cast[ByteAddress](s64[1].addr) mod 8
```

Which produces the expected good output, all seqs are 8-byte aligned:

```
uint8
0
1

uint16
0
2

uint32
0
4

uint64
0
0
```


Having said that, you can always use `swapEndian/toBE/...` in a for loop if you really need that level of control (and if it really helps).


> the set of functions that I'd consider critical is toXEndian(uintX): array[X: bytes] and fromXEndian(array[X, byte]): uintX - this because you often need to have a specific endianness (in binary protocols) and you want the code to work generically for all platforms. again, we're using value semantics because that results in more efficient code which is reasonable for low-level code like this.

Well, functions accepting openArrays are more versatile and as I said I'm not convinced they incur a performance penalty (but I haven't inspected the assembly output). I'm not quite following this line of thought; if an array passed in to an openArray argument behaves just like an array, won't the compiler just be able to apply the same optimisations?

But again, I think a library should be convenient to use, with the possibily to go really low-level if you want to, but it shouldn't force you to do that, because most of the time it's just not worth it.  Quite often you use the endianness conversions before/after IO and you're not manipulating gigabytes of data anyway, so the difference from the "theoretical best performance" (I'm using this term on purpose) simply doesn't matter at all. And when it matters, most often you'll end up with some horrible looking code anyway and quite possibly you know what you're doing and you'll need to write some custom code or assembly anyway. But... this is a library. I guess many parts of the stdlib could be more "more efficient" for specialised cases by specialised code, but usually they're fast enough for normal scenarios so it's rare that you need to write those specialised versions.

> swapEndian taking int as a type is problematic: either the input or the output is not an integer

This is just a case of unnecessary pedanticism over practicality. The fact is, I can read
a value from a stream that is not in the native byte order, then it does exist in
my variable (or CPU register), so why should I be forbidden to run a function on it to swap
the byte order?

lib/pure/endians.nim
arnetheduck: just found a bug related to this that could easily have been avoided with good API: https://github.com/status-im/nim-eth/blob/bc6c9819166fd2aba242aee06bd7ea59725ebcac/eth/p2p/auth.nim#L195

notice how the address of an `int` is passed in - this works on little endian platforms but not on big endian platforms, which is extra funny

lib/pure/endians.nim
```diff
@@ -98,16 +124,734 @@ else:
...
+                 buf: pointer) {.inline, since: (1,1).} =
+  ## Swaps the byte order of a number at a memory location in-place. The type
+  ## of the number must be passed in as an argument.
+  runnableExamples:
+    var i = 0xdeadbeef'i32
+    swapEndian(int32, i.addr)
+    doAssert i == 0xefbeadde'i32
+
+  var value = cast[ptr T](buf)[]
+  cast[ptr T](buf)[] = swapEndian(value)
```
arnetheduck, line 234: here's the alignment bug, from what I can tell: 

if `buf` points to an address that is not aligned for T, it's not valid to write a T-sized value to it. unaligned writes often happen when you're putting things in network buffers that are packed - the very use case you're trying to be convenient for.

this is one way in which this API is unsafe - the other is obviously buffer overruns - I'd prefer to see that the std library is kept memory-safe as far as possible.

arnetheduck:
> align

see https://github.com/nim-lang/Nim/issues/9206 - this is not the first time this bug appears in this library.

you cannot test alignment issues on Intel processors, they allow unaligned memory access (albeit slower, if it falls on a cache line boundary). other architectures don't.

> compared to plain arrays.

arrays are value types, so depending on compiler impl, they could the same underlying representation as a normal `uint32` - when you pass one into a function it's possible to implement without pointers / indirection. It's only when you introduce `openArray` that you need to start using pointers. the current nim impl uses a C array which decomposes into a single pointer when passed to a function, but it doesn't have to be this way, and it seems easier to fix the compiler in a few months, than all the code that will come to depend on this.

when experimenting with this in wasm, I had to use an uint32 instead of an array to get an optimal version with the current nim compiler, but if I was writing a standard library, I'd rather ensure that the compiler generated good code instead of working around it in the std lib.

the difference was significant for the use case - smart contracts on blockchains - where space is at a premium and endian conversion is common. what happens is that instead of passing the value directly, a local variable has to be allocated, then a pointer to it has to be passed to the function, then the function has to dereference the variable and perform the operations on the resulting integer - this sequence of operations is much longer and cannot be optimised as easily, because the compiler must consider buffer pointers that are not aligned properly. 

if you're using openArrays, they have dynamic length - at the very least, you'll be passing pointer+length instead of just a pointer - this is a (tiny) performance cost, but more importantly, any access to the openArray will be length-checked at runtime. Some C compilers may be able to constant-propagate and remove the extra checking and raising. better be clear in the API how many bytes you require and let the caller ensure that the right number of bytes is there.

> The fact is, I can read a value from a stream that is not in the native byte order,

you can, but should a standard library encourage this kind of unsafe code? the whole point of using a typed language is to represent types accurately - and when changing byte order, you're really looking at an array of bytes.

that said, I can see there is a convenience argument. we left the uint-to-uint variants in `endians2` for this reason (but that's not a standard lib).

one last thing: int in nim is range-checked - this leads to some really interesting issues after endianness conversion - generally, since endians are typically used in binary protocols, I'd recommend that _no_ `int` versions be present and only unsigned integers supported. callers can easily cast to the unsigned variant if they feel the need - specially when value-semantics are used and the endian conversion functions return a value instead of writing it to a given buffer.

arnetheduck:
> all seqs are 8-byte aligned:

indeed they are, on your computer. 

the memory layout of a seq is two `int` followed by the data. the memory to the seq is always allocated on at least a sizeof(int) alignment because otherwise said `int` values could not be accessed, thus it follows that your data is aligned as well.

but this is typically not how buffers are used, specially when used with binary data - typically you have some packed binary format that defy normal alignment rules that the compiler works hard to comply to. if you're writing some kind of binary streaming library on top, you should be specially careful about this.

this is another reason `toBytesBE(x: uint32): array[4, byte]` is a better api: when you copy the data from the resulting array, the compiler will know to generate alignment-safe access - either copying byte-by-byte, or sometimes when it can prove that the data is sufficiently aligned already, a more efficient variation.

we've had these alignment issues in the standard library several times now, and I keep finding them in libraries that promote these kinds of unsafe API - sadly, often in the libraries that are meant to be dealing with packed binary streams that modules like these are supposed to serve - I suspect this is the case mostly because the alignment issues merely cause slowdowns on x86 and not crashes.

arnetheduck:
> The main reason we wrote endians2 at Status was to allow compile-time endianness manipulation.

well, to avoid the unsafe and cumbersome API with pointers as well - I should know because I wrote it ;) having the converted value returned instead of written to a pointer buffer is often more ergonomic, depending on surrounding code - it is certainly safer in all kinds of ways and more optimizeable.

compare `let x = 42'u32.toBytesBE()` vs:
```
var x: array[4, byte]
42'u32.toBytesBE(x)
```

I'd say the former is easier to read, shorter, has no exception or buffer overrun issues. if you want to assign to a seq, you already can do: `someSeq[0..<3] = 42'u32.toBytesBE()` which more clearly demonstrates what will happen in terms of memory accesses and required buffer sizes.

> big integers

@mratsim I think these are out of scope for this std module, which otherwise only operates on native integer sizes

> In reality there is only swap for the sizes 2, 4, 8. That's 3 different procs. 

as others have pointed out, the most common use case for dealing with endians is when you go from a specific endianness (big endian/network order) to platform endian - it would be tedious to have to check platform endianness every time before swap.

lib/pure/endians.nim
arnetheduck: this is why we wrote `endians2` as well: the existing API is needlessly unsafe for no good reason

arnetheduck:
fwiw, a thing to consider would be adding `endians2` as-is in the std lib - it has already gone through some battle testing and has proven useful in finding and avoiding bugs. we created `stew` so that we could experiment with API and make sure they're solid before proposing anything to std lib - this feels like a more natural progression, rather than experimenting in the std lib. think boost.

If you are writing something based on `stew/endians2` and find the API lacking, we'd be curious to hear back (with code samples) what didn't work well. We can then solve it there, continue experimenting and not burden the wider Nim community with poorly tested interfaces.

It lacks the compatibility section simply because I found it easier to create a new module than trying to juggle the two api's, but that's easy to solve - or the two modules can be kept forever and the old one can just be deprecated - this has the advantage of guaranteed 100% compatibility.

Araq:
> fwiw, a thing to consider would be adding endians2 as-is in the std lib - it has already gone through some battle testing and has proven useful in finding and avoiding bugs

Yes, was about to propose the same.

arnetheduck:
> Yes, was about to propose the same.

I'd wait for a bit though - I mean, it's there in stew, people are free to us it and we probably missed some important cases. 

One thing I'm missing for example, to make the current `endians2` interface convenient, is this function: https://github.com/status-im/nim-stew/pull/18/files#diff-d4c04f942120772be663d48efa847ce6R28 - without it, it's inconvenient to make slice assignments from the array that's returned - notably, if we're assigning to an array from an array or a compile-time sized slice, the range checks should all be made at compile time - the function I refer to above is missing that.

johnnovak:
@arnetheduck First of all, thanks for your detailed answer and patiently explaining these things, now it all makes sense. Yes, I'm only using Nim on Intel hardware for desktop/terminal apps so crashes due to unaligned access is news to me.

Re-reading `endians2` again after your explanation puts it in a very different light. Now I too agree that it should be included in the stdlib instead of my attempt. Well, I tried to "improve" on it because I didn't like the array return types, but that was when I didn't understand the reasons why it was done that way.

I also think I kind of conflated the endianness conversions with some sort of a byte buffer implementation in my functions, which should actually be a separate module. But don't worry, I won't come up with a `bytebuffer` module tomorrow :)

All in all, I have learned a lot from this discussion and I will think about these issues a bit more before presenting my `binstreams` module (at least, I'll use it for a bit longer in my projects).

arnetheduck:
a bit off-topic, but @Araq, thoughts on `struct { char x[N]; };` for arrays? that should give it just barely enough value semantics to be usable in parameters etc - but I guess there are issues as well? I remember reading strong language somewhere in the cgen about the current approach ;)

Araq:
> a bit off-topic, but @Araq, thoughts on struct { char x[N]; }; for arrays? that should give it just barely enough value semantics to be usable in parameters etc - but I guess there are issues as well? I remember reading strong language somewhere in the cgen about the current approach ;)

Indeed I'm considering it.

johnnovak:
So what about this PR? You guys want me to close it, or shall we keep it open for discussions? But I think we kinda agreed that `endians2` is the way to go anyway?

Araq:
Let's close it for now.

arnetheduck:
an alternative here is that we move `endians2` to its own nimble package, if people don't want to use it from within `stew` - this would serve many purposes, including clean upgrade paths, "regular" version management with stablility etc - the `endians` module in the std lib would then simply be deprecated with a notice that the new package is the recommended place to go.

johnnovak:
@arnetheduck That seems like a good idea. Right now I'm just depending on `stew` in my projects, a separate `endians2` package would be even better.

arnetheduck:
also, for [nim-result](https://github.com/arnetheduck/nim-result/blob/master/to_stew) I have my own experimental copy that I sync with stew occasionally.. for these simple one-file libraries, it's quite feasible. 

we don't use nimble at status - another easy way to use simple libs like endians2 and result is to copy them to your project - this has many advantages also

PR #13394, opened 2020-02-12
Fix to asynchttpserver form data/body broken with #13147

mrhdias:
This fix add a new server option that allow the request body to be processed outside the asynchttpserver library to break big files into chunks of data. This change does not break anything and works with the jester and the rosencrantz.

The new server option "stream" when it is true uses the Future Streams otherwise all body is readed at once. The "stream" option by default is set to false.

The idea of this option I took from the python requests module (https://requests.readthedocs.io/en/v1.2.3/api/).
"When stream=True is set on the request, this avoids reading the content at once into memory for large responses. The chunk size is the number of bytes it should read into memory. This is not necessarily the length of each item returned as decoding can take place."

Example:
```nim
let server = newAsyncHttpServer(stream = true) # if true use Future Streams
waitFor server.serve(Port(8080), cb)
```

Issue #6576, opened 2017-10-23
Async iterators


dom96:
This is related to ``FutureStream``s. It sure would be neat to be able to do something like this:

```nim
import asyncdispatch, httpclient
let client = newAsyncHttpClient()

let resp = client.get("http://google.com")
for data in resp.bodyStream:
  # Do something with data.
```

This might not be possible.

zielmicha:
reactor.nim has asyncFor: https://github.com/zielmicha/reactor.nim/blob/ef6afa3de8946b4959d5525de04ea42202541892/reactor/async/asyncmacro.nim#L197

Example:
```
asyncFor item in src:
  echo "piping ", item
  await dst.send(item)
```

It's implemented as a very simple macro (see source).

With support from `{.async.}` macro (e.g. translate stmtFor into call to `asyncFor`) it's possible to achieve `for` syntax you want. Of course then stdlib version of `asyncFor` macro would also need to handle normal iterators.

jackhftang:
FYI, a working example as inspired by reactor.nim. 
```nim
import asyncdispatch, httpclient, macros

macro asyncFor*(iterClause: untyped, body: untyped): untyped =
  if iterClause.kind != nnkInfix or iterClause[0] != ident("in"):
    error("expected `x in y` after for")

  let stream = iterClause[2]
  let itemName = iterClause[1]
  result = quote do:
    while true:
      let future = read(`stream`)
      yield future
      if future.failed: 
        raise future.readError()
      let tu = future.read()
      if not tu[0]: break
      let `itemName` {.inject.} = tu[1]
      `body`

proc main() {.async.} = 
  let client = newAsyncHttpClient()
  let resp = await client.get("http://google.com")
  asyncFor s in resp.bodyStream:
    echo s[1..20]

waitFor main()
```

The may not be possible part is that there is no true `await` (which is a macro transformed to future iterator), so it is not possible to have proc `items()` on `FutureStream[T]`, which is sync, with async effect. Here is an example with implemented with `waitFor`, it is not so right, but it works.

```nim
import asyncdispatch, httpclient

iterator items[T](stream: FutureStream[T]): T =
  while true:
    let future = stream.read()
    let tu = waitFor future
    if not tu[0]: break
    yield tu[1]

proc main() {.async.} =
  let client = newAsyncHttpClient()
  let resp = await client.get("http://google.com")
  for data in resp.bodyStream:
    echo data[..20]
    
waitFor main()
```

[...]

PR #13468, opened 2020-02-22
Revert broken asynchttpserver FutureStream additions.

dom96:
As discussed in #13394, these changes cannot work. Reverted via

```
git revert --no-commit 5bf571f061d53d35aab727f420afd9f415987723
git revert --no-commit abd660c407d00d0c4f2129ff11bfc69badda8ece
git revert --no-commit 955465e5f42b1353f69f3bd884908a7ef91ce13b
git commit
```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -72,8 +72,6 @@
 
 ## Library changes
 
-- `asynchttpserver` added an iterator that allows the request body to be read in
-   chunks of data when new server "stream" option is set to true.
 - `asyncdispatch.drain` now properly takes into account `selector.hasPendingOperations`
   and only returns once all pending async operations are guaranteed to have completed.
 - `asyncdispatch.drain` now consistently uses the passed timeout value for all
--- a/lib/pure/asynchttpserver.nim
+++ b/lib/pure/asynchttpserver.nim
@@ -29,129 +29,43 @@
 ##      await req.respond(Http200, "Hello World")
 ##
 ##    waitFor server.serve(Port(8080), cb)
-##
-## Basic Post request handle
-## =========================
-##
-## This example will create an HTTP server on port 8080. The server will
-## respond with a page with the actual and expected body length after
-## submitting a file.
-##
-## .. code-block::nim
-##    import asynchttpserver, asyncdispatch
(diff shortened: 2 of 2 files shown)


Araq:
Pinq @mrhdias. As I said elsewhere, better than reverting this is fixing the feature so that it works well. It's an important feature.

mrhdias:
> Pinq @mrhdias. As I said elsewhere, better than reverting this is fixing the feature so that it works well. It's an important feature.

Thank you @Araq. The idea of using an iterator is not new.
https://github.com/nim-lang/Nim/issues/6576

I replace "let data = request.client.recv(readSize)" with "let data = waitFor request.client.recv(readSize)" to wait for the data.

So, reading the data is clear:
```nim
for data in req.bodyStream(8*1024):
  echo data
```
Iterator code
```nim
when (NimMajor, NimMinor) >= (1, 0):
  iterator bodyStream*(
    request: Request,
    chunkSize: int = 8*1024): string =
    ## The chunkSize parameter is optional and default value is 8*1024 bytes.
    ##
    ## This iterator return a tuple with the length of the data that was read
    ## and a future.
    var remainder = request.contentLength
    while remainder > 0:
      let readSize = min(remainder, chunkSize)
      let data = waitFor request.client.recv(readSize)
      if data.len != readSize:
        raise newException(ValueError, "Error reading POST data from client.")
      yield data
      remainder -= readSize
```

mrhdias:
Another solution forum (Thanks Hlaaftana): https://forum.nim-lang.org/t/5971

```nim
template forStream(req: Request, chunkSize: int, name, body: untyped): untyped =
  var remainder = req.contentLength
  while true:
    let `name` = await req.client.recv(min(remainder, chunkSize))
    body
    remainder -= len(`name`)
    if remainder == 0:
      break

req.forStream(8*1024, d):
  echo d
```

dom96:
It is better to fix this, but it's going to be much easier to review if we start fresh IMO. Can we merge this?

dom96:
Okay, I am merging this since there was no feedback in 12 days. Please ping me to review any new PRs on this.

PR #13473, opened 2020-02-23
Replace the iterator that reads the chunks of data with a cleaner one.

mrhdias:
I replace "let data = request.client.recv(readSize)" with "let data = waitFor request.client.recv(readSize)" to wait for the future. I don't know if I can or is correct use the "waitFor" here, but it works. I tested it with a 900MB file and work like a charm.

References:
https://github.com/nim-lang/Nim/issues/6576

Asynchronous Iteration Support
Stream reader supports asynchronous iteration over BODY.
https://docs.aiohttp.org/en/stable/streams.html

Code to test:
```nim
import asynchttpserver, asyncdispatch
import strutils, strformat

const stream = true # for test purposes switch from true to false

proc htmlpage(contentLength, bodyLength: int): string =
    return &"""
<!Doctype html>
<html lang="en">
<head><meta charset="utf-8"/></head>
<body>
<form action="/" method="post" enctype="multipart/form-data">
File: <input type="file" name="testfile" accept="text/*"><br />
<input style="margin:10px 0;" type="submit">
</form><br />
Expected Body Length: {contentLength} bytes<br />
Actual Body Length: {bodyLength} bytes
</body>
</html>
"""

  var
    contentLength = 0
    bodyLength = 0
  if req.reqMethod == HttpPost:
    contentLength = req.headers["Content-length"].parseInt
    if stream:
      # Read 8*1024 bytes at a time
      # optional chunkSize parameter. The default is 8*1024
      for data in req.bodyStream(8*1024):
        bodyLength += data.len
      if bodyLength != contentLength:
        await req.respond(Http400,
          "Bad Request. Data read has a different length than the expected.")
        return
    else:
      bodyLength += req.body.len

  await req.respond(Http200, htmlpage(contentLength, bodyLength))

let server = newAsyncHttpServer(maxBody = 1048576000, stream = stream)
waitFor server.serve(Port(8080), cb)
```

### DIFF
--- a/lib/pure/asynchttpserver.nim
+++ b/lib/pure/asynchttpserver.nim
@@ -68,15 +68,12 @@
 ##        if stream:
 ##          # Read 8*1024 bytes at a time
 ##          # optional chunkSize parameter. The default is 8*1024
-##          for length, data in req.bodyStream(8*1024):
-##            let content = await data
-##            if length == content.len:
-##              bodyLength += content.len
-##            else:
-##              # Handle exception
-##              await req.respond(Http400,
-##                "Bad Request. Data read has a different length than the expected.")
-##              return
+##          for data in req.bodyStream(8*1024):
+##            bodyLength += data.len
+##          if bodyLength != contentLength:
+##            await req.respond(Http400,
+##              "Bad Request. Data read has a different length than the expected.")
+##            return
 ##        else:
 ##          bodyLength += req.body.len
 ##      await req.respond(Http200, htmlpage(contentLength, bodyLength))
@@ -217,7 +214,7 @@ proc sendStatus(client: AsyncSocket, status: string): Future[void] =
 when (NimMajor, NimMinor) >= (1, 1):
   iterator bodyStream*(
     request: Request,
-    chunkSize: int = 8*1024): (int, Future[string]) =
+    chunkSize: int = 8*1024): string =
     ## The chunkSize parameter is optional and default value is 8*1024 bytes.
     ##
     ## This iterator return a tuple with the length of the data that was read
@@ -225,10 +222,10 @@ when (NimMajor, NimMinor) >= (1, 1):
     var remainder = request.contentLength
     while remainder > 0:
       let readSize = min(remainder, chunkSize)
-      let data = request.client.recv(readSize)
-      if data.failed:
+      let data = waitFor request.client.recv(readSize)
+      if data.len != readSize:
         raise newException(ValueError, "Error reading POST data from client.")
-      yield (readSize, data)
+      yield data
       remainder -= readSize
 
 proc processRequest(
--- a/lib/pure/asynchttpserver.nim
+++ b/lib/pure/asynchttpserver.nim
@@ -68,8 +68,10 @@
 ##        if stream:
 ##          # Read 8*1024 bytes at a time
 ##          # optional chunkSize parameter. The default is 8*1024
-##          for data in req.bodyStream(8*1024):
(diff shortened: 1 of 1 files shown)


dom96 (changes requested):
> I replace "let data = request.client.recv(readSize)" with "let data = waitFor request.client.recv(readSize)" to wait for the future.

This is most likely incorrect, but I don't see it among your changes. You shouldn't use `waitFor` in async procs, use `await`.

As for your changes here, you are just reimplementing what `FutureStream` already does by allowing a callback to be passed to the `bodyStream` proc. I think that at this point we should revert what is there and start fresh, it's unclear what this change achieves with regards to how the rest of this is implemented.

mrhdias:
> > I replace "let data = request.client.recv(readSize)" with "let data = waitFor request.client.recv(readSize)" to wait for the future.
> 
> This is most likely incorrect, but I don't see it among your changes. You shouldn't use `waitFor` in async procs, use `await`.
> 
> As for your changes here, you are just reimplementing what `FutureStream` already does by allowing a callback to be passed to the `bodyStream` proc. I think that at this point we should revert what is there and start fresh, it's unclear what this change achieves with regards to how the rest of this is implemented.

I just made minor changes to the library. They can be seen by the version.
when (NimMajor, NimMinor) >= (1, 1)

https://github.com/nim-lang/Nim/blob/c1ef26c5788532cb8af52aea231a3e13eea90f94/lib/pure/asynchttpserver.nim

These changes have no impact on the code that already exists (only activated when flag stream = true). You can test the example below and modify the code or tell me what I have to do to make this work as you expect.

Code to test:
```nim
import asynchttpserver, asyncdispatch
import strutils, strformat

const stream = true # for test purposes switch from true to false

proc htmlpage(contentLength, bodyLength: int): string =
  return &"""
<!Doctype html>
<html lang="en">
  <head><meta charset="utf-8"/></head>
  <body>
    <form action="/" method="post" enctype="multipart/form-data">
      File: <input type="file" name="testfile" accept="text/*"><br />
      <input style="margin:10px 0;" type="submit">
    </form><br />
    Expected Body Length: {contentLength} bytes<br />
    Actual Body Length: {bodyLength} bytes
  </body>
</html>
"""
  var
    contentLength = 0
    bodyLength = 0
  if req.reqMethod == HttpPost:
    contentLength = req.headers["Content-length"].parseInt
    if stream:
      # Read 8*1024 bytes at a time
      # optional chunkSize parameter. The default is 8*1024
      await req.bodyStream(chunkSize = 8*1024,
        callback = proc(data: string) {.async.} =
          bodyLength += data.len
      )
      if bodyLength != contentLength:
        await req.respond(Http400,
          "Bad Request. Data read has a different length than the expected.")
        return
    else:
      bodyLength += req.body.len
  await req.respond(Http200, htmlpage(contentLength, bodyLength))

let server = newAsyncHttpServer(maxBody = 1073741824, stream = stream) # 1GB
waitFor server.serve(Port(8080), cb)
```

dom96:
This is more so about API design than it is about the implementation. Once you let an API into the stdlib you cannot change it because that will break compatibility.

You have implemented a proc that takes a callback, and calls that callback every time there is new data received. This is exactly what `FutureStream` abstracts, you write data into it and whatever callback is set on it is called. It's by no means perfect, and there are issues for it that we planned to fix, but introducing an ad-hoc way to achieve the same thing is not the solution.

I think it would be better to start again because then I can properly review your PR. It's hard to review when you've made other changes which I cannot see.

PR #13483, opened 2020-02-24
Add more JS stuff to dom.nim

treeform:
* LocalStorage
* activeElement
* innerText
* some Style properties.

### DIFF
--- a/lib/js/dom.nim
+++ b/lib/js/dom.nim
@@ -41,7 +41,7 @@ type
     onunload*: proc (event: Event) {.nimcall.}
 
   DomEvent* {.pure.} = enum
-    ## `https://developer.mozilla.org/en-US/docs/Web/Events`_
+    ## `<https://developer.mozilla.org/en-US/docs/Web/Events>`_
     Abort = "abort",
     BeforeInput = "beforeinput",
     Blur = "blur",
@@ -217,7 +217,7 @@ type
     offsetLeft*: int
     offsetTop*: int
 
-  ValidityState* = ref ValidityStateObj ## `https://developer.mozilla.org/en-US/docs/Web/API/ValidityState`_
+  ValidityState* = ref ValidityStateObj ## `<https://developer.mozilla.org/en-US/docs/Web/API/ValidityState>`_
   ValidityStateObj {.importc.} = object
     badInput*: bool
     customError*: bool
@@ -231,23 +231,24 @@ type
     valid*: bool
     valueMissing*: bool
 
-  Blob* = ref BlobObj ## `https://developer.mozilla.org/en-US/docs/Web/API/Blob`_
+  Blob* = ref BlobObj ## `<https://developer.mozilla.org/en-US/docs/Web/API/Blob>`_
   BlobObj {.importc.} = object of RootObj
     size*: int
     `type`*: cstring
 
-  File* = ref FileObj ## `https://developer.mozilla.org/en-US/docs/Web/API/File`_
+  File* = ref FileObj ## `<https://developer.mozilla.org/en-US/docs/Web/API/File>`_
   FileObj {.importc.} = object of Blob
     lastModified*: int
     name*: cstring
 
-  TextAreaElement* = ref object of ElementObj ## `https://developer.mozilla.org/en-US/docs/Web/API/HTMLTextAreaElement`_
+  TextAreaElement* = ref TextAreaElementObj ## `<https://developer.mozilla.org/en-US/docs/Web/API/HTMLTextAreaElement>`_
+  TextAreaElementObj {.importc.} = object of Element
     value*: cstring
     selectionStart*, selectionEnd*: int
     selectionDirection*: cstring
     rows*, cols*: int
 
-  InputElement* = ref InputElementObj ## `https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement`_
+  InputElement* = ref InputElementObj ## `<https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement>`_
   InputElementObj {.importc.} = object of Element
     # Properties related to the parent form
     formAction*: cstring
@@ -329,7 +330,7 @@ type
     text*: cstring
     value*: cstring
 
-  FormElement* = ref FormObj ## `https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement`_
+  FormElement* = ref FormObj ## `<https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement>`_
   FormObj {.importc.} = object of ElementObj
     acceptCharset*: cstring
     action*: cstring
@@ -462,7 +463,7 @@ type
     AtTarget,
     BubblingPhase
 
-  Event* = ref EventObj ## `https://developer.mozilla.org/en-US/docs/Web/API/Event`_
+  Event* = ref EventObj ## `<https://developer.mozilla.org/en-US/docs/Web/API/Event>`_
   EventObj {.importc.} = object of RootObj
     bubbles*: bool
     cancelBubble*: bool
@@ -475,12 +476,12 @@ type
     `type`*: cstring
     isTrusted*: bool
 
-  UIEvent* = ref UIEventObj ## `https://developer.mozilla.org/en-US/docs/Web/API/UIEvent`_
+  UIEvent* = ref UIEventObj ## `<https://developer.mozilla.org/en-US/docs/Web/API/UIEvent>`_
   UIEventObj {.importc.} = object of Event
     detail*: int64
     view*: Window
 
-  KeyboardEvent* = ref KeyboardEventObj ## `https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent`_
+  KeyboardEvent* = ref KeyboardEventObj ## `<https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent>`_
   KeyboardEventObj {.importc.} = object of UIEvent
     altKey*, ctrlKey*, metaKey*, shiftKey*: bool
     code*: cstring
@@ -489,7 +490,7 @@ type
     keyCode*: int
     location*: int
 
-  KeyboardEventKey* {.pure.} = enum ## `https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values`_
+  KeyboardEventKey* {.pure.} = enum ## `<https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values>`_
     # Modifier keys
     Alt,
     AltGraph,
@@ -843,7 +844,7 @@ type
     FourthButton = 8,
     FifthButton = 16
 
-  MouseEvent* = ref MouseEventObj ## `https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent`_
+  MouseEvent* = ref MouseEventObj ## `<https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent>`_
   MouseEventObj {.importc.} = object of UIEvent
     altKey*, ctrlKey*, metaKey*, shiftKey*: bool
     button*: int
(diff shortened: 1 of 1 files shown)


lib/js/dom.nim
```diff
@@ -238,6 +246,13 @@ type
     lastModified*: int
     name*: cstring
 
+  # https://developer.mozilla.org/en-US/docs/Web/API/HTMLTextAreaElement
```
timotheecour: nit: might as well put this link as a doc comment instead of comment? ditto elsewhere; that way users can get more info on each field by just clicking this link

treeform: I don't know why it had these comments there, it seems obvious where to go get the docs. I am just trying to be consistent. 

I would like to remove all comments of that type?

timotheecour: having them in docs removes friction

treeform:
Let's hope BSD test will not fail for some intermittent reason again related to io selectors.

timotheecour:
> Let's hope BSD test will not fail for some intermittent reason again related to io selectors.

you can simply rebase, I fixed the CI build.sh failure in https://github.com/nim-lang/Nim/pull/13564/files

> pkg: wrong architecture: FreeBSD:12.0:amd64 instead of FreeBSD:12:amd64

lib/js/dom.nim
treeform: I think it just adds extra noise to parse the definitions - packed with easy to find URLs.

timotheecour: if you're gonna delete all those links at the very least add 1 link in the existing module level doc comment on top:

```nim
## Declaration of the Document Object Model for the `JavaScript backend
## <backends.html#backends-the-javascript-target>`_.
## See  also https://developer.mozilla.org/en-US/docs/Web/API
```
(added 3rd line)

treeform:
Ok rebased and all tests pass.

lib/js/dom.nim
```diff
@@ -165,10 +169,13 @@ type
     parentNode*: Node
     previousSibling*: Node
     innerHTML*: cstring
+    innerText*: cstring
+    textContent *: cstring
```
dom96: nit, no space before `*`

lib/js/dom.nim
```diff
@@ -226,19 +232,22 @@ type
     valid*: bool
     valueMissing*: bool
 
-  # https://developer.mozilla.org/en-US/docs/Web/API/Blob
   Blob* = ref BlobObj
   BlobObj {.importc.} = object of RootObj
     size*: int
     `type`*: cstring
 
-  # https://developer.mozilla.org/en-US/docs/Web/API/File
```
dom96, line 235: Why remove these?

lib/js/dom.nim
```diff
@@ -1136,7 +1144,7 @@ proc createAttribute*(d: Document, identifier: cstring): Node
 proc getElementsByName*(d: Document, name: cstring): seq[Element]
 proc getElementsByTagName*(d: Document, name: cstring): seq[Element]
 proc getElementsByClassName*(d: Document, name: cstring): seq[Element]
-proc getSelection*(d: Document): cstring
+proc getSelection*(d: Document): Selection
```
dom96: This is a breaking change, add a `converter` which implicitly converts `Selection` to `cstring`.

dom96 (changes requested):

lib/js/dom.nim
timotheecour: converters are a little bit evil but that aside:
http://help.dottoro.com/ljqxhfte.php

> The document.getSelection method works differently in Google Chrome, Safari and Internet Explorer than in Firefox and Opera. It returns a string in Firefox and Opera, and returns a selectionRange object in Google Chrome, Safari and Internet Explorer (the document.getSelection method is identical to the window.getSelection method in Google Chrome, Safari and Internet Explorer). Because of that, this method should not be used.

lib/js/dom.nim
timotheecour: see https://github.com/nim-lang/Nim/pull/13483#discussion_r383606456 seems like most reviewers agree these comments were valuable; https://github.com/nim-lang/Nim/pull/13483#discussion_r388564243 otherwise would be a good compromise

dom96: yeah, it's useful to have a reference to what the definition is based on.

lib/js/dom.nim
Araq: Should be used or not, it's a breaking change, please introduce ``proc getSelectionEnum`` or similar.

treeform: I have reverted getSelection to the previous cstring behavior. No changes to the API.

lib/js/dom.nim
treeform: I have made all comments pointing to documentation be ##doc comments as requested.

lib/js/dom.nim
treeform: Fixed, thanks!

lib/js/dom.nim
treeform: I have made all comments pointing to documentation be ##doc comments are requested.

treeform:
I think have made the requested changes. Please take a look again when you can.

lib/js/dom.nim
timotheecour: how about turning code into:
```nim
proc getSelection*(d: Document): cstring {.importc: "getSelection", deprecated: "incompatible across browsers, use instead XYZ".}
proc getSelectionAsObject*(d: Document): Selection {.importc: "getSelection", deprecated: "incompatible across browsers, use instead XYZ".}
```

lib/js/dom.nim
```diff
@@ -832,7 +853,7 @@ type
     FourthButton = 8,
     FifthButton = 16
 
-  # https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent
+  ## https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent
```
timotheecour: doc comment at wrong place, generates the links for the wrong node in rendered html (eg when running `nim doc --outdir:/tmp/d55 lib/js/dom.nim`)

image

* these doc comments need to be moved so the render properly
* these should display as clickable links eg like this:
```
Nim's standard library only covers the basics, check
out `<https://nimble.directory/>`_ for a list of 3rd party packages.
```

treeform: Sorry I don't know what to do in this case:

```nim
  ## `https://developer.mozilla.org/en-US/docs/Web/API/Blob`_
  Blob* = ref BlobObj
  BlobObj {.importc.} = object of RootObj
    size*: int
    `type`*: cstring
```

Most of the cases involve double objects.

Do I put them on both? 

```nim
  Blob* = ref BlobObj ## `https://developer.mozilla.org/en-US/docs/Web/API/Blob`_  
  BlobObj {.importc.} = object of RootObj
    ## `https://developer.mozilla.org/en-US/docs/Web/API/Blob`_
    size*: int
    `type`*: cstring
```

Or on one? Do you have a preference on which one?

```nim
  Blob* = ref BlobObj 
  BlobObj {.importc.} = object of RootObj
    ## `https://developer.mozilla.org/en-US/docs/Web/API/Blob`_
    size*: int
    `type`*: cstring
```

Thank you for the quick response, I just want to get the commit to the state you are happy with.

timotheecour: the doc comment should be attached to the thing that's exported:

```nim
  Blob* = ref BlobObj ## https://developer.mozilla.org/en-US/docs/Web/API/Blob
  BlobObj {.importc.} = object of RootObj
    size*: int
    `type`*: cstring
```
(when in doubt just run `nim doc` to see the output)

more generally, even if both were exported, I'd prefer to attach it to the one that's most likely to be used (the ref type) and not repeat the doc comment; docs should be DRY just like code (otherwise you always wonder what subtle difference in doc comment you missed)

treeform: I made everything match now:
image

lib/js/dom.nim
treeform: Yeah, I want to do some thing like that in another commit.

lib/js/dom.nim
```diff
@@ -226,20 +231,23 @@ type
...
-  # https://developer.mozilla.org/en-US/docs/Web/API/File
-  File* = ref FileObj
+  File* = ref FileObj ## `https://developer.mozilla.org/en-US/docs/Web/API/File`_
   FileObj {.importc.} = object of Blob
     lastModified*: int
     name*: cstring
 
-  # https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement
-  InputElement* = ref InputElementObj
+  TextAreaElement* = ref object of ElementObj ## `https://developer.mozilla.org/en-US/docs/Web/API/HTMLTextAreaElement`_
```
timotheecour: question for other reviewers:
I notice other types here use this pattern:
```nim
type
 FormElement* = ref FormObj
  FormObj {.importc.} = object of ElementObj
```
however here it's without the importc:
```nim
TextAreaElement* = ref object of ElementObj
```
and, unless I'm missing something, it seems to work too;

what's the point of `{.importc.}` **for types** with `nim js` , and what difference does it make when we omit importc?

lib/js/dom.nim
```diff
@@ -40,8 +40,8 @@ type
     onsubmit*: proc (event: Event) {.nimcall.}
     onunload*: proc (event: Event) {.nimcall.}
 
-  # https://developer.mozilla.org/en-US/docs/Web/Events
   DomEvent* {.pure.} = enum
+    ## `https://developer.mozilla.org/en-US/docs/Web/Events`_
```
timotheecour: You have to run nim doc (and inspect html) to check whether output looks good after your modifications (including clicking on links).

The link format you're using is wrong and gives broken links;  I just checked these work with nim doc, and generate clickable links:
```nim
    ## see [docs](https://developer.mozilla.org/en-US/docs/Web/Events)       (markdown format)
    ## see `docs<https://developer.mozilla.org/en-US/docs/Web/Events>`_    (RST format)
    ## see https://developer.mozilla.org/en-US/docs/Web/Events                  (lazy format, text shown = link)
```
however, I just realized right now that all these only render clickable links in certain contexts, specifically at top-level of a declaration (eg type);
```nim
type
  Gook* = object ## ok at declaration top-level, parsed as link: https://developer.mozilla.org/en-US/docs/Web/Events
    jsHeapSizeLimit*: float ## not ok, not parsed as link: https://developer.mozilla.org/en-US/docs/Web/Events
```
(hence my original comment https://github.com/nim-lang/Nim/pull/13483#discussion_r390724719 where i recommeded using the RST format, even though this was a red herring, the culprit being this nim doc bug)

## TLDR
simply use the lazy format, but at at top-level of type declaration like in above example for `Gook`, and check that nim doc renders links ok

lib/js/dom.nim
Araq: Reason: Legacy code.

timotheecour: you mean, it was there for historical reasons but isn't needed anymore and can be omitted (safe even for boostrapping from latest nim form csources) ?
or you mean something else?

lib/js/dom.nim
treeform: I made the links be in the format of "see <docs>" as requested.
image
I made sure the links click. Sorry very rusty on my RST formatting rules.

lib/js/dom.nim
treeform: I don't know why we are doing the Obj thing. But I made it consistent:
```nim
  TextAreaElement* = ref TextAreaElementObj 
  TextAreaElementObj {.importc.} = object of Element
```

timotheecour (approved):

timotheecour:
* great, LGTM. subsequent PRs could add doc links for other procs (eg https://developer.mozilla.org/en-US/docs/Web/API/Performance/now)
* @dom96 I see you often use the "change requested" github feature (in other PRs) but IMO that's overkill, as it often stays forever in that state even long after the contributor has addressed the issue (and contributor has no way to undo that state, only you can). IMO the following is 100% sufficient assuming no abuse:
  * reviewer writes a comment
  * contributor replies, and, using best judgement, marks as resolved if comment is fully addressed
  * unless reviewer disagrees and un-resolves that comment, that's the end of the story

dom96:
@timotheecour huh? Where do you see that state reported? I am asking for changes, so I select "Changes Requested", I don't see how that negatively affects the PRs.

dom96 (approved):
LGTM

lib/js/dom.nim
Araq: It was written before ``ref object`` was added to Nim, so yes, now it can be omitted.

PR #13507, opened 2020-02-26
Fix docgen snippet numbering

genotrance:
- Fix snippet file name - everything was 101 - also print snippet if error
- Reduce noise in snippet and example compilation
- Reduce specific warnings, hints in koch docs

### DIFF
--- a/compiler/docgen.nim
+++ b/compiler/docgen.nim
@@ -191,10 +191,10 @@ proc newDocumentor*(filename: AbsoluteFile; cache: IdentCache; conf: ConfigRef, 
   initStrTable result.types
   result.onTestSnippet =
     proc (gen: var RstGenerator; filename, cmd: string; status: int; content: string) =
+      inc(gen.id)
       var d = TDocumentor(gen)
       var outp: AbsoluteFile
       if filename.len == 0:
-        inc(d.id)
         let nameOnly = splitFile(d.filename).name
         outp = getNimcacheDir(conf) / RelativeDir(nameOnly) /
                RelativeFile(nameOnly & "_snippet_" & $d.id & ".nim")
--- a/lib/system/inclrtl.nim
+++ b/lib/system/inclrtl.nim
@@ -49,7 +49,7 @@ when defined(nimlocks):
 else:
   {.pragma: benign, gcsafe.}
 
-template since(version, body: untyped) {.dirty.} =
+template since(version, body: untyped) {.dirty, used.} =
   ## limitation: can't be used to annotate a template (eg typetraits.get), would
   ## error: cannot attach a custom pragma.
   when (NimMajor, NimMinor) >= version:


compiler/docgen.nim
```diff
@@ -159,6 +159,7 @@ proc getOutFile2(conf: ConfigRef; filename: RelativeFile,
   else:
     result = getOutFile(conf, filename, ext)
 
+var gIdx = 101
```
Araq: This introduces global state and I don't understand why it's needed.

compiler/docgen.nim
```diff
@@ -211,9 +212,11 @@ proc newDocumentor*(filename: AbsoluteFile; cache: IdentCache; conf: ConfigRef,
       writeFile(outp, importStmt & content)
       let c = if cmd.startsWith("nim "): os.getAppFilename() & cmd.substr(3)
               else: cmd
-      let c2 = c % quoteShell(outp)
+      let subst = if "-r" in c: "-r" else: "$1"
+      let c2 = c.replace(subst, "--verbosity:0 --hints:off " & subst) % quoteShell(outp)
       rawMessage(conf, hintExecuting, c2)
       if execShellCmd(c2) != status:
+        echo importStmt & content
```
Araq: No, the compiler doesn't use ``echo`` for error reporting and this PR is about making the compiler less noisy, not moreso.

tools/kochdocs.nim
```diff
@@ -5,7 +5,7 @@ import os, strutils, osproc, sets
 const
   gaCode* = " --doc.googleAnalytics:UA-48159761-1"
   # --warning[LockLevel]:off pending #13218
-  nimArgs = "--warning[LockLevel]:off --hint[Conf]:off --hint[Path]:off --hint[Processing]:off -d:boot --putenv:nimversion=$#" % system.NimVersion
+  nimArgs = "--warning[LockLevel]:off --warning[Deprecated]:off --warning[UnusedImport]:off --warning[RedefinitionOfLabel]:off --hint[Conf]:off --hint[Path]:off --hint[Processing]:off --hint[XDeclaredButNotUsed]:off -d:boot --putenv:nimversion=$#" % system.NimVersion
```
Araq: This line seems to be the only necessary line to accomplish what your PR title says.

compiler/docgen.nim
genotrance: Fixed

compiler/docgen.nim
genotrance: Idea was to print the snippet code that failed, makes it easier to debug the problematic snippet instead of having to search the snippet numerically in source file.

Removed for now.

compiler/docgen.nim
```diff
@@ -211,7 +211,8 @@ proc newDocumentor*(filename: AbsoluteFile; cache: IdentCache; conf: ConfigRef,
       writeFile(outp, importStmt & content)
       let c = if cmd.startsWith("nim "): os.getAppFilename() & cmd.substr(3)
               else: cmd
-      let c2 = c % quoteShell(outp)
+      let subst = if "-r" in c: "-r" else: "$1"
+      let c2 = c.replace(subst, "--verbosity:0 --hints:off " & subst) % quoteShell(outp)
```
Araq: Why not ``let c2 = (c & "--verbosity:0 --hints:off") % quoteShell(outp)"``?

genotrance: Can't put flags after the filename.

Araq: Ok, but ``let subst = if "-r" in c: "-r" else: "$1"`` is suspicious. Why does ``-r`` lack the ``$1`` part?

genotrance: Not all tests are run so -r isn't common. But when it is, these flags have to go before it, hence looking for both.

genotrance:
Only freebsd is failing in unrelated test.

compiler/docgen.nim
timotheecour: -r seems fragile, eg would trigger for `--rangeChecks:on`
a more robust solution would identify the position right before the filename in `nim cmd -flag1 -flag2 filename arg`, ideally by reusing same code compiler uses instead of duplicating logic

genotrance: Open to ideas but like I said, these params have to go before -r. How about `" -r "`?

timotheecour: > these params have to go before -r

anything between nim binary and filename can go anywhere:
```
nim c --rangeChecks:on -r main
nim c  -r --rangeChecks:on main
nim --rangeChecks:on -r c main
```
> How about " -r "

not robust to things like:
```
nim c -r -d:mystr:"foo -r bar" main
```
(there are more such cases, eg --passC, --passL etc)

sure, these may be rare/edge cases, but I think it's just better to reuse (not re-implement) the logic nim compiler itself uses

timotheecour: ~~how about instead adding `--verbosity:0 --hints:off` to here:
`tools/kochdocs.nim:235:27:    commands[i] = nim & " rst2html $# --git.url:$# -o:$# --index:on $#" %`~~
=> see comment below for a better idea

timotheecour:
I like the spirit of this PR but not the current implementation.
* Some of the warnings are legit and we should fix the code instead of masking the warnings, eg:
```
/Users/timothee/git_clone/nim/Nim_prs/lib/system/channels.nim(154, 22) Warning: pragma before generic parameter list is deprecated [Deprecated]
    Channel* {.gcsafe.}[TMsg] = RawChannel ## a channel for thread communication
/Users/timothee/git_clone/nim/Nim_prs/lib/system/channels.nim(269, 37) Warning: Number of spaces around '*%' is not consistent [Spacing]
                  cast[pointer](s +% i*% mt.base.size), mt.base, t, mode)
```

* we need to identify which warnings don't make sense for a compiler target, eg: for `rst2html` `RedefinitionOfLabel` does not make sense so we should fix it at the compiler level:

compiler/main.nim:
```
  of "rst2html":
    conf.notes.excl warnRedefinitionOfLabel # similar to issue #13218
```

## here's all the warnings produced:
https://gist.github.com/timotheecour/397948fa2017e147e62e1b4044561060

we can do analysis on this eg:
```
cat $logsmisc_D/kochdocs.log | grep Warning: | grep -v RedefinitionOfLabel| sort | uniq
cat $logsmisc_D/kochdocs.log | grep Warning: | grep -v RedefinitionOfLabel| sort | uniq | wc -l
```


=> see https://github.com/nim-lang/Nim/pull/13550 which addresses at the compiler level some of those irrelevant warnings; IMO it's a better approach, eg it doesn't suppress `UnusedImport` from $nim_prs_D/doc/docgen_sample.nim unlike this PR (which does `--warning[UnusedImport]:off`), but it does suppress them from things like nimrtl (because of `--app:lib`); ditto with other flags

genotrance:
I'm fine if this can be fixed in compiler as well as by fixing the actual warnings, there's too many though. Fact is that nim doc warnings are not really attended to, nim doc isn't run in the main CI so there's too many warnings flooding the logs. Nightlies are already a long process.

I can trim this PR to just fixing the snippet numbering.

timotheecour:
> I'm fine if this can be fixed in compiler as well as by fixing the actual warnings, there's too many though.

https://github.com/nim-lang/Nim/pull/13550 will fix a lot of these; subsequent PRs to fix remaining ones are welcome

> Fact is that nim doc warnings are not really attended to, nim doc isn't run in the main CI so there's too many warnings flooding the logs. Nightlies are already a long process.

out of sight, out of mind; they're much more likely to be attended to if we surface them. Indeed, many tests are only run via `nim doc` (eg runnableExamples), and simply gagging warnings (including relevant ones) will make it really hard to fix those.

> I can trim this PR to just fixing the snippet numbering.

you mean `inc(gen.id)` ? ya, after reading your whole PR, I think that's the best approach; we can followup after https://github.com/nim-lang/Nim/pull/13550 with more warning slayings as needed.

`template since(version, body: untyped) {.dirty, used.} =`, 

maybe that one might be ok, I don't care either way for that specific one

genotrance:
I've cleaned up this PR to just fix the docgen snippet numbering.

PR #13533, opened 2020-02-28
Add posix_utils.isSsd

juancarlospaco:
- Add `posix_utils.isSsd()` proc that returns `true` if disk is SSD (Solid).
- Can work at compile-time, so allows to optimize for slow or fast I/O.
- Documentation with example.
- proc is single line, easy merge.

### DIFF
--- a/lib/posix/posix_utils.nim
+++ b/lib/posix/posix_utils.nim
@@ -106,4 +106,4 @@ when defined(linux):
     ## .. code-block:: nim
     ##   echo isSsd('a') ## 'a' for /dev/sda, 'b' for /dev/sdb, ...
     ##
-    try: readFile("/sys/block/sd" & diskLetter & "/queue/rotational") == "0\n" except: false
+    try: readFile("/sys/block/sd" & $diskLetter & "/queue/rotational") == "0\n" except: false
--- a/lib/posix/posix_utils.nim
+++ b/lib/posix/posix_utils.nim
@@ -99,3 +99,11 @@ proc mkdtemp*(prefix: string): string =
     raise newException(OSError, $strerror(errno))
   return $tmpl
 
+when defined(linux):
+  proc isSsd*(diskLetter: char): bool {.inline.} =
+    ## Returns ``true`` if disk is SSD (Solid). Linux only.
+    ##
+    ## .. code-block:: nim
+    ##   echo isSsd('a') ## 'a' for /dev/sda, 'b' for /dev/sdb, ...
+    ##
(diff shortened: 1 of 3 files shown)


lib/posix/posix_utils.nim
```diff
@@ -99,3 +99,11 @@ proc mkdtemp*(prefix: string): string =
...
   return $tmpl
 
+when defined(linux):
+  proc isSsd*(diskLetter: char): bool {.inline.} =
+    ## Returns ``true`` if disk is SSD (Solid). Linux only.
+    ##
+    ## .. code-block:: nim
+    ##   echo isSsd('a') ## 'a' for /dev/sda, 'b' for /dev/sdb, ...
+    ##
+    try: readFile("/sys/block/sd" & $diskLetter & "/queue/rotational") == "0\n" except: false
```
alaviss, line 109: `sda`, `sdb` aren't the only things that can be SSD. There are also NVMe drives too.

Also, I don't know how useful would this proc actually be in practice to warrant an addition to the stdlib, given that you have to extract the device letter from whatever an user might give (ie. `/dev/sda`, etc.), and I'm not sure what kind of optimization would target only SSD.

tests/stdlib/tposix.nim
```diff
@@ -17,3 +17,8 @@ when not defined(windows):
   writeLine(stdout, u.nodename)
   writeLine(stdout, u.release)
   writeLine(stdout, u.machine)
+
+  when defined(linux):
+    block:
+      # lib/posix/posix_utils.nim
+      doAssert posix_utils.isSsd(diskLetter = 'a') is bool
```
alaviss: Please don't do this kind of test.

lib/posix/posix_utils.nim
juancarlospaco: But proc is about disk mechanical or solid.

I/O intensive stuff mostly, but the difference may worth it.

If you find benchmarks of mechanical disk outperforming solid disk link it, I want to read it.
😃

alaviss: I am not convinced that this should be in the stdlib. It appears to be too specialized for a vague use case of "I/O optimization", and it's Linux-only to boot.

I suspect that anyone who want this might already have their own function that do the same, given how trivial it is. IMO this would belong to a 3rd-party package that provides detailed disk information or something, not the stdlib.

juancarlospaco: Yes is Linux-only, but according to the last official Survey, most people are using Linux anyway. 
:man_shrugging:

Araq:
So add it to ``linux.nim`` or something.

PR #13542, opened 2020-02-29
Add deques.peekFirst/Last(var Deque[T]) -> var T

metagn:


Araq:
Please add a ``.since`` annotation and it's ready to go.

metagn:
Done

metagn:
Actually done this time

Clyybber:
The pragma way should work too. The issue here is something else..

metagn:
updated for 1.3, pragma way works now

PR #13544, opened 2020-02-29
sink parameter inference for types that have destructors

Araq:


Araq:
@cooldome Please tell me if this PR causes trouble for you, it's about to get merged.

PR #13570, opened 2020-03-03
Document as infix operator

### LINKED ISSUE #13409: `as` is usable as infix operator but its existence and precedence are not documented
https://nim-lang.org/docs/manual.html#precedence https://nim-lang.org/docs/manual.html#lexical-analysis-operators

### Example
```nim
template `as`[T, U](x: T, _: typedesc[U]): U = U(x)

echo(65 as char)
```

### Current Output
```
A
```

### Possible Solution

* List it as an operator in the manual and add it to the operators with precedence 5

```
$ nim -v
Nim Compiler Version 1.0.6
```

metagn:
Fixes #13409

### DIFF
--- a/doc/manual.rst
+++ b/doc/manual.rst
@@ -559,7 +559,7 @@ following characters::
 defined here.)
 
 These keywords are also operators:
-``and or not xor shl shr div mod in notin is isnot of``.
+``and or not xor shl shr div mod in notin is isnot of as``.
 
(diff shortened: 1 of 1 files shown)


Araq:
No [ci skip] here! Docs are checked too!

Clyybber:
Luckily the [ci skip] was ignored here, presumably as its the first commit.

Issue #13574, opened 2020-03-03
asyncHttpServer does not parse repeated headers


andreaferretti:
Related to https://github.com/nim-lang/Nim/issues/13573. `httpcore` was recently refactored to add support for repeated headers - in particular we have types `HttpHeaders` and `HttpHeaderValues`, where before headers were just a `stringTable`.

But the server does not make use of this, as seen [here](https://github.com/nim-lang/Nim/blob/devel/lib/pure/asynchttpserver.nim#L325):

```nim
let (key, value) = parseHeader(lineFut.mget)
request.headers[key] = value
```

The correct thing to do would be

```nim
let (key, value) = parseHeader(lineFut.mget)
request.headers.add(key, value)
```
which allows to accumulate multiple values for the headers.

Found this while investigating https://github.com/andreaferretti/rosencrantz/issues/28

andreaferretti:
[Turns out](https://github.com/nim-lang/Nim/pull/13575#issuecomment-595377444), this was a misunderstanding on my part

Issue #12211, opened 2019-09-17
proposed changes to HttpHeaders as in HttpClient


disruptek:
It seems silly to write a long list of grievances with such a minor implementation, but the fact is that they add up to time wasted debugging and work-arounds in real code.  :frowning_face: 

Here's [the section of the RFC we're trying to represent](https://tools.ietf.org/html/rfc7230#section-3.2).

Here's a list of changes I'd like to make which stem from actual misbehavior of client requests:

1. One shalt be the number of unique header fields thou shalt send in any request/response, and the number of the sending shall be one.  Two shalt thou not send, nor either send thou zero, except in the event that the user has specified a header, in which case thou then proceed to send one.  Five, as you have by now guessed, is right out.  :wink:

1. The `Set-Cookie` header is an exception to the above rule, though the client shouldn't have cause to send it.

1. The `Cookie` header is **not** an exception to the above rule, and multiple values for this header, should they exist, should be separated by `; `.

1. The string representation of multiple header values should otherwise be equal to the concatenation of the string representation of each such value, separated with a `,`.

1. The string representation of a header name or value should be equal to that value with leading and trailing whitespace stripped.

1. A header with an empty value is not equivalent to a header with no value.

1. Two headers are equivalent if _and only if_ they specify the same header field names and the same values in the same order.

1. User-specified header values should replace _and not supplant_ header values otherwise computed automatically by the `httpclient` implementation.

Here's a list of a changes I'd like to make which stem from perceived faults of the stdlib API:

1. The `toString` converter should be removed, or at least yield all values of a given header separated by `,` as expected by the RFC.  It currently yields only the first value.

1. The `contains` implemented for header _values_ should be case-sensitive except _perhaps_ in the case of fields for which the protocol specifies case-insensitivity.  To put this another way, the protocol specifies case-insensitivity for some values, and the protocol does not specify case-insensitivity for all values.  `contains` is currently case-insensitive for all header values.

1. A `$` implemented for `HttpHeaders` should yield a reasonable _if not ideal_ approximation of the header section in an HTTP request/response.  It currently yields the `$` of the underlying table.

1. `httpcore` should use a `distinct` type for `HttpHeaderValues` either everywhere, or nowhere.  If it uses a `distinct` type, then it should implement most/all semantics for the base type and accept base types in addition to its native `distinct` type for all operations, and it should yield the `distinct` type by default.

[...]

PR #13575, opened 2020-03-03
Fix #13573 and #13574

### LINKED ISSUE #13573: newHttpHeaders does not accept repeated headers
In `httpcore`, the constructor `newHttpHeaders` does not accept repeated headers. `httpcore` was recently refactored to add support for repeated headers - in particular we have types `HttpHeaders` and `HttpHeaderValues`, where before headers were just a `stringTable`.

Still, the constructor `newHttpHeaders` assumes that each header is passed a single time, as seen [here](https://github.com/nim-lang/Nim/blob/devel/lib/pure/httpcore.nim#L106-L112):

```nim
proc newHttpHeaders*(keyValuePairs:
    openArray[tuple[key: string, val: string]]): HttpHeaders =
  var pairs: seq[tuple[key: string, val: seq[string]]] = @[]
  for pair in keyValuePairs:
    pairs.add((pair.key.toLowerAscii(), @[pair.val]))
  new result
  result.table = newTable[string, seq[string]](pairs)
```

To check this behaviour it is enough to do

```nim
import httpcore, strutils

let h = newHttpHeaders({"a": "1", "a": "2"})

assert seq[string](h["a"]).join(",") == "1,2"
```

andreaferretti:


andreaferretti:
I added a small test for httpcore - I couldn't find testament tests for asynchttpclient

mrhdias:
I don't know if it's correct. I think that according to the specifications the following headers with the same field-name that already exist, the values must be concatenated and separated by commas.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)]. **It MUST be possible to combine the multiple header fields into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field-value to the first, each separated by a comma.** The order in which header fields with the same field-name are received is therefore significant to the interpretation of the combined field value, and thus a proxy MUST NOT change the order of these field values when a message is forwarded.

andreaferretti:
@mrhdias If I understand correctly, this means that the semantics of

```
Field1: Val1
Field1: Val2
Field1: Val3
```

is the same as

```
Field1: Val1,Val2,Val3
```

but I am not really an expert. If this was the case, the implementation should be ok. I would like to add tests for asynchttpserver, but I couldn't find any in testament.

Semi-related to this: I see some failures in CI, but I made sure to pass tests before pushing. What is the correct way to check that everything is ok with koch? I did

```
./koch boot
./koch test
```

Araq:
> What is the correct way to check that everything is ok with koch? I did

Seems fine. I use ``koch boot -d:release`` since all the checks are still in the resulting compiler.

disruptek:
No, RFC7230 _obsoletes_ 2616.  You should not refer to 2616 at all.

mrhdias:
https://tools.ietf.org/html/rfc7230#page-24

Basically the text is the same, it just creates an exception for the "Set-Cookie" header.

   header section is received, since later header fields might include
   conditionals, authentication credentials, or deliberately misleading
   duplicate header fields that would impact request processing.

   A sender MUST NOT generate multiple header fields with the same field
   name in a message unless either the entire field value for that
   header field is defined as a comma-separated list [i.e., #(values)]
   or the header field is a well-known exception (as noted below).

   A recipient MAY combine multiple header fields with the same field
   name into one "field-name: field-value" pair, without changing the
   semantics of the message, by appending each subsequent field value to
   the combined field value in order, separated by a comma.  The order
   in which header fields with the same field name are received is
   therefore significant to the interpretation of the combined field
   value; a proxy MUST NOT change the order of these field values when
   forwarding a message.

      Note: In practice, the "Set-Cookie" header field ([RFC6265]) often
      appears multiple times in a response message and does not use the
      list syntax, violating the above requirements on multiple header
      fields with the same name.  Since it cannot be combined into a
      single field-value, recipients ought to handle "Set-Cookie" as a
      special case while processing header fields.  (See Appendix A.2.3
      of [Kri2001] for details.)

disruptek:
Right.  See also #12211 for more backstory.  Point being that

```
Field1: Val1
Field1: Val2
Field1: Val3
```

is not the same as
```
Field1: Val1,Val2,Val3
```
simply because the second is permitted by 7230 while the first is not.  :wink:

andreaferretti:
Ok, so #13574 was not a bug after all. I am going to change this PR to only fix #13573

dom96 (approved):
LGTM

PR #13576, opened 2020-03-03
allow category nimble-packages to test a single package

krux02:
After this PR, `nimble cat nimble-packages nimx` will test `nimx` only. I used it to reproduce a CI failure locally.

### DIFF
--- a/testament/testament.nim
+++ b/testament/testament.nim
@@ -717,7 +717,6 @@ proc main() =
   of "c", "cat", "category":
     skips = loadSkipFrom(skipFrom)
     var cat = Category(p.key)
-    p.next
     processCategory(r, cat, p.cmdLineRest.string, testsDir, runJoinableTests = true)
   of "pcat":
     skips = loadSkipFrom(skipFrom)
--- a/testament/categories.nim
+++ b/testament/categories.nim
@@ -480,7 +480,7 @@ proc makeSupTest(test, options: string, cat: Category): TTest =
   result.options = options
   result.startTime = epochTime()
 
-proc testNimblePackages(r: var TResults, cat: Category) =
+proc testNimblePackages(r: var TResults; cat: Category; packageFilter: string) =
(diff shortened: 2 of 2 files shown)


testament/testament.nim
```diff
@@ -717,7 +717,6 @@ proc main() =
   of "c", "cat", "category":
     skips = loadSkipFrom(skipFrom)
     var cat = Category(p.key)
-    p.next
```
Araq, line 720: Looks really harmful to remove this line. What if I want to do ``testament cat foo -d:foobar``? is ``-d:foobar`` then your package filter?

krux02: Well for me this `p.next` looks like I bug that I fixed. `p.next` caused the `cmdLineRest` to be empty. Only without calling `p.next` it actually contains something. 

> What if I want to do ``testament cat foo -d:foobar``? is ``-d:foobar`` then your package filter?

Well category `foo` does not exist, so ``testament cat foo -d:foobar`` won't do anything. But if you mean ``testament cat nimble-packages -d:foobar``, then yes here ``-d:foobar`` will be the package filter.

Araq: > Well for me this p.next looks like I bug that I fixed. 

Well but it isn't a "bug you fixed", you introduced a new bug as I tried to explain.

krux02: Well to what I tested so far, if you call ``testament cat foo -d:foobar`` with current `testatment`, the argument `-d:foobar` is ignored and no warning about it is printed. The call is identical to ``testament cat foo``. If that is intentional behavior, then ok I revert this part and find a different solution, but to me this looks like a bug. 

With the deletion of `p.next` `-d:foobar` will be part of the options string that is passed down to the category. Correct me if I am wrong, but I think this is what should happen?

Araq: Ok, my bad.

PR #13580, opened 2020-03-03
Nimpretty Fix negative indent breaks code

juancarlospaco:
```console
$ cat example.nim 
proc f() =
    echo 42

try:
   echo 42
except:
   discard

$ nimpretty --maxLineLen:-9 --indent:-9 example.nim

$ cat example.nim 
proc f() =
echo 42

try:
echo 42
except:
discard

$ nim c -r /home/juan/code/Nim/nimpretty/nimpretty.nim --indent:-9 example.nim
Hint: nimpretty --indent:-9 example.nim [Exec]
Error: unhandled exception: value out of range: -9 [RangeError]

$
```

- 2 Line fix, easy merge, feel free to try it yourself with code above.

### DIFF
--- a/nimpretty/nimpretty.nim
+++ b/nimpretty/nimpretty.nim
@@ -44,8 +44,8 @@ proc writeVersion() =
 
 type
   PrettyOptions = object
-    indWidth: int
-    maxLineLen: int
+    indWidth: Natural
+    maxLineLen: Positive
 
 proc prettyPrint(infile, outfile: string, opt: PrettyOptions) =
   var conf = newConfigRef()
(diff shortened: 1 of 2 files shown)


nimpretty/nimpretty.nim
```diff
@@ -44,8 +44,8 @@ proc writeVersion() =
 
 type
   PrettyOptions = object
-    indWidth: int
-    maxLineLen: int
+    indWidth: Natural
```
Varriount, line 47: Shouldn't this be positive too?

Varriount (changes requested):

nimpretty/nimpretty.nim
juancarlospaco: No, `0` is Auto-Detect.

PR #13590, opened 2020-03-05
Removed simpleGetOrDefault

yglukhov:
Changed `simpleGetOrDefault` rewrite rule to `{}` single-arg overload. This removes annoying `Hint: simpleGetOrDefault` in the logs, and potentially increases compilation speed (but not sure).


juancarlospaco:
Add `{.inline.}` to be closer to `template` performance, it is a tiny proc anyway.
Needs a changelog entry, because term-rewriting --> proc

Clyybber:
Doesn't need a changelog entry if its done behind the scenes.

PR #13593, opened 2020-03-05
cycle breaker

Araq:
Do not merge.


lib/system/cyclebreaker.nim
```diff
@@ -0,0 +1,195 @@
...
+we found a cycle, it only means the node is shared.
+
+
+   a -------> b <----- c
+   |          ^        ^
+   +----------+        |
+   |                   |
+   +-------------------+
+
+If we simply remove all links to already processed nodes we end up with:
```
zah, line 43: Err, how is the memory going to be freed in this simple program? It creates a cycle without having any "shared" object.

```nim
type Node = ref object
  another: Node

proc createCycle =
  var x = new Node
  x.another = new Node
  x.another.another = x
  x = nil
```

Also, I'm not sure I understand the general concept. Nim breaks the cycles for me by setting some pointers in my program to `nil`? Does it restore them to their previous values afterwards? If not, this doesn't sound like something that's safe to do with the user's data structures.

Araq: Simple, call ``thinout(x)`` before ``x = nil``.

Clyybber: It doesn't. You must manually call thinout x before doing that. Unfortunately that will set x.another.another to nil, so it changed our structure (this isn't bad in this example, because we don't acces x.another afterwards as it will be destroyed, but it is bad in case its referenced from outside).
This is why I'm not sold on this approach in yet (I think it may suffice for async tho)

zah: Hrm, indeed `thinout` applies this traversal from the perspective of a single root, but it doesn't prevent other roots from pointing to the same graph of objects (which will be now damaged after the `thinout` operation).

Also, am I supposed to call `thinout` manually? We'll be back into the land of memory leaks caused by programming mistakes.

@Clyybber, maybe async needs a more [radical redesign](https://github.com/status-im/nim-chronos/issues/2) to work well in the ARC world.

Araq: Well it's still in the "experimental" phase.

PR #13600, opened 2020-03-06
added range functionality for runes

sealmove:


juancarlospaco:
Needs changelog entry and more than 1 assert test.
I would make it `func` and `{.inline.}` I dont know.

lib/pure/unicode.nim
```diff
@@ -1256,7 +1258,11 @@ proc isTitle*(s: string): bool {.noSideEffect, procvar, rtl, extern: "nuc$1Str",
     elif rune.isWhiteSpace():
       firstRune = true
 
+proc `..`(`from`, to: Rune): RuneRange =
```
kaushalmodi: These would need export markers too, right?

SolitudeSF:
There is a builtin type for ranges: `HSlice`. Why not use that one instead of introducing new one?

```nim
func contains*(range: HSlice[Rune, Rune], rune: Rune): bool =
  rune >=% range.a and rune <=% range.b
```
this is sufficient for your test to pass.

sealmove:
> There is a builtin type for ranges: `HSlice`. Why not use that one instead of introducing new one?
> 
> ```nim
> func contains*(range: HSlice[Rune, Rune], rune: Rune): bool =
>   rune >=% range.a and rune <=% range.b
> ```
> 
> this is sufficient for your test to pass.

Because HSlice generally works with ordinal types which Rune isn't.
For example `len`, `[]`, `[]=` won't work out of the box, and I don't know how to define them.

sealmove:
Ok... so there is a confict about the `..` operator.
`expression '[toRune("a") .. toRune("z"), toRune("A") .. toRune("Z")]' is of type: array[0..1, RuneRange]`

If Rune is not an ordinal type, and `..` operator can't work, then I think this PR is pointless after all... Maybe it's best if we force the user to simply use the `<%` operator?

Araq:
Unclear benefits so far. Sorry, rejected for now.

PR #13625, opened 2020-03-11
fix #13310, Deque misbehaves on VM

### LINKED ISSUE #13310: Deque misbehaves on VM
It seems all hashsets in a Deque of hashsets are empty after it reallocates. It misbehaves in devel, but works as expected in Nim 1.0.6.

### Example
```nim
import deques
import sets

proc main() =
  var q = initDeque[HashSet[int16]](2)
  q.addFirst([1'i16].toHashSet)
  q.addFirst([2'i16].toHashSet)
  # [{2}, {1}]
  echo q
  q.addFirst([3'i16].toHashSet)
  # [{3}, {}, {}]
  echo q

static:
  main()
```

### Current Output
```
[{2}, {1}]
[{3}, {}, {}]
```

### Expected Output
```
[{2}, {1}]
[{3}, {2}, {1}]
```

### Additional Information

* Works well in Nim 1.0.6

```
$ nim -v
Compiled at 2020-01-23

```

narimiran:


Araq:
This is slower and a workaround for a VM bug. To make it a decent workaround, use a ``when nimVM`` block.

narimiran:
> To make it a decent workaround, use a when nimVM block.

Done. Unrelated `freebsd` CI failure (`tests/async/tioselectors.nim`)

lib/pure/collections/deques.nim
```diff
@@ -274,8 +274,9 @@ proc expandIfNeeded[T](deq: var Deque[T]) =
   if unlikely(deq.count >= cap):
     var n = newSeq[T](cap * 2)
     var i = 0
-    for x in mitems(deq): # don't use copyMem because of the GC and because it's slower.
-      n[i] = move(x)
+    for x in mitems(deq):
+      when nimVM: n[i] = x # workaround for VM bug
+      else: n[i] = move(x)
       inc i
```
Clyybber, line 280: Could this be changed to
```nim
for i, x in mpairs(deq):
  when nimVM: n[i] = x
  else: n[i] = move(x)
```
?

Araq: Probably but not important enough. Plus I don't really like the ``mpairs`` / ``pairs`` design for seqs and arrays.

PR #13650, opened 2020-03-14
introduce getPeerCertificates, fixes #13299

### LINKED ISSUE #13299: net/asyncnet: Unable to access peer's certificate chain
### Summary
I'm using asyncnet to establish TLS connections and would like to access the verified certificate chain of the peer I'm connected with (either client or server).

### Description
Currently it's not possible to retrieve the certificate chain from an OpenSSL-wrapped ``AsyncSocket`` after the TLS-Handshake has completed. In the [chat server example](https://nim-lang.org/docs/asyncnet.html#examples-chat-server) from the asyncnet module this would be in ``processClient``::

``` nim
...
var sslContext = newContext()

proc processClient(client: AsyncSocket) {.async.} =
  wrapConnectedSocket(sslContext, client, handshakeAsServer)
  while true:
    let line = await client.recvLine()
    let certs = client.getPeerCertificates() # not possible
    if line.len == 0: break
    for c in clients:
      await c.send(line & "\c\L")
...
```

OpenSSL provides ``SSL_get0_verified_chain``, but it needs the ``SSL`` pointer of the current connection, which I can't access. There are two possible solutions that I can think of:

1. Export the ``sslHandle`` field of ``AsyncSocket`` / ``Socket``, so users can call ``SSL_get0_verified_chain`` on it

2. add a ``proc getPeerCertificates`` to the standard library, e.g.:
``` nim
type
  PSTACK_OF_X509* = SslPtr
  Certificate* = string

proc CRYPTO_free*(p: pointer, file: cstring, line: int) {.cdecl, dynlib: DLLSSLName, importc.}

proc i2d_X509*(x: PX509, output: ptr ptr cchar): int {.cdecl, dynlib: DLLSSLName, importc.}

proc OPENSSL_sk_num*(stack: SslPtr): int {.cdecl, dynlib: DLLSSLName, importc.}

proc OPENSSL_sk_value*(stack: SslPtr, index: int): pointer {.cdecl, dynlib: DLLSSLName, importc.}

proc SSL_get0_verified_chain*(ssl: SslPtr): PSTACK_OF_X509 {.cdecl, dynlib: DLLSSLName, importc.}

proc getPeerCertificates*(socket: AsyncSocket): seq[Certificate] =
  result = newSeq[Certificate]()
  let stack = SSL_get0_verified_chain(socket.sslHandle)
  if stack == nil:
    return
  let length = OPENSSL_sk_num(stack)
  if length == 0:
    return
  for i in 0 .. length - 1:
    var der: ptr cchar = nil
    let derLen = i2d_X509(cast[PX509](OPENSSL_sk_value(stack, i)), addr der)
    var cert = newString(derLen)
    copyMem(addr cert[0], der, derLen)
    CRYPTO_free(der, instantiationInfo().filename, instantiationInfo().line)
    result.add(cert)
```

### Alternatives
I tried to register a callback using ``SSL_CTX_set_info_callback`` to retrieve the ``SSL`` pointer, but it seemed unnessecarily hard since it is already stored in the socket.

### Additional Information
[OpenSSL's ``SSL_get0_verified_chain``](https://www.openssl.org/docs/man1.1.1/man3/SSL_get0_verified_chain.html)

royneary:
getPeerCertificates retrieves the verified certificate chain of the peer
we are connected to through an SSL-wrapped Socket/AsyncSocket. This
introduces the new type ``Certificate`` which stores a DER-encoded X509 certificate.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -119,6 +119,9 @@ echo f
 - Added a new module, `std / compilesettings` for querying the compiler about
   diverse configuration settings.
 - `base64` adds URL-Safe Base64, implements RFC-4648 Section-7.
+- Added `net.getPeerCertificates` and `asyncnet.getPeerCertificates` for
+  retrieving the verified certificate chain of the peer we are connected to
+  through an SSL-wrapped `Socket`/`AsyncSocket`.
 
 
 ## Library changes
--- a/lib/pure/asyncnet.nim
+++ b/lib/pure/asyncnet.nim
@@ -95,6 +95,8 @@
 ##   runForever()
 ##
 
+include "system/inclrtl"
+
 import asyncdispatch
 import nativesockets
 import net
@@ -742,6 +744,17 @@ when defineSsl:
       sslSetConnectState(socket.sslHandle)
     of handshakeAsServer:
       sslSetAcceptState(socket.sslHandle)
+
+  proc getPeerCertificates*(socket: AsyncSocket): seq[Certificate] {.since: (1, 1).} =
+    ## Returns the certificate chain received by the peer we are connected to
+    ## through the given socket.
+    ## The handshake must have been completed and the certificate chain must
+    ## have been verified successfully or else an empty sequence is returned.
+    ## The chain is ordered from leaf certificate to root certificate.
+    if not socket.isSsl:
+      result = newSeq[Certificate]()
+    else:
+      result = getPeerCertificates(socket.sslHandle)
 
 proc getSockOpt*(socket: AsyncSocket, opt: SOBool, level = SOL_SOCKET): bool {.
   tags: [ReadIOEffect].} =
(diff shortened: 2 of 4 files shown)


Araq:
Add a ``.since`` annotation, a changelog entry and rebase since Azure pipelines got the flu and didn't test anything.

royneary:
Done.

In the meanwhile 5b854442448af96d57135ba7328b0c21f1f80f40 introduced the wrapper ``proc i2d_X509*(cert: PX509): string`` which is useful for ``getPeerCertificates``. Unfortunately it was defined in a ``when not defined(nimDisableCertificateValidation)`` block. My code is not about certificate validation, so it should compile independently from the ``nimDisableCertificateValidation`` flag.

To solve this, I moved ``i2d_X509`` out of the block. This should be ok since the reason for ``nimDisableCertificateValidation`` (that some symbols are not available in older OpenSSL versions) does not apply to ``i2d_X509`` which has been "available in all versions of SSLeay and OpenSSL" [1].

For consistency I moved the reverse proc ``d2i_X509`` too.

[1] https://linux.die.net/man/3/i2d_x509

PR #13651, opened 2020-03-14
fixes #12747 [backport]

### LINKED ISSUE #12747: Arrays are passed by copy to iterators, causing crashes, unnecessary allocations and slowdowns
```nim
type
    ABC = ref object
        arr: array[0x40000, pointer]

let a = ABC()

for a in a.arr:
    assert a == nil
```
This code crashes with an unspecified error:
```
Error: execution of an external program failed: 'c:\Users\Kemal\Desktop\workframe\arg.exe '
```

If I interpret the C code correctly, the reason for this that the whole array is copied onto the stack, which can't hold the 2mb object.

cooldome:


compiler/transf.nim
```diff
@@ -563,7 +563,7 @@ proc putArgInto(arg: PNode, formal: PType): TPutArgInto =
       return paDirectMapping    # XXX really correct?
                                 # what if ``arg`` has side-effects?
   case arg.kind
-  of nkEmpty..nkNilLit:
+  of nkEmpty..nkNilLit, nkDotExpr:
```
Araq: What if it's ``sideEffectHere().field``? Then we cannot use a direct mapping, can we?

cooldome: Yep, fixed

cooldome:
I have also added nkObjConstr handling as well and also noticed that nkExprColonExpr is not handled for nkPar, nkTupleConstr.

Araq:
Please rebase

cooldome:
Ready for review

Clyybber (approved):

PR #13665, opened 2020-03-16
Add 2 modules

juancarlospaco:
- Add 2 new modules, `xpm.nim` and `netpbm.nim`.
- Half the code is Tests approx, all functions are asserted.
- Only uses `system.nim`, no imports, does not use std lib, easy to maintain.
- `runnableExamples` with `static: ` and `doAssert`.
- Documentation with Examples and Links. `changelog.md` entry.
- Works at compile-time and JavaScript.
- All functions are `func`, except 1 that does the I/O.
- Tiny `func` of 3 ~ 5 lines, simple code, simple API.
- Functionality can be extended with `sequtils` by the user.
- Examples render 6 small fractals at compile-time to mimic a real-world like usage.
- Only adds ~300 lines where half are test, adding 5 image formats, so I think it worth it.

#### Random example use cases

- Generate Icons of arbitrary sizes instead of bundling them in binaries and installers.
- Generate patterns on client-side browser instead of transferring them over internet.
- Render QR codes, Bar codes, Icons, visual reports, simple diagrams for documentation, etc.
- Plot charts with data of any kind on any target.
- Anything image related, good addition to SVG and HTML that std lib already can do.

#### Format

Image format is loss-less and human-readable, UTF-8 but fits ASCII,
is also valid C source code, can be embedded in C, supports millions of colors and transparency.

#### Links

- Rust lang one https://docs.rs/image/latest/image/pnm/struct.PNMEncoder.html
- Python also has a module to draw images on std lib.
- Spec https://en.wikipedia.org/wiki/X_PixMap
- Spec https://en.wikipedia.org/wiki/Netpbm_format

# Examples with Std Lib only

fractalito

fractalito_hexaflake

fractals

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -103,6 +103,9 @@ echo f
 - Added `times.isLeapDay`
 - Added a new module, `std / compilesettings` for querying the compiler about
   diverse configuration settings.
+- Added `netpbm` module with NetPBM raster bitmap image implementation for image rendering.
+- Added `xpm` module with XPM X-Pixmap raster image implementation for image rendering.
+
 
 ## Library changes
 
--- a/lib/std/netpbm.nim
+++ b/lib/std/netpbm.nim
@@ -1,0 +1,356 @@
+## NetPBM raster bitmap image standard implemented in pure Nim.
+## Has image manipulation funcs, patterns, crop, expand, fill, save to file, etc.
+## Uses only ``system.nim`` (without any imports), works on any compiler target,
+## including compile-time, JavaScript and NimScript, etc.
+## Should work on any embedded devices that can handle ``string``.
+## Almost all functions work in-place and are side-effects free.
+## Format is loss-less uncompressed and human-readable, UTF-8 but fits ASCII.
+## Pixels are just ``byte``, Bitmap is a ``seq[seq[byte]]``. Unstable API.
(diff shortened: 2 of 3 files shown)


krux02:
* You don't need `var` for `ref` types (unless you want to set the pointer itself). Also the `var` prefix binds stronger than `|`, so the ast for your arguments looks like this: ``(Infix
     (Infix (Ident "|") (VarTy (Ident "PBM")) (Ident "PGM")))``, not as you might think``(VarTy (Infix (Ident "|") (Ident "PBM") (Ident "PGM")))``. 
 * Using types composed with `|` makes the proc implicitly generic with code instantiation per type, even though the implementation is completely identical, no generic code instantiation needed, except for the `$` and the `writeFile` proc. Please don't use generics for this module. 
 * Please don't use `ref` types if you don't need to, and I  don't see the need for ref types here.
 * Please don't use `seq[seq[byte]]` for image data. This has too much allocation overhead. Use one flat `seq[byte]` and compute the indices. This is also the format that for example OpenGL needs when you pass image data to it.

I Didn't even start yet to look into the procedures and question if they are worth to be added to the standard library. Why don't you make this just a nimble package?

dom96 (changes requested):
I don't think this belongs to the stdlib.

PR #13666, opened 2020-03-16
fix sets of scoped imported enums

ghost:
fails because `typedef NU8 tySet_namespaced::Enum;` is invalid.


Araq:
The fix is fine but how come the CIs are yet again messed up...

PR #13669, opened 2020-03-17
updated builds.sr.ht script according to their email

Araq:


Araq:
builds.sr.ht job is green, good enough.

PR #13684, opened 2020-03-18
add `move` to `tables` to prevent warnings when compiled with `--gc:arc`

narimiran:


lib/pure/collections/tables.nim
```diff
@@ -2947,17 +2947,17 @@ when isMainModule:
     p1, p2: Person
   p1.firstName = "Jon"
   p1.lastName = "Ross"
-  salaries[p1] = 30_000
+  salaries[move(p1)] = 30_000
```
Araq: Do not touch these test cases, you change the test! Now after ``move(p1)`` ``p1`` will be empty!

PR #13697, opened 2020-03-19
Implement SSL certificate verification

FedericoCeratto:
To allow a smoother adoption curve, certificate verification is not enabled by default in httpclient.
It's also not supported on Windows due to an old OpenSSL version. 
See #6664 for details


.github/workflows/ci_ssl.yml
```diff
@@ -0,0 +1,87 @@
+name: Nim SSL CI
+on:
+  pull_request:
```
alaviss, line 3: You might want to duplicate this for `push` too.

.github/workflows/ci_ssl.yml
```diff
@@ -0,0 +1,87 @@
...
+      - 'tests/untestable/thttpclient_ssl*'
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-18.04, macos-10.15, windows-2019]
+        cpu: [amd64]
+    name: '${{ matrix.os }} (${{ matrix.cpu }})'
```
alaviss, line 20:
```suggestion
    name: '${{ matrix.os }}'
```

We aren't doing 32 bit testing here, so this could be cut

.github/workflows/ci_ssl.yml
```diff
@@ -0,0 +1,87 @@
...
+        uses: actions/checkout@v2
+
+      - name: 'Checkout csources'
+        uses: actions/checkout@v2
+        with:
+          repository: nim-lang/csources
+          path: csources
+
+      - name: 'Install dependencies (Linux amd64)'
+        if: runner.os == 'Linux' && matrix.cpu == 'amd64'
```
alaviss, line 33:
```suggestion
      - name: 'Install dependencies (Linux)'
        if: runner.os == 'Linux'
```

.github/workflows/ci_ssl.yml
```diff
@@ -0,0 +1,87 @@
...
+        run: |
+          sudo apt-fast update -qq
+          DEBIAN_FRONTEND='noninteractive' \
+            sudo apt-fast install --no-install-recommends -yq \
+              libcurl4-openssl-dev libgc-dev libsdl1.2-dev libsfml-dev \
+              valgrind libc6-dbg
+      - name: 'Install dependencies (macOS)'
+        if: runner.os == 'macOS'
+        run: brew install boehmgc make sfml
```
alaviss:
```suggestion
        run: brew install make
```

.github/workflows/ci_ssl.yml
```diff
@@ -0,0 +1,87 @@
...
+          path: csources
+
+      - name: 'Install dependencies (Linux amd64)'
+        if: runner.os == 'Linux' && matrix.cpu == 'amd64'
+        run: |
+          sudo apt-fast update -qq
+          DEBIAN_FRONTEND='noninteractive' \
+            sudo apt-fast install --no-install-recommends -yq \
+              libcurl4-openssl-dev libgc-dev libsdl1.2-dev libsfml-dev \
+              valgrind libc6-dbg
```
alaviss: the `DEBIAN_FRONTEND` part should've been put after sudo or else it won't inherit, but I don't think we even need it.

Can you trim these dependencies so that they are just enough to run the tests?

.github/workflows/ci_ssl.yml
```diff
@@ -0,0 +1,87 @@
...
+          'macOS')
+            ncpu=$(sysctl -n hw.ncpu)
+            ;;
+          'Windows')
+            ncpu=$NUMBER_OF_PROCESSORS
+            ;;
+          esac
+          [[ -z "$ncpu" || $ncpu -le 0 ]] && ncpu=1
+
+          make -C csources -j $ncpu CC=gcc ucpu='${{ matrix.cpu }}'
```
alaviss, line 73:
```suggestion
          make -C csources -j $ncpu CC=gcc
```

.github/workflows/ci_ssl.yml
```diff
@@ -0,0 +1,87 @@
...
+        shell: bash
+        run: nim c koch
+
+      - name: 'Build the real compiler'
+        shell: bash
+        run: ./koch boot
+
+      - name: 'Run SSL integration tests'
+        shell: bash
+        run: ./bin/nim c -d:ssl -p:. --threads:on -r tests/untestable/thttpclient_ssl.nim
```
alaviss:
```suggestion
        run: nim c -d:ssl -p:. --threads:on -r tests/untestable/thttpclient_ssl.nim
```

nim is in PATH.

.travis.yml
```diff
@@ -20,6 +20,7 @@ addons:
     - libsfml-dev
     - libc6-dbg
     - valgrind
+    - libssl-dev
```
alaviss: Not necessary, travis only runs docgen.

.github/workflows/ci_ssl.yml
FedericoCeratto: People might want to add it back

.github/workflows/ci_ssl.yml
FedericoCeratto: Done

.github/workflows/ci_ssl.yml
FedericoCeratto: Done

narimiran:
@FedericoCeratto After talking with @Araq, I've pushed a commit to your branch which disables this on Windows.

.travis.yml
FedericoCeratto: Removed

.github/workflows/ci_ssl.yml
FedericoCeratto: Fixed

PR #13710, opened 2020-03-20
remove deprecated things in macros except for callsite

ghost:
From what I can tell these have all been deprecated for 2 years or more.
I'll leave this here for when it is time to clean them up.

callsite's future is uncertain (see: https://github.com/nim-lang/Nim/pull/11864) so didn't touch it.

### DIFF
--- a/compiler/ast.nim
+++ b/compiler/ast.nim
@@ -664,8 +664,8 @@ type
     mNccValue, mNccInc, mNcsAdd, mNcsIncl, mNcsLen, mNcsAt,
     mNctPut, mNctLen, mNctGet, mNctHasNext, mNctNext,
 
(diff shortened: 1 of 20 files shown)


krux02:
Of course these deprecated symbols are used a lot in packages, because back then when I marked them as deprecated I also just introduced the substitution. If you want to make this PR test green, I am sorry I think you have to make a lot of PR for all the important packages as well. If you make a PR with just the changes where you don't use the deprecated symbols anymore (but without the removal of any feature), it will be valuable and it will be approved quickly (at least from me).

Regarding ``callsite``: callsite is only used to get the line information of the callsite. A specific magic proc for that could take that over.

timotheecour:
@dumjyl just curious whether you closed because you don't have time to work on this (which is fine) or for some other reason?
as @krux02 said, a less ambitious PR would work great in the meantime and represent less work for you

PR #13711, opened 2020-03-20
fixes #13708

### LINKED ISSUE #13708: ARC: variables are no copied on the thread spawn causing crashes
The test case that works in single threaded mode, but not in multithreaded mode.
Compile with `--gc:arc --threads:on`
```nim
import threadpool
import strutils
import os

proc work(s: string) =
  let (abs_dir, file_ext) = splitPath(s)
  echo file_ext # needed to make task heavier to produce race conditions
  doAssert(file_ext.endsWith(".xml"))

proc thread_run(p: string, singleThread = false) =
  let path = p
  if singleThread:
    work(path)
  else:
    spawn(work(path))

var i = 0
for path in walkDirRec(r"c:\"):
  if splitFile(path).ext == ".xml":
    echo path
    thread_run(path, singleThread = false) # works  if singleThread = true
    i.inc
  if i >= 1000:
    break

sync()
```

cooldome:
fixes #13708

### DIFF
--- a/compiler/lowerings.nim
+++ b/compiler/lowerings.nim
@@ -312,10 +312,10 @@ proc indirectAccess*(a: PNode, b: PSym, info: TLineInfo): PNode =
 proc indirectAccess*(a, b: PSym, info: TLineInfo): PNode =
   result = indirectAccess(newSymNode(a), b, info)
 
-proc genAddrOf*(n: PNode): PNode =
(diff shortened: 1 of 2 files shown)


cooldome:
fix works for c but not for cpp, don't merge yet

cooldome:
now works with cpp as well

PR #13727, opened 2020-03-22
swap sighash for string in idOrSig

disruptek:


Araq:
Didn't work out, thanks for the experiment. I'm keeping this solution in mind for future code generators.

PR #13734, opened 2020-03-23
Enable GitHub CI

FedericoCeratto:


FedericoCeratto:
ARM is advertised as supported but apparently it's not.

PR #13741, opened 2020-03-24
Fix vm.nim for --gc:arc

Clyybber:
Compiling the compiler with `--gc:arc` now passes the Nim stage, but makes the C compiler complain.


Clyybber:
This is not the right way..

PR #13748, opened 2020-03-25
fix deprecations and other warnings

narimiran:


narimiran:
Unrelated fail on freebsd.

PR #13752, opened 2020-03-25
Fix typeSym.getImpl for ref types

zah:


zah:
To clarify the issue, since the type AST was not attached to the `T:ObjectType` symbol created for ref types, magics such as `getImpl` were returning `nil` nodes for such types breaking basic functionality such as listing all the fields.

Araq:
This breaks  tests/stdlib/tjsonmacro.nim

PR #13754, opened 2020-03-25
docs: make the syntax for generics easy to look up

roryokane:
edits https://nim-lang.org/docs/tut2.html#generics

### DIFF
--- a/doc/tut2.rst
+++ b/doc/tut2.rst
@@ -469,7 +469,8 @@ Generics
 ========
 
 Generics are Nim's means to parametrize procs, iterators or types
-with `type parameters`:idx:. They are most useful for efficient type safe
+with `type parameters`:idx:. Generic parameters are written within square
+brackets, for example ``Foo[T]``. They are most useful for efficient type safe
 containers:
 
 .. code-block:: nim


doc/tut2.rst
```diff
@@ -469,7 +469,8 @@ Generics
 ========
 
 Generics are Nim's means to parametrize procs, iterators or types
-with `type parameters`:idx:. They are most useful for efficient type safe
+with `type parameters`:idx:. Generic parameters are written within square
+brackets, for example ``Foo[T]``. They are most useful for efficient type safe
```
juancarlospaco, line 473: `They are most useful for efficient type safe` weird grammar,
I would just link https://en.wikipedia.org/wiki/Generic_programming instead. :thinking:

roryokane: This file uses hard wrap, so that sentence continues on the next line. The full sentence:

> They are most useful for efficient type safe containers:

I think the wording of that sentence is fine. The following paragraph is a Nim code example that implements an efficient type-safe container using generics, so that sentence serves a purpose beyond describing generics in general.

Also, note that I didn’t write the sentence we are discussing. The only change I made was adding this sentence:

> Generic parameters are written within square brackets, for example ``Foo[T]``.

juancarlospaco: You are right.

PR #13759, opened 2020-03-25
Add Data URI Base64, implements RFC-2397

juancarlospaco:
- `base64` adds Data URI Base64, implements RFC-2397.
- Documentation with links and examples, Changelog, Test, `since`, `runnableExamples: static:`.
- Follow up from https://github.com/nim-lang/Nim/pull/13672#issue-389779067


ghost:
I'm not really into web development, but does this need to belong to base64 module? Maybe it's better to put it into some http-related module, not really sure.

dom96:
Maybe put it in the `uri` module?

dom96:
Do other languages have a similar proc? What do they call it? (`dataUri` sounds to me like it should be `getDataUri` at least)

PR #13769, opened 2020-03-26
make nim_temp compile with --gc:arc --sinkInference:off

cooldome:
This is the changes I had to make to compile nim_temp with `--gc:arc --sinkInference:off`.
Compiled nim_temp doesn't work yet there are runtime issues.

### DIFF
--- a/compiler/cgen.nim
+++ b/compiler/cgen.nim
@@ -425,7 +425,7 @@ proc resetLoc(p: BProc, loc: var TLoc) =
 
 proc constructLoc(p: BProc, loc: var TLoc, isTemp = false) =
   let typ = loc.t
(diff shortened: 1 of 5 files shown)


Clyybber:
Ha, I just recently did the same here (the vm fix only): https://github.com/nim-lang/Nim/pull/13741, I switched it to use `ptr UncheckedArray` as to save one pointer indirection.

cooldome:
@Clyybber , your approach is more efficient so I merged it. This is now a codegen only fix

Clyybber:
Nice, I think @Araq wanted to delay merging my PR until 1.2 is released tho :D

narimiran:
@cooldome no need to restart the build. The fail on `freebsd` is unrelated, and is happening on some other PRs/commits too.

PR #13770, opened 2020-03-26
Add long arrow to sugar

juancarlospaco:
- Add `==>` to `sugar`, literally the same implementation as `=>` but generates an inlined `func`.
- Documentation with examples, changelog, `since`, `runnableExamples`, etc.

I did not touch the implementation, just added a `when` with the pragmas.

```nim
import macros

static:
  expandMacros:
     let works = (a: int) ==> a + 2
     let goods = (a: int) ==> debugEcho a
```

```console
$ nim c -f example.nim
let works = proc (a: int): auto {.inline, noSideEffect.} = result = a + 2
let goods = proc (a: int): auto {.inline, noSideEffect.} = debugEcho a

```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -123,6 +123,8 @@ echo f
   retrieving the verified certificate chain of the peer we are connected to
   through an SSL-wrapped `Socket`/`AsyncSocket`.
 - Added `distinctBase` overload for values: `assert 12.MyInt.distinctBase == 12`
+- Added `==>` to `sugar`, same as `=>` but generates an anonymous inlined func.
+
 
 ## Library changes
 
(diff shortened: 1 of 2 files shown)


Clyybber:
Too specific/minor to receive its own operator IMO.

juancarlospaco:
I agree, so does `=>`, but thats why is named *Syntax Sugar* by definition.

Clyybber:
No, `=>` has a common usecase.

juancarlospaco:
I know, seen people coming from TS/JS/Flow liking it a lot, thats why it can be improved.

narimiran:
With six thumbs down in just this short amount of time, I have to reject this, sorry.

PR #13778, opened 2020-03-27
new minor feature: macros for proc types, to be documented

Araq:


lib/pure/asyncmacro.nim
```diff
@@ -212,6 +212,12 @@ proc verifyReturnType(typeName: string) {.compileTime.} =
 proc asyncSingleProc(prc: NimNode): NimNode {.compileTime.} =
   ## This macro transforms a single procedure into a closure iterator.
   ## The ``async`` macro supports a stmtList holding multiple async procedures.
+  if prc.kind == nnkProcTy:
+    result = copyNimTree(prc)
```
krux02: you never need to call ``copyNimTree`` in macros. The argument is already a fresh copy.

krux02 (changes requested):
There is no test for the new feature.
There is no specification for the new feature.

PR #13784, opened 2020-03-28
fixes #13781, fixes #13805

### LINKED ISSUE #13781: gc:arc cannot fully support threadpool with FlowVar
# Example
```nim
import threadpool, os

proc thread(): string =
  os.sleep(1000)
  return "ok"

var fv = spawn thread()
sync()
echo ^fv
```

### Current Output
```
Error: cannot bind another '=destroy' to: FlowVarObj; previous declaration was constructed here implicitly: d:\nim\lib\pure\concurrency\threadpool.nim(211, 48)
```

### Expected Output
```
ok
```

$ nim -v
Compiled at 2020-03-27```

### LINKED ISSUE #13805: arc & threads: doesn't pass the reference forwarding test
The object is not destroyed:
```nim
import threadpool, os
type
  MyObj = object
    p: int

proc `=destroy`(x: var MyObj) = 
  if x.p != 0:
    echo "destroyed"

proc thread4(x: ref MyObj): ref MyObj =
  os.sleep(1000)
  result = x

proc ref_forwarding_test = 
  var x = new(MyObj)
  x[].p = 2
  var y = spawn thread4(x)
  sync()

ref_forwarding_test()
```

Destructor is not invoked.

cooldome:
fixes #13781 and #13805

### DIFF
--- a/compiler/injectdestructors.nim
+++ b/compiler/injectdestructors.nim
@@ -44,7 +44,7 @@ type
     graph: ModuleGraph
     emptyNode: PNode
     otherRead: PNode
-    inLoop, hasUnstructuredCf, inDangerousBranch: int
+    inLoop, inSpawn, hasUnstructuredCf, inDangerousBranch: int
     declaredVars: IntSet # variables we already moved to the top level
     uninit: IntSet # set of uninit'ed vars
     uninitComputed: bool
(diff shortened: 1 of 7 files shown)


Araq:
tests/arc/tthread.nim fails with this patch.

cooldome:
Added fix for #13805.
Added ability to spawn a function with sink parameter

Along the way I have to fix an issue that for proc:
```nim
proc test(x: ref MyObj)
```
hasDestructor(x.typ) returned false simply because createTypeBoundOps was not invoked for it.
createTypeBoundOps is also needed for proc's return types. Triggered change in sempass2.

cooldome:
Closed another PR is submitted instead

PR #13816, opened 2020-03-31
Unwind just the "pseudorandom probing" part of recent sets,tables changes

c-blake:
Unwind just the "pseudorandom probing" (whole hash-code-keyed variable stride double hashing) part of recent sets & tables changes (which has still been causing bugs over a month later (e.g., two days ago https://github.com/nim-lang/Nim/issues/13794) as well as still having several "figure this out" implementation question comments in them (see just diffs of this PR).

This topic has been discussed in many places:
    https://github.com/nim-lang/Nim/issues/13393
    https://github.com/nim-lang/Nim/pull/13418
    https://github.com/nim-lang/Nim/pull/13440
    https://github.com/nim-lang/Nim/issues/13794

Alternative/non-mandatory stronger integer hashes (or vice-versa opt-in identity hashes) are a better solution that is more general (no illusion of one hard-coded sequence solving all problems) while retaining the virtues of linear probing such as cache obliviousness and age-less tables under delete-heavy workloads (still untested after a month of this change).

The only real solution for truly adversarial keys is a hash keyed off of data unobservable to attackers.  That all fits better with a few families of user-pluggable/define-switchable hashes which can be provided in a separate PR more about `hashes.nim`.

This PR carefully preserves the better (but still hard coded!) probing of the  `intsets` and other recent fixes like `move` annotations, hash order invariant tests, `intsets.missingOrExcl` fixing, and the move of `rightSize` into `hashcommon.nim`.

### DIFF
--- a/lib/pure/collections/hashcommon.nim
+++ b/lib/pure/collections/hashcommon.nim
@@ -30,8 +30,8 @@ proc nextTry(h, maxHash: Hash): Hash {.inline.} =
   result = (h + 1) and maxHash
 
 proc mustRehash[T](t: T): bool {.inline.} =
-  assert(t.data.len > t.counter)
-  result = (t.data.len * 2 < t.counter * 3) or (t.data.len - t.counter < 4)
+  assert(t.dataLen > t.counter)
+  result = (t.dataLen * 2 < t.counter * 3) or (t.dataLen - t.counter < 4)
 
 proc rightSize*(count: Natural): int {.inline.} =
   ## Return the value of ``initialSize`` to support ``count`` items.
--- a/lib/pure/collections/hashcommon.nim
+++ b/lib/pure/collections/hashcommon.nim
@@ -18,87 +18,43 @@ when not defined(nimHasDefault):
     var v: T
     v
 
-const freeMarker = 0
-const deletedMarker = -1
-
-type UHash = uint
-
 # hcode for real keys cannot be zero.  hcode==0 signifies an empty slot.  These
 # two procs retain clarity of that encoding without the space cost of an enum.
-proc isFilledAndValid(hcode: Hash): bool {.inline.} =
-  result = hcode != 0 and hcode != deletedMarker
-    # performance: we could use bit magic if needed
+proc isEmpty(hcode: Hash): bool {.inline.} =
+  result = hcode == 0
 
 proc isFilled(hcode: Hash): bool {.inline.} =
   result = hcode != 0
 
-
-proc translateBits(a: UHash, numBitsMask: int): UHash {.inline.} =
-  result = (a shr numBitsMask) or (a shl (UHash.sizeof * 8 - numBitsMask))
-
-proc nextTry(h, maxHash: Hash, perturb: var UHash): Hash {.inline.} =
-  # FACTOR between hashcommon.nextTry, intsets.nextTry
-  # an optimization would be to use `(h + 1) and maxHash` for a few iterations
-  # and then switch to the formula below, to get "best of both worlds": good
-  # cache locality, except when a collision cluster is detected (ie, large number
-  # of iterations).
-  const PERTURB_SHIFT = 5 # consider tying this to `numBitsMask = fastLog2(t.dataLen)`
-  result = cast[Hash]((5*cast[uint](h) + 1 + perturb) and cast[uint](maxHash))
-  perturb = perturb shr PERTURB_SHIFT
+proc nextTry(h, maxHash: Hash): Hash {.inline.} =
+  result = (h + 1) and maxHash
 
 proc mustRehash[T](t: T): bool {.inline.} =
-  # FACTOR between hashcommon.mustRehash, intsets.mustRehash
-  let counter2 = t.counter + t.countDeleted
-  let length = t.dataLen
-  assert(length > counter2)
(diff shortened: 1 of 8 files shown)


lib/pure/collections/hashcommon.nim
```diff
@@ -18,87 +18,43 @@ when not defined(nimHasDefault):
...
-  let counter2 = t.counter + t.countDeleted
-  let length = t.dataLen
-  assert(length > counter2)
-  result = (length * 2 < counter2 * 3) or (length - counter2 < 4) # synchronize with `rightSize`
+  assert(t.data.len > t.counter)
+  result = (t.data.len * 2 < t.counter * 3) or (t.data.len - t.counter < 4)
 
 proc rightSize*(count: Natural): int {.inline.} =
-  ## Return the value of `initialSize` to support `count` items.
+  ## Return the value of ``initialSize`` to support ``count`` items.
```
Clyybber, line 37: Was this by accident? I don't think this is necessary.

lib/pure/collections/setimpl.nim
```diff
@@ -69,6 +68,11 @@ template containsOrInclImpl() {.dirty.} =
     rawInsert(s, s.data, key, hc, -1 - index)
     inc(s.counter)
 
+template doWhile(a, b) =
+  while true:
+    b
+    if not a: break
```
Clyybber, line 74: I think this should be inlined as its only used once and its name doesn't really reflect what it does (runs b once even if a is false already)

lib/pure/collections/hashcommon.nim
c-blake: Well, it was on purpose because I thought/think doc comments should be RST not markdown quoting.  No?

lib/pure/collections/setimpl.nim
c-blake: I don't really disagree.  In fact I inlined it just as you say in my https://github.com/c-blake/adix package (e.g. in `lpset.nim`).

I left it that way because the original code being restored was that way.  So, minimizing the pre-Tombstone to post-restored-no-Tombstone diff.

It is named after the C do-while construct, but because of indentation based block structure there's not really a better way to name it, IMO.  I agree it could go, but that should maybe be a follow-on clean-up PR.

Clyybber: I think its fine to inline it in this PR :)

lib/pure/collections/hashcommon.nim
Clyybber: Asked on irc; it doesn't really matter here, but single quotes are more common.

Clyybber: Quoting Araq: 
> Single if it refers to a parameter
> single or double otherwise ;-)

c-blake: Hmm.  I don't really care.  I was just trying to do what's right.  Should I change it back or not?  The `initialSize` part refers to the parameter of *another* far flung `proc` which makes this kind of an unusual case.

Clyybber:
```suggestion
  ## Return the value of `initialSize` to support `count` items.
```
Yeah doesn't really matter. It just threw me off because I've never seen `` used in the stdlib docs (but I also don't look at them that often :p ).

lib/pure/collections/setimpl.nim
c-blake: Let's do one thing at a time for simplicity.  I have a couple clean-ups and API enhancements in mind.

c-blake: While I have your attention, do you think `getCap` is a good name for the `data.len`?  `len` itself is already used for `count`.  I have been using `getCap` and `setCap` in https://github.com/c-blake/adix.  Both are kind of missing by oversight in both `sets` and `tables`.

c-blake:
All the tests passed.  @Araq  Re: https://gist.github.com/dom96/f84dd692f42a35c30bbee14309529d9e#gistcomment-3231894 and @dom96 - please take a look.  Thanks.

juancarlospaco:
I think that for core stuff like this all `assert` need to be changed to `doAssert`,
and add a descriptive string message for if someday eventually it does raise.

c-blake:
That sounds like a fine clean-up.  Maybe for the same follow-on PR that inlines `doWhile`?  (There are actually also a bunch in `heapqueue.nim`, `deques.nim`, `lists.nim`, etc.  Someone besides me might be better to do all that.)

juancarlospaco:
I was talking about the `assert` touched on this PR, but Ok.  :P

narimiran:
When the only two remarks are "should this be a single or double backtick" and "should this be `assert` or `doAssert`", you know you did a good job! :)

Joking aside, I've reviewed it locally, and this is a nice, clean, precise PR (when one manually diffs it with `8c22518d6^`)!

Good job @c-blake, and looking forward to your future hash-related PRs!

c-blake:
Well, I tried to follow up with one alternative hash as promised.  (I may have made the git somewhat messier than ideal..sorry if so.)

lib/pure/collections/setimpl.nim
Clyybber: I think `getCap`/`setCap` is fine. Its kind of analog to seq's capacity.

PR #13835, opened 2020-04-01
Add browsers.openDefaultBrowser without URL, implements RFC-6694

juancarlospaco:
- Add `browsers.openDefaultBrowser()` without URL.
- Implements IETF RFC-6694 Section-3 https://tools.ietf.org/html/rfc6694#section-3
- Code is literally the same, moved into `template` to DRY.
- Can *not* use empty string because https://github.com/nim-lang/Nim/pull/13818#issue-396473868
- Documentation with example and links.

### DIFF
--- a/lib/pure/browsers.nim
+++ b/lib/pure/browsers.nim
@@ -13,6 +13,7 @@
 ## Unstable API.
 
 import strutils
+include "system/inclrtl"
 
 when defined(windows):
   import winlean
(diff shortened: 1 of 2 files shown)


lib/pure/browsers.nim
```diff
@@ -54,3 +43,38 @@ proc openDefaultBrowser*(url: string) =
...
+  ##
+  ## Under Windows, ``ShellExecute`` is used. Under Mac OS X the ``open``
+  ## command is used. Under Unix, it is checked if ``xdg-open`` exists and
+  ## used if it does. Otherwise the environment variable ``BROWSER`` is
+  ## used to determine the default browser to use.
+  ##
+  ## This proc doesn't raise an exception on error, beware.
+  ##
+  ## .. code-block:: nim
+  ##   block: openDefaultBrowser("https://nim-lang.org")
```
narimiran, line 59: Why does this example have to be inside of a `block`?

(This is how it looks rendered: https://nim-lang.github.io/Nim/browsers.html#openDefaultBrowser%2Cstring)

PR #13845, opened 2020-04-02
feature/containsIt

decanus:
Implements the function `containsIt` allows to check whether an iterator contains items matching the predicate.

### DIFF
--- a/lib/pure/collections/sequtils.nim
+++ b/lib/pure/collections/sequtils.nim
@@ -572,7 +572,7 @@ template filterIt*(s, pred: untyped): untyped =
   result
 
 since (1, 1):
-  template containsIt*(s, pred: untyped): Bool =
+  template containsIt*(s, pred: untyped): bool =
     ## Checks if an item fulfills the predicate.
     ##
     ## The predicate needs to be an expression using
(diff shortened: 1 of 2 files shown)


lib/pure/collections/sequtils.nim
```diff
@@ -571,6 +571,26 @@ template filterIt*(s, pred: untyped): untyped =
     if pred: result.add(it)
   result
 
+since (1, 1):
+  template containsIt*(s, pred: untyped): bool =
```
metagn, line 575: This template doesn't work because it has `return`. Also anyIt exists https://nim-lang.org/docs/sequtils.html#anyIt.t%2Cuntyped%2Cuntyped

lib/pure/collections/sequtils.nim
```diff
@@ -571,6 +571,26 @@ template filterIt*(s, pred: untyped): untyped =
...
+      let numbers = @[-3, -2, -1, 0, 1, 2, 3, 4, 5, 6]
+      iterator iota(n: int): int =
+        for i in 0..<n: yield i
+      assert numbers.containsIt(it < 0)
+      assert containsIt(iota(10), it < 2)
+      assert not numbers.containsIt(it > 6)
+
+    for it {.inject.} in s:
+      if not pred:
+        return false
```
narimiran, line 591: You can check your `runnableExamples` by running `nim doc lib/pure/collections/sequtils.nim`. If you do that, you'll see that your examples don't work because of this line here; you get `Error: 'return' not allowed here`.

narimiran: But even if you fix that, the logic seems wrong.

Issue #13866, opened 2020-04-04
asynchttpserver documentation example causes SIGSEGV


supakeen:
The documentation example at https://nim-lang.org/docs/asynchttpserver.html causes a SIGSEGV when ran:

### Example
```nim
import asynchttpserver, asyncdispatch

var server = newAsyncHttpServer()
proc cb(req: Request) {.async.} =
  await req.respond(Http200, "Hello World")

waitFor server.serve(Port(8080), cb)
```

### Current Output
After performing ``curl http://localhost:8080/``.
```
Traceback (most recent call last)
/home/user/test.nim(7)   test
/home/user/bin/nim-1.2.0/lib/pure/asyncdispatch.nim(1886) waitFor
/home/user/bin/nim-1.2.0/lib/pure/asyncdispatch.nim(1576) poll
/home/user/bin/nim-1.2.0/lib/pure/asyncdispatch.nim(1340) runOnce
/home/user/bin/nim-1.2.0/lib/pure/asyncdispatch.nim(210) processPendingCallbacks
/home/user/bin/nim-1.2.0/lib/pure/asyncmacro.nim(37) processRequestNimAsyncContinue
/home/user/bin/nim-1.2.0/lib/pure/asynchttpserver.nim(258) processRequestIter
/home/user/bin/nim-1.2.0/lib/pure/asyncmacro.nim(319) cb
/home/user/bin/nim-1.2.0/lib/pure/asyncmacro.nim(34) cbNimAsyncContinue
/home/user/test.nim(5)   cbIter
/home/user/bin/nim-1.2.0/lib/pure/asynchttpserver.nim(103) respond
/home/user/bin/nim-1.2.0/lib/pure/httpcore.nim(180) hasKey
SIGSEGV: Illegal storage access. (Attempt to read from nil?)
```

### Expected Output
No exception.

### Possible Solution

* After a chat on IRC Yardanico indicated that this might have been introduced by: https://github.com/nim-lang/Nim/pull/13846/

### Additional Information

* This seems to be a regression.

```
$ nim -v
Nim Compiler Version 1.2.0
```

Yardanico:
Simplest way to fix it is to add another check in respond in asynchttpclient, although this might be simplified (but I don't know how :P)
```nim
  var msg = "HTTP/1.1 " & $code & "\c\L"

  if headers != nil:
    msg.addHeaders(headers)
    
  if headers != nil and not headers.hasKey("Content-Length"):
[...]

PR #13867, opened 2020-04-04
Add isNil check to custom Content-Length.

supakeen:
Related to #13866, adds isNil check to headers to prevent SIGSEGV.


ghost:
I guess this needs to be backported, cc @narimiran, just add "[backport]" prefix to the commit right?

PR #13875, opened 2020-04-05
Documentation add DrNim to 'Tools available with Nim'

juancarlospaco:
- Add DrNim to `Tools available with Nim`, Documentation only.

### DIFF
--- a/doc/tools.rst
+++ b/doc/tools.rst
@@ -38,5 +38,5 @@ The standard distribution ships with the following tools:
     has logging, can generate HTML reports, skip tests from a file and more,
     so can be useful to run your tests, even the most complex ones.
 
-- | `DrNim <https://nim-lang.org/docs/drnim.html>`_
+- | `DrNim <drnim.html>`_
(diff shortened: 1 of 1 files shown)


Araq:
DrNim isn't documented nor announced for a reason. ;-) This has to wait.

juancarlospaco:
Reopen considering is now part of stable official documentation for a while now (?)
https://nim-lang.org/docs/theindex.html & https://nim-lang.org/docs/drnim.html

Araq:
It should be removed from the documentation instead.

juancarlospaco:
@Araq 
But been out for a while now, theres even blog post related, if you "hide" it from public looks weird or even suspicious...,
I really dont understand the idea of intentionally hiding Documentation from potential users and companies wanting to use Nim,
you can just add a big warning "This is not production ready" or similar  ...and most people will understand it.
:thinking:

Araq:
It's not that it isn't production ready, it's really unfinished. :-(

PR #13883, opened 2020-04-05
Add barebones asynchttpserver tests

supakeen:
The regression in https://github.com/nim-lang/Nim/issues/13866 showed that there were no tests for ``asynchttpserver`` so I took a stab at some barebones tests which at least try to:

- Perform a request
- Perform a request with a different status code
- Perform a request with a custom headers response.
- Perform a request with a custom content-length header in the response.

While the ``request`` proc in the tests could be simplified with another template I think future test cases might want to perform multiple requests.

Since this is my first 'serious' Nim code aside from the PR for the abovementioned ticket (a one line change) I'd like some help getting it up to par.

### DIFF
--- a/tests/stdlib/tasynchttpserver.nim
+++ b/tests/stdlib/tasynchttpserver.nim
@@ -1,0 +1,119 @@
+discard """
+  cmd: "nim c --threads:on $file"
+  exitcode: 0
+  output: "OK"
+  disabled: false
+"""
+
+import strutils
+from net import TimeoutError
+
+import httpclient, asynchttpserver, asyncdispatch, asyncfutures
+
+template runTest(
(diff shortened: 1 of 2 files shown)


tests/stdlib/tasynchttpserver.nim
```diff
@@ -0,0 +1,119 @@
+discard """
+  cmd: "nim c --threads:on $file"
```
ghost, line 2: Is threads:on needed here though? I don't see you using threads anywhere in the file

Araq: Still it's a good idea.

Araq:
I like it. We'll later see if it's a "flaky" test or not.

Issue #9942, opened 2018-12-12
Generic static int parameters with default values cannot be used as an array size indicator


zah:
The following valid code currently fails to compile:

```nim
const
  pageSize = 4096

type
  Foo*[T; size: static[int] = pageSize] = object
    buffer: array[size, byte] # Error: ordinal type expected

var f1: Foo[int]
var f2: Foo[int, 1024]

```

Omitting `static[int]` in the signature of `Foo` solves the problem:

```nim
  Foo*[T; size = pageSize] = object
```

c-blake:
Since @bung87 seems to be working on this, a related problem is that, though it compiles as @zah mentions, the default value is ignored even if you drop the `static[int]`:
```Nim
const pageSize = 4096
type Foo*[T; size = pageSize] = object
  buffer: array[size, byte]
var a: Foo[int]      ; assert a.sizeof == 4096 # as expected
var b: Foo[int, 1024]; assert b.sizeof == 4096 # as, erm, UNEXPECTED
```
(Tested on Nim 0.20.2 and nim-devel.)  Same condition fails (by which I mean assert to wrong value succeeds) if you replace `.sizeof` with `.buffer.len` (i.e. compile-time with run-time.)  Compiling but not using the default value is, well, less than ideal.

***Meanwhile***, for "type values" (instead of ints like `pageSize`), default values do appear to work in types like the `Foo[T]` above, but for a regular old generic `proc` they seem to fail:
```Nim
proc f[A, B = int](x: A, y: int) =
  when B is int: echo "hi"
  echo x, y
f[float,int](1.0, 2)    # fully specified instantiation..
f[float,float](3.0, 4)  #..seems to work ok.(just 1 "hi")
#f(5.0, 6) # 'f[float]'  # cannot instantiate: 'B'
#f[float](5.0, 6)        # 'f[float]'; got 1 typeof(s) but expected 2
```
Above also fails to compile (@ first instantiation) if you edit `B = int` to `B: SomeInteger = int` which is maybe a 3rd distinct bug on this general topic.

PR #13890, opened 2020-04-06
[WIP, help needed] introduce BTree-based Tables

narimiran:
This is a WIP pull request which introduces `Tables` implemented via BTrees.

This is based on [my sorta package](https://github.com/narimiran/sorta), and it tries to mimic current API of the existing hash-based `Tables`.
BTrees allow for sorted order of keys, which means we can now have few new iterators: `keysFrom`, `keysBetween`, and the same for `values` and `pairs`(`valuesFrom`, etc.) — which will iterate from a given key to the end of the table or until another provided key, respectively. (See [here](https://github.com/nim-lang/Nim/compare/miran-btree-tables?expand=1#diff-835aaafa7c680f0108b4e421628c9498R691-R702))

This is in a very early state and it needs to be optimized/improved and tested more thoroughly.
Ideally, somebody else (@c-blake maybe?) would take over and help with this....


c-blake:
I don't see much evolution of this impl from the early `sorta` release and I did raise a couple issues there to try to guide you toward a more general kind of B-Tree.  The way you want to do a cursor is a "tree path"..basically just a `seq[tuple[Node,ix]]` filled as search from the root to some key.

Along with "rank augmentation" (Knuth V3 shows how to do this for AVL trees) this lets you do things like `proc insert*[T](x: var seq[T], item: sink T, i = 0.Natural)` and `proc delete*[T](x: var seq[T], i: Natural)`, but with `BTree[T]` instead of `seq[T]`.  It also lets a B-tree function as an "order statistics" collection (like `adix/bist.nim` but for more general keys than small integers)..kind of a "HistoTree" instead of a histogram, if you will.  It jus generally makes the API more like a "sorted seq" for anything you might want to do, but optimized via hierarchy.

I have a pile of C code written in my own C-generic/templating that "does Btrees right" that I have been meaning to port into Nim/`adix`.  I got held up with CI tests failing that I could not reproduce locally for that `hashWangYi1` switch.  If you guys help with that (is there some Azure image I can download locally so I can test in that environment without git-pushing a bunch of random tweaks?  Or even more help would be figuring out why tests are failing) then this week I can do a hard push to do my port and next week we can maybe try to integrate into stdlib.  I think that would be a better starting basis than what you have here (no insult intended..BTrees have many corner cases, generalization opportunities, and specialization opportunities that are all a pain to get right and not "taught well", IMO).

c-blake:
I should also say this ranked tree approach (or a sorted seq for that matter) create a "multi-path" kind of indexing where one single kind of `[]` cannot work (at least if `int` keys are possible).  Because only ranked trees really support `[position]` and `[key]`, and because they are also enough harder to implement than hashing/arrays, they are never "early" in the series of data structures new prog langs support. Because of lateness, ignorance, and a desire to have associative structures "look like" positional access structures like `seq`, this "multipath interface" detail is usually overlooked.  Honestly, that ranked feature of trees is usually even "late" in tree impls because it's hard enough to get the basic stuff right, but efficient operation availability not implementation complexity should drive these choices.

To my mind the most natural thing would be to have `x{key}` mean keyed-access and `x[key]` mean positional access.  It might even be possible to do this in a backward compatible way in Nim where `{}` means the same thing as `[]` for things like hash tables where access-by-position is not well defined.  This would make BTree tables less "drop-in" compatible..but also perhaps highlight this less-well taught flexibility.  With the above proposal, using `btree[]` means you're using it like a `seq` while using `btree{}` means using it like a hash table.  The converse approach would be to keep `[]` as keyed access and have `{}` or `[[pos]]` or something be for positional access when both are supported.  That's a little confusing, too, since `specialDataStructure[[pos]]` becomes equivalent to `seq[T] [pos]`.  None of that is a big deal in a non-operator framework where you might just have a `seekByKey` and `seekByRank` procs that "fill in a tree path/cursor", but it probably deserves a little thought at this point in the process, especially by @Araq, because I think what is least confusing (in an operator framework) may be subjective.  I kind of think adding `{}` for hash tables and using `{}` for key and `[]` for position makes the most sense.

I should also mention that while the overhead for these `rank` procs is large for binary trees (1 counter space and ~lg N counter updates per link update), it is actually small for B-trees (just 1 counter per big Node, and log_M updates).  This is why `seekByRank` or the equivalent is so much more common in B-tree libraries/APIs than binary tree libraries.  Even so, the overhead is not *zero* and so some will want to be able to turn it off, and once they do things are no longer multi-path which means that *only* `x{key}` might be available.  That would be fine in my above proposal, but some will be confused `x[key]` doesn't exist and need to have all this re-explained.

`{}`/`[]` is a kind of nice, explicit but terse feature, but never caught on.  The prog.lang origin of this not existing earlier is clearly "fixed syntax/extensible semantics" operator overloading from the 1980s C++/Python/etc. zeitgeist.  One bonus of Nim's "define your own operators" to me has always seemed fixing this old error by having different things look different.  An optimized "key only" B-tree won't have `[]`, but that's good since it means your instantiation and usage don't match.  Rank-only/position-only b-trees won't have `{}` - also good since there isn't even a "key type".

Anyway, this particular issue is probably broadly interesting/opinionated enough that it warrants an RFC -- "`Table`s should grow `{}` for keyed access to match a split `{}`/`[]` for trees".  The hard part in all this is just broad programmer ignorance of this subtlety.  We could ditch operators and force seekByRank/seekByKey, but that would also be unpopular.  Maybe I should have brought all this up 5 years ago.  I'm not sure the questions/answers/problems would have been any different then though.

Vindaar:
Sorry for offtopic, but:

@c-blake You have so much knowledge about stuff I don't even know I know nothing about (that is as someone who didn't study IT). You should write a book, I'd definitely read it. :)

c-blake:
Thanks.  Maybe someday.  I think many books cover most of this stuff.  Knuth The Art Of Computer Programming, especially Volume3: Sorting And Searching, covered almost everything I talk about before most of us were born.  Courses, blogs, "community/received" wisdom just "drops the ball", but I'm sure this is not unique to data structures.  B-trees in particular had this whole phase in the 1970s/1980s where if you were good and did a nice one you could start a company.  So, some wisdom was lost to capitalism.

If you want one more "mind bender" that is less relevant in the days of giant RAMs and NVMe's, the "node" of a B-tree could be a Red-Black Tree/AVL tree/Skip list.  So, if you wanted a 1000s+-way B-tree you could keep operations on those nodes efficient.  In truth, latency/bandwidth ratios are such that you probably don't want more than nodes of size about 100 for the trees being proposed in the stdlib.  So, this is only really relevant with spinning rust backing stores.  Even so, Winchester disks are still in wide use.

So, the "ideal implementation" of a B-tree would be abstract over node representation.  That might be more advanced than we get to initially and would not really impact much in the way of interfaces to client code in Nim (or elsewhere).  Someone once told me that Sybase used Skip lists for their nodes, but being commercial/closed source I never got confirmation (nor did I inspect-bytes-on-the-disk/reverse engineer).  That is mostly for "insert/delete" heavy workloads as binary search on a node array is already as fast as trees.  It's only the memmove/shifting around that scales badly.

c-blake:
(Or, of course, one could do a "nested B-tree" where the branching ratio for the nodes is small while the nodes themselves hold a lot of elements.)

c-blake:
Oh, I did bring up multipath access 2.5 years ago.  Maybe earlier too.  The forum search only goes back so far. https://forum.nim-lang.org/t/3358#21171  I didn't highlight obvious follow-on operator questions, and often I wrote so much that some details get lost. (EDIT like in that forum case equivalence between dynamic quantiles and keyless seq-like access.)

c-blake:
Ok.  I opened an RFC on syntax questions for this.  https://github.com/nim-lang/RFCs/issues/207  Let's see what people say.

c-blake:
I am back to working on this now, btw.  I had been thinking of using generics like `BTree[K,V,M=32,Rank=true]` where  `K is void` would mean `seq`-like-only-style although one giant template or two dozen small ones with various default parameters can also work.  There seem to be several open issues related to generic parameter defaults (e.g. were I to try to put `M: static[int]=32` in that `M` slot [it would break](https://github.com/nim-lang/Nim/issues/9942)).  There may also be other params we wind up adding.  Any opinions?  I am kind of leaning towards one giant template.

Araq:
I would also go for the "gigant template" solution. :-)

Araq:
This is now in fusion.

Issue #13889, opened 2020-04-06
Critical: 1 completed Future, multiple await: Only 1 await will be awakened (the last one)


eliezedeck:
When having multiple `await` on a **single future**, when the future is `complete`d, only the *last* `await` will be awakened/triggered.

### Example

https://play.nim-lang.org/#ix=2gNI

```nim
import asyncdispatch

var
  promise = newFuture[bool]()

proc believers(name: string) {.async.} =
  let v = await promise
  echo "believer " & name & " is saved:" & $v

asyncCheck believers("Foo")
asyncCheck believers("Bar")
asyncCheck believers("Baz")

proc savior() {.async.} =
  await sleepAsync(1500)
  complete(promise, true)
  await sleepAsync(1500) # give enough time to see who was saved

waitFor(savior())
```

### Current Output
```
believer Baz is saved:true
```

### Expected Output
```
believer Foo is saved:true
believer Bar is saved:true
believer Baz is saved:true
```

### Possible Solution

* As suggested by @dom96 in the chatroom, it has something to do about using `addCallback` instead of `callback=`. That is: the last call to `await` overwrites the other that were already there (or the last one that was there)
* I can contribute to a fix/PR, but I need direction as to where in the source I should be looking at

### Additional Information

* All Nim versions are plagued by this bug
* This issue is blocking my project which depends on the proper behavior of future completion
* I strongly believe that other projects are also affected by this, and a fix should be backported, at least to v1.0.x

```
$ nim -v
Nim Compiler Version 1.2.0 [MacOSX: amd64]
Compiled at 2020-04-03
Copyright (c) 2006-2020 by Andreas Rumpf

active boot switches: -d:release
```

eliezedeck:
I've tested my above PR and it is working great. My project is now working nicely as intended.

@Araq please merge. Thanks @dom96 for the pointer.

PR #13893, opened 2020-04-06
Fix multiple await on one future (https://github.com/nim-lang/Nim/issues/13889)

eliezedeck:
This fix the critical bug on `async` / `await`, reported at https://github.com/nim-lang/Nim/issues/13889.

In addition to this, please, show the compile time. It will help newcomers to trying to contribute to Nim easily see that their compiled Nim version is indeed being used.

### DIFF
--- a/compiler/commands.nim
+++ b/compiler/commands.nim
@@ -49,7 +49,7 @@ type
 
 const
   HelpMessage = "Nim Compiler Version $1 [$2: $3]\n" &
-      "Compiled at $4\n" &
+      "Compiled at $4 - $5\n" &
       "Copyright (c) 2006-" & copyrightYear & " by Andreas Rumpf\n"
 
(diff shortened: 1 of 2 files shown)


eliezedeck:
I believe these CI issues have something to do with the tooling (nodejs to be exact), and nothing to do with the actual PR.

dom96 (changes requested):
This fix makes sense to me, although I haven't thought about the possible repercussions of this. Can you add a test to make sure this continues to work? Also, please run `./koch tests cat async` (or whatever the command is these days) to double check that nothing broke (I don't trust the Nim CI nowadays).

Also, please don't mix two different things into a single PR, it's always better to have two PRs when you're making two completely unrelated changes. Can you split this up?

Clyybber:
Cherrypicked and added testcase in https://github.com/nim-lang/Nim/pull/13896

PR #13896, opened 2020-04-06
Fix #13889 with testcase

### LINKED ISSUE #13889: Critical: 1 completed Future, multiple await: Only 1 await will be awakened (the last one)

Clyybber:


Clyybber:
Test failures unrelated (nodejs doesn't install for some reason)

dom96:
I'm curious if you tested this without the `sleepAsync`'s, reading this again it seems to me like they are not required at all.

Clyybber:
Yeah they don't seem to be required, but we might as well test them here.

PR #13897, opened 2020-04-06
Step2: fixes #13781, fixes #13805 

### LINKED ISSUE #13781: gc:arc cannot fully support threadpool with FlowVar

### LINKED ISSUE #13805: arc & threads: doesn't pass the reference forwarding test

cooldome:
Spawn for gc arc fixes


compiler/injectdestructors.nim
```diff
@@ -791,6 +791,12 @@ proc p(n: PNode; c: var Con; mode: ProcessMode): PNode =
       if mode == normal and isRefConstr:
         result = ensureDestruction(result, c)
     of nkCallKinds:
+      let inSpawn = c.inSpawn
+      if n[0].kind == nkSym and n[0].sym.magic == mSpawn:
+        c.inSpawn.inc
+      elif c.inSpawn > 0:
```
Araq, line 797: Suspicious logic. Why not at the end of the ``of nkCallkinds`` handling code like

```nim

if n[0].kind == nkSym and n[0].sym.magic == mSpawn:
  c.inSpawn.dec
```

?

cooldome: The idea is that spawn can be written in the following way:
```nim
spawn myproc(call1(), (let x = call2(); call3(x, 0))
```
I want  `c.inSpawn` to be greater than zero only for `myproc` call but not for `call2` and `call3` in example above.

cooldome:
I can't replicate CI test failures. I have tried 3 machines, 2 platforms and 3 compilers. All tests pass.
Don't know how to progress further

Araq:
Well BSD says:

Category: arc
Name: tests/arc/tthread.nim C
Action: run
Result: reExitcodesDiffer
-------- Expected -------
exitcode: 0
--------- Given  --------
exitcode: 1

Output:

-------------------------

cooldome:
Ready for review

lib/pure/concurrency/threadpool.nim
```diff
@@ -234,43 +236,42 @@ proc awaitAndThen*[T](fv: FlowVar[T]; action: proc (x: T) {.closure.}) =
...
+    blockUntil(fv[])
     result = fv.blob
-  finished(fv)
+    finished(fv[])
+
+else:
+  proc `^`*[T](fv: FlowVar[T]): T =
+    ## Blocks until the value is available and then returns this value.
+    blockUntil(fv[])
+    when T is string or T is seq or T is ref:
```
metagn: Why not get rid of the 2 different procs in each when/else branch and put `when defined(nimV2) and (T is string or T is seq or T is ref):` here?

cooldome: done

PR #13911, opened 2020-04-07
fix #13894, httpclient hang on Http204

### LINKED ISSUE #13894: httpclient hangs if it recieves an HTTP 204 (No Content)
If you use the httpclient in the stdlib to make a call to a service which results in a HTTP 204 being returned, the 204 status code is ignored, and the library still attempts to read data from the wire, in my case it hangs at line 696 of `httpclient.nim`:
```nim
let recvLen = await client.recvFull(4000, client.timeout, true)
```

HTTP 204 explicitly says 'No Content' so no attempts to process a body should happen after seeing that status code.

This bug appears to exist in 1.2 and 1.0.6 (only versions I've tested).

### Possible Solution

Line 810 of httpclient.nim should be changed to read:
```nim
if getBody and result.status != Http204:
```

```
$ nim -v
```

narimiran:


dom96 (approved):
LGTM.

A great improvement would be a testing framework for httpclient, we could then add tests more easily for these kinds of cases.

PR #13925, opened 2020-04-08
fix compiler crash in isUnresolvedSym [backport]

cooldome:
I have experienced a compiler crash today in isUnresolvedSym.
I have quickly checked what the sym  was, it was a module sym. 
Module doesn't have a type.


Araq:
Needs the ``[backport]`` tag.

Issue #11518, opened 2019-06-17
`func` should be allowed in type descriptions


khogeland:
If you're using `func`, you're probably writing functional code that is frequently going to have functors as arguments or fields, where you now have to use `proc(): foo {.noSideEffect.}`. The utility seems lost if you still need to use the pragma in type descriptions.

Is there a better fix for #10838 that would allow `func` in type descriptions? If not, is `func` really still useful?

Araq:
Are you really asking if proc/func declarations are more common than callback type definitions? Yes, they are!

khogeland:
I didn't say that. Supporting `func` in type descriptions makes it easier to write functional code where higher-order functions are common.

Araq:
Alright.

Grinshpon:
Has there been any progress on this?

Araq:
No. But the workaround is ``type P = proc (args...) {.noSideEffect.}``

Grinshpon:
I made a small feature request (#13947) and a pr (#13948) that adds two macro operators to the sugar module in the standard library, which allows you to easily create side effect free function types and lambdas. It's a convenient workaround I think.

Issue #2716, opened 2015-05-14
The `func` keyword for "pure" functions


ozra:
As I'm investigating this interesting language, somewhere, on some page, I saw a mention that a `func` keyword as more convenient syntax for declaring a procedure 'purely functional' was considered.
I couldn't find a issure refering to it, so I add it here, since I find it an essential feature.

Nim is ofcourse a systems language (which is what I'm after!), still purely functional coding is proven to be safer, so enabling and pushing for purer functionalistic coding style while still allowing everything would be a good thing.

Also, many backend (out of Nim perspective) compilers (C/C++) has pragmas and annotations for "pure" or "const" ("really pure") functions which can be leveraged and hints and helps the backend compiler generate even faster code. 

So:
- Already on the table for Nim (as far as I understand)
- Expresses coders intent
- Intent verifiable
- Possibly increased performance in final binary.

reactormonk:
So what would be the difference to `proc {.noSideEffect.}`?

ozra:
None. Just more sugar on the porridge. Cleaner syntax, promoting its use.

reactormonk:
Ok, feel free to implement it. Lexer for the keyword, parser for the pragma stuff.

dom96:
Araq mentioned that he was considering making the 'func' keyword something that could be library-defined. So by default it would mean noSideEffects but a library could potentially change that meaning.

ozra:
@reactormonk cool, I'll just have to get more aquainted with Nim first, gonna code a couple of utils for myself to get warm.
@dom96 - Some thoughts: I think `func` has a pretty obvious meaning, so instead of over generalizing the language, perhaps it's good to cut some things in stone.
Whether a piece of code is procedural or purely functional is of great value both to the code reviewer / programmer and the compiler to know. So the semantic value is high imo.

- When it comes to backend optimizability, gcc has the pragmas `pure` and `const` for "functions"  - I think opting for the gcc-`const`-pragma-def would be a good thing (which essentially boils down to "purely pure". If I remember correctly the `pure` pragma still allows globals, but not mutating arguments. `const` allows no side effects. 
- For readability and intent: I think seeing `func` lends naturally, in relation to having the `proc`-keyword, that it refers to a pure function - having behavior redefined by libraries would just be confusing, would be better to use another keyword with a more open meaning for that perhaps?
[...]

PR #13948, opened 2020-04-10
Add macros |=> and |-> to create procedures and procedure types with noSideEffect pragma…

### LINKED ISSUE #13947: Adding macros for creating anonymous `func`s and `func` types in sugar
### Summary

The `->` and `=>` macros can only be used to create procedures with possible side effects. This proposal adds the corresponding `|->` and `|=>` macros that perform a similar task, but add the `{.noSideEffect.}` pragma to enable easier functional programming capabilities.

### Description

While Nim is not a functional language, it borrows functional concepts, such as the `func` which is a `proc` that doesn't allow mutation of any global state (at least that isn't referenced in the function's parameters with a `var` or `ref`). It also supports first class functions, passing in functions as arguments and creating anonymous/lambda functions.

Using first class functions is made easier thanks to the sugar library, with the `->` and `=>` macros. However, these macros will describe a procedure that can have side effects. If you want to bridge first class functions with `func`, you cannot use the macros. Adding the `|->` and `|=>` macros bridges the functional language concepts and makes it convenient and easy to write safer and functional code.

### Alternatives

Technically no new features are introduced in this request. It is still possible to do things the verbose way, but just like `func` is convenient sugar for the `{.noSideEffect.}` pragma, so is this simply convenient sugar.

### Additional Information

For the syntax, I chose to prepend `|` to the two existing operators to create the new operators, as a sort of pun. In many functional contexts, the `|` symbol is called a "guard", so I thought that `|=>` and `|->` are like `=>` and `->` but they guard against side effects. 

As an example, say I'm working on a library that can have the user to provide a custom distance metric, for pathfinding or nearest neighbor search, etc. The distance metric should be a function that simply returns a distance, and should not have any side effects. To enforce that, I can simply write `type DistMetric = (Point, Point) |-> float`

Grinshpon:
Resolves #13947

Reposting the small feature request:

### Summary

The `->` and `=>` macros can only be used to create procedures with possible side effects. This proposal adds the corresponding `|->` and `|=>` macros that perform a similar task, but add the `{.noSideEffect.}` pragma to enable easier functional programming capabilities.

### Description

While Nim is not a functional language, it borrows functional concepts, such as the `func` which is a `proc` that doesn't allow mutation of any global state (at least that isn't referenced in the function's parameters with a `var` or `ref`). It also supports first class functions, passing in functions as arguments and creating anonymous/lambda functions.

Using first class functions is made easier thanks to the sugar library, with the `->` and `=>` macros. However, these macros will describe a procedure that can have side effects. If you want to bridge first class functions with `func`, you cannot use the macros. Adding the `|->` and `|=>` macros bridges the functional language concepts and makes it convenient and easy to write safer and functional code.

### Alternatives

Technically no new features are introduced in this request. It is still possible to do things the verbose way, but just like `func` is convenient sugar for the `{.noSideEffect.}` pragma, so is this simply convenient sugar.

### Additional Information

For the syntax, I chose to prepend `|` to the two existing operators to create the new operators, as a sort of pun. In many functional contexts, the `|` symbol is called a "guard", so I thought that `|=>` and `|->` are like `=>` and `->` but they guard against side effects. 

As an example, say I'm working on a library that can have the user to provide a custom distance metric, for pathfinding or nearest neighbor search, etc. The distance metric should be a function that simply returns a distance, and should not have any side effects. To enforce that, I can simply write `type DistMetric = (Point, Point) |-> float`

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,7 @@
 
 ## Standard library additions and changes
 
+- Add `|=>` and `|->` macro operators to the sugar module which behave like `=>` and `->` respectively, but guarding against side effects.
 
 ## Language changes
 
@@ -14,4 +15,3 @@
 - The `define` and `undef` pragmas have been de-deprecated.
 
 ## Tool changes
-
--- a/lib/pure/sugar.nim
+++ b/lib/pure/sugar.nim
@@ -15,7 +15,7 @@ include system/inclrtl
 import macros
 import typetraits
 
-proc createProcType(p, b: NimNode): NimNode {.compileTime.} =
+proc createProcPrototype(p, b: NimNode): NimNode {.compileTime.} =
   #echo treeRepr(p)
   #echo treeRepr(b)
   result = newNimNode(nnkProcTy)
@@ -45,23 +45,24 @@ proc createProcType(p, b: NimNode): NimNode {.compileTime.} =
     formalParams.add identDefs
 
   result.add formalParams
-  result.add newEmptyNode()
+  #result.add newEmptyNode()
   #echo(treeRepr(result))
   #echo(result.toStrLit())
 
-macro `=>`*(p, b: untyped): untyped =
-  ## Syntax sugar for anonymous procedures.
-  ##
-  ## .. code-block:: nim
-  ##
-  ##   proc passTwoAndTwo(f: (int, int) -> int): int =
-  ##     f(2, 2)
-  ##
-  ##   passTwoAndTwo((x, y) => x + y) # 4
+proc createProcType(p, b: NimNode): NimNode {.compileTime.} =
+  result = createProcPrototype(p,b)
+  result.add newEmptyNode()
 
-  #echo treeRepr(p)
-  #echo(treeRepr(b))
-  var params: seq[NimNode] = @[newIdentNode("auto")]
+proc createFuncType(p, b: NimNode): NimNode {.compileTime.} =
+  var pragma = newNimNode(nnkPragma)
+  pragma.add newIdentNode("noSideEffect")
+
+  result = createProcPrototype(p,b)
+  result.add pragma
+  #result.add newEmptyNode()
+
+proc createParams(p: NimNode): seq[NimNode] {.compileTime.} =
+  result = @[newIdentNode("auto")]
 
   case p.kind
   of nnkPar, nnkTupleConstr:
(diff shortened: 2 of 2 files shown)


juancarlospaco:
https://github.com/nim-lang/Nim/pull/13770#issue-394157402  :shrug:

Grinshpon:
@juancarlospaco it's a shame your pr was rejected but I'm hoping I can make a convincing argument that the functional programming concepts in Nim should be bridged together, not just for their own sake but because there exist many use cases, and these additions will lead to safer code being written.

juancarlospaco:
This file has lots of commented out `echo` as abandoned debug code that should be removed
(not your code).

Grinshpon:
I noticed those too, but decided to leave them because imo that would be a separate issue

zetashift:
FWIW I like this, (somewhat like ==>) better, but just the comment in the previous issue mentioned: it's quite specific. 
Maybe one could get best of both worlds and PR this to something like a library like this one: https://github.com/vegansk/nimfp ?

Grinshpon:
It's as specific as `func` is, which I feel justifies this being in the language, considering `func` is in the language.

zetashift:
@Grinshpon but `func` didn't get in as easy, and the post also said it's mostly for `{.noSideEffects.}` not to enforce FP(though it's a consequence): https://forum.nim-lang.org/t/657 and https://github.com/nim-lang/Nim/issues/2716

It's a nice addition to sugar but I think it's fine if it's added in a FP library first then see how it's used. I think one who likes FP should get something like nimfp or https://github.com/alehander92/gara recommended by the community anyway. 
They ease the entry to Nim for them by that much and depending on their usage stdlib additions can be made.

edit: also an addition is is if you add it to a library I think you can just add it as `=>` which makes it a lot cleaner.

Grinshpon:
One problem with putting these in an external library is that I'd essentially have to copy-paste some of the code from sugar.nim, because the implementation for the new macros are 99% identical to the old ones apart from adding the pragma at the end (you can see in my pr that I moved the common code between them to a separate function to adhere to DRY). Since they're so similar in implementation and fulfill a purpose introduced by the inclusion of `func`, I believe it's appropriate for this to be part of the main language's sugar module.

If this is rejected then my next plan is supporting an external library like nimfp, but that'd basically require reinventing the wheel, since sugar already has the groundwork and most of the implementation.

Araq:
1. ``|=>`` is not intuitive.
2. We should support ``type F = func (...)`` first IMO. Yeah, it's a parser/grammar/sem'check addition, but it's so natural that nobody will complain.
3. "I cannot use external Nimble packages" is not how we grow the stdlib, we only add non-trivial helpers that *are* required for helping us design a better stdlib. (Like ``dup`` -- with its help we can trim down the stdlib, that's why it's not in a Nimble package.)

Grinshpon:
@Araq 

1. The syntax can be changed to long arrows (`==>` and `-->`) which, I admit, seem more intuitive than what I came up with.

2. That's a good point, I would like to see #11518 be resolved (and help out in any way I can)

3. I never said I can't use external packages, but rather that it'd be like reinventing the wheel. Like I said, the sugar module already has 99% of the groundwork and implementation for making these work, so while implementing them is trivial, I believe the addition to the stdlib would be a significant improvement for the users.

Araq:
2. Search for ``nkProcTy`` inside the compiler, add ``nkFuncTy`` (**append** to ``ast.TNodeKind`` for binary compat) and mimic the logic.

Araq:
Sorry, rejected for the reason that I cannot imagine an operator symbol that would not be overly cryptic. Also, let's make ``func`` work first.

PR #13949, opened 2020-04-10
Times refactorings

GULPF:
Some cleanups in preparation for future improvements...

- Reorder the procs so they are grouped by subject and to avoid forward declarations
- Remove the unexported type `TimesMutableTypes` and replace the procs that used it. It looks like this type was just used as a shortcut to avoid writing a bunch of similar looking concrete procs, but the procs that used it ended up looking confusing and where not correct anyway. Example:
```nim
proc `+=`*[T, U: TimesMutableTypes](a: var T, b: U) =
  a = a + b
```
This proc will accept invalid usage like `datetime += datetime`. Of course it will fail anyway since there won't be a `+(DateTime, DateTime)`, but what's the point of `TimesMutableTypes` if it doesn't prevent incorrect usage. I've replaced these procs with concrete ones for the cases that are allowed.

Is there a policy for when deprecated procs can be removed? The times module has a bunch of procs that has been deprecated for over 2 years (v0.18) and it would be nice to get rid of them.

### DIFF
--- a/lib/pure/times.nim
+++ b/lib/pure/times.nim
@@ -283,19 +283,6 @@ type
     dSat = "Saturday"
     dSun = "Sunday"
 
-when defined(nimHasStyleChecks):
-  {.push styleChecks: off.}
-
-type
-  DateTimeLocale* = object
-    MMM*: array[mJan..mDec, string]
-    MMMM*: array[mJan..mDec, string]
-    ddd*: array[dMon..dSun, string]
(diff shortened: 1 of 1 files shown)


juancarlospaco:
+1 to remove all "2 years deprecated" and older proc.

Araq:
The policy is not clear but IMO "deprecated for 2 versions or longer --> remove" is fine.

PR #13950, opened 2020-04-10
added extended msg for failed library loads w/ incorrect DLL formats

awr1:
Common issue that I've seen in the Nim IRC is the `could not load:` error with DLLs with the wrong architecture, hopefully this pushes people in the right direction.

### DIFF
--- a/lib/system/dyncalls.nim
+++ b/lib/system/dyncalls.nim
@@ -19,18 +19,31 @@ const
 
 proc nimLoadLibraryError(path: string) =
   # carefully written to avoid memory allocation:
-  cstderr.rawWrite("could not load: ")
+  const prefix = "could not load: "
+  cstderr.rawWrite(prefix)
   cstderr.rawWrite(path)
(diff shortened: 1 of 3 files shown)


awr1:
Message box works:
image

Varriount:
Hm, keep in mind that (as paradoxical as it may seem) there _are_ circumstances where "GUI" applications can't/shouldn't create windows. The most common situation is when an application is running as a Windows service (Linux daemon equivalent).

ghost:
@Varriount but the code only shows MessageBox with "when defined(guiapp):", otherwise it's just "cstderr.rawWrite(badExe)"

awr1:
> 
> 
> Hm, keep in mind that (as paradoxical as it may seem) there _are_ circumstances where "GUI" applications can't/shouldn't create windows. The most common situation is when an application is running as a Windows service (Linux daemon equivalent).

The message boxes on fatal crashes aren't new and only show up with `--app:gui` for Windows

PR #13957, opened 2020-04-11
Treat zig like clang/gcc wrt integer arithmetic.

hessammehr:
Compiling Nim using `zig cc` is currently broken on nightly because gcc/clang implementation of integer arithmetic ops are not used (see https://github.com/nim-lang/Nim/pull/13757). Since `zig cc` should behave identically to clang, this PR fixes the issue by asking for the same integer code.

### DIFF
--- a/lib/system/integerops.nim
+++ b/lib/system/integerops.nim
@@ -19,7 +19,7 @@ proc raiseDivByZero {.compilerproc, noinline.} =
 
 {.pragma: nimbaseH, importc, nodecl, noSideEffect, compilerproc.}
 
-when (defined(gcc) or defined(clang)) and not defined(nimEmulateOverflowChecks):
+when (defined(gcc) or defined(clang) or defined(zig)) and not defined(nimEmulateOverflowChecks):
   # take the #define from nimbase.h
 
   proc nimAddInt(a, b: int, res: ptr int): bool {.nimbaseH.}


Araq:
Instead zig should define ``clang`` since there is no difference.

hessammehr:
Fair point, that makes more sense.

Out of interest, I tried defining clang using `nim c -d:clang` but it didn't help. Is this to be expected?

andrewrk:
> Instead zig should define `clang` since there is no difference.

it does:

```
andy@ark ~/tmp> cat test.c
#ifndef __clang__
#error expected to define clang
#endif
andy@ark ~/tmp> zig cc -c test.c
andy@ark ~/tmp> 
```

Araq:
@andrewrk Not my point, the Nim define shall be ``clang`` too.

Araq:
> Out of interest, I tried defining clang using nim c -d:clang but it didn't help. Is this to be expected?

It's surprising for me too. :-)

timotheecour:
that's because -d:clang doesn't `--passc:-D__clang__`, no?
```#if (!defined(_MSC_VER) || defined(__clang__)) && !defined(NIM_EmulateOverflowChecks)...```

**Comment by hessammehr** [comment]
Does that mean `nim c -d:__clang__ ` would work?

**Comment by timotheecour** [comment]
no, but try with the flag i showed (and read docs for --passc)

**Comment by alaviss** [comment]
> that's because -d:clang doesn't `--passc:-D__clang__`, no?
> `#if (!defined(_MSC_VER) || defined(__clang__)) && !defined(NIM_EmulateOverflowChecks)...`

It doesn't really matter. What matter is that the compiler (nim itself) has to define the symbol else it won't use the builtin functions.

**Comment by hessammehr** [comment]
@timotheecour `--passc:-D__clang__` didn't help either. Like @alaviss, I think the nim compiler needs to do one of the following for this to work:

1. Add a case for `defined(zig)` (as implemented in this PR).
2. Define `clang` when `--cc=zig` is used. 

Edit: To clarify, I also believe the issue is not in nimbase.h as your snippet suggests but in  integerops.nim.

```nim
when (defined(gcc) or defined(clang)) and not defined(nimEmulateOverflowChecks):
  # take the #define from nimbase.h

  proc nimAddInt(a, b: int, res: ptr int): bool {.nimbaseH.}
  proc nimSubInt(a, b: int, res: ptr int): bool {.nimbaseH.}
  proc nimMulInt(a, b: int, res: ptr int): bool {.nimbaseH.}

  proc nimAddInt64(a, b: int64; res: ptr int64): bool {.nimbaseH.}
  proc nimSubInt64(a, b: int64; res: ptr int64): bool {.nimbaseH.}
  proc nimMulInt64(a, b: int64; res: ptr int64): bool {.nimbaseH.}
```

Issue #13864, opened 2020-04-04
crash in nimDestroyAndDispose


yglukhov:
The following code works with default gc, but not with `--gc:arc`.
```nim
type Variant* = object
    refval: ref RootObj

proc newVariant*[T](val: T): Variant =
    let pt = T.new()
    pt[] = val
    result = Variant(refval: cast[ref RootObj](pt))

var v = newVariant(@[1, 2, 3])
```
```
nim c -r --gc:arc test.nim
```
```
Traceback (most recent call last)
~/test.nim(7) test
~/nim/lib/system/refs_v2.nim(119) nimDestroyAndDispose
SIGSEGV: Illegal storage access. (Attempt to read from nil?)
```
**Nim version**: 1.2

cooldome:
It is caused by the fact that `seq` doesn't inherit from `RootObj`. This is more of ref in arc design question. In principal it is illegal cast, but I understand it is a common trick in old runtime.

@Araq: do we plan to support such use case?

yglukhov:
I am happy to work this around. If there any other type erasure technique for seq?

Araq:
> do we plan to support such use case?

Uh, no...

cooldome:
As a discussion point. 

Looks like with `gc:arc` there is no way to have a ref to anything anymore, it only ptr to anything left. Sounds like a limitation. I understand we have reference to final objects and non final objects. Reference to final objects are more efficient because they don't have to keep runtime type info. Possibly we can have a procedure to promote/convert ref on final object to ref on non final object, it will not be free but it will exist. Something like automatic packing into another object with NTI as first element.  Problem with alignment might appear though.

@Araq, What do you think?

Araq:
I don't see the point. With ``move`` it's easy to now code this without hacks:

```nim

type
  Base[T] = object of RootObj
    val: T

proc newVariant*[T](val: sink T): Variant =
    let pt = Base[T](val: val)
    result = Variant(refval: pt)

```

cooldome:
make sense

PR #13962, opened 2020-04-12
error msg for #13864

cooldome:
#13864 is closed as won't fix, but I have added proper sem error message for it instead of runtime crash


Araq:
Acceptable but stricting speaking the compiler frontend shouldn't be specific to the selected memory management strategy. ``cast`` assumes that you know what you're doing...

PR #13963, opened 2020-04-12
Do not use clone(2) if a process can setuid(2).

kubo39:
please see: https://wiki.sei.cmu.edu/confluence/pages/viewpage.action?pageId=87152373
I hope nobody uses this flag..

cc: @alaviss, because I guess you're familiar with *nix.

### DIFF
--- a/lib/posix/linux.nim
+++ b/lib/posix/linux.nim
@@ -26,3 +26,7 @@ proc clone*(fn: pointer; child_stack: pointer; flags: cint;
             ctid: ptr Pid): cint {.importc, header: "<sched.h>".}
 
 proc pipe2*(a: array[0..1, cint], flags: cint): cint {.importc, header: "<unistd.h>".}
+
+proc getresuid*(ruid: ptr Uid, euid: ptr Uid, suid: ptr Uid): cint {.importc, header: "<unistd.h>".}
+
+proc getresgid*(rgid: ptr Gid, egid: ptr Gid, sgid: ptr Gid): cint {.importc, header: "<unistd.h>".}
--- a/lib/pure/osproc.nim
+++ b/lib/pure/osproc.nim
@@ -976,16 +976,41 @@ elif not defined(useNimRtl):
       var pid: Pid
       var dataCopy = data
 
+      # TODO: deprecate this flag?
       when defined(useClone):
-        const stackSize = 65536
-        let stackEnd = cast[clong](alloc(stackSize))
-        let stack = cast[pointer](stackEnd + stackSize)
-        let fn: pointer = startProcessAfterFork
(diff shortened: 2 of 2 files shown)


alaviss:
`useClone` is the default if you're on Linux.

Now, I don't think we're affected by this at all, because we don't `setuid`, ever. See `startProcessAfterFork` for how the process spawning after `clone` is done.

alaviss:
Now, there are a lot of problems with `vfork` that's unrelated to this `setuid` issue that we are certain to hit. From reading this [bug report](https://sourceware.org/bugzilla/show_bug.cgi?id=10354) it appears that musl libc has cracked the formula to be safe while using `vfork` and I believe we should learn from them.

kubo39:
> Now, I don't think we're affected by this at all, because we don't setuid, ever. See startProcessAfterFork for how the process spawning after clone is done.

Ahh, first I thought startProcess uses `chidr` between fork and exec, but I'm still afraid that because the child's stack is allocated by the parent process.

> Now, there are a lot of problems with vfork that's unrelated to this setuid issue that we are certain to hit. From reading this bug report it appears that musl libc has cracked the formula to be safe while using vfork and I believe we should learn from them.

Exactly so, especially for signal handlers I described in comments.
glibc hit the problem: https://sourceware.org/bugzilla/show_bug.cgi?id=14750

kubo39:
Hmm, startProcessAfterFork has race condition when the parent process in multi-threaded situation?

alaviss:
> Hmm, startProcessAfterFork has race condition when the parent process in multi-threaded situation?

How so?

kubo39:
Sorry, it's in startProcessAuxSpawn, not startProcessAfterFork.

alaviss:
Aside from the `chdir` call, I don't think there's any problem?

Too bad `posix_spawn` API can't do `chdir` :P

kubo39:
> Too bad posix_spawn API can't do chdir :P

AFAIK Solaris and glibc >= 2.29 only has API for it, but in case there's no portable way.

> Aside from the chdir call, I don't think there's any problem?

Hmm... I think this is bad and may cause an unexpected behavior when another threads see chdir()'ed directory silently, but I couldn't come up with a certain problem now.

alaviss:
> Hmm... I think this is bad and may cause an unexpected behavior when another threads see chdir()'ed directory silently, but I couldn't come up with a certain problem now.

The bigger problem might be that we don't even `chdir` back to the old one :p 

Also, I think this bug in [python](https://bugs.python.org/issue35823) describes pretty well the shortcomings of `posix_spawn`.

Araq:
I see nothing to "fix" with our code.

PR #13968, opened 2020-04-13
Add the ability to pass title case headers to an HTTP server

wonderix:
Some older HTTP servers (especially those used in embedded devices) expect title case HTTP headers. Normally HTTP headers should be case insensitive. To also support the communication to these HTTP servers, I added a flag `titleCase` to the `HttpHeaders`
This was also added as feature to other programming languages (e.g. [rust](https://docs.rs/reqwest/0.9.22/reqwest/struct.ClientBuilder.html#method.http1_title_case_headers))

### DIFF
--- a/lib/pure/httpcore.nim
+++ b/lib/pure/httpcore.nim
@@ -17,7 +17,7 @@ import tables, strutils, parseutils
 type
   HttpHeaders* = ref object
     table*: TableRef[string, seq[string]]
-    convert: proc (str: string): string
+    isTitleCase: bool
 
   HttpHeaderValues* = distinct seq[string]
 
@@ -108,23 +108,25 @@ proc toTitleCase(s: string): string =
     result[i] = if upper: toUpperAscii(s[i]) else: toLowerAscii(s[i])
     upper = s[i] == '-'
 
+proc toCaseInsensitive(headers: HttpHeaders, s: string): string {.inline.} =
+  return if headers.isTitleCase: toTitleCase(s) else: toLowerAscii(s)
 
-proc newHttpHeaders*(titleCase =false): HttpHeaders =
+proc newHttpHeaders*(titleCase=false): HttpHeaders =
   ## Returns a new ``HttpHeaders`` object. if ``titleCase`` is set to true, 
   ## headers are passed to the server in title case (e.g. "Content-Length")
   new result
   result.table = newTable[string, seq[string]]()
-  result.convert = if titleCase: toTitleCase else : toLowerAscii
+  result.isTitleCase = titleCase
 
 proc newHttpHeaders*(keyValuePairs:
-    openArray[tuple[key: string, val: string]], titleCase =false): HttpHeaders =
+    openArray[tuple[key: string, val: string]], titleCase=false): HttpHeaders =
   ## Returns a new ``HttpHeaders`` object from an array. if ``titleCase`` is set to true, 
   ## headers are passed to the server in title case (e.g. "Content-Length")
   new result
   result.table = newTable[string, seq[string]]()
-  result.convert = if titleCase: toTitleCase else : toLowerAscii
+  result.isTitleCase = titleCase
   for pair in keyValuePairs:
-    let key = result.convert(pair.key)
+    let key = result.toCaseInsensitive(pair.key)
     if key in result.table:
       result.table[key].add(pair.val)
     else:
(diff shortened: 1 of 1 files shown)


lib/pure/httpcore.nim
```diff
@@ -17,6 +17,7 @@ import tables, strutils, parseutils
 type
   HttpHeaders* = ref object
     table*: TableRef[string, seq[string]]
+    convert: proc (str: string): string
```
Yardanico, line 20: Is this needed? Why not just use toTitleCase procedure?

lib/pure/httpcore.nim
```diff
@@ -100,16 +101,30 @@ const
...
-proc newHttpHeaders*(): HttpHeaders =
+proc toTitleCase(s: string): string =
+  result = newString(len(s))
+  var upper = true
+  for i in 0..len(s) - 1:
+    result[i] = if upper: toUpperAscii(s[i]) else: toLowerAscii(s[i])
+    upper = s[i] == '-'
+
+
+proc newHttpHeaders*(titleCase: bool =false): HttpHeaders =
```
ghost: you can just write "titleCase = false" here

lib/pure/httpcore.nim
```diff
@@ -100,16 +101,30 @@ const
...
+  ## Returns a new ``HttpHeaders`` object. if ``titleCase`` is set to true, 
+  ## headers are passed to the server in title case (e.g. "Content-Length")
   new result
   result.table = newTable[string, seq[string]]()
+  result.convert = if titleCase: toTitleCase else : toLowerAscii
 
 proc newHttpHeaders*(keyValuePairs:
-    openArray[tuple[key: string, val: string]]): HttpHeaders =
+    openArray[tuple[key: string, val: string]],titleCase: bool =false): HttpHeaders =
```
ghost: same here, also add a space like "openArray[tuple[key: string, val: string]], titleCase"

ghost:
I think like this indeed might be a useful feature, but it needs more thought on how to do it properly

lib/pure/httpcore.nim
wonderix: OK. Changed

lib/pure/httpcore.nim
wonderix: OK. Changed.

ghost:
About your Rust example - see https://github.com/hyperium/hyper/pull/1497, we only need to convert headers to TitleCase when actually sending them, not every time you try to assign keys or something like that.

wonderix:
I also thought about putting my changes in [httpclient.nim](https://github.com/nim-lang/Nim/blob/3036ec9c4456f6a5edb0a5bbfbb3e3c79fba7d23/lib/pure/httpclient.nim#L512). But I decided against it, because in this case the conversion has to be done twice. (`key`->`toLowerAscii` -> `toTitleCase`)

I also decided against simply replacing `toLowerAscii` by `toTitleCase` because this could be a breaking change.

lib/pure/httpcore.nim
```diff
@@ -100,16 +101,30 @@ const
...
+proc newHttpHeaders*(titleCase =false): HttpHeaders =
+  ## Returns a new ``HttpHeaders`` object. if ``titleCase`` is set to true, 
+  ## headers are passed to the server in title case (e.g. "Content-Length")
   new result
   result.table = newTable[string, seq[string]]()
+  result.convert = if titleCase: toTitleCase else : toLowerAscii
 
 proc newHttpHeaders*(keyValuePairs:
-    openArray[tuple[key: string, val: string]]): HttpHeaders =
+    openArray[tuple[key: string, val: string]], titleCase =false): HttpHeaders =
```
dom96: Nit
```suggestion
    openArray[tuple[key: string, val: string]], titleCase=false): HttpHeaders =
```

lib/pure/httpcore.nim
```diff
@@ -100,16 +101,30 @@ const
 const httpNewLine* = "\c\L"
 const headerLimit* = 10_000
 
-proc newHttpHeaders*(): HttpHeaders =
+proc toTitleCase(s: string): string =
```
dom96, line 104: separate PR to add this into strutils would be nice :)

lib/pure/httpcore.nim
dom96: I don't think we really need this, it just adds extra complexity. Just store `isTitleCase`.

dom96 (changes requested):

lib/pure/httpcore.nim
wonderix: I had a look at other files and it seems that normally a space is used as separator before and after `=`. Therefore I added a space after the `=`

lib/pure/httpcore.nim
wonderix: I changed this accordingly.

Araq:
Or instead: You write HTTP servers that adhere to the spec and are not case sensitive...

PR #13985, opened 2020-04-15
Make bitand, bitor, bitxor varargs-friendly 

awr1:
This should make the standard bitops easier to use, e.g. instead of `bitor(bitor(a, b), c)`, one should be able to now do `bitor(a, b, c)`.

The reason the parameters are of the form `(x, y: T; z: varargs[T])` is to enforce a lower bound of two arguments (and also to avoid a more complicated implementation in `bitand` specifically)

This *shouldn't* break the preexisting API.


Araq:
We should make it a macro so that we can guarantee it has no overhead.

alaviss:
Wouldn't just `{.inline.}` work here? Macros are a PITA when you step through code in debugging.

Clyybber:
We could make the macro expand to ```x `op` y `op` e1 `op` e2 ...```

Araq:
If ``inline`` worked I wouldn't have proposed a macro. ;-)

awr1:
> 
> 
> We should make it a macro so that we can guarantee it has no overhead.

Do you mean a template or a macro that composes, e.g.

```nim
internalBitor(internalBitor(internalBitor(x, y), z[0]), z[1] ...
```

Araq:
Yes.

PR #13993, opened 2020-04-16
Copy info to sym for copyLineInfo

alehander92:
Otherwise i can't change an iterator's lineinfo with `copyLineInfo` : as it seems cgen takes the sym lineinfo.

The other option is to provide a copyLineInfoToSym


krux02:
If you intend to use this feature, you should add a test for it. But you get my approval.

krux02 (approved):

Araq:
> as it seems cgen takes the sym lineinfo.

So can we please fix the codgen instead? Overwriting a symbol's point of definition via the macro system is wrong.

Araq:
This is not a good fix for anything, sorry. Closing.

alehander92:
sorry, i'll try to look at cgen !

PR #14000, opened 2020-04-16
Add --randomize to Testament

juancarlospaco:
- Add `--randomize` to Testament. Tiny diff.

# Why?

On Unittests a "Unit" must not depend on previous unit,
and next unit must not depend on the current unit, and so on,
random order for tests helps against unwanted unit cross-dependency,
when something on a test depends on something on a previous test,
like bad Tear-Downs, state, file or folder existence, leaked data, etc.

### DIFF
--- a/testament/testament.nim
+++ b/testament/testament.nim
@@ -10,7 +10,7 @@
 ## This program verifies Nim against the testcases.
 
 import
-  strutils, pegs, os, osproc, streams, json,
+  strutils, pegs, os, osproc, streams, json, random,
   backend, parseopt, specs, htmlgen, browsers, terminal,
   algorithm, times, md5, sequtils, azure
 
@@ -40,6 +40,7 @@ Options:
   --print                   also print results to the console
   --simulate                see what tests would be run but don't run them (for debugging)
   --failing                 only show failing/ignored tests
+  --randomize               Random order for tests (unwanted unit crossdependency)
   --targets:"c c++ js objc" run tests for specified targets (default: all)
   --nim:path                use a particular nim executable (default: $$PATH/nim)
   --directory:dir           Change to directory dir before reading the tests or doing anything else.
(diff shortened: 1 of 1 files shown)


Araq:
The idea is nice but are you certain this does the randomization you're after?

juancarlospaco:
@Araq 
I `echo` before and after several times, and is working, obviously I removed debug `echo` for PR. 

I always used similar construct on Python for years, sometimes it catches bugs too,
also ensures tests are not dependent of each others, correct setups/teardowns.

More stuff can be randomized if wanted on the future, but this is a good base feature,
some testing framework do stuff like randomized delays between tests, etc etc,
is up to creativity, if you see something else that should be randomized feel free to comment,
if I know how to do it can be implemented on a future PR, to keep Diff tiny.

alaviss:
Are you certain this is what you want? As far as this goes, it only randomize the order of `categories` tested.

juancarlospaco:
I know, still better than nothing, what do you want to randomize?, individual files?,
really dont know where the files are on a sequence form on the code to shuffle them up, 
if you can tell me where the list of files is on the code I can add that too. :shrug: 

If you thinking of compiler targets?, I dont think was too important, but maybe can be added too. :shrug:

Araq:
Sorry, rejected. I cannot imagine this to find any real bugs given that nimcache is project specific already.

juancarlospaco:
@Araq 
This is `nimcache` agnostic, can be confirmed by the fact that still works on JS target. :shrug:

disruptek:
I like the idea.  Can you explain what you mean by reference to the nimcache, @Araq?

Araq:
The nimcache is project specific so there is no overlap between projects. This PR's randomization basically tests for this fact and only for this fact. But we know it to be true and it's a tradeoff between "test what is super unlikely to ever get broken" and "add more code to testament to maintain".

PR #14006, opened 2020-04-17
fixes #14003

### LINKED ISSUE #14003: simple 'var openarray[char]' assignment crash when the openarray source is a local string and using gc:arc
This simple `nonce[1] = char(65)` crash. it works with default `gc` but not with `gc:arc`

### Example
```nim
proc cryptCTR*(nonce: var openArray[char]) =
  nonce[1] = char(65)

proc main() =
  var nonce = "0123456701234567"
  cryptCTR(nonce.toOpenArray(0, nonce.len-1))
  debugEcho nonce
  
main()
```

### Current Output
```
```

### Expected Output
```
0A23456701234567
```

```
$ nim -v
Compiled at 2020-04-09

git hash: b4e9f8e814373fc38741736197d88475663ce758
```

cooldome:
fixes #14003

### DIFF
--- a/compiler/ccgcalls.nim
+++ b/compiler/ccgcalls.nim
@@ -113,7 +113,7 @@ proc fixupCall(p: BProc, le, ri: PNode, d: var TLoc,
 
 proc genBoundsCheck(p: BProc; arr, a, b: TLoc)
 
-proc openArrayLoc(p: BProc, n: PNode): Rope =
+proc openArrayLoc(p: BProc, formalType: PType, n: PNode): Rope =
   var a: TLoc
 
   var q = skipConv(n)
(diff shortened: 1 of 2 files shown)


jangko:
I've tried it with more than one call to openarray proc, and it generates a call to `nimPrepareStrMutationV2` everytime before the literal string passed to the proc.

isn't  multiple call to `nimPrepareStrMutationV2` like this will create another bug?
shouldn't it only called once during local var initialization?

cooldome:
nimPrepareStrMutationV2 does nothing if string is already mutable. I think it is ok for now.
Another solution would require control flow graph and sounds like overkill for now.

Araq to decide

Araq:
Good enough for me.

Issue #3203, opened 2015-08-12
Should --app:gui ensure linking to user32 on Windows?


skyfex:
To build a GUI app on Windows using VCC, it seems we have to do:

```
nim c --app:gui --passL:user32.lib
```

This actually took me a bit of time to find out. Should the Nim compiler perhaps add user32.lib by default when the --app:gui flag is set?

Araq:
yeah.

Leu-Gim:
Is it needed for programs using `--app:gui` just to suppress console window (servers, etc.)?

skyfex:
Leu-Gim: No, I don't believe it is. So that might be a valid use case. But the switch seems to apply that it's meant for gui apps. Maybe a different switch could be used for apps that just want to be invisible.

Varriount:
This is only for VCC, right? Is there something that has to be done for GCC, Intel, etc?

dom96:
Is this really an installation issue?

clarkeaa:
To help other people find this issue I'll describe how I found this issue:

If you forget to add `--app:gui` then when running your exe an extra console window gets brought up with your GUI window.

If you add `--app:gui` but leave out the linker argument, you get the following linker error:
```
... stdlib_system.nim.cpp.obj : error LNK2019: unresolved external symbol __imp_MessageBoxA referenced in function "void __cdecl writeToStdErr__a2kDfqdSc1eYf0ZCWOGinQ(char *)" (?writeToStdErr__a2kDfqdSc1eYf0ZCWOGinQ@@YAXPEAD@Z)
```

ringabout:
The issues seem to be fixed by https://github.com/nim-lang/Nim/pull/14013

PR #14013, opened 2020-04-18
Add LTO support for most compilers and do some VCC fixes

Keithcat1:
I added /link to the vcc link options so that passL works properly.
Clang_cl: Always use lld linker. It always seems to be provided on Windows, so shouldn't be a problem.
Add "user32.lib" when linking GUI apps with vcc to resolve missing import MessageBoxA.
You can define lto to enable link time optimization, It doesn't support all compilers yet.
You can define -d:lto_incremental to enable lto and additionally enable incremental linking for compilers that support it (currently just VCC, but it's very fast).
Added -d:strip to strip debug symbols at link time, for those compilers like GCC that embed COFF symbols and you don't want them.
Add /nologo to VCC compile and link tmpl in extccomp.nim.
Added /Md to compileTmpl in extccomp.nim to reduce binary size by dynamically linking to runtime.

### DIFF
--- a/compiler/extccomp.nim
+++ b/compiler/extccomp.nim
@@ -1,1207 +1,1 @@
-#
-#
-#           The Nim Compiler
-#        (c) Copyright 2013 Andreas Rumpf
-#
-#    See the file "copying.txt", included in this
-#    distribution, for details about the copyright.
-#
-
-# Module providing functions for calling the different external C compilers
-# Uses some hard-wired facts about each C/C++ compiler, plus options read
-# from a lineinfos file, to provide generalized procedures to compile
-# nim files.
-
-import
-  ropes, os, strutils, osproc, platform, condsyms, options, msgs,
-  lineinfos, std / sha1, streams, pathutils, sequtils, times, strtabs
-
-type
-  TInfoCCProp* = enum         # properties of the C compiler:
-    hasSwitchRange,           # CC allows ranges in switch statements (GNU C)
-    hasComputedGoto,          # CC has computed goto (GNU C extension)
-    hasCpp,                   # CC is/contains a C++ compiler
-    hasAssume,                # CC has __assume (Visual C extension)
-    hasGcGuard,               # CC supports GC_GUARD to keep stack roots
-    hasGnuAsm,                # CC's asm uses the absurd GNU assembler syntax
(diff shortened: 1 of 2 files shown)


Araq:
You changed the line endings, please undo that so that I can review your PR.

Keithcat1:
OK. Should be fixed now.

ghost:
@Keithcat1 you broke them even further now :D

ghost:
Here

Keithcat1:
Tried to fix it,but if it's not fixed now, I have no idea lol.
Sorry, windows user here.

ghost:
Note: this PR will also solve https://github.com/nim-lang/Nim/issues/3203

compiler/extccomp.nim
```diff
@@ -157,12 +157,12 @@ compiler vcc:
     optSize: " /O1 ",
     compilerExe: "cl",
     cppCompiler: "cl",
-    compileTmpl: "/c$vccplatform $options $include /Fo$objfile $file",
-    buildGui: " /link /SUBSYSTEM:WINDOWS ",
+    compileTmpl: "/c$vccplatform $options $include /nologo /MD /Fo$objfile $file",
```
Araq: ``/MD`` is an undocumented change and I'd prefer static linking too.

Keithcat1:
How about I turn it off and add a define "dynamic_runtime" which turns it back on?

Keithcat1:
Removed it.
I'd like to add the dynamic_cruntime define, but I don't know enough abut non-windows compilers.

ghost:
@Keithcat1 I've hit an error while trying to use -d:lto with Clang 9 on Linux, do you know anything about that?:
```clang-9: error: invalid linker name in argument '-fuse-ld=lld'```

ghost:
However when I pass "-flto" or "-flto=thin" myself directly via passC/passL LTO works with clang just fine

Keithcat1:
Hi,
Do you have the LLVM LLD linker installed?

ghost:
@Keithcat1 no, but clang's LTO and ThinLTO work without it

Keithcat1:
Pretty sure that's the problem then. It uses lld by default.

PR #14028, opened 2020-04-20
refactor system.$ for objects a little; refs #13398

Araq:


arnetheduck:
just for the record, the original PR removed nil checks because the implementation was broken - not because it's  a bad feature per se - in fact, it would be nice to have something like it that works (reliably)

Araq:
Well yeah, work on ``not nil`` is on-going via DrNim for the time being.

PR #14037, opened 2020-04-20
Fix broken async httpclient example

0atman:
As the async httpclient is almost certainly the first async example beginners will want to try, we OWE it to them to give them a real example.

Example repeated here for clarity:

```nim
import asyncdispatch, httpclient

proc asyncProc(): Future[string] {.async.} =
  return await client.getContent("http://example.com")

echo waitFor asyncProc()
```

This is my first Nim contribution, please let me know if the code is right. (it runs on my machine, but may not be the best example)

### DIFF
--- a/lib/pure/httpclient.nim
+++ b/lib/pure/httpclient.nim
@@ -25,15 +25,19 @@
 ## ``AsyncHttpClient``:
 ##
 ## .. code-block:: Nim
-##   import httpClient
+## import asyncdispatch, httpclient
+## 
+## proc asyncProc(): Future[string] {.async.} =
 ##   var client = newAsyncHttpClient()
-##   echo await client.getContent("http://google.com")
+##   return await client.getContent("http://example.com")
+## 
+## echo waitFor asyncProc()
 ##
 ## The functionality implemented by ``HttpClient`` and ``AsyncHttpClient``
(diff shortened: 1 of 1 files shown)


0atman:
This certainly tripped me up when I was trying to learn Nim (last month!)

0atman:
Question: Why don't the Nim docs use the awesome `runnableExample` feature? That way this bug would never would have happened, as the docs wouldn't have compiled.

ghost:
Well I wouldn't say that it's "broken", the docs tell you that you need to run async stuff in async procedures

dom96 (approved):

0atman:
@Yardanico I think the gold standard for code examples is if the *compiler* thinks it's broken! :laughing: 

Working examples should be our default, but I'd be interested to hear alternative viewpoints.

0atman:
Thanks for merging this in, @dom96 :ok_hand:

Araq:
> Question: Why don't the Nim docs use the awesome runnableExample feature? That way this bug would never would have happened, as the docs wouldn't have compiled.

Because some code predates the runnableExample feature.

0atman:
@Araq Can't argue with that. Is this the right way to fix that? https://github.com/nim-lang/Nim/pull/14045

Issue #12639, opened 2019-11-11
`mapIt` can't be chained when used to generated func-s


foldl:
Two `mapIt` can't be chained together when used to generated func-s.

### Example
```nim
import sequtils

static:
  let m = @[func (x: int): int = x + 100, func (x: int): int = x + 200]
  echo m.mapIt(func (x: int): int = it(x)).mapIt(it 1)
```

### Current Output
```
Error: illformed AST: proc (x: int): int = result = it(x)
```

### Expected Output
```
@[101, 201]
```

### Possible Solution

Introduce a variable for the first `mapIt` to break them into two statements could be a possible 
workaround.

But, for this specific example, there is another issue (#12625), so we can't get the correct output.

### Additional Information

N/A

```
$ nim -v
Nim Compiler Version 1.0.2 [Windows: amd64]
```

Issue #12625, opened 2019-11-08
IndexError when use a list of funcs generated by`mapit`


foldl:
Use `mapit` to generate a list of functions. When calling these functions, Index error occur.

### Example
```nim
import sequtils, sugar
static:
  # following map & mapit should be equivalent
  echo [1, 2].map((it) => func (x: int): int = it + x).map((f) => f(100))
  echo [1, 2].mapit(func (x: int): int = it + x).map((f) => f(100))
```

### Current Output
```
@[101, 102]
fatal.nim(39)            sysFatal
Error: unhandled exception: index 23 not in 0 .. 2 [IndexError]
```

### Expected Output
```
@[101, 102]
@[101, 102]
```

### Possible Solution

N/A.

### Additional Information

N/A

```
$ nim -v
Nim Compiler Version 1.0.2 [Windows: amd64]
Compiled at 2019-10-22
Copyright (c) 2006-2019 by Andreas Rumpf

git hash: 193b3c66bbeffafaebff166d24b9866f1eaaac0e
active boot switches: -d:release
```

narimiran:
Regarding the expected output, running the example outside of the static block gives:
```
@[101, 102]
@[102, 102]
```

foldl:
Here is an example that can't give the correct result running in `isMainModule`. 
It seems like that only m[1] is used in `mapIt`.

```nim
if isMainModule:
  let m = @[func (x: int): int = x + 100, func (x: int): int = x + 200]
  let l = m.mapIt(func (x: int): int = it(x))
  echo l.mapIt(it 1)
```

## Current Output

```
@[201, 201]
```

`` Expected Output

```
@[101, 201]
```

alehander92:
hm, is `if isMainModule` valid , i thought one can only `when`

narimiran:
> hm, is if isMainModule valid , i thought one can only when

It doesn't matter, I originally reproduced the problem without it.

foldl:
My finding it that: *`it`* is not captured by the generated proc of `mapIt`, but defined as
a global variable.

```nim
if isMainModule:
  let m = @[func (x: int): int = x + 100, func (x: int): int = x + 200]
  # The "simplified" generated C code of this line is listed below
  let l = m.mapIt(func (x: int): int = it(x))  
  echo l.mapIt(it 1)
```

```C
it__MpefkJ9cQAwQD9aL17D5DVvw; // *it* is a global variable, which is wrong!
[...]

PR #14041, opened 2020-04-21
fix mapIt issues #12625 & #12639

foldl:
1. fallback to call `map` when the result of `op` is a closure;
2. use `items(s)` in the for loop;
3. add a testcase.

### DIFF
--- a/lib/pure/collections/sequtils.nim
+++ b/lib/pure/collections/sequtils.nim
@@ -915,20 +915,18 @@ template mapIt*(s: typed, op: untyped): untyped =
     assert strings == @["4", "8", "12", "16"]
 
   when defined(nimHasTypeof):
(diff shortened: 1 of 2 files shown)


Araq:
Isn't this a workaround for a compiler bug? Also, I think it's wiser to error out instead of using ``map`` because of the different performance characteristics.

foldl:
It is not a compiler bug. It's limitations of `mapIt`.

On fallback to `map` or report error: maybe it's better to fallback to `map` and give a hint on performance? For me, in this fallback case, `mapIt` means less code than `map`, while performance are
the same.

Araq:
Well since you understand the problem and I don't, please add a good comment to the source code helping me and others in the review process.

PR #14045, opened 2020-04-21
New runnableExample for `newAsyncHttpClient()`

0atman:
I've added an example that uses example.com, the contents of which are static, so it should be safe in testing.

Some context in #14037


0atman:
Thanks @Araq! I note that the CI hasn't been running on the dev branch for 21 days, which includes my code: https://travis-ci.org/github/nim-lang/Nim/builds/669252721 is everything all right?

Araq:
Which CI? We moved from travis to azure pipelines...

0atman:
Oh! Then a readme update is needed: The travis badge is still there, which confused me. Merge this https://github.com/nim-lang/Nim/pull/14062

Thanks!

PR #14046, opened 2020-04-21
base `parseEnum` on a case statement, fixes #14030

### LINKED ISSUE #14030: [regression] `parseEnum` does not work anymore for enums with holes
Already mentioned on [IRC](https://irclogs.nim-lang.org/20-04-2020.html#08:47:17), but obviously worth an issue.

The fix https://github.com/nim-lang/Nim/commit/9874981e757f0e2658baa3a871450509cee9204c for #14001 broke `parseEnum` for enums with holes, since the current implementation of `parseEnum` is based on iterating over the enum using `low(T) .. high(T)`.

It may be fine to break `for x in MyEnum`, but imo there's no reason why `parseEnum` taking a string should be affected.

### Example
```nim
import strutils

type
  HoledEnum = enum
    A = 1
    B = 4
    C = 8

let x = parseEnum[HoledEnum]("A")
doAssert x == A
```

### Current Output
```
/tmp/broken.nim(9, 18) template/generic instantiation of `parseEnum` from here
/home/basti/src/nim/nim_git_repo/lib/pure/strutils.nim(1249, 18) template/generic instantiation of `..` from here
/home/basti/src/nim/nim_git_repo/lib/system/iterators_1.nim(85, 12) Error: type mismatch: got <HoledEnum>
but expected one of: 
proc inc[T: Ordinal](x: var T; y = 1)
  required type for x: var T: Ordinal
  but expression 'res' is of type: HoledEnum

expression: inc(res)
```

### Expected Output
Should compile and work.

### Possible Solution

Either as mentioned by @timotheecour in #14001 by adding magics similar to the `fields` iterator or by adding a macro similar to:
```nim
proc listEnumImpl(enumType: NimNode,
                  returnFields: bool,
                  returnValues: bool): NimNode =
  # `getTypeInst` to unpack e.g. a generic `T: enum` from a symbol
  let impl = enumType.getTypeInst[1].getImpl[2]
  expectKind(impl, nnkEnumTy)
  result = nnkBracket.newTree()
  doAssert returnFields or returnValues
  for el in impl:
    case el.kind
    of nnkEmpty: discard # first field of enumTy
    of nnkEnumFieldDef:
      if returnFields and returnValues:
        result.add nnkPar.newTree(el[0], el[1])
      elif returnFields:
        result.add el[0]
      else:
        result.add el[1]
    else: doAssert false, "shouldn't exist in enum type"

macro listFields*(enumType: typed): untyped =
  ## Returns all fields of the given enum (with or without holes) as an array.
  result = listEnumImpl(enumType, true, false)
```
and using that for the loop in `parseEnum` instead of `low(T) .. high(T)`.

The latter has the major disadvantage that one suddenly has to import macros in `strutils` / import a new module. Adding those to `system.nim` is also not a good idea. 

Finally, by adding magics one can also support `for x in HoledEnum` again (even if via explicit iterators). 

I tried to add a new magic, but I fail to even get that magic e.g.:
```nim
iterator fields*[T: enum](x: T): RootObj {.magic: "EnumFields", noSideEffect.}
```
to match an enum in the first place, because I don't understand how to match `type MyEnum`. 

### Additional Information

Worked before the aforementioned commit https://github.com/nim-lang/Nim/commit/9874981e757f0e2658baa3a871450509cee9204c.

Vindaar:
Changes the `parseEnum` implementation to be based on a case statement comparing the given normalized string value to the normalized field of the enum.

@cooldome: I'm not quite sure what you meant with:
```nim
nnkOfBranch.newTree(normalize(str), enumVal)
```
I don't understand why this should work, nor could I get it to work.

In any case, I had to change the implementation a little bit anyways. While writing a test, I put an enum into a `block`, which broke the code, since the fields weren't known anymore. So instead I determine the correct field value and use `T(fieldValue)` instead.

### DIFF
--- a/lib/pure/strutils.nim
+++ b/lib/pure/strutils.nim
@@ -75,6 +75,7 @@
 import parseutils
 from math import pow, floor, log10
 from algorithm import reverse
+import macros # for `parseEnum`
 
 when defined(nimVmExportFixed):
   from unicode import toLower, toUpper
@@ -1229,6 +1230,53 @@ proc parseBool*(s: string): bool =
   of "n", "no", "false", "0", "off": result = false
   else: raise newException(ValueError, "cannot interpret as a bool: " & s)
 
+proc addOfBranch(n, field, enumType, norm: NimNode): NimNode =
+  result = nnkOfBranch.newTree(
+    nnkDotExpr.newTree(n, norm), # `<field>.normalize`
+    nnkCall.newTree(enumType, field)) # `T(<fieldValue>)`
+
+macro genEnumStmt(n, argIdent: typed,
+                  default: typed): untyped =
+  # generates a case stmt, which assigns the correct enum field given
+  # a normalized string comparison to the `argIdent` input.
+  # NOTE: for an enum with fields Foo, Bar, ... we cannot generate
+  # `of "Foo".normalize: Foo`.
+  # This will fail, if the enum is not defined at top level (e.g. in a block).
+  # Thus we check for the field value of the (possible holed enum) and convert
+  # the integer value to the generic argument `n`.
+  let impl = n.getTypeInst[1].getImpl[2]
+  expectKind impl, nnkEnumTy
+  let norm = bindSym"normalize"
+  result = nnkCaseStmt.newTree(nnkDotExpr.newTree(argIdent, norm))
+  var fieldNum: BiggestInt = 0
+  for f in impl:
+    case f.kind
+    of nnkEmpty: continue # skip first node of `enumTy`
+    of nnkSym, nnkIdent:
+      result.add addOfBranch(f.toStrLit, newLit fieldNum, n, norm)
+    of nnkEnumFieldDef:
+      case f[1].kind
+      of nnkStrLit: result.add addOfBranch(f[1], newLit fieldNum, n, norm)
+      of nnkTupleConstr:
+        let enumFieldNum = f[1][0]
+        result.add addOfBranch(f[1][1], enumFieldNum, n, norm)
+        fieldNum = enumFieldNum.intVal
+      of nnkIntLit:
+        let enumFieldNum = f[1]
+        result.add addOfBranch(f[0].toStrLit, f[1], n, norm)
+        fieldNum = enumFieldNum.intVal
+      else: error("Invalid tuple syntax!")
+    else: error("Invalid node for enum type!")
+    inc fieldNum
+  # finally add else branch to raise or use default
+  expectKind(default, nnkSym)
+  if default == newLit false:
+    let raiseStmt = quote do:
+      raise newException(ValueError, "Invalid enum value: " & $`argIdent`)
+    result.add nnkElse.newTree(raiseStmt)
+  else:
+    result.add nnkElse.newTree(default)
+
 proc parseEnum*[T: enum](s: string): T =
   ## Parses an enum ``T``.
   ##
@@ -1245,11 +1293,7 @@ proc parseEnum*[T: enum](s: string): T =
     doAssert parseEnum[MyEnum]("second") == second
     doAssertRaises(ValueError):
       echo parseEnum[MyEnum]("third")
-
-  for e in low(T)..high(T):
-    if cmpIgnoreStyle(s, $e) == 0:
-      return e
-  raise newException(ValueError, "invalid enum value: " & s)
+  genEnumStmt(T, s, default = false)
 
 proc parseEnum*[T: enum](s: string, default: T): T =
   ## Parses an enum ``T``.
@@ -1267,10 +1311,7 @@ proc parseEnum*[T: enum](s: string, default: T): T =
     doAssert parseEnum[MyEnum]("second") == second
     doAssert parseEnum[MyEnum]("last", third) == third
 
-  for e in low(T)..high(T):
-    if cmpIgnoreStyle(s, $e) == 0:
-      return e
-  result = default
+  genEnumStmt(T, s, default)
 
 proc repeat*(c: char, count: Natural): string {.noSideEffect,
   rtl, extern: "nsuRepeatChar".} =
--- a/lib/pure/strutils.nim
+++ b/lib/pure/strutils.nim
@@ -280,12 +280,12 @@ proc capitalizeAscii*(s: string): string {.noSideEffect, procvar,
   else: result = toUpperAscii(s[0]) & substr(s, 1)
 
 proc nimIdentNormalize*(s: string): string =
-  ## Normalizes the string `s`as Nim identifier.
+  ## Normalizes the string `s` as a Nim identifier.
   ##
-  ## That means to convert to lower case and remove any '_' on all characters except
-  ## first one.
+  ## That means to convert to lower case and remove any '_' on all characters
+  ## except first one.
   runnableExamples:
-    doAssert normalize("Foo_bar") == "Foobar"
+    doAssert nimIdentNormalize("Foo_bar") == "Foobar"
   result = newString(s.len)
   if s.len > 0:
     result[0] = s[0]
--- a/lib/pure/strutils.nim
+++ b/lib/pure/strutils.nim
@@ -279,6 +279,26 @@ proc capitalizeAscii*(s: string): string {.noSideEffect, procvar,
   if s.len == 0: result = ""
   else: result = toUpperAscii(s[0]) & substr(s, 1)
 
+proc nimIdentNormalize*(s: string): string =
+  ## Normalizes the string `s`as Nim identifier.
+  ##
+  ## That means to convert to lower case and remove any '_' on all characters except
+  ## first one.
+  runnableExamples:
+    doAssert normalize("Foo_bar") == "Foobar"
+  result = newString(s.len)
+  if s.len > 0:
+    result[0] = s[0]
+  var j = 1
+  for i in 1..len(s) - 1:
+    if s[i] in {'A'..'Z'}:
+      result[j] = chr(ord(s[i]) + (ord('a') - ord('A')))
+      inc j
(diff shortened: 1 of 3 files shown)


lib/pure/strutils.nim
```diff
@@ -1245,11 +1293,7 @@ proc parseEnum*[T: enum](s: string): T =
     doAssert parseEnum[MyEnum]("second") == second
     doAssertRaises(ValueError):
       echo parseEnum[MyEnum]("third")
-
-  for e in low(T)..high(T):
-    if cmpIgnoreStyle(s, $e) == 0:
-      return e
-  raise newException(ValueError, "invalid enum value: " & s)
+  genEnumStmt(T, s, default = false)
```
cooldome: bool type is also enum, likely this example will trigger incorrect behavior:
```nim
parseEnum[bool]("xx", false) # exception instead of `false`
```
I suggest as default to use `nil` value

lib/pure/strutils.nim
```diff
@@ -1229,6 +1230,53 @@ proc parseBool*(s: string): bool =
...
+      of nnkStrLit: result.add addOfBranch(f[1], newLit fieldNum, n, norm)
+      of nnkTupleConstr:
+        let enumFieldNum = f[1][0]
+        result.add addOfBranch(f[1][1], enumFieldNum, n, norm)
+        fieldNum = enumFieldNum.intVal
+      of nnkIntLit:
+        let enumFieldNum = f[1]
+        result.add addOfBranch(f[0].toStrLit, f[1], n, norm)
+        fieldNum = enumFieldNum.intVal
+      else: error("Invalid tuple syntax!")
```
cooldome: ```error("Invalid tuple syntax!", f[1])```

it will help with displaying lineinfo  next to error

lib/pure/strutils.nim
```diff
@@ -1229,6 +1230,53 @@ proc parseBool*(s: string): bool =
...
+    else: error("Invalid node for enum type!")
```
cooldome: error("Invalid node for enum type!", f)

cooldome:
Appologies in my example with `normalize` I didn't convert `NimNode` to string. I also realized that getImpl() gives you idents of enums, not symbols. IMO, it should be fixed in future.

This should work
```nim
proc addOfBranch(s: string, field, enumType: NimNode): NimNode =
  result = nnkOfBranch.newTree(newLit normalize(s),
    nnkDotExpr.newTree(enumType, field)) # `T.<fieldValue>`

macro genEnumStmt(typ, argSym: typed, default: typed): untyped =
  # generates a case stmt, which assigns the correct enum field given
  # a normalized string comparison to the `argIdent` input.
  # NOTE: for an enum with fields Foo, Bar, ... we cannot generate
  # `of "Foo".normalize: Foo`.
  # This will fail, if the enum is not defined at top level (e.g. in a block).
  # Thus we check for the field value of the (possible holed enum) and convert
  # the integer value to the generic argument `n`.
  let typ = typ.getTypeInst[1]
  let impl = typ.getImpl[2]
  expectKind impl, nnkEnumTy
  let norm = bindSym"normalize"
  result = nnkCaseStmt.newTree(nnkDotExpr.newTree(argSym, norm))
  for f in impl:
    case f.kind
    of nnkEmpty: continue # skip first node of `enumTy`
    of nnkSym, nnkIdent:
      result.add addOfBranch(f.strVal, f, typ)
    of nnkEnumFieldDef:
      case f[1].kind
      of nnkStrLit: 
        result.add addOfBranch(f[1].strVal, f[0], typ)
      of nnkTupleConstr:
        result.add addOfBranch(f[1][1].strVal, f[0], typ)
      of nnkIntLit:
        result.add addOfBranch(f[0].strVal, f[0], typ)
      else: error("Invalid tuple syntax", f[1])
    else: error("Invalid node for enum type", f)
  # finally add else branch to raise or use default
  if default == nil:
    let raiseStmt = quote do:
      raise newException(ValueError, "Invalid enum value: " & $`argSym`)
    result.add nnkElse.newTree(raiseStmt)
  else:
    expectKind(default, nnkSym)
    result.add nnkElse.newTree(default)

proc parseEnum*[T: enum](s: string): T =
  ## Parses an enum ``T``.
  ##
  genEnumStmt(T, s, default = nil)
```

lib/pure/strutils.nim
Vindaar: Good catch, I forgot that that's how bools work.

edit: well, nevermind. Using `parseEnum` with `bool` results in a "cannot instantiate: `parseEnum[bool]`" error. Better to use `nil` anyways though.

lib/pure/strutils.nim
```diff
@@ -1229,6 +1230,54 @@ proc parseBool*(s: string): bool =
   of "n", "no", "false", "0", "off": result = false
   else: raise newException(ValueError, "cannot interpret as a bool: " & s)
 
+proc addOfBranch(s: string, field, enumType: NimNode): NimNode =
+  result = nnkOfBranch.newTree(
+    newLit normalize(s),
+    nnkCall.newTree(enumType, field) # `T(<fieldValue>)`
+  )
+
+macro genEnumStmt(typ, argSym: typed,
```
cooldome: For better clarity we can more accurately specify arg type: `typ: typedesc`

lib/pure/strutils.nim
```diff
@@ -1229,6 +1230,54 @@ proc parseBool*(s: string): bool =
   of "n", "no", "false", "0", "off": result = false
   else: raise newException(ValueError, "cannot interpret as a bool: " & s)
 
+proc addOfBranch(s: string, field, enumType: NimNode): NimNode =
+  result = nnkOfBranch.newTree(
+    newLit normalize(s),
+    nnkCall.newTree(enumType, field) # `T(<fieldValue>)`
```
cooldome, line 1256: With https://github.com/nim-lang/Nim/pull/14048 we can use field syminstead of `nnkCall.newTree(enumType, field)`

Vindaar:
The test of `nimx` fails on MacOS, because of this enum here:

https://github.com/yglukhov/darwin/blob/master/darwin/foundation/nsnumber.nim#L8-L21

So this kind of means we cannot apply `normalize` to the argument and a field if the field has a string value attached to it, no?

edit: the old `parseEnum` implementation would have simply returned the first matching element (which may not be the correct one, if the argument was e.g. `"C"`! So parsing such enums was definitely broken with the old `parseEnum`). How should we handle this? @yglukhov, what do you say?

cooldome:
I think it is a bug in `normalize`. The first character of identifier is case sensitive while rest are not.

Vindaar:
The doc string for `normalize` explicitly states though:

```
That means to convert it to lower case and remove any '_'. This
should NOT be used to normalize Nim identifier names.
```

But still, I'm not sure if it's reasonable to argue that capitalization should only make sense for the first character of the string values? For the actual enum field names, sure. 

There don't seem to be well defined rules how string values are treated, which makes me believe that they are just considered as they are?

And well, apparently even this is allowed:
```nim
type
  Enum = enum
    A = "A"
    B = "A"
    C = "A"

let a = A
let b = B
let c = C
```

cooldome:
IMO,
```nim
type
  Enum = enum
    A = "A"
    B = "A"
    C = "A"
```
should give a reasonable compile time error user can understand. You can't parse this one.

For the rest we should use standard nim identifier convention:
```nim
proc nimIdentNormalize*(s: string): string =
  ## Normalizes the string `s`as Nim identifier.
  ##
  ## That means to convert to lower case and remove any '_' on all characters except
  ## first one.
  runnableExamples:
    doAssert normalize("Foo_bar") == "Foobar"
  result = newString(s.len)
  if s.len > 0:
    result[0] = s[0]
  var j = 0
  for i in 1..len(s) - 1:
    if s[i] in {'A'..'Z'}:
      result[j] = chr(ord(s[i]) + (ord('a') - ord('A')))
      inc j
    elif s[i] != '_':
      result[j] = s[i]
      inc j
  if j != s.len: setLen(result, j)
```

Vindaar:
Ah, thanks to GitHub being messed up again, my commit didn't appear.

I used your `nimIdentNormalize` impl @cooldome (well, fixed it `j = 1` first) and now track which field names have already been encountered. If duplicates are found a CT error is thrown.

lib/pure/strutils.nim
```diff
@@ -1229,8 +1250,65 @@ proc parseBool*(s: string): bool =
...
+      of nnkStrLit: fStr = f[1].strVal
+      of nnkTupleConstr:
+        fStr = f[1][1].strVal
+        fNum = f[1][0].intVal
+      of nnkIntLit:
+        fStr = f[0].strVal
+        fNum = f[1].intVal
+      else: error("Invalid tuple syntax!")
+    else: error("Invalid node for enum type!")
+    # add field if string not already added
```
cooldome, line 1291: Once call to nimIdentNormalize(fStr) can be saved:
```fStr = nimIdentNormalize```
and then use `fStr`

cooldome:
Looks good, I made a couple of minor comments on style in code review. Please have a look.
Otherwise very good work

Vindaar:
I think your comments may have fallen victim to GitHub's outage. Can't see them.

cooldome:
I hope you see my comments now

lib/pure/strutils.nim
Vindaar: Just tried it. Unfortunately that does not solve the problem if the enum is defined in a block. Still results in the same "undeclared identifer" error.

Vindaar:
Yes, I could see your comments now. I rebased on current devel and applied the changes, except the one to use the symbol directly. That still doesn't work if the enum isn't defined at top level.

cooldome:
Test failures are unrelated, merging

PR #14065, opened 2020-04-22
use newer nodejs on Azure Pipelines

narimiran:


Varriount:
@narimiran Wouldn't 10 make more sense, since it's the oldest long-term support release? Do you know if we have a statement on what versions of Node Nim supports?

Araq:
We don't support nodejs at all, we merely use it for testing.

PR #14069, opened 2020-04-22
change some Exceptions to CatchableError or Defect, fixes #10288

### LINKED ISSUE #10288: Exception types in the stdlib should inherit from `CatchableError` or `Defect`, not `Exception`
Exception types should always inherit from `CatchableError` or `Defect` (directly or indirectly) to indicate if they should be catched or not. Here is a list of exception types in the stdlib that breaks this rule:

- [ ] `nre.RegexError`
- [ ] `asyncfutures.FutureError`
- [ ] `net.SslError`
- [ ] `net.TimeoutError`
- [ ] `punycode.PunyError`
- [ ] `selectors.IOSelectorsException`

metagn:
fixes #10288

### DIFF
--- a/lib/impure/nre.nim
+++ b/lib/impure/nre.nim
@@ -198,7 +198,7 @@ type
   Captures* = distinct RegexMatch
   CaptureBounds* = distinct RegexMatch
 
-  RegexError* = ref object of Exception
+  RegexError* = ref object of CatchableError
 
   RegexInternalError* = ref object of RegexError
     ## Internal error in the module, this probably means that there is a bug
--- a/lib/pure/asyncfutures.nim
+++ b/lib/pure/asyncfutures.nim
@@ -33,7 +33,7 @@ type
 
   FutureVar*[T] = distinct Future[T]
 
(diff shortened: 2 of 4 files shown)


lib/impure/nre.nim
```diff
@@ -198,7 +198,7 @@ type
   Captures* = distinct RegexMatch
   CaptureBounds* = distinct RegexMatch
 
-  RegexError* = ref object of Exception
+  RegexError* = ref object of CatchableError
 
   RegexInternalError* = ref object of RegexError
```
hlaaftana, line 203: I debated making RegexInternalError extend Defect instead of RegexError but didn't do it, should I do it?

metagn, line 203: I debated making RegexInternalError extend Defect instead of RegexError but didn't do it, should I do it?

arnetheduck:
1.2 backport?

lib/impure/nre.nim
Araq: RegexErrors are ValueErrors

PR #14070, opened 2020-04-22
fix #14064 xmltree should allow create text node with raw text(non-es…

### LINKED ISSUE #14064: xmltree need add raw text, when add style element
https://github.com/nim-lang/Nim/blob/dd0e823a0ca94ada3a8e155838da07f76e223012/lib/pure/xmltree.nim#L648

bung87:
…caped) eg. html style element's text


Araq:
Can we name it ``VerbatimText``? It also needs a changelog entry. And a ``.since`` annotation though that's probably gonna be too difficult...

bung87:
yeah , remind me `rawText` conflict with `proc rawText`, but how to these two task,
`invalid pragma: since: (1, 2)` it gives me.

about change log:

add to here,right? https://github.com/nim-lang/Nim/blob/devel/changelogs/changelog_1_2_0.md

Araq:
```nim

include inclrtl

proc newVerbatimText*(text: string): XmlNode {.since: (1, 3).} =
  ...
```

PR #14099, opened 2020-04-24
Faster readStr()

### LINKED ISSUE #13857: streams.readStr() allocates new string on each call
### Summary
We have

`streams.readLine(s: var string)`

but we also have

`streams.readStr(length: int): string`

which allocates a new string on each call.

### Description
Usually, the purpose of `readStr()` is to be called repeatedly, on same-length or similar-length records, so I'd like to re-use the allocated string memory. It's a minor optimization.

### Alternatives
The alternative is to avoid the **streams** library.

### Additional Information
I'm happy to code this change myself. Would it be accepted?

cdunn2001:
Avoid extra memory allocations, similar to existing `readLine(stream, length, str)`

resolves #13857

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -37,6 +37,9 @@
   and this can now throw in edge cases where `getCurrentDir` throws.
   `relativePath` also now works for js with `-d:nodejs`.
 
+- Added `streams.readStr` and `streams.peekStr` overloads to
+  accept an existing string to modify, which avoids memory
+  allocations, similar to `streams.readLine` (#13857).
 
 ## Language changes
 - In newruntime it is now allowed to assign discriminator field without restrictions as long as case object doesn't have custom destructor. Discriminator value doesn't have to be a constant either. If you have custom destructor for case object and you do want to freely assign discriminator fields, it is recommended to refactor object into 2 objects like this:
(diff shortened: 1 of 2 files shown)


Araq:
Needs a .since annotation and a changelog entry.

pb-cdunn:
Ok. Done.

cdunn2001:
I don't understand the new nim-docs error (after my most recent commit):
```
Hint: streams [Processing]
/home/runner/work/Nim/Nim/lib/pure/streams.nim(863, 10) Error: type mismatch: got <Stream, int, TaintedString>
but expected one of: 
proc readStr(s: Stream; length: int): TaintedString

expression: readStr(s, length, result)
```
Something related to overloading? The code tests are fine.

alaviss:
A `since` annotation restrict the `proc` to only be available in that nim version. In here, `{.since: (1, 3).}` make it only available for compiler newer than version 1.3, but the compiler is bootstrapped from `csources` which is `0.20`, so the `proc` wasn't available then.

You should move the implementation into a hidden `template`/`proc` then use it to implement the functions.

cdunn2001:
Done!

PR #14105, opened 2020-04-24
add FileReader Web API to js dom

cooldome:
FileReader is web api that is used to display content of file before uploading it.

Details of API can be found here:
   https://developer.mozilla.org/en-US/docs/Web/API/FileReader
   https://w3c.github.io/FileAPI/#dfn-filereader

All new types and procs are within `since(1,3)`


lib/js/dom.nim
```diff
@@ -39,6 +39,10 @@ type
     onselect*: proc (event: Event) {.nimcall.}
     onsubmit*: proc (event: Event) {.nimcall.}
     onunload*: proc (event: Event) {.nimcall.}
+    when (NimMajor, NimMinor) >= (1, 3):
```
Clyybber:
```suggestion
    since (1, 3):
```

cooldome: You can't use templates insode object type definitions, only direct when statements are allowed

cooldome: In the end I had to remove `when` as well. It doesn't work, semfold chokes.

lib/js/dom.nim
```diff
@@ -1318,6 +1345,29 @@ proc offsetLeft*(e: Node): int {.importcpp: "#.offsetLeft", nodecl.}
 since (1, 3):
   func newDomParser*(): DOMParser {.importcpp: "new DOMParser()".}
     ## DOM Parser constructor.
-
   func parseFromString*(this: DOMParser; str: cstring; mimeType: cstring): Document {.importcpp.}
     ## Parse from string to `Document`.
+
+  proc newDomException*(): DomException {.importcpp: "new DomException()", constructor.}
```
juancarlospaco, line 1351:
```suggestion
  proc newDomException*(): DomException {.importcpp: "(new DomException())", constructor.}
```

lib/js/dom.nim
```diff
@@ -1318,6 +1345,29 @@ proc offsetLeft*(e: Node): int {.importcpp: "#.offsetLeft", nodecl.}
...
+    ## DOM Exception constructor
+  proc message*(ex: DomException): cstring {.importcpp: "#.message", nodecl.}
+    ## https://developer.mozilla.org/en-US/docs/Web/API/DOMException/message
+  proc name*(ex: DomException): cstring  {.importcpp: "#.name", nodecl.}
+    ## https://developer.mozilla.org/en-US/docs/Web/API/DOMException/name
+
+  proc newFileReader*(): FileReader {.importcpp: "new FileReader()", constructor.}
```
juancarlospaco, line 1358:
```suggestion
  proc newFileReader*(): FileReader {.importcpp: "(new FileReader())", constructor.}
```

cooldome: I don't think you right, wrapping in the parenthesis is not required, have a look how all other objects in this module are implemented

lib/js/dom.nim
cooldome: I don't think you right, wrapping in the parenthesis is not required, have a look how all other objects in this module are implemented

lib/js/dom.nim
```diff
@@ -39,6 +39,9 @@ type
     onselect*: proc (event: Event) {.nimcall.}
     onsubmit*: proc (event: Event) {.nimcall.}
     onunload*: proc (event: Event) {.nimcall.}
+    onloadstart*: proc (event: Event) {.nimcall.}
```
Araq, line 42: well now this can work inside a ``when``.

cooldome: ok, let's try

cooldome:
Ready for review

cooldome:
tried. Ci can't pickup nim changes that quickly. reverted last change.

PR #14107, opened 2020-04-24
Allow vm in nkRecWhen

cooldome:
The fix is to use `forceBool(c, c.semConstExpr(c, n))` instead of `c.semConstBoolExpr(c, n)`.
This allows for `when (NimMajor, NimMinor) >= (1,1)` in object definitions.  

Also noticed that semConstBoolExpr is not used often so I am trying to remove it and use `semConstExpr` everywhere


cooldome:
alternative PR is proposed

Issue #7422, opened 2018-03-27
Async networking and blocking host name resolving / DNS


zevv:
I believe the current mechanism used for hostname lookups interferes with truly asynchronous networking. The async dial() ends up using the getaddrinfo() C library call, which can show serious blocking behaviour.

This is a common problem in network programming, with no obvious solutions, but it would be nice to offer a truly async experience for the Nim user without unexpected blocking when doing high level networking, like AsyncHttpClient.

Some ideas:

- Can Nim wrap the getaddrinfo() call in a separate thread to do the lookups? This requires threading to be enabled, oviously
- Should Nim implement (a subset of) getaddrinfo() asynchronously in pure Nim?
- Should Nim be able to rely on an external library for lookups, like libc-ares?

dom96:
This also extends to our handling of asynchronous file reading. Currently ``asyncfile`` isn't truly async as far as I remember. I think Node solves this issue by using a separate thread, sadly doing so in Nim is a bit tough.

zevv:
Are there any guidelines on combining async and threads in Nim? I'm trying to build a local workaround with a resolver thread, but currently I need to loop doing a channel poll() and a sleepAsync() until the resolver is done.

euantorano:
I've had a quick look, and it looks like there are a few platform specific APIs to resolve names asynchronously:

- **Windows**: [`DnsQueryEx`](https://msdn.microsoft.com/en-us/library/windows/desktop/hh447188(v=vs.85).aspx)
- **glibc**: [`getaddrinfo_a`](http://man7.org/linux/man-pages/man3/getaddrinfo_a.3.html)
- **OpenBSD**: [`getaddrinfo_async`](https://man.openbsd.org/asr_run.3)

Some interesting links on the topic:

- https://medium.com/where-the-flamingcow-roams/asynchronous-name-resolution-in-c-268ff5df3081
- https://github.com/clementfarabet/async/blob/master/luv/libuv/src/unix/getaddrinfo.c

zevv:
My preferred solution would be something like the first option above: wrap the standard getaddrinfo() into a thread, and provide minimal infrastructure for hooking this into asyncdispatch. My only problem is that for this I either need a poll/sleep loop (not nice) or posix pipes (not portable), unless there is some mechanism I don't know about that can connect threads and async. (On IRC, dom96 mentioned FlowVar for this?)

euantorano:
[...]

PR #4683, opened 2016-08-31
Add async IO operations with buffers on files and sockets

vegansk:
Sometimes it's better to use read/write operations on buffers. Syncronous IO already had that functions (for example http://nim-lang.org/docs/system.html#readBuffer,File,pointer,Natural). This PR adds them to asynchronous IO.

PR #14109, opened 2020-04-25
added high level sendTo and recvFrom to std/asyncnet (UDP functionality)

### LINKED ISSUE #6956: asyncnet missing udp functionality
At the very least it is missing sendTo and recvFrom.

rockcavera:
Fix [#6956](https://github.com/nim-lang/Nim/issues/6956)

### DIFF
--- a/lib/pure/asyncnet.nim
+++ b/lib/pure/asyncnet.nim
@@ -780,6 +780,122 @@ proc isClosed*(socket: AsyncSocket): bool =
   ## Determines whether the socket has been closed.
   return socket.closed
 
+proc sendTo*(socket: AsyncSocket, data: pointer, dataSize: int, address: string,
+             port: Port, flags = {SocketFlag.SafeDisconn}): owned(Future[void])
+            {.async.} =
+  ## This proc sends ``data`` to the specified ``address``, which may be an IP
+  ## address or a hostname. If a hostname is specified this function will try
+  ## each IP of that hostname. The returned future will complete once all data
+  ## has been sent.
+  ## 
+  ## If an error occurs an OSError exception will be raised.
+  assert(socket.protocol != IPPROTO_TCP, "Cannot `sendTo` on a TCP socket")
+  assert(not socket.closed, "Cannot `sendTo` on a closed socket")
+
+  let
+    aiList = getAddrInfo(address, port, socket.domain, socket.sockType,
+                         socket.protocol)
+    retFuture = newFuture[void]("sendTo")
+
+  var
+    it = aiList
+    success = false
+    lastException: ref Exception
+  
+  while it != nil:
+    let fut = sendTo(socket.fd.AsyncFD, data, dataSize, it.ai_addr,
+                   it.ai_addrlen.SockLen, flags)
+    
+    yield fut
+
+    if not fut.failed:
+      success = true
+
+      break
+    
+    lastException = fut.readError()
+
+    it = it.ai_next
+  
+  freeaddrinfo(aiList)
+
+  if not success:
+    if lastException != nil:
+      retFuture.fail(lastException)
+
+    else:
+      retFuture.fail(newException(
+            IOError, "Couldn't resolve address: " & address))
+
+  else:
+    retFuture.complete()
+
+proc sendTo*(socket: AsyncSocket, data, address: string, port: Port):
+             owned(Future[void]) {.async.} =
+  ## This proc sends ``data`` to the specified ``address``, which may be an IP
+  ## address or a hostname. If a hostname is specified this function will try
+  ## each IP of that hostname. The returned future will complete once all data
+  ## has been sent.
+  ## 
+  ## If an error occurs an OSError exception will be raised.
+  await sendTo(socket, cstring(data), len(data), address, port)
+
+proc recvFrom*(socket: AsyncSocket, data: pointer, size: int,
+               address: FutureVar[string], port: FutureVar[Port],
+               flags = {SocketFlag.SafeDisconn}): owned(Future[int]) {.async.} =
+  ## Receives a datagram data from ``socket`` into ``data``, which must be at
+  ## least of size ``size``. The address and port of datagram's sender will be
+  ## stored into ``address`` and ``port``, respectively. Returned future will
+  ## complete once one datagram has been received, and will return size of
+  ## packet received.
+  ## 
+  ## If an error occurs an OSError exception will be raised.
+  template awaitRecvFromInto() =
+    var lAddr = sizeof(sAddr).SockLen
+    
+    result = await recvFromInto(AsyncFD(getFd(socket)), data, size,
+                                cast[ptr SockAddr](addr sAddr), addr lAddr)
+                                
+    address.mget.add(getAddrString(cast[ptr SockAddr](addr sAddr)))
+
+    address.complete()
+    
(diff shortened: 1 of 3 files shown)


Araq:
Needs ``.since`` annotation and a changelog entry.

rockcavera:
> Needs `.since` annotation and a changelog entry.

it's done.

lib/pure/asyncnet.nim
```diff
@@ -780,6 +780,123 @@ proc isClosed*(socket: AsyncSocket): bool =
   ## Determines whether the socket has been closed.
   return socket.closed
 
+proc sendTo*(socket: AsyncSocket, data: pointer, dataSize: int, address: string,
```
Varriount, line 783: Does this only work with UDP packets? What happens if this is a TCP socket?

Varriount (changes requested):

lib/pure/asyncnet.nim
rockcavera: I am no expert on this, but looking at the low-level code (asyncdispatch) on Windows, WSASendTo is used, in which the documentation specifies that it is generally used for connectionless sockets. However, there is a note saying that: if used on a connected socket, the address will be ignored and WSASendTo will behave the same as WSASend.

I haven't tested it either. I will test the behavior on Windows. Do you have a solution in mind or something to add?

rockcavera: I tested it now on Windows and the behavior was as expected. That is, when compiled without "-d:danger" it displays the error message caused by the assert. When compiled with "-d:danger", regardless of the address and port used in sendTO, the message was correctly forwarded by the socket to the destination used in connect for TCP protocol.

rockcavera: I will do the same test on linux.

rockcavera: I tested on linux and it happens the same as in Windows. That is, there is no problem using sendTo on TCP connections, however, it is not the most recommended.

Can we end this?

lib/pure/asyncnet.nim
```diff
@@ -780,6 +780,123 @@ proc isClosed*(socket: AsyncSocket): bool =
...
+  ## address or a hostname. If a hostname is specified this function will try
+  ## each IP of that hostname. The returned future will complete once all data
+  ## has been sent.
+  ## 
+  ## If an error occurs an OSError exception will be raised.
+  assert(socket.protocol != IPPROTO_TCP, "Cannot `sendTo` on a TCP socket")
+  assert(not socket.closed, "Cannot `sendTo` on a closed socket")
+
+  let
+    aiList = getAddrInfo(address, port, socket.domain, socket.sockType,
```
Varriount, line 799: I guess my main concern is that clients using TCP connections will use this instead of `send`, and thus incur the cost of DNS lookup each time they send data (it wouldn't be immediately obvious either, since the behavior is to ignore the hostname). This is in contrast to the C version, which requires the user to pass in a sockaddr data structure.

Additionally, I also find it conceptually confusing (and this is also a criticism of the original C functions) that `sendTo` works with TCP and UDP, while `recvFrom` works with just UDP.

That being said, I'm not super-familiar with UDP and sendTo/recvFrom, so perhaps there's a reason that TCP works for sending data.

lib/pure/asyncnet.nim
```diff
@@ -780,6 +780,123 @@ proc isClosed*(socket: AsyncSocket): bool =
...
+  ## has been sent.
+  ## 
+  ## If an error occurs an OSError exception will be raised.
+  await sendTo(socket, cstring(data), len(data), address, port)
+
+proc recvFrom*(socket: AsyncSocket, data: pointer, size: int,
+               address: FutureVar[string], port: FutureVar[Port],
+               flags = {SocketFlag.SafeDisconn}): owned(Future[int])
+              {.async, since: (1, 3).} =
+  ## Receives a datagram data from ``socket`` into ``data``, which must be at
```
Varriount, line 852: I know that the word `datagram` is used (referring to UDP), but I feel that putting a statement saying that this should only be used with UDP sockets, and that `reecv` should be used instead, would be helpful.

lib/pure/asyncnet.nim
rockcavera: So, this DNS query cost, if a hostname is sent, is a cost that the person will have, if they use the less appropriate proc. The assert is there, but I can improve the message, as well as add comments in the proc to use the send() proc instead of sendTo().

lib/pure/asyncnet.nim
rockcavera: I will improve the description here.

dom96:
This API really isn't ideal, asyncnet should offer as little APIs providing `pointer` as possible. @Araq let's keep at least the high level modules memory safe...

Can we change the `recvFrom` to at least return a string instead of taking a `pointer`? Seriously, this is just a recipe for disaster.

rockcavera:
@dom96, I knew that the getAddrInfo() part was not ideal, but connect() and dial() asyncnet also use it, as it does not have an asynchronous getAddrInfo. Even have this issue open [#7422](https://github.com/nim-lang/Nim/issues/7422)

Regarding the use of pointer I did not know, as I saw other procs using asyncnet. But I thought of putting recvFrom returning string. Now I don't remember why I didn't do it in the second recvFrom.

Araq:
@rockcavera Please address dom96's remarks, I was too quick to merge this PR.

dom96:
> @dom96, I knew that the getAddrInfo() part was not ideal, but connect() and dial() asyncnet also use it, as it does not have an asynchronous getAddrInfo. Even have this issue open #7422

Oh, I don't mind that. It's definitely not ideal but we're pretty far off having an async getAddrInfo and once we do we can plop it in fairly easily.

API issues on the other hand cannot be easily fixed once they make it into the stdlib, which is why I am so critical of them.

From my side I never wanted these `pointer`-based APIs (you can see my comments to that effect in the PR that added them #4683). Personally I'd like to just deprecate them and eventually get rid of them as I cannot see them giving much of a performance benefit (they're still allocating a string + a future, they're just doing it in a much less memory safe way).

rockcavera:
@dom96, I believe I understand what you want. Do not use the pointer type in the parameters, at least in the high level API.

One question I had: what about the use of pointers in the middle of the procs?

I believe sendTO looks like this:
``` nim
proc sendTo*(socket: AsyncSocket, data, address: string, port: Port): owned(Future[void]) {.async, since: (1, 3).} =
```

That is, for what uses a pointer in here.

RecvFrom instead of returning the content received by a parameter, use a tuple. Staying like this:
``` nim
proc recvFrom*(socket: AsyncSocket, size: int): owned(Future[tuple[read, address: string, port: Port]]) {.async, since: (1, 3).} =
```

That is, the procs that use pointers are removed. Is there a problem with the tuple? I believe it is easier than creating an object.

rockcavera:
@dom96, I made the changes. Now have two procs, sendTo and recvFrom, and not four as before. Without pointers, of course.

However, there was a doubt: I am using `newFuture()` for tracking, I believe it is advisable. However, if I want to `raise`, should I do it by `future.read()`, after `complete()`, or a direct `raise` in the asynchronous proc?

Before I had the false perception that `future.fail()` already launched a `raise`. Living and learning.

dom96:
``raise`` in an async proc will call `fail` on the resulting future. If you are not in an async proc (but are completing futures manually) then you can use `fail` directly.

rockcavera:
I opened a new pull.

PR #14135, opened 2020-04-27
Fix issue with stringifying incomplete types

### LINKED ISSUE #14066: nim jsondoc -o:doc.json filename.nim fails when sequences without a type are used
Running "nim jsondoc -o:doc.json filename.nim" results in:
Error: unhandled exception: index out of bounds, the container is empty [IndexError]

### Example
```nim
proc hashesToStr(hashes: seq): string = 
  var h:string
  if hashes.len == 0:
      result = ""
  else:
    for hash in hashes:
      h = h & hash & "|"
    h = h[0 .. len(h)-2] # cuts of the last |
    result = h
```

### Additional Information
Already discussed and found a workaround with PMunch on IRC.

```
$ nim -v
Nim Compiler Version 1.2.0
```

PMunch:
When a type such as `seq` is used without the type argument, which is
allowed in a template for example, the jsondoc document generation would
fail. This fixes #14066

### DIFF
--- a/compiler/types.nim
+++ b/compiler/types.nim
@@ -623,7 +623,7 @@ proc typeToString(typ: PType, prefer: TPreferedDesc = preferName): string =
     of tyOpenArray:
       result = "openArray"
       if t.len > 0:
-        result = "[" & typeToString(t[0]) & ']'
+        result &= "[" & typeToString(t[0]) & ']'
     of tyDistinct:
       result = "distinct " & typeToString(t[0],
         if prefer == preferModuleInfo: preferModuleInfo else: preferTypeName)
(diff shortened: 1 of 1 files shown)


narimiran:
> This fixes #14066

Since it looks like you'll have to make some additional changes, can you please add `fixes #14066` to the commit message, otherwise closing of that issue won't be automatic, AFAIK.

PMunch:
> > This fixes #14066
> 
> Since it looks like you'll have to make some additional changes, can you please add `fixes #14066` to the commit message, otherwise closing of that issue won't be automatic, AFAIK.

That line is enough to auto-close the issue when merged

narimiran:
Unrelated fails. Merging.

PR #14137, opened 2020-04-27
Add RSA key reading and encrypt/decrypt to openssl

PMunch:
This adds the procedures to read RSA keys and encrypt/decrypt messages
with both public and private keys.


PMunch:
Pretty sure none of those issues are caused by this change. It simply adds some more procedures to the openssl module, it doesn't modify anything, so nothing should break from it (unless somewhere there are procedures by the same names in other modules).

Araq:
Indeed, irrelevant CI failure, merging.

PR #14147, opened 2020-04-28
Update grammar.txt with `func` and `as`

metagn:


Araq:
Ah bah, but this is a generated file...

hlaaftana:
Maybe generate again?

zah:
BTW, what's the plan for the `as` operator? We recently considered starting overloading it as a "explicit type conversion" operator.

metagn:
Maybe a call to another symbol like `in` calls `contains`. Something like:

```nim
template `as`(x: untyped, T: typedesc): T =
  var converted: T
  convert(x, converted)
  converted
```

and things like serialization/game libraries can overload it. There should be an RFC about it.

Edit: Closest RFC is https://github.com/nim-lang/RFCs/issues/84, I'll make a comment there about it

Araq:
I have no plans for ``as``, it's yours to take. :-)

PR #14149, opened 2020-04-28
Fix regression: hashes module doesn't compile with vcc

cooldome:
This is showstopper regression.
Code doesn't compile because `_umul128` is in `intrin.h` and this header file is not included.

### DIFF
--- a/lib/pure/hashes.nim
+++ b/lib/pure/hashes.nim
@@ -99,7 +99,10 @@ proc hiXorLo(a, b: uint64): uint64 {.inline.} =
     elif defined(gcc) or defined(llvm_gcc) or defined(clang):
       {.emit: """__uint128_t r = a; r *= b; `result` = (r >> 64) ^ r;""".}
     elif defined(windows) and not defined(tcc):
-      {.emit: """a = _umul128(a, b, &b); `result` = a ^ b;""".}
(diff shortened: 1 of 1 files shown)


Clyybber:
cc @c-blake

c-blake:
How did the pre-fixing version pass the Azure CI?  Does the Windows branch of that CI not use `vcc` for some reason? Is it a `vcc` version skew thing?

Anyway, the fix looks fine, but also are we sure that `defined(windows) and not defined(tcc)` is basically selecting `vcc`?  Should the guard be `defined(vcc)` or maybe `defined(windows) and defined(vcc)` instead?  A quick search in the whole Nim repo yields varying answers.

ghost:
@c-blake yeah, seems a bit wrong since you can have GCC (mingw), Clang on Windows too

c-blake:
Those should be (?) caught by the first `elif defined(gcc) or defined(llvm_gcc) or defined(clang)` and that `__uint128_t` fork should work ok (at least or recent versions...Not entirely sure when `__uint128_t` was introduced).

Anyway, it could definitely use some multi-compiler testing on Windows.  I was never sure how much the CI tested, exactly.

Araq:
@c-blake VCC isn't tested at all, unfortunately.

PR #14154, opened 2020-04-29
fix sendTo and recvFrom in asyncnet

rockcavera:
* As discussed in pull [#14109](https://github.com/nim-lang/Nim/pull/14109), the pointers on parameters of the high-level procs were removed.
* Now there are only 2 procs. Before there were 4.

Open to suggestions.


lib/pure/asyncnet.nim
```diff
@@ -780,9 +780,9 @@ proc isClosed*(socket: AsyncSocket): bool =
   ## Determines whether the socket has been closed.
   return socket.closed
 
-proc sendTo*(socket: AsyncSocket, data: pointer, dataSize: int, address: string,
-             port: Port, flags = {SocketFlag.SafeDisconn}): owned(Future[void])
-            {.async, since: (1, 3).} =
+proc sendTo*(socket: AsyncSocket, data: string, address: string, port: Port,
+             flags = {SocketFlag.SafeDisconn}): owned(Future[void])
+            {.async, since: (1, 3, 1).} =
```
dom96: Pretty sure this `, 1` isn't necessary.

lib/pure/asyncnet.nim
```diff
@@ -780,9 +780,9 @@ proc isClosed*(socket: AsyncSocket): bool =
   ## Determines whether the socket has been closed.
   return socket.closed
 
-proc sendTo*(socket: AsyncSocket, data: pointer, dataSize: int, address: string,
-             port: Port, flags = {SocketFlag.SafeDisconn}): owned(Future[void])
-            {.async, since: (1, 3).} =
+proc sendTo*(socket: AsyncSocket, data: string, address: string, port: Port,
```
dom96: Please change order of arguments so that it matches the ones in `net`: https://nim-lang.org/docs/net.html#sendTo%2CSocket%2Cstring%2CPort%2Cstring

lib/pure/asyncnet.nim
```diff
@@ -824,88 +822,58 @@ proc sendTo*(socket: AsyncSocket, data: pointer, dataSize: int, address: string,
...
-  ## This proc sends ``data`` to the specified ``address``, which may be an IP
-  ## address or a hostname. If a hostname is specified this function will try
-  ## each IP of that hostname. The returned future will complete once all data
-  ## has been sent.
+      raise newException(IOError, "Couldn't resolve address: " & address)
+
+proc recvFrom*(socket: AsyncSocket, size: int,
+               flags = {SocketFlag.SafeDisconn}):
+              owned(Future[tuple[data: string, address: string, port: Port]])
+              {.async, since: (1, 3, 1).} =
```
dom96: This is fine, but if you're up for it, a `recvFrom` taking a `FutureVar` and matching the net.recvFrom proc (https://nim-lang.org/docs/net.html#recvFrom%2CSocket%2Cstring%2Cint%2Cstring%2CPort%2Cint32) would be nice. Might be better to create it in a separate PR.

lib/pure/asyncnet.nim
```diff
@@ -824,88 +822,58 @@ proc sendTo*(socket: AsyncSocket, data: pointer, dataSize: int, address: string,
...
-
-    awaitRecvFromInto()
+    yield fut
 
-    port.complete(ntohs(sAddr.sin6_port).Port)
-    
-  else:
-    var sAddr: Sockaddr_in
+    if fut.failed():
+      raise fut.readError()
```
dom96: Can you use ``await`` here instead, it has the same semantics (exceptions are re-thrown).

dom96 (approved):
Overal looks good, but please fix the `sendTo` API.

lib/pure/asyncnet.nim
rockcavera: It's done.

lib/pure/asyncnet.nim
rockcavera: It's done.

lib/pure/asyncnet.nim
rockcavera: It's done.

lib/pure/asyncnet.nim
rockcavera: Let me see if I understand, because my English is not the best.

Create another proc recvFrom (using net.recvFrom as a model), but using FutureVars in the parameters? In that, would be two recvFrom in asyncnet?

dom96: yep, that's correct. You can do this in a separate PR.

PR #14172, opened 2020-04-30
Fix compilations using tcc in nim 1.2+ [backport]

### LINKED ISSUE #14171: Compiling with tcc needs -DNIM_EmulateOverflowChecks [nim 1.2.x+]
Compilations break when compiling with `--cc:tcc`.

### Example
```nim
echo 0
```

### Current Output
```
CC: stdlib_io.nim
CC: stdlib_system.nim
CC: t.nim
/tmp/kmodi/.cache/nim/t_d/stdlib_io.nim.c:56: error: invalid type
Error: execution of an external compiler program 'tcc -c -w   -I/home/kmodi/usr_local/apps/6/nim/devel/lib -I/home/kmodi/sandbox/nim/tcc -o /tmp/kmodi/.cache/nim/t_d/stdlib_io.nim.c.o /tmp/kmodi/.cache/nim/t_d/stdlib_io.nim.c' failed with exit code: 1


/tmp/kmodi/.cache/nim/t_d/stdlib_system.nim.c:457: error: invalid type
```

**Here is the full [`stdlib_io.nim.c`][2]**.

### Expected Output
```
0
```

### Possible Solution

* Change the `tcc.options.always` in nim's default `nim.cfg` to `tcc.options.always = "-w -DNIM_EmulateOverflowChecks"` (**Credit**: @c-blake )

### Additional Information

- Apparently compilations using `tcc` worked just fine earlier, but it probably broke in nim 1.2.0+ (based on [its release notes][1]).
- Discovered this as I was trying to get the fastest possible nim compilations in https://github.com/c-blake/cligen/issues/139#issuecomment-621883694 

```
$ nim -v
Compiled at 2020-04-30

git hash: d6e1b5c070bf58449bf4af210ec099418d496aa5
```

[1]: https://nim-lang.org/blog/2020/04/03/version-120-released.html

[2]: http://ix.io/2kbp/c

kaushalmodi:
Realistically, `tcc` is never going to grow all those gcc/clang
`__builtin_*`s.

Credit: @c-blake

Fixes https://github.com/nim-lang/Nim/issues/14171 .

### DIFF
--- a/lib/nimbase.h
+++ b/lib/nimbase.h
@@ -566,7 +566,7 @@ NIM_STATIC_ASSERT(sizeof(NI) == sizeof(void*) && NIM_INTBITS == sizeof(NI)*8, ""
 #define nimModInt(a, b, res) (((*res) = (a) % (b)), 0)
 #define nimModInt64(a, b, res) (((*res) = (a) % (b)), 0)
 
-#if (!defined(_MSC_VER) || defined(__clang__)) && !defined(NIM_EmulateOverflowChecks)
+#if (!defined(_MSC_VER) || defined(__clang__)) && !defined(NIM_EmulateOverflowChecks) && !defined(__TINYC__)
   /* these exist because we cannot have .compilerProcs that are importc'ed
     by a different name */
 


kaushalmodi:
@Araq Does this look good to backport?

Araq:
No, looks wrong to me and in fact, we need to redo this differently.

c-blake:
If only `gcc`/`clang` support those `__builtin_*_overflow` then you might want to just have a "purely positive filter".. `#if (defined(gcc) || defined(clang)) && !defined(NIM_EmulateOverflowChecks)` sorta thing.  Then ZigCC/whatever backends will work, too.

Seems like one could make that positive filter idea even more precise.  I don't know about `clang`, but `gcc` has `__GNUC__` cpp macros to test `gcc` versions.  Not sure why the release notes for Nim-1.2 say checking for `__builtin_*_overflow` could not be made reliable.  Might need to do `#if __GNUC__ > 5` or just whatever the first major version of `gcc` supporting them was (and similarly for `clang`).  Other parts of `nimbase.h` use `__GNUC__`.  (Hopefully the gcc/clang guys did not only define them for just some CPUs or something whacky like that...)

c-blake:
The respective versions of gcc and clang when `__builtin_*_overflow` were introduced are gcc-5.0 and clang-3.5.  So, something like this should work:
```C
#if !defined(NIM_EmulateOverflowChecks) && ( \
    __GNUC__ >= 5 || __clang_major__ * 100 + __clang_minor__ >= 305)
...
#endif
```
I understand it's not as elegant as `__has_builtin`, but it should work reliably even in cross compiling environments/whatnot.  If another compiler ever grows support that can be added easily enough.

cooldome:
I like this fix. @Araq, what would you like to change in this fix?

Araq:
This is how to do it, https://github.com/nim-lang/Nim/pull/14213

PR #14176, opened 2020-04-30
discardable async procs are now an error

### LINKED ISSUE #11912: Add undiscardable pragma and forbid Futures from being discarded
### Summary
The docs say that we shouldn't discard futures, but Nim should really catch that on compile time if possible.

### Description
The easiest way should be by adding an {.undiscardable.} pragma for types, using it for Future and making nim generate a compile time error discarding such types on sem checking: I can open a PR, if this is accepted

### Alternatives
Just forbidding discarding Future, but not sure how this will work without somehow hardcoding that (the user might have different Future type)

### LINKED ISSUE #12912: Pragmas async+discardable should be a Syntax Error
Pragmas `{.async.}` and `{.multisync.}` combined with `{.discardable.}` 
should be a Syntax Error or at least a compile time Warning.

### Example
```nim
import asyncdispatch

proc foo(): Future[void] {.async, discardable.} = discard
proc bar(): Future[void] {.multisync, discardable.} = discard

```

Similar but not the same https://github.com/nim-lang/Nim/issues/11912#issue-478298948
recommends adding an  `{.undiscardable.}` pragma for `discard asyncProc()`, 
this bug is about the combination of `{.async, discardable.}` should be a Syntax Error.
I guess should not be too complicated to get a list of pragmas and if both found together thrown error.

metagn:
fixes #12912 

Also had discard as an error (#11912) but decided against it.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -61,6 +61,9 @@
 - `paramCount` & `paramStr` are now defined in os.nim instead of nimscript.nim for nimscript/nimble.
 - `dollars.$` now works for unsigned ints with `nim js`
 
+- Discarding async futures is now an error caught at compile time thanks to term rewriting macros.
+  `asyncCheck` should be used instead.
+
 ## Language changes
 - In newruntime it is now allowed to assign discriminator field without restrictions as long as case object doesn't have custom destructor. Discriminator value doesn't have to be a constant either. If you have custom destructor for case object and you do want to freely assign discriminator fields, it is recommended to refactor object into 2 objects like this:
   ```nim
--- a/changelog.md
+++ b/changelog.md
@@ -61,9 +61,6 @@
 - `paramCount` & `paramStr` are now defined in os.nim instead of nimscript.nim for nimscript/nimble.
 - `dollars.$` now works for unsigned ints with `nim js`
 
-- Discarding async futures is now an error caught at compile time thanks to term rewriting macros.
-  `asyncCheck` should be used instead.
-
 ## Language changes
 - In newruntime it is now allowed to assign discriminator field without restrictions as long as case object doesn't have custom destructor. Discriminator value doesn't have to be a constant either. If you have custom destructor for case object and you do want to freely assign discriminator fields, it is recommended to refactor object into 2 objects like this:
   ```nim
--- a/doc/manual.rst
+++ b/doc/manual.rst
@@ -2568,7 +2568,8 @@ Example:
   discard p(3, 4) # discard the return value of `p`
 
 The ``discard`` statement evaluates its expression for side-effects and
-throws the expression's resulting value away.
+throws the expression's resulting value away, and should only be used
+when ignoring this value is known not to cause problems.
 
 Ignoring the return value of a procedure without using a discard statement is
 a static error.
--- a/lib/pure/asyncfutures.nim
+++ b/lib/pure/asyncfutures.nim
@@ -431,6 +431,9 @@ proc asyncCheck*[T](future: Future[T]) =
       raise future.error
   future.callback = asyncCheckCallback
 
+template discardedFuture*{discard fut}(fut: Future) {.
+  error: "cannot discard future, use asyncCheck instead".} = discard
+
 proc `and`*[T, Y](fut1: Future[T], fut2: Future[Y]): Future[void] =
   ## Returns a future which will complete once both ``fut1`` and ``fut2``
   ## complete.
--- a/lib/pure/asyncfutures.nim
+++ b/lib/pure/asyncfutures.nim
@@ -9,6 +9,8 @@
 
 import os, tables, strutils, times, heapqueue, options, deques, cstrutils
 
+include "system/inclrtl"
+
 # TODO: This shouldn't need to be included, but should ideally be exported.
 type
   CallbackFunc = proc () {.closure, gcsafe.}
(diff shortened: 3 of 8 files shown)


metagn:
Test failures because of https://github.com/nim-lang/Nim/commit/3d2459bdc0b6d6236a2cd9209ed81c965ee411a5, unrelated

metagn:
Still unrelated? Those GitHub CI errors are weird. Will try rebasing with devel

Araq:
Tough decision, TR macros only exist for optimization purposes.

metagn:
Well `asyncCheck` is some kind of optimization. We just force people to use it instead of automatically using it.

> optimization: an act, process, or methodology of making something (such as a design, system, or decision) as fully perfect, functional, or effective as possible

It's functional or effective if you check exceptions inside futures.

disruptek:
So you had to remove the successful use of `discard` from the tests...  Was that not a hint that this is a reasonable use of `discard` against a `Future`?

metagn:
Warning instead of error then? I don't really know what the test does, I just assumed since it's a test that it's not a real world example. If not you can still `let _ = `.

disruptek:
I think a warning is fair but I think documentation improvements are the correct patch.  I'm sorry if my previous message came across as harsh; it was composed in haste! :pray:

Araq:
I think  #11912 should be closed as "won't fix", ``discard x`` is a dangerous construct, you clearly throw away information, you clearly should know what you're doing. Comparable to Nim's ``cast``.

metagn:
Custom warning here is a bit problematic since you can't suppress specific user errors without suppressing all user errors (the warning category User, you cant make your own warning category). You can opt out of this term rewriting macro altogether by doing `import asyncfutures except discardedFutureWarning` though. Maybe `-d:asyncSuppressDiscardedFutureWarning` too. This is too complicated though, I'll just remove it.

The problem here is people do `discard fut` without realizing it's wrong. ["Use asyncCheck"](https://nim-lang.org/docs/asyncdispatch.html#discarding-futures) could not be more clear in the docs, some people just don't read them or they learn through trial and error.

If `discard` is some kind of unsafe/unsound operation then we should start mentioning that in documentation. "Only use discard when you know what will happen next". Otherwise there is no clue to that at all.

disruptek:
The point is that it isn't unsafe.  It's behavior by design, and to do otherwise is a slippery (not to mention inconsistent) slope; a "race to the bottom" as @Araq says.  :grin:

My personal opinion is that the emphasis on `asyncCheck` is misplaced and the source of confusion for newcomers to Nim's async, but you know what they say about opinions...

dom96:
> So you had to remove the successful use of discard from the tests... Was that not a hint that this is a reasonable use of discard against a Future?

Funnily enough it is not a reasonable use of `discard`. You should never discard a future, it really is that simple. That test needs to be fixed because it leaks exceptions!

disruptek:
If I want to leak exceptions, why should Nim deny me?  We don't forbid leaks like some other languages.  Where does it end?

dom96:
> If I want to leak exceptions, why should Nim deny me? We don't forbid leaks like some other languages. Where does it end?

Then you can still do it, just not using `discard`. This is like saying "If I want to concatenate strings and integers, why should Nim deny me?".

D-Nice:
> > So you had to remove the successful use of discard from the tests... Was that not a hint that this is a reasonable use of discard against a Future?
> 
> Funnily enough it is not a reasonable use of `discard`. You should never discard a future, it really is that simple. That test needs to be fixed because it leaks exceptions!

This really doesn't change that, I can still discard futures (on latest devel branch git hash: baa80ca9d4c5bef4b90aabf4a666b2300c62b5b7), this PR just seems to stop me from potentially identifying an async proc as being acceptable to discard, the behaviour you consider dangerous is still quite possible, so I am confused by this PR as it leaves the root of the apparent problem intact.

A warning would seem more appropriate if discard on futures are used, because there are still instances where I may run a bunch of async things, but don't care to know what happens with them or that they block other things from running. And from my experimentation, there are no memory leaks or the like for not catching these exceptions.

metagn:
@D-Nice The old version of the PR used term rewriting macros to error on `discard fut` where `fut` is of type `Future[T]`. I changed it after someone suggested the current behavior. Term rewriting macros aren't meant for gatekeeping language features or library defined errors.

D-Nice:
@hlaaftana I see, thanks for the context! So is this PR moreso to then require a verbose discard, over before being able to "hide" the discarding?

I am still wondering, what is so bad if you wantonly and knowingly ignore Future exceptions in Nim, especially if you don't want other async events affected by one raising exceptions. I understand it's against best practices in general, but there are cases where one wants this purposefully, for example creating something that stress tests a webserver, you may want it to keep hitting that webserver at the same speed even if requests start being excepted, rather than having to start to handle each exception, and lowering the load.

dom96:
> I am still wondering, what is so bad if you wantonly and knowingly ignore Future exceptions in Nim, especially if you don't want other async events affected by one raising exceptions. I understand it's against best practices in general, but there are cases where one wants this purposefully, for example creating something that stress tests a webserver, you may want it to keep hitting that webserver at the same speed even if requests start being excepted, rather than having to start to handle each exception, and lowering the load.

The answer to your question is the same as the answer to "What's so bad about handling all exceptions and silently dropping them?"

Hopefully you know the answer. Sure, you may not care and still want to do it. But the problem with `discard`ing a future is that people aren't aware that this is what they are doing (discarding potential exceptions).

PR #14182, opened 2020-05-01
change 'iff' to 'if' to stop "corrections" once and for all

narimiran:
I can't take it anymore.

### DIFF
--- a/lib/pure/collections/critbits.nim
+++ b/lib/pure/collections/critbits.nim
@@ -54,7 +54,7 @@ proc rawGet[T](c: CritBitTree[T], key: string): Node[T] =
       return if it.key == key: it else: nil
 
 proc contains*[T](c: CritBitTree[T], key: string): bool {.inline.} =
-  ## Returns true iff `c` contains the given `key`.
+  ## Returns true if `c` contains the given `key`.
   runnableExamples:
     var c: CritBitTree[void]
     incl(c, "key")
@@ -156,7 +156,7 @@ proc excl*[T](c: var CritBitTree[T], key: string) =
   discard exclImpl(c, key)
 
 proc missingOrExcl*[T](c: var CritBitTree[T], key: string): bool =
-  ## Returns true iff `c` does not contain the given `key`. If the key
+  ## Returns true if `c` does not contain the given `key`. If the key
   ## does exist, c.excl(key) is performed.
   ##
   ## See also:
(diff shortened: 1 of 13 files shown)


ghost:
lol

lib/pure/htmlparser.nim
```diff
@@ -988,7 +988,7 @@ proc entityToRune*(entity: string): Rune =
     "DoubleRightArrow": Rune(0x021D2)
   of "dArr", "Downarrow", "DoubleDownArrow": Rune(0x021D3)
   of "hArr", "Leftrightarrow", "DoubleLeftRightArrow",
-    "iff": Rune(0x021D4)
+    "if": Rune(0x021D4)
```
ghost: Is the change correct here?

narimiran: Probably not :) Thanks!

lib/pure/htmlparser.nim
```diff
@@ -988,7 +988,7 @@ proc entityToRune*(entity: string): Rune =
     "DoubleRightArrow": Rune(0x021D2)
   of "dArr", "Downarrow", "DoubleDownArrow": Rune(0x021D3)
   of "hArr", "Leftrightarrow", "DoubleLeftRightArrow",
-    "if": Rune(0x021D4)
```
Araq: Er, no.

narimiran: Already fixed.

lib/pure/collections/critbits.nim
```diff
@@ -156,7 +156,7 @@ proc excl*[T](c: var CritBitTree[T], key: string) =
   discard exclImpl(c, key)
 
 proc missingOrExcl*[T](c: var CritBitTree[T], key: string): bool =
-  ## Returns true iff `c` does not contain the given `key`. If the key
+  ## Returns true if `c` does not contain the given `key`. If the key
```
bluenote10, line 159: I really like the clarity of "if and only if", maybe we could spell it out explicitly in the docstrings?

There are many instances where the following sentence contains a sentence with "if not" semantics. If the first sentence says "if and only if" it becomes clear that the second sentence is only a refining information.

E.g. here if we only read the weaker first sentence:

> Returns true if `c` does not contain the given `key`.

and then 

> If the key does exist, c.excl(key) is performed.

the question raises what is returned in that negated case? Can it be `true` as well? Maybe that depends on whether `c.excl(key)` is successful or so?

Reading a strong "if and only if" first avoids that.

hlaaftana: Returning implies there is no need for "and only if", the proc is done. "If and only if the key does exist, c.excl(key) is performed" makes sense, though it's painful to read

disruptek: I think these questions are answered more simply:

> Returns true if `c` contains the given `key`.  Otherwise, equivalent to ``c[key] = val``.

Programmers are reading this.  Lets give them a mental model of the control-flow.

PR #14187, opened 2020-05-01
Document that proc named fooTask is created for every foo task [backport]

kaushalmodi:
Ref: https://irclogs.nim-lang.org/01-05-2020.html#15:18:03

/cc @genotrance

### DIFF
--- a/lib/system/nimscript.nim
+++ b/lib/system/nimscript.nim
@@ -379,11 +379,26 @@ when not defined(nimble):
   template `==?`(a, b: string): bool = cmpIgnoreStyle(a, b) == 0
   template task*(name: untyped; description: string; body: untyped): untyped =
     ## Defines a task. Hidden tasks are supported via an empty description.
+    ##
     ## Example:
     ##
     ## .. code-block:: nim
     ##  task build, "default build is via the C backend":
     ##    setCommand "c"
+    ##
+    ## For a task named ``foo``, this template generates a ``proc`` named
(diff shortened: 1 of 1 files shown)


kaushalmodi:
@narimiran Looks like an unrelated failure

https://builds.sr.ht/~araq/job/199613#task-test-827

```
Error: execution of an external program failed: '/usr/home/build/Nim/testament/testament --pedantic all -d:nimCoroutines'
FAILURE
+ nim c -r tools/ci_testresults.nim
```

lib/system/nimscript.nim
```diff
@@ -379,11 +379,26 @@ when not defined(nimble):
...
     ##  task build, "default build is via the C backend":
     ##    setCommand "c"
+    ##
+    ## For a task named ``foo``, this template generates a ``proc`` named
+    ## ``fooTask``.  This is useful if you need to call one task in
+    ## another in your Nimscript.
+    ##
+    ## Example:
+    ##
+    ## .. code-block:: nim      # > nim foo
```
Araq: So you commented on the ``.. code-block`` itself? Come on, did you even check how this renders in the HTML?

kaushalmodi: No .. let me check. ~A comment breaks that? Is that a bug?~

🤦‍♂️ Oh, I'll fix it.

kaushalmodi: Now, fixed and confirmed as well.

image

PR #14188, opened 2020-05-01
move since from inclrtl to std/private/since

metagn:
refs https://github.com/nim-lang/RFCs/issues/215

SSL failures unrelated, some other checks might fail for unrelated reasons

### DIFF
--- a/lib/core/macros.nim
+++ b/lib/core/macros.nim
@@ -8,6 +8,7 @@
 #
 
 include "system/inclrtl"
+import std/private/since
 
 ## This module contains the interface to the compiler's abstract syntax
 ## tree (`AST`:idx:). Macros operate on this tree.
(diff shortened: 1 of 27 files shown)


metagn:
The test failure (ignoring SSL) is tests/dll/nimhcr_integration failing with "system module needs: programResult". It's because I was importing `std/private/since` before `programResult` was defined, causing in [this line](https://github.com/nim-lang/Nim/blob/9c33bcac5c73f8188e1c017ba8b2a14db5d5a4fe/compiler/cgen.nim#L2036) to fail. This is a minor inconvenience and this PR works without it, I can just move the import down below, but for future reference, that check (`m.hcrOn`) needs an `and` clause that makes it fail if it's still processing `system`. I won't fix it personally because I don't know how to check if it's processing the system module, just a heads up if someone wants to do it or they need `since` to be higher in `system`.

Araq:
Superb work, as usual. Thanks!

PR #14213, opened 2020-05-04
fixes #14209 [backport:1.2]

### LINKED ISSUE #14209: Use -d:nimEmulateOverflowChecks by default?
### Summary

Since 1.2.0, nim starts to require the `__builtin_saddll_overflow` compiler builtin by default. I believe this builtin was added to gcc in 5.x. Unfortunately, CentOS/RHEL 6/7, which is widely used in academia due to their stability, comes with gcc-4.8.5 that doesn't support this builtin. The requirement of this builtin means we can't build programs with the default `nim c` on CentOS/RHEL 6/7. This is inconvenient especially given that `-d:nimEmulateOverflowChecks` is long to type and is not exposed to command line help.

How much does `__builtin_saddll_overflow` help performance? If the improvement is minor, I guess users would prefer to disable it by default for portability. Thank you.

EDIT: perhaps the better solution is to test gcc version with macros.

Araq:


kaushalmodi:
**Update**: Fixed my misunderstanding below.

---

This PR says that `ccHasSaneOverflow` will return true if any of the below is true:
- compiler is clang
- **compiler is gcc** and the `--version` switch is valid for the compiler and the compiler version is > 5

For everything else, return `false`.

~I think it's very cryptic. Why not do that version check only if the compiler is `gcc`?~

---

*Also it would be awesome to have that version string to (major, minor, micro) tuple parser as a separate proc in the stdlib somewhere.*

/cc @c-blake

ghost:
Yeah, also the "else" check should account for other compilers like ccZig

Araq:
With ccZig the code cannot be backported as 1.2 doesn't have that support. I might also remove ``--cc:zig`` again because you can simply use ``--clang.exe=zig`` instead. On a related note, I find naming the work of the clang team "zig" pure presumptuousness.

Araq:
> The --version switch is valid for the compiler and the compiler version is > 5

That's not what my code does though, read it more carefully please.

alaviss:
I think trial compilation would work better (ie. generate a C file that uses the built in and test if that file compiles), than parsing for the compiler version.

Araq:
> I think trial compilation would work better (ie. generate a C file that uses the built in and test if that file compiles), than parsing for the compiler version.

I considered it, but it's slower and more work to implement.

c-blake:
Maybe I'm being dense here, but I don't see what's wrong with my positive filter based on compiler C PreProcessor version macros idea which seems lightest weight of all.  Every version of any compiler with these special builtins has those version macros.  Was this feature "phased in" for only some CPUs or something on the backend compilers so CPP version macros don't work?  I get their uglier/less reliable than `declared` but it's just backend compatibility `#if`'ing.  There's probably *some* version floor after which they reliably work and if Nim'ers want the speed-up they can use a vaguely recent backend compiler.

Anyway, as time goes on and more such builtins get used auto-optionally, it's obviously a general problem.  So, whatever  style @Araq is happiest to maintain is fine with me as long as it works.  This bug was stealthily blocking my `tcc` use for months until I was prodded to dive deeper.

Araq:
> Maybe I'm being dense here, but I don't see what's wrong with my positive filter based on compiler C PreProcessor version macros idea which seems lightest weight of all.

The problem is that it doesn't work. ;-) I want to know in the Nim layer whether these builtins are available and so whatever is done inside ``nimbase.h`` happens much too late.

Look at https://github.com/nim-lang/Nim/blob/devel/lib/system/integerops.nim#L22 

Maybe there are ways to do that via ``.emit`` instead but that's even uglier.

c-blake:
Ah.  Ok.  Totally fair, then.

PR #14235, opened 2020-05-05
new nimscript impl

Araq:
WIP; do not merge

### DIFF
--- a/compiler/commands.nim
+++ b/compiler/commands.nim
@@ -901,6 +901,8 @@ proc processSwitch*(switch, arg: string, pass: TCmdLinePass, info: TLineInfo;
     conf.globalOptions.incl optSourcemap
     conf.options.incl optLineDir
     # processOnOffSwitchG(conf, {optSourcemap, opt}, arg, pass, info)
+  of "nimscript2":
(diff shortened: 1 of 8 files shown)


kaushalmodi:
cool! Looking forward to details on what's new!

Araq:
It's a much faster implementation with different quirks. :-)

compiler/commands.nim
```diff
@@ -901,6 +901,8 @@ proc processSwitch*(switch, arg: string, pass: TCmdLinePass, info: TLineInfo;
     conf.globalOptions.incl optSourcemap
     conf.options.incl optLineDir
     # processOnOffSwitchG(conf, {optSourcemap, opt}, arg, pass, info)
+  of "nimscript2":
```
kaushalmodi, line 904: I don't understand this commit 100%.. but looks like --nimconfig2 is the "new nimscript". If so, should this be "nimconfig2" instead of "nimscript2"? Or is "nimscript2" different than "nimconfig2"? :)

Araq: It's all WIP, don't think about it too hard.

Araq:
Bad way to do it this way, there is a better way.

PR #14237, opened 2020-05-05
add a second asyncnet.recvFrom

rockcavera:
As requested by @dom96 on pull [#14154](https://github.com/nim-lang/Nim/pull/14154).

This recvFrom() uses FutureVar in the parameters and returns the number of bytes received.

### DIFF
--- a/lib/pure/asyncnet.nim
+++ b/lib/pure/asyncnet.nim
@@ -822,6 +822,55 @@ proc sendTo*(socket: AsyncSocket, address: string, port: Port, data: string,
     else:
       raise newException(IOError, "Couldn't resolve address: " & address)
 
+proc recvFrom*(socket: AsyncSocket, data: FutureVar[string], size: int,
+               address: FutureVar[string], port: FutureVar[Port],
+               flags = {SocketFlag.SafeDisconn}): owned(Future[int])
+              {.async, since: (1, 3).} =
+  ## Receives a datagram data from ``socket`` into ``data``, which must be at
+  ## least of size ``size``. The address and port of datagram's sender will be
+  ## stored into ``address`` and ``port``, respectively. Returned future will
+  ## complete once one datagram has been received, and will return size of
+  ## packet received.
+  ##
+  ## If an error occurs an OSError exception will be raised.
+  ## 
+  ## This proc is normally used with connectionless sockets (UDP sockets).
+  template adaptRecvFromToDomain(domain: Domain) =
+    var lAddr = sizeof(sAddr).SockLen
+    
+    result = await recvFromInto(AsyncFD(getFd(socket)), cstring(retData), size,
+                                cast[ptr SockAddr](addr sAddr), addr lAddr,
+                                flags)
+    
+    retData.setLen(result)
+
+    data.mget() = retData
+    data.complete()
+
+    address.mget() = getAddrString(cast[ptr SockAddr](addr sAddr))
+    address.complete()
+
+    when domain == AF_INET6:
+      port.complete(ntohs(sAddr.sin6_port).Port)
+    else:
+      port.complete(ntohs(sAddr.sin_port).Port)
+
+  assert(socket.protocol != IPPROTO_TCP,
+         "Cannot `recvFrom` on a TCP socket. Use `recv` or `recvInto` instead")
+  assert(not socket.closed, "Cannot `recvFrom` on a closed socket")
+
+  var retData = newString(size)
+  
+  case socket.domain
+  of AF_INET6:
+    var sAddr: Sockaddr_in6
+    adaptRecvFromToDomain(AF_INET6)
+  of AF_INET:
+    var sAddr: Sockaddr_in
+    adaptRecvFromToDomain(AF_INET)
+  else:
+    raise newException(ValueError, "Unknown socket address family")
+
 proc recvFrom*(socket: AsyncSocket, size: int,
                flags = {SocketFlag.SafeDisconn}):
               owned(Future[tuple[data: string, address: string, port: Port]])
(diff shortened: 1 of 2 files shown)


lib/pure/asyncnet.nim
```diff
@@ -822,40 +822,44 @@ proc sendTo*(socket: AsyncSocket, address: string, port: Port, data: string,
...
+               address: FutureVar[string], port: FutureVar[Port],
+               flags = {SocketFlag.SafeDisconn}): owned(Future[int])
               {.async, since: (1, 3).} =
-  ## Receives a datagram data from ``socket``, which must be at least of size
-  ## ``size``. Returned future will complete once one datagram has been received
-  ## and will return tuple with: data of packet received; and address and port
-  ## of datagram's sender.
-  ## 
+  ## Receives a datagram data from ``socket`` into ``data``, which must be at
+  ## least of size ``size``. The address and port of datagram's sender will be
```
dom96, line 830: > least of size ``size``

I would emphasise this with its own line like is done for these docs: https://nim-lang.org/docs/net.html#recv%2CSocket%2Cstring%2Cint

lib/pure/asyncnet.nim
```diff
@@ -822,40 +822,44 @@ proc sendTo*(socket: AsyncSocket, address: string, port: Port, data: string,
...
     else:
-      result.port = ntohs(sAddr.sin_port).Port
+      port.complete(ntohs(sAddr.sin_port).Port)
 
   assert(socket.protocol != IPPROTO_TCP,
          "Cannot `recvFrom` on a TCP socket. Use `recv` or `recvInto` instead")
   assert(not socket.closed, "Cannot `recvFrom` on a closed socket")
 
-  var data = newString(size)
+  var retData = newString(size)
```
dom96: You're allocating a new string here, when your doc comment states that the user of this API should have the data pre-allocated. Please also add an assert to ensure the allocation is done correctly.

dom96 (commented):
This looks fine in general, but I'm a little bit concerned about how you're handling the need to pre-allocated `data`. Right now you're reallocating it anyway, but even if you fix that, the behaviour will be inconsistent with ``address``, but for that I am not sure there is a way to pre-allocate which sucks a little.

@Araq any thoughts on the best design here?

lib/pure/asyncnet.nim
rockcavera: Here you speak to add a note? Example: `data` must be initialized to the length of `size`.

lib/pure/asyncnet.nim
rockcavera: Here I should remove this newString and let the user initialize FutureVar with data.mget().setlen()?

I thought about adding this assert:
``` nim
assert(size == len(data.mget()),
         "`date` was not initialized correctly. `size` != `len(data.mget())`")
```

rockcavera: Regarding `address`, initialize it to the maximum size that an IPv6 can have, which is 39, and then truncate it to the final size.

rockcavera:
Regarding `address`, initialize it to the maximum size that an IPv6 can have, which is 39, and then truncate it to the final size.

Edit:
I had forgotten that: address.mget.add()

lib/pure/asyncnet.nim
rockcavera: I believe I did not understand your comment.

Araq:
> @Araq any thoughts on the best design here?

Nope, I'm clueless.

rockcavera:
@dom96 , do I need to make any more changes? Or some other reason for not accepting the PR?

lib/pure/asyncnet.nim
```diff
@@ -822,40 +822,44 @@ proc sendTo*(socket: AsyncSocket, address: string, port: Port, data: string,
...
+                                flags)
     
-    data.setLen(fut)
-    
-    result.data = data
-    result.address = getAddrString(cast[ptr SockAddr](addr sAddr))
+    data.mget().setLen(result)
+    data.complete()
+
+    address.mget.add(getAddrString(cast[ptr SockAddr](addr sAddr)))
```
dom96: This isn't right. You'll be adding new data into the ``address`` every time this is called. Instead you should ask the caller to prealloc the maximum IPv6 length and do the same as you do for `data`.

dom96 (changes requested):
Just one thing, I think the rest looks fine.

lib/pure/asyncnet.nim
rockcavera: The problem with pre-allocating `address` is that `getAddrString()` will allocate another string and not write the IP address in the memory already allocated in `address`. And to pass the result of `getAddrString()` the memory allocated in `address` will be discarded. That is, this pre-allocation is unnecessary.

`address.mget.add()` is in fact not the best solution, as it will add new data in `address` to each call, when the user does not send an empty variable. However, you can add an assertion to check that `0 == len(address.mget())`.

Another way would be the one that was used first and you didn't like it: `address.mget() = getAddrString(cast[ptr SockAddr](addr sAddr))`

In my view, the only way to pre-allocate `address` and be able to write to that memory would be to redo `getAddrString()`.

@dom96, what's wrong with using: `address.mget() = getAddrString(cast[ptr SockAddr](addr sAddr))`? Could you explain it to me, because I'm a bit of a layman on the subject?

dom96: The whole point of this overload is to reduce string allocations. Calling ``getAddrString`` will allocate a string, so you do need to create an overload for it to take a `var string` as well and use it here IMO.

rockcavera: I hope that's what you expect.

lib/pure/asyncnet.nim
```diff
@@ -822,40 +822,47 @@ proc sendTo*(socket: AsyncSocket, address: string, port: Port, data: string,
...
+                                cast[ptr SockAddr](addr sAddr), addr lAddr,
+                                flags)
     
-    data.setLen(fut)
+    data.mget().setLen(result)
+    data.complete()
     
-    result.data = data
-    result.address = getAddrString(cast[ptr SockAddr](addr sAddr))
+    address.complete(getAddrString(cast[ptr SockAddr](addr sAddr)))
```
dom96: This is still wrong, you need  `getAddrString` variant that will write to a `var` argument (in the same way that recvFromInto does with the `data.mget()` argument). Here you are still copying the string.

rockcavera: @dom96 , I don't know if it's the best way, but I've been thinking about doing it, because it takes advantage of the pre-allocated memory:
```nim
    let addressTmp = getAddrString(cast[ptr SockAddr](addr sAddr))

    for i in 0 ..< len(addressTmp):
      address.mget()[i] = addressTmp[i]

    address.mget().setLen(len(addressTmp))    
    address.complete()
```
Because, in my view, it is safer than `moveMem()` and `copyMem()`, since the user may not correctly pre-allocate `address`.

Creating a new `getAddrString()` would fall into the same problem as `address = getAddrString()`, unless something is missed from me.

dom96: Why don't you create a `getAddrString` like: `proc getAddrString(sAddr: ptr SockAddr, out: var string)`? That will solve the problem here.

rockcavera: @dom96, sorry for my ignorance. I was able to understand and see how it works, now with more time.

Now there is a question as to how this should be done. Make a new `proc getAddrString(sockAddr: ptr SockAddr, strAddress: var string)` in the `asyncnet` package or `nativesockets`? Should it be `asyncnet`, should it be marked as exportable? Should another PR be done to first add this new `getAddrString`?

Also, there is the possibility, which I already did for testing, to use `winlean.inet_ntop()` or `posix.inet_ntop()` directly in the body of `recvFrom()`.

What better designer solution?

I also checked that the pre-allocation size of the `address FutureVar[string]` must be 46.

dom96: I would put it in `nativesockets`, if you want to export it then it definitely shouldn't live in `asyncnet`.

> Also, there is the possibility, which I already did for testing, to use winlean.inet_ntop() or posix.inet_ntop() directly in the body of recvFrom().

I guess you can do this, but won't you just be redoing what getAddrString already implements? Might as well do it in its own function and make that available for others.

rockcavera: I made the new `getAddrString` using the existing one as a base. However, I used `inet_ntop()` for both IPv4 and IPv6.

I tested on Windows 10 and Debian 10.

dom96 (approved):
Looks good. Thank you!

Issue #13764, opened 2020-03-26
set[T] construction is not typechecked for variables


alaviss:
This bug caused erroneously written code to compile and execute if the value inside the variables were small enough.

### Example
```nim
#var a: set[int] # Error: set is too large

var b = 1_000_000
let a = {b} # Compiles
```

### Current Output
Compiles, crashes at runtime (expected). If you change `b` into a number smaller than `high(uint8)`, then no error occurs.

### Expected Output
Error about the type of the set being too big? `b` was resolved to be `int` before the set was even made.

### Additional Information
```
$ nim -v
Nim Compiler Version 1.1.1 [Linux: amd64]
Compiled at 2020-02-24
Copyright (c) 2006-2019 by Andreas Rumpf

git hash: 3dad130034e3212159dda5d61988ca2b18959b24
active boot switches: -d:release -d:useLinenoise -d:nativeStackTrace
```

bung87:
problem here is it doesn't even check for int literal, simply pass `isOrdinalType` where `baseKinds = {tyChar, tyInt..tyInt64, tyUInt..tyUInt64, tyBool, tyEnum}`

metagn:
The original code in the issue still only errors at runtime but whatever, since #21659 it gives a warning

I know `{b}` will give a warning but unsure about `{123, b}`, that might implicitly convert `b` to `range[0..255]`

PR #14252, opened 2020-05-06
net: remove more erroneous set constructions

alaviss:
Refs #13764


alaviss:
Should also be backported.

PR #14265, opened 2020-05-07
nativesockets: add missing inheritable pass-through

alaviss:


timotheecour:
@Araq has anything changed in the way this PR was merged? every other PR was merged by squash-and-merge to avoid tangled git history (refs https://github.com/nim-lang/Nim/issues/8664)

image

Clyybber:
Githubs mobile site UI does that by default AFAIR, so thats probably why.

timotheecour:
even on mobile phone UI it gives you the option to pick squash and merge; hence https://github.com/nim-lang/Nim/issues/8664 which proposed to disable other options by default to prevent such mistakes:

> in nim-lang/Nim/settings, disable "Allow merge commits", as follows:

and in the < 1% cases where a squash+merge is not desirable, this option can be temporarily re-enabled.

The damage here was limited (merged commit had recent common ancestor) but it can be more damaging in other cases when ppl merge instead of rebase their PR's and common ancestor is old.

Araq:
Well nothing changed but I was not cautious enough this time.

PR #14287, opened 2020-05-09
Fix #13946

### LINKED ISSUE #13946: htmlparser.parseHtml behaves differently using --gc:arc or --gc:orc
Function `htmlparser.parseHtml` seems to behave differently when using `arc` or `orc` as GC mechanisms.

### Example
```nim
import htmlparser
import xmltree

let html = """
    <div>
        <img src="something.png" />
        Hello World
    </div>
"""
let doc = parseHtml(html)
echo $doc
```
When ran with any  of these configurations:
```bash
nim c -r --gc:refc test.nim
nim c -r --gc:markAndSweep test.nim
nim c -r --gc:boehm test.nim
nim c -r --gc:none test.nim
```
The output is correct:
```
<document>    <div>
        <img src="something.png" />
        Hello World
    </div>
</document>
```

**But** when using `arc` or `orc`:
```bash
nim c -r --gc:arc test.nim
nim c -r --gc:orc test.nim
```
The output is missing the "Hello World" part:
```
<document>    <div>
        <img src="something.png" /></div>
</document>
```
Seems to work properly in `arc` and `orc` when the `` tag isn't self-closed, like this:
```nim
import htmlparser
import xmltree

let html = """
    <div>
        <img src="something.png">
        Hello World
    </div>
"""
let doc = parseHtml(html)
echo $doc
```
```bash
nim c -r --gc:arc test.nim
nim c -r --gc:orc test.nim
```
```
<document>    <div>
        <img src="something.png" />
        Hello World
    </div>
</document>
```

## Additional information
```
$ nim -v
Compiled at 2020-04-03

```

Clyybber:


lib/pure/parsexml.nim
```diff
@@ -299,15 +299,15 @@ template piRest*(my: XmlParser): string =
   assert(my.kind == xmlPI)
   my.b
 
-proc rawData*(my: var XmlParser): string {.inline.} =
+proc rawData*(my: sink XmlParser): string {.inline.} =
   ## returns the underlying 'data' string by reference.
```
Varriount, line 303: Since the file is already being modified, perhaps the documentation should be updated to state *why* this is unsafe as well?

cooldome:
I haven't checked, but possibly this would work better:
```nim
proc rawData*(my: XmlParser): lent string {.inline.} 
proc rawData2*(my: XmlParser): lent string {.inline.} 
```
 One implementation for old and new runtime without `move`, without `shallowCopy` and without `when defined(gcDestructors)` inside.

lib/pure/parsexml.nim
Clyybber: With this change its actually safe with `--gc:arc`.
But in general its unsafe because it returns it by reference/shallow copy.

PR #14299, opened 2020-05-11
Fix move message

planetis-m:
The "use 'move($1)' to prevent it" part of the hint message was reworded.
See forum [thread](https://forum.nim-lang.org/t/6309)


planetis-m:
@Araq i made these changes just now:

```nim
proc isCapturedVar(n: PNode): bool =
  n.kind == nkDotExpr and n[0].kind == nkHiddenDeref and
      n[0][0].kind == nkSym and n[0][0].sym.name.s[0] == ':'
```

shall i make a new pr?

Araq:
Yes.

PR #14322, opened 2020-05-12
New "ping-pong" DFA

### LINKED ISSUE #13456: Still strange ARC issue with Combinatorics Module of R. Behrends (very low priority)
We tried to discuss it on forum already, but unfortunately the issue is much more difficult and strange.
(https://forum.nim-lang.org/t/5937)

To allow installation of gintro when installer script is compiled with --gc:arc we have to copy the generated combinations in a temporary seq:

https://github.com/StefanSalewski/gintro/blob/master/tests/gen.nim#L2353

```nim
          #for t in unp.combinations(a + 1): # works fine with default GC
          var hhh = toSeq(unp.combinations(a + 1)) # nedded for --gc:arc
          for t in hhh:
```
Direct iterator usage result in 

```
$ ./gen
GC is arc
Generating bindings for GTK3...
/home/stefan/gintrotest/tests/gen.nim(2600) gen
/home/stefan/gintrotest/tests/gen.nim(2535) launch
/home/stefan/gintrotest/tests/combinatorics.nim(76) main
/home/stefan/Nim/lib/system/fatal.nim(45) sysFatal
/home/stefan/Nim/lib/system/fatal.nim(33) sysFatal
Error: unhandled exception: index out of bounds, the container is empty [IndexError]
```

The iterator

https://github.com/StefanSalewski/gintro/blob/master/tests/combinatorics.nim#L55

is difficult as two seqs are involved.

Of course that is not really a problem, no one will use --gc:arc for the nimble install of packages, and it works fine with this temporary seq. It is very hard to debug and strange, as simple basic tests with that iterator just works fine. But maybe we should remember this issue and try it again in some months, I will close this issue tomorrow and maybe reopen in some months or when I have more insights.

Clyybber:
Fixes #13456


cooldome:
I had a quick run on my project. Unfortunately, compiler hangs in endless loop during cfg traversal. I will be able to provide more details later today.

Clyybber:
@cooldome Yeah, I'm currently investigating that. (also occurs with norm pkg but its hard to reduce the testcase)

Clyybber:
Test failure unrelated (flaky async test)

Clyybber:
@cooldome Should be fixed now, can you recheck?

cooldome:
Boosting the rocket into space

PR #14335, opened 2020-05-13
fixes #14318

### LINKED ISSUE #14318: Raises pragma and generic error/exception types compiler crash
Given this piece of code:
```nim
type MyValueError*[T] = object of ValueError

proc failure() {.raises: [MyValueError].} =
  raise newException(MyValueError[int], "x")

failure()
```

The compiler will crash and except  (version: 1.2.0):
```
Error: unhandled exception: types.nim(1181, 10) `b.kind in {tyObject} + skipPtrs`  [AssertionError]
```

Expected:
No crash and a better error message stating that type specialization is required.
Or actually handle them generically (preferable?)

Araq:


narimiran:
@Araq is there a reason why this hasn't been merged?

Araq:
@zah claimed it's not a correct patch. But maybe it's better than nothing?

PR #14351, opened 2020-05-14
Improve nimeval, changes some defaults

PMunch:
This improves nimeval to make it easier to use NimScript as an embedded
scripting languages. It does change some of the defaults, so in case
anything used the old version and depended on the old behaviour these
would have to be changed.

### DIFF
--- a/compiler/nimeval.nim
+++ b/compiler/nimeval.nim
@@ -99,14 +99,16 @@ proc findNimStdLibCompileTime*(): string =
 
 proc createInterpreter*(scriptName: string;
                         searchPaths: openArray[string];
-                        flags: TSandboxFlags = {}): Interpreter =
+                        flags: TSandboxFlags = {},
+                        defines = @[("nimscript", "true")],
+                        registerOps = true): Interpreter =
   var conf = newConfigRef()
   var cache = newIdentCache()
   var graph = newModuleGraph(cache, conf)
   connectCallbacks(graph)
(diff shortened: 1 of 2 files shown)


compiler/nimeval.nim
```diff
@@ -99,14 +99,16 @@ proc findNimStdLibCompileTime*(): string =
 
 proc createInterpreter*(scriptName: string;
                         searchPaths: openArray[string];
-                        flags: TSandboxFlags = {}): Interpreter =
+                        flags: TSandboxFlags = {},
+                        defines = @[("nimscript", "true")],
```
Araq, line 103: Huh? "true"? Only pass the defined symbols instead.

PMunch: But what if you want to do the equivalent of `-d:someflag:100`?

Araq: That doesn't work with your approach either iirc, the "true" is caused by a silly implementation.

PMunch: Just tested and it works fine. I added a `const testDefine* {.intdefine.} = 100` in my script, then called `createInterpreter(scriptname, libpaths, defines = @[("nimscript", "true"), ("testDefine", "42")])` and then in my script I `echo`'d out the value and it came out as 42.

Araq: Ok.

Issue #14032, opened 2020-04-20
coroutine tests fail on OpenBSD with segmentation faults


euantorano:
The coroutine tests ` tests/coroutines/tgc.nim` and `tests/coroutines/twait.nim` both fail on OpenBSD with segmentaiton faults:

```
nim c -r tests/coroutines/tgc.nim
Hint: used config file '/home/euant/Nim/config/nim.cfg' [Conf]
Hint: used config file '/home/euant/Nim/config/config.nims' [Conf]
Hint: used config file '/home/euant/Nim/tests/config.nims' [Conf]
Hint: used config file '/home/euant/Nim/tests/coroutines/tgc.nim.cfg' [Conf]
Hint: 15794 LOC; 0.020 sec; 17.086MiB peakmem; Debug build; proj: tgc; out: tgc [SuccessX]
Hint: /home/euant/Nim/tests/coroutines/tgc  [Exec]
Segmentation fault (core dumped)
Error: execution of an external program failed: '/home/euant/Nim/tests/coroutines/tgc '
```

```
nim c -r tests/coroutines/twait
Hint: used config file '/home/euant/Nim/config/nim.cfg' [Conf]
Hint: used config file '/home/euant/Nim/config/config.nims' [Conf]
Hint: used config file '/home/euant/Nim/tests/config.nims' [Conf]
Hint: used config file '/home/euant/Nim/tests/coroutines/twait.nim.cfg' [Conf]
Hint: system [Processing]
Hint: widestrs [Processing]
Hint: io [Processing]
Hint: twait [Processing]
Hint: coro [Processing]
Hint: os [Processing]
Hint: strutils [Processing]
Hint: parseutils [Processing]
Hint: math [Processing]
Hint: bitops [Processing]
Hint: macros [Processing]
Hint: algorithm [Processing]
Hint: unicode [Processing]
Hint: pathnorm [Processing]
Hint: osseps [Processing]
Hint: posix [Processing]
Hint: times [Processing]
Hint: options [Processing]
Hint: typetraits [Processing]
Hint: lists [Processing]
Hint: clang -c  -w -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=0   -I/home/euant/Nim/lib -I/home/euant/Nim/tests/coroutines -o /home/euant/.cache/nim/twait_d/amd64.S.o /home/euant/Nim/lib/arch/x86/amd64.S [Exec]
Hint:  [Link]
[...]

PR #14353, opened 2020-05-14
add OpenBSD MAP_STACK for coroutines bug #14032

j-bm:
OpenBSD requires stacks to have the MAP_STACK attribute.

This partly fixes coroutines #14032 except I've seen titerators fail with this patch occasionally.  (Another issue).


Araq:
But we don't want all heap allocations to be potentially used as a stack. Our coroutine implementation should use a special ``allocStack`` proc.

j-bm:
I was somewhat expecting this comment (mixed stack and heap).  At the least this allows to run while your allocStack approach is considered.

PR #14358, opened 2020-05-15
Speed up testing of some packages

ghost:
See https://github.com/status-im/nim-bncurve/blob/master/bncurve.nimble.
Also "nim c -r tvectors.nim" tested only a single file which took around 10 minutes (!) to complete because it was compiled in debug mode.


ghost:
Do not merge yet, I'll try to find other packages which can be tested faster (which don't use -d:release in arguments)

PR #14362, opened 2020-05-15
asyncdispatch, asyncnet: add inheritance control

alaviss:


lib/pure/asyncdispatch.nim
```diff
@@ -228,6 +228,17 @@ proc initCallSoonProc =
   if asyncfutures.getCallSoonProc().isNil:
     asyncfutures.setCallSoonProc(callSoon)
 
+template implementSetInheritable() {.dirty.} =
```
dom96, line 231: Remove this template and define the `setInheritable` here?

lib/pure/asyncdispatch.nim
```diff
@@ -1484,16 +1505,26 @@ else:
...
     var retFuture = newFuture[tuple[address: string,
         client: AsyncFD]]("acceptAddr")
     proc cb(sock: AsyncFD): bool =
       result = true
       var sockAddress: Sockaddr_storage
       var addrLen = sizeof(sockAddress).SockLen
-      var client = accept(sock.SocketHandle,
-                          cast[ptr SockAddr](addr(sockAddress)), addr(addrLen))
+      var client =
+        when declared(accept4):
```
dom96, line 1518: Since `accept4` is only available in newer Linux kernels, let's make it possible to disable it's usage using this flag:

```suggestion
        when declared(accept4) and defined(nimInheritHandles):
```

lib/pure/asyncdispatch.nim
```diff
@@ -1484,16 +1505,26 @@ else:
...
+          accept4(sock.SocketHandle, cast[ptr SockAddr](addr(sockAddress)),
+                  addr(addrLen), if inheritable: 0 else: SOCK_CLOEXEC)
+        else:
+          accept(sock.SocketHandle, cast[ptr SockAddr](addr(sockAddress)),
+                 addr(addrLen))
+      when declared(setInheritable) and not declared(accept4):
+        if client != osInvalidSocket and not setInheritable(client, inheritable):
+          close client
+          client = osInvalidSocket
```
dom96: I'm not too happy with this logic as it looked wrong to me initially. Please at least add a comment that this is so that the below error code path is reached, or better yet consider refactoring the error handling code into a separate proc and calling it from both code paths.

lib/pure/asyncdispatch.nim
```diff
@@ -1824,12 +1858,15 @@ proc withTimeout*[T](fut: Future[T], timeout: int): owned(Future[bool]) =
   return retFuture
 
 proc accept*(socket: AsyncFD,
-    flags = {SocketFlag.SafeDisconn}): owned(Future[AsyncFD]) =
+             flags = {SocketFlag.SafeDisconn},
+             inheritable = defined(nimInheritHandles)): owned(Future[AsyncFD]) =
   ## Accepts a new connection. Returns a future containing the client socket
   ## corresponding to that connection.
+  ## If ``inheritable`` is false (the default), the resulting client socket
+  ## will not be inheritable by child processes.
```
dom96, line 1871: Nit: 
```suggestion
  ##
  ## If ``inheritable`` is false (the default), the resulting client socket
  ## will not be inheritable by child processes.
  ##
```

lib/pure/asyncnet.nim
```diff
@@ -458,14 +476,16 @@ proc acceptAddr*(socket: AsyncSocket, flags = {SocketFlag.SafeDisconn}):
...
         retFuture.complete(resultTup)
   return retFuture
 
 proc accept*(socket: AsyncSocket,
     flags = {SocketFlag.SafeDisconn}): owned(Future[AsyncSocket]) =
   ## Accepts a new connection. Returns a future containing the client socket
   ## corresponding to that connection.
+  ## If ``inheritable`` is false (the default), the resulting client socket will
+  ## not be inheritable by child processes.
   ## The future will complete when the connection is successfully accepted.
```
dom96, line 491:
```suggestion
  ##
  ## If ``inheritable`` is false (the default), the resulting client socket will
  ## not be inheritable by child processes.
  ## The future will complete when the connection is successfully accepted.
  ##
```

dom96 (approved):
Some comments, but overall looks good.

lib/pure/asyncdispatch.nim
alaviss: I would if I could, but we don't have `AsyncFD` defined til later.

lib/pure/asyncdispatch.nim
alaviss: I'd say kernel 2.6.28 is old enough :P, but sure, I'll add something to disable `accept4` itself from `posix`.

lib/pure/asyncdispatch.nim
dom96: Move this proc to the bottom of the file then? forward the procedure if you must

lib/pure/asyncdispatch.nim
dom96: 2008 is indeed old, but if we can support the previous versions easily we should. You never know who might need it.

lib/pure/asyncdispatch.nim
alaviss: This wouldn't do, I'll end up duplicating the signature multiple times for no good reason.

lib/pure/asyncdispatch.nim
alaviss: I'll make some changes to facilitate introducing a new flag: `nimOldPosix` for when these newer procs are not supported.

But adding the flag itself will have to be done in an another PR, since it'd touch many other parts of the stdlib.

PR #14372, opened 2020-05-16
fix #14364 - turn on the TLS emulation when using Boehm

### LINKED ISSUE #14364: Boehm GC does not scan thread-local storage
That's how you get surprise deallocations for targets of pointers that are only in threadvars.

https://www.hboehm.info/gc/gcinterface.html :

"Thread users should also be aware that on many platforms objects reachable only from thread-local variables may be prematurely reclaimed. Thus objects pointed to by thread-local variables should also be pointed to by a globally visible data structure."

gc-8.0.4/README.QUICK :

"The collector does not guarantee to scan thread-local storage (e.g. of the kind accessed with pthread_getspecific()).  The collector does scan thread stacks though, so generally the best solution is to ensure that any pointers stored in thread-local storage are also stored on the thread's stack for the duration of their lifetime."

---
I found that out the hard way (Valgrind output):

```text
==20803== Jump to the invalid address stated on the next line
==20803==    at 0x10: ???
==20803==    by 0x2C38DF: zonedTimeFromTime__zT9bMgAwfe1o0G71u2tO24Q (times.nim:1064)
==20803==    by 0x2C38DF: inZone__o59cT0ssqq4BC09cuUGSdJiA (times.nim:1093)
==20803==    by 0x2C38DF: local__JDUz4bfwLSZvJxYV9aiNG9bg (times.nim:1235)
==20803==    by 0x2C38DF: now__YmAcAi9bz5G8OQqVFWHN2uA (times.nim:1263)
==20803==    by 0x38F0FE: rfcTimestamp__i2AqQvvKFsZl0RRd5ysaHA (log_output.nim:426)
==20803==    by 0x38F0FE: initLogRecord__AdFvJuSUXMySj9bot5WYdCQ (log_output.nim:438)
==20803==    by 0x58C8DF: generateDeposits__u9cJTyqoU2khEGAmzCsRniA (chronicles.nim:305)
==20803==    by 0x5AB7D3: main__kYRn29bPOsjfiAecYiv9b2ng (beacon_node.nim:1007)
==20803==    by 0x5B56C1: NimMain (mainchain_monitor.nim:490)
==20803==    by 0x12AD8C: main (mainchain_monitor.nim:497)
==20803==  Address 0x10 is not stack'd, malloc'd or (recently) free'd
==20803== 
```

lib/pure/times.nim:
```nim
var localInstance {.threadvar.}: Timezone
```

And Timezone is a ref object:

```nim
  Timezone* = ref object ## \
      ## Timezone interface for supporting `DateTime <#DateTime>`_\s of arbitrary
      ## timezones. The ``times`` module only supplies implementations for the
      ## systems local time and UTC.
    zonedTimeFromTimeImpl: proc (x: Time): ZonedTime
        {.tags: [], raises: [], benign.}
    zonedTimeFromAdjTimeImpl: proc (x: Time): ZonedTime
        {.tags: [], raises: [], benign.}
    name: string
```

stefantalpalaru:
Turn on the TLS emulation when using Boehm, since it doesn't scan the
real TLS.

### DIFF
--- a/compiler/commands.nim
+++ b/compiler/commands.nim
@@ -493,6 +493,7 @@ proc processSwitch*(switch, arg: string, pass: TCmdLinePass, info: TLineInfo;
       of "boehm":
         conf.selectedGC = gcBoehm
         defineSymbol(conf.symbols, "boehmgc")
+        incl conf.globalOptions, optTlsEmulation # Boehm GC doesn't scan the real TLS
       of "refc":
         conf.selectedGC = gcRefc
       of "v2":


Araq:
This would be a serious BoehmGC bug, can we confirm this somehow?

stefantalpalaru:
> This would be a serious BoehmGC bug, can we confirm this somehow?

It is documented repeatedly in the Boehm GC archive, in "README.md", "README.QUICK", "doc/debugging.md", "doc/gcinterface.md", etc.

See the rather explicit quotes in https://github.com/nim-lang/Nim/issues/14364

PR #14395, opened 2020-05-18
Fix #14394

### LINKED ISSUE #14394: Bad codegen/emit for hashes.hiXorLo in some contexts.
Using devel branch, the C target, using gcc.
```
Compiled at 2020-05-18

git hash: 5beed56d0a5d9827680b34e7964da2031611dd6d
```

In some contexts, the C code that gets generated for hashes.hiXorLo is invalid, and gcc fails with syntax errors.  The variables produced in a emit section don't match the generated parameters (a vs a_0):

```C
static N_INLINE(NU64, hiXorLo__wWqo9alrSZddOyokrApfNaQhashes)(NU64 a_0, NU64 b_0) {
        NU64 result;
        nimfr_("hiXorLo", ".../Nim/lib/pure/hashes.nim");
        result = (NU64)0;
        nimln_(100, ".../Nim/lib/pure/hashes.nim");
        __uint128_t r = a; r *= b; result = (r >> 64) ^ r;    // <<<<< From the emit in hashes.nim
        popFrame();
        return result;
}
```

This is the object and hash function this code was generated for:
```nim
type
  # In another module, V2X[T] = tuple[x, y: T]
  ShortCoord* = V2X[int16]
  # Used as a value for a table, ie: Table[ShortCoord,ASNodeInfo]
  ASNodeInfo = tuple[estScore, curScore: float32; parent: ShortCoord]

proc hash*(sc: ShortCoord) : Hash = 
  result = result !& hash(sc.x)
  result = result !& hash(sc.y)
  result = !$result
```

The bad news is extracting out this code from my module does not reproduce the problem.  My hope is this is enough to see what's going on.  If not, I can do some more work to either extract a reproducible case, bisect to where it stopped working, or push the project up so it can be compiled by someone else.

Clyybber:


Araq:
@narimiran This needs to be backported.

PR #14406, opened 2020-05-20
Marked for export net.parseIPv4Address and net.parseIPv6Address

rockcavera:
I don't know why `net.parseIPv4Address` and `net.parseIPv6Address` were not marked as exportable. However, I see these two procedures as useful for use and not just within `net.parseIpAddress` and `net.isIpAddress`, which are marked as exportable.

If there is a reason for not marking them as exportable, I would like to know.

I don't know if it is necessary to add an annotation to the changelog.


alaviss:
Now that they are exported, we should give them some proper documentation too.

rockcavera:
> Now that they are exported, we should give them some proper documentation too.

In the case of these procs, what do you classify as adequate documentation?

Before adding the pr I had already looked at `net.parseIpAddress`, to know what kind of information the documentation brought. And it brings nothing that is not already in `net.parseIPv4Address` and `net.parseIPv6Address`.

But if something more is needed, I’m willing to do it.

Change the documentation to:
Parses a IPvX address contained in `addressStr`.
If it is not a valid IPvX address, ValueError is raised.

lib/pure/net.nim
```diff
@@ -247,7 +247,7 @@ proc newSocket*(domain: Domain = AF_INET, sockType: SockType = SOCK_STREAM,
     raiseOSError(osLastError())
   result = newSocket(fd, domain, sockType, protocol, buffered)
 
-proc parseIPv4Address(addressStr: string): IpAddress =
+proc parseIPv4Address*(addressStr: string): IpAddress =
   ## Parses IPv4 addresses
   ## Raises ValueError on errors
```
dom96, line 252: You should at least fix how this is rendered (check via `nim doc net.nim`). You'll see that it's just one sentence which is wrong, you need to add an empty ``##`` between them to ensure the line break is there in the docs.

But really, ideally, we would like runnable examples here too.

dom96 (commented):
Please explain your use case. It's hard for me to imagine why you cannot simply use `parseIpAddress`.

rockcavera:
> Please explain your use case. It's hard for me to imagine why you cannot simply use `parseIpAddress`.

An example of use would be for an `isIpv6Address()`, which would return true or false if the IP entered is an IPv6 and only IPv6, that is, it would return false for a valid IPv4. The same for `isIpv4Address()`.

This is necessary when we need specific functions to validate IPv4 or IPv6 and not a generic IP address validation function like `net.isIpAddress()`.

So the greatest use, for me, would be in these cases. Specific use for a specific IP protocol and not the generic one, which is now adopted by the `net` library.

rockcavera:
I know it can be done like this:
```nim
proc isIpv6Address(ip: string): bool =
  try:
    let parsed = parseIpAddress(ip)

    if parsed.family == IpAddressFamily.IPv6:
      result = true
  except:
    discard

proc isIpv4Address(ip: string): bool =
  try:
    let parsed = parseIpAddress(ip)

    if parsed.family == IpAddressFamily.IPv4:
      result = true
  except:
    discard
```

If you find it unnecessary, you can close the PR.

rockcavera:
Closing the PR, because there is a lack of motivation. At the moment, there is the `parseIpAddress()` proc that meets all expectations. And with a little creativity, you can get separate results for both IP protocols.

If, in the future, a greater motivation appears, I will be happy to reopen this PR.

PR #14408, opened 2020-05-20
add bindParams to db_sqlite

### LINKED ISSUE #9453: db misses an API to precompile the query and bind data later.
dbFormat is used to substitute the param values. 
the question mark (?) is matched on the entire sql-string instead
of binding by the database. This would harden the API and the
database is able to cache the (already parsed) prepared statements.
At the moment the database does a hard-parse  on every invocation.

#### example for sqlite3 to demonstrate the issue:
```nim
import db_sqlite

const create_table1_sql = sql"""
create table if not exists ttab(
  id integer primary key,
  question text,
  answer text
);"""

const insert_sql = sql"""
insert into ttab (id,question,answer) values (?,'how do you do ?',?)
 """

let db = open(":memory:","","","") 
db.exec(create_table1_sql)
db.exec(insert_sql,1,"fine!") # index out of bounds is raised here
db.close

#[ runtime outcome is:
  db_sqlite.nim(139)       exec
  db_sqlite.nim(129)       tryExec
  db_sqlite.nim(119)       dbFormat
  system.nim(2939)         sysFatal
  Error: unhandled exception: index out of bounds [IndexError]
]#
```

#### Solution:
DbType already present within db_common.nim; for db_postres.nim
there is also a "prepare" proc already present but
unfortunately the parameter dbType is missing.

The API could be extended a little bit (introduce new procs without affecting the 
present API) so that the caller is able to promote the dbTypes. 
Or introduce a preparedStatement object (better solution I think) and bind the values like this:  ps.setString(1,"fine!")
Binding (driver layer) would be not that difficult then.

### Additional Information

this affects the entire db-driver layer.

Discussion can be found here:
https://forum.nim-lang.org/t/4320

related discussion here
https://github.com/nim-lang/Nim/issues/5768

also in the past some issues relating this problem
bubbled up:
https://github.com/nim-lang/Nim/issues/3328
https://github.com/nim-lang/Nim/issues/6571

bung87:
partial fix https://github.com/nim-lang/Nim/issues/9453


lib/impure/db_sqlite.nim
```diff
@@ -113,7 +113,7 @@
 ## * `db_mysql module <db_mysql.html>`_ for MySQL database wrapper
 ## * `db_postgres module <db_postgres.html>`_ for PostgreSQL database wrapper
 
-import sqlite3
+import sqlite3,typeinfo
```
alaviss:
```suggestion
import sqlite3, typeinfo
```

lib/impure/db_sqlite.nim
```diff
@@ -166,6 +168,7 @@ proc dbFormat(formatstr: SqlQuery, args: varargs[string]): string =
       add(result, c)
 
 proc prepare*(db: DbConn; q: string): SqlPrepared =
+
```
alaviss: I'm pretty sure having docs here would be more useful than newlines :p

lib/impure/db_sqlite.nim
```diff
@@ -716,8 +719,64 @@ proc setEncoding*(connection: DbConn, encoding: string): bool {.
 proc finalize*(sqlPrepared:SqlPrepared){.discardable.} = 
   discard finalize(sqlPrepared.PStmt)
 
+proc bindInt32*( ps : SqlPrepared, paramIdx: int,
+  val: int32 ):int =
```
alaviss: Please follow the style already used in this module.

```suggestion
proc bindInt32*(ps: SqlPrepared, paramIdx: int, val: int32): int = 
```

And we should just have one API: `bindParam(prepared, paramIdx, T)`, where for each `T` we use an overload, ie.

```nim
proc bindInt32*(ps: SqlPrepared, paramIdx: int, val: int32)
proc bindString*(ps: SqlPrepared, paramIdx: int, val: string)
```

becomes

```nim
proc bindParam*(ps: SqlPrepared, paramIdx: int, val: int32)
proc bindParam*(ps: SqlPrepared, paramIdx: int, val: string)
```

lib/impure/db_sqlite.nim
```diff
@@ -716,8 +719,64 @@ proc setEncoding*(connection: DbConn, encoding: string): bool {.
 proc finalize*(sqlPrepared:SqlPrepared){.discardable.} = 
   discard finalize(sqlPrepared.PStmt)
 
+proc bindInt32*( ps : SqlPrepared, paramIdx: int,
+  val: int32 ):int = 
+  ## Binds a int32 float to the specified paramIndex.
+  result = bind_int( ps.PStmt, paramIdx.int32, val )
```
alaviss: Ditto

```suggestion
  result = bind_int(ps.PStmt, paramIdx.int32, val)
```

lib/impure/db_sqlite.nim
```diff
@@ -716,8 +719,64 @@ proc setEncoding*(connection: DbConn, encoding: string): bool {.
...
+  ## Binds a 64bit float to the specified paramIndex.
+  result = bind_double(ps.PStmt, paramIdx.int32, val)
+
+proc bindNull*( ps : SqlPrepared, paramIdx: int) : int {.inline.}  =
+  ## Sets the bindparam at the specified paramIndex to null 
+  ## (default behaviour by sqlite).
+  result = bind_null(ps, paramIdx.int32) 
+
+proc bindString*( ps: SqlPrepared, paramIdx : int,
+                   val: string,) : int =
```
alaviss: Why is there a trailing `,` here?

lib/impure/db_sqlite.nim
```diff
@@ -716,8 +719,64 @@ proc setEncoding*(connection: DbConn, encoding: string): bool {.
...
+          discard
+      of akString:
+        if SQLITE_OK != ps.bindString(idx,v.getString):
+          discard
+      of akCString:
+        if SQLITE_OK != ps.bindBlob(idx,v.getCString):
+          discard
+      else:
+        discard
+    inc idx
```
alaviss: This should be a template/macro, then it can bind the type using the info available at compile time.

Using runtime information like this is dangerous.

lib/impure/db_sqlite.nim
bung87: hmm, I can copy from somewhere, I have no confidence with my english skill.

lib/impure/db_sqlite.nim
bung87: ok, very nice, let me change it in next commit.

lib/impure/db_sqlite.nim
bung87: ok , let me try

lib/impure/db_sqlite.nim
```diff
@@ -716,8 +719,60 @@ proc setEncoding*(connection: DbConn, encoding: string): bool {.
...
+          discard
+    elif v is string:
+        if SQLITE_OK != ps.bindParam(idx,v):
+          discard
+    elif v is cString:
+        if SQLITE_OK != ps.bindParam(idx,v):
+          discard
+    else:
+      discard
+    inc idx
```
alaviss: If done this way, then only one Nim parameter can be bound. Instead, turn this into a macro that takes a `vararg[untyped]`, then just generate code that calls `bindParam()` on each entry.

The syntax I have in mind is like this:

`statement.bindParams(10, "foo", 42.0)`

Which should generates:

```nim
  statement.bindParam(10, 0)
  statement.bindParam("foo", 1)
  statement.bindParam(42.0, 2)
```

IMO this would feel more intuitive.

lib/impure/db_sqlite.nim
```diff
@@ -716,8 +719,60 @@ proc setEncoding*(connection: DbConn, encoding: string): bool {.
...
+
+proc bindParam*(ps: SqlPrepared, paramIdx: int,val: int64):int =
+  ## Binds a int64  to the specified paramIndex.
+  result = bind_int64(ps.PStmt, paramIdx.int32, val)
+
+proc bindParam*(ps: SqlPrepared, paramIdx: int, val: float64):int =
+  ## Binds a 64bit float to the specified paramIndex.
+  result = bind_double(ps.PStmt, paramIdx.int32, val)
+
+proc bindParam*(ps: SqlPrepared, paramIdx: int):int =
```
alaviss: This one should still be `bindNull`, since it's a special case.

lib/impure/db_sqlite.nim
```diff
@@ -716,8 +719,60 @@ proc setEncoding*(connection: DbConn, encoding: string): bool {.
 proc finalize*(sqlPrepared:SqlPrepared){.discardable.} = 
   discard finalize(sqlPrepared.PStmt)
 
+proc bindParam*(ps: SqlPrepared, paramIdx: int, val: int32):int =
```
alaviss: The `db_*` family of modules tries to abstract away vendor-specific error codes. So instead of returning the error code, you should throw an exception here with the error.

lib/impure/db_sqlite.nim
bung87: I turn it to template and handle iterable. make logic simple, otherwise I should considering `bindParams(10,[2,3],3)`

lib/impure/db_sqlite.nim
bung87: I add a `dbBindParamError` template,
but I dunno how to display val, val could be seq[char] and very long.
`e.msg = "error binding param in position " & $paramIdx `

alaviss: You don't have to, just say the position and the vendor-specific error message.

bung87: In that proc, does not need db context, it's hard to decide pass db context only for get db error message.

lib/impure/db_sqlite.nim
```diff
@@ -716,8 +719,64 @@ proc setEncoding*(connection: DbConn, encoding: string): bool {.
...
+  e.msg = "error binding param in position " & $paramIdx 
+  raise e
+
+proc bindParam*(ps: SqlPrepared, paramIdx: int, val: int32) =
+  ## Binds a int32  to the specified paramIndex.
+  if SQLITE_OK != bind_int(ps.PStmt, paramIdx.int32, val):
+    dbBindParamError(paramIdx,val)
+
+proc bindParam*(ps: SqlPrepared, paramIdx: int, val: int) =
+  ## Binds a int32  to the specified paramIndex.
```
alaviss: The `int` version can perform a simple detection to determine which `int` size should be used by checking `sizeof(int)`. Truncating `int` (which would be `int64` on 64-bit platforms) to `int32` can be a potential source of hidden bugs.

lib/impure/db_sqlite.nim
```diff
@@ -716,8 +719,64 @@ proc setEncoding*(connection: DbConn, encoding: string): bool {.
...
+  if SQLITE_OK != bind_int(ps.PStmt, paramIdx.int32, val.int32):
+    dbBindParamError(paramIdx,val)
+
+proc bindParam*(ps: SqlPrepared, paramIdx: int,val: int64) =
```
alaviss:
```suggestion
proc bindParam*(ps: SqlPrepared, paramIdx: int, val: int64) =
```

Nit, just some cosmetic.

lib/impure/db_sqlite.nim
```diff
@@ -716,8 +719,64 @@ proc setEncoding*(connection: DbConn, encoding: string): bool {.
...
+  if SQLITE_OK != bind_double(ps.PStmt, paramIdx.int32, val):
+    dbBindParamError(paramIdx,val)
+
+proc bindNull*(ps: SqlPrepared, paramIdx: int) =
+  ## Sets the bindparam at the specified paramIndex to null 
+  ## (default behaviour by sqlite).
+  if SQLITE_OK != bind_null(ps.PStmt, paramIdx.int32):
+    dbBindParamError(paramIdx,val)
+
+proc bindParam*(ps: SqlPrepared, paramIdx: int,val: string) =
```
alaviss:
```suggestion
proc bindParam*(ps: SqlPrepared, paramIdx: int, val: string) =
```

lib/impure/db_sqlite.nim
```diff
@@ -716,8 +719,64 @@ proc setEncoding*(connection: DbConn, encoding: string): bool {.
...
+  ## Binds a string to the specified paramIndex.
+  if SQLITE_OK != bind_text(ps.PStmt, paramIdx.int32,val.cstring,-1.int32 , SQLITE_STATIC):
+    dbBindParamError(paramIdx,val)
+
+proc bindParam*(ps: SqlPrepared, paramIdx: int,val: cstring) =
+  ## binds a blob to the specified paramIndex.
+  let len = val.len
+  if SQLITE_OK != bind_blob(ps.PStmt, paramIdx.int32, val[0].unSafeAddr, len.int32 , SQLITE_STATIC):
```
alaviss:
```suggestion
  if SQLITE_OK != bind_blob(ps.PStmt, paramIdx.int32, val, len.int32 , SQLITE_STATIC):
```

It's already a cstring, you don't have to do the `val[0].unsafeAddr` dance

lib/impure/db_sqlite.nim
```diff
@@ -716,8 +719,64 @@ proc setEncoding*(connection: DbConn, encoding: string): bool {.
...
+
+proc bindNull*(ps: SqlPrepared, paramIdx: int) =
+  ## Sets the bindparam at the specified paramIndex to null 
+  ## (default behaviour by sqlite).
+  if SQLITE_OK != bind_null(ps.PStmt, paramIdx.int32):
+    dbBindParamError(paramIdx,val)
+
+proc bindParam*(ps: SqlPrepared, paramIdx: int,val: string) =
+  ## Binds a string to the specified paramIndex.
+  if SQLITE_OK != bind_text(ps.PStmt, paramIdx.int32,val.cstring,-1.int32 , SQLITE_STATIC):
```
alaviss: This should be `bind_blob` too, since it's not atypical for us to have people using strings as byte buffers.

lib/impure/db_sqlite.nim
```diff
@@ -716,8 +719,64 @@ proc setEncoding*(connection: DbConn, encoding: string): bool {.
...
+template bindParams*(ps: SqlPrepared, params:untyped) = 
+  var idx:int = 1
+  when params is tuple or params is object:
+    for v in params.fields:
+      ps.bindParam(idx,v)
+      inc idx
+  else:
+    for v in params:
+      ps.bindParam(idx,v)
+      inc idx
```
alaviss: This template limits the current use case, since it forces you to use a tuple and/or objects when you want a set of differently typed arguments.

```nim
ps.bindParams((0, 42.0, "string"))
ps.bindParams((value: 0, str: "foo"))
```

This can also cause confusion if users just unknowingly bind an object thinking that it will bind the object as one parameter.

Instead if we use a macro like this:

```nim
macro bindParams(ps: SqlPrepared, params: varargs[untyped]): untyped =
  let bindParam = bindSym("bindParam", brOpen)
  let preparedStatement = genSym()
  result = newStmtList()
  # Store `ps` in a temporary variable. This prevents `ps` from being evaluated every call.
  result.add newNimNode(nnkLetSection).add(newIdentDefs(preparedStatement, newEmptyNode(), ps))
  for idx, param in params:
    result.add newCall(bindParam, preparedStatement, newIntLit idx, param)
```

Then binding will simply use the typically expected API of:

```nim
ps.bindParams(1, 10, 10000, "string", 42.0)
```

Not to mention that it will transparently work with the case where only one argument is used:

```nim
ps.bindParams("string")
```

lib/impure/db_sqlite.nim
```diff
@@ -716,8 +719,64 @@ proc setEncoding*(connection: DbConn, encoding: string): bool {.
...
+  if SQLITE_OK != bind_null(ps.PStmt, paramIdx.int32):
+    dbBindParamError(paramIdx,val)
+
+proc bindParam*(ps: SqlPrepared, paramIdx: int,val: string) =
+  ## Binds a string to the specified paramIndex.
+  if SQLITE_OK != bind_text(ps.PStmt, paramIdx.int32,val.cstring,-1.int32 , SQLITE_STATIC):
+    dbBindParamError(paramIdx,val)
+
+proc bindParam*(ps: SqlPrepared, paramIdx: int,val: cstring) =
+  ## binds a blob to the specified paramIndex.
```
alaviss, line 771: The typical Nim standard for a blob is a sequence of bytes or `openArray[byte]`. We never use `cstring` for these things.

If you want to provide a special bind for `cstring`, make it so that the user has to specify the size, since we don't know if the `cstring` buffer is a NUL-terminated one to pass to `len()`

lib/impure/db_sqlite.nim
alaviss: `sqlite3_errstr` should work, though I can see that the interface is rather new :/

Maybe we can skip on the vendor-specific error message for now.

lib/impure/db_sqlite.nim
bung87: then how to implement bind_text, string , cstring is used for blob, I have no idea

alaviss: Do we need `bind_text`? or can sqlite work this out on it's own?

AFAIK `ormin` uses `bind_blob` and it works pretty well: https://github.com/Araq/ormin/blob/master/ormin/ormin_sqlite.nim#L36-L38

lib/impure/db_sqlite.nim
bung87: hmm, I thought db c api support bind param by index or name, so tuple and object ,array, seq are good choice. when `params` becoming  `varargs[untyped]` I dont know how to handle unexpected objects. passing user defined type object just ignored.

lib/impure/db_sqlite.nim
bung87: ah , that's a nice api, considering in next PR.

lib/impure/db_sqlite.nim
alaviss: > I dont know how to handle unexpected objects.

If no `bindParam` proc was defined for a particular type, this would simply error at compile time, same as with how your code works atm.

> db c api support bind param by index or name.

Then we should expose single-param versions of those, then we can think of using something like `openArray[(string, untyped)]` to enable the `{"param1": 10, "foo": "bar"}` syntax. The JSON module already did this, so we can learn from the logic there.

lib/impure/db_sqlite.nim
bung87: there is also `sqlite3_bind_text16`,`sqlite3_bind_text64` , developer may handle compressed data or runes.

lib/impure/db_sqlite.nim
```diff
@@ -125,6 +125,8 @@ type
   InstantRow* = PStmt ## A handle that can be used to get a row's column
                       ## text on demand.
   SqlPrepared* = distinct PStmt ## a identifier for the prepared queries
+  SQLite3UnbindCb = Tbind_destructor_func ## called by sqlite3 to release the string/blob ptr
+  RCode* = tuple[vendorcode: int, errStr: string]
```
alaviss: What is this type for?

lib/impure/db_sqlite.nim
alaviss: We can actually go ahead and store `DbConn` inside `SqlPrepared` (which will become an object, preferably with the fields exported).

It would just be easier if sqlite provides access to the db handle associated with the prepared statement...

lib/impure/db_sqlite.nim
bung87:
``` nim
when sizeof(val) != 4:
    {.error: "val should be size of 4".}
```
how does this looks?

lib/impure/db_sqlite.nim
bung87: may handle error in lib level, as `db_postgres` already declared `SqlPrepared* = distinct string` ,and easier unify the dbs api.

lib/impure/db_sqlite.nim
bung87: oh, I know it now, proc only declared when sizeof(int) ==4

lib/impure/db_sqlite.nim
alaviss: > developer may handle compressed data or runes.

Not sure what you mean by this.

lib/impure/db_sqlite.nim
```diff
@@ -716,8 +717,63 @@ proc setEncoding*(connection: DbConn, encoding: string): bool {.
...
+proc bindParam*(ps: SqlPrepared, paramIdx: int, val: int32) =
+  ## Binds a int32  to the specified paramIndex.
+  if SQLITE_OK != bind_int(ps.PStmt, paramIdx.int32, val):
+    dbBindParamError(paramIdx,val)
+
+when sizeof(int) == 4:
+  proc bindParam*(ps: SqlPrepared, paramIdx: int, val: int) =
+    ## Binds a int32  to the specified paramIndex.
+    if SQLITE_OK != bind_int(ps.PStmt, paramIdx.int32, val.int32):
+      dbBindParamError(paramIdx,val)
```
alaviss: You could have just implement this so that it calls `bindParam(val.int32)` when `sizeof(int) == 4` and `bindParam(val.int64)` otherwise...

lib/impure/db_sqlite.nim
```diff
@@ -716,8 +717,63 @@ proc setEncoding*(connection: DbConn, encoding: string): bool {.
...
+  ## (default behaviour by sqlite).
+  if SQLITE_OK != bind_null(ps.PStmt, paramIdx.int32):
+    dbBindParamError(paramIdx,val)
+
+proc bindParam*(ps: SqlPrepared, paramIdx: int, val: string) =
+  ## Binds a string to the specified paramIndex.
+  if SQLITE_OK != bind_text(ps.PStmt, paramIdx.int32,val.cstring,-1.int32 , SQLITE_STATIC):
+    dbBindParamError(paramIdx,val)
+
+proc bindParam*(ps: SqlPrepared, paramIdx: int,val: openArray[byte]) =
```
alaviss:
```suggestion
proc bindParam*(ps: SqlPrepared, paramIdx: int, val: openArray[byte]) =
```

Please correct all the spacing inconsistency in this module.

lib/impure/db_sqlite.nim
bung87: ok, I checked ormin , it always using bind int64

bung87: cstring cant store compressed data right? since there's many zero terminers inside the data.
so the binding api should have pass string version

lib/impure/db_sqlite.nim
bung87: cstring cant store compressed data right? since there's many zero terminers inside the data.
so the binding api should have pass string version

alaviss: You shouldn't have to implement `bind_text`. Probably `bind_blob` is good enough for our use. Is there any practical difference between `bind_blob` and `bind_text`?

lib/impure/db_sqlite.nim
```diff
@@ -716,8 +719,67 @@ proc setEncoding*(connection: DbConn, encoding: string): bool {.
...
+
+proc bindParam*(ps: SqlPrepared, paramIdx: int, val: string) {.since: (1, 3).} =
+  ## Binds a string to the specified paramIndex.
+  if SQLITE_OK != bind_text(ps.PStmt, paramIdx.int32, val.cstring, -1.int32, SQLITE_STATIC):
+    dbBindParamError(paramIdx, val)
+
+proc bindParam*(ps: SqlPrepared, paramIdx: int,val: openArray[byte]) {.since: (1, 3).} =
+  ## binds a blob to the specified paramIndex.
+  let len = val.len
+  if SQLITE_OK != bind_blob(ps.PStmt, paramIdx.int32, val[0].unsafeAddr, len.int32, SQLITE_STATIC):
```
alaviss: Is `SQLITE_STATIC` the way to go here? or should be use `SQLITE_TRANSIENT`?

Also, can you follow the general style in this module and compare with `SQLITE_OK` on the right hand side?

lib/impure/db_sqlite.nim
```diff
@@ -716,8 +719,67 @@ proc setEncoding*(connection: DbConn, encoding: string): bool {.
...
+    dbBindParamError(paramIdx, val)
+
+macro bindParams(ps: SqlPrepared, params: varargs[untyped]): untyped =
```
alaviss: This one should get the `since` treatment too.

changelog.md
```diff
@@ -3,6 +3,7 @@
 
 
 ## Standard library additions and changes
+- Add `bindParams`,`bindParam` to `db_sqlite`.
```
alaviss: Maybe describe a bit on what these new procs do?

lib/impure/db_sqlite.nim
bung87: ok, I change it to right hand, SQLITE_TRANSIENT will cause sqlite copy the value.

alaviss: Yea but if someone modifies the array right after they call `bindParam`, then would it keep using the old array as the caller would expects?

lib/impure/db_sqlite.nim
bung87: If the fourth parameter to sqlite3_bind_text() or sqlite3_bind_text16() is negative, then the length of the string is the number of bytes up to the first zero terminator.

 https://sqlite.org/c3ref/bind_blob.html
bind_text is zero terminate for current implementation , and no need count size,
bind_blob is  NUL terminated , prefill a size

alaviss: Just pass the length to it? Our strings keep track of it's length, we do absolutely no computation on `len()` of a string.

lib/impure/db_sqlite.nim
bung87: I have not go that further, if someone modify the param, the prepared can store right value without through bind_* ? if so I assuming developer know what they do when they change the bound value

alaviss: Then maybe add a warning stating that the value will not be copied?

An another factor that worries me is when the GC collects the memory, for example:

```nim
proc bindFoo(ps: SqlPrepared, idx: int) =
  var s = "foo"
  ps.bindParam(idx, s)

  # Once control flow exit the proc, `s` might be collected and
  # the data will be invalid!
```

No copy will only be safe if you're in control of the data for the entire period of bind->execution->reset.

bung87: hmm, for now I think just give comment note, maybe implement param pass by value version, or type that has destructor, and also a template runnning entire period in the future

lib/impure/db_sqlite.nim
```diff
@@ -290,7 +293,8 @@ iterator fastRows*(db: DbConn, query: SqlQuery,
     if finalize(stmt) != SQLITE_OK: dbError(db)
 
 iterator fastRows*(db: DbConn, stmtName: SqlPrepared,
-                   args: varargs[string, `$`]): Row {.tags: [ReadDbEffect,WriteDbEffect].} =
+                   args: varargs[string, `$`]): Row 
+                  {.tags: [ReadDbEffect,WriteDbEffect], since: (1, 3).} =
```
alaviss, line 292: Why are you adding `since` to this proc? It's not a new proc.

We use `since` to signify when an API is added in a new version of Nim.

changelog.md
```diff
@@ -3,6 +3,7 @@
 
 
 ## Standard library additions and changes
+- Add `bindParams`,`bindParam` to `db_sqlite`,call `prepare` get `SqlPrepared` then `bindParams` or `bindParam` to it then `exec` then `finalize` the `SqlPrepared`.
```
alaviss:
```suggestion
- Added `bindParams`, `bindParam` to `db_sqlite` for binding parameters into a `SqlPrepared` statement.
```

I think this should work better.

lib/impure/db_sqlite.nim
```diff
@@ -713,11 +720,70 @@ proc setEncoding*(connection: DbConn, encoding: string): bool {.
...
+
+
+proc bindParam*(ps: SqlPrepared, paramIdx: int, val: int) {.since: (1, 3).} =
+  ## Binds a int  to the specified paramIndex.
+  when sizeof(int) == 8:
+    if bind_int64(ps.PStmt, paramIdx.int32, val.int64) != SQLITE_OK:
+      dbBindParamError(paramIdx, val)
+  else:
+    if bind_int(ps.PStmt, paramIdx.int32, val.int32) != SQLITE_OK:
+      dbBindParamError(paramIdx, val)
```
alaviss: Instead of duplicating code, move this proc below the `int64` overload, then call each of the overloads depending on the size of `int`

lib/impure/db_sqlite.nim
```diff
@@ -713,11 +720,70 @@ proc setEncoding*(connection: DbConn, encoding: string): bool {.
...
+
+proc bindNull*(ps: SqlPrepared, paramIdx: int) {.since: (1, 3).} =
+  ## Sets the bindparam at the specified paramIndex to null 
+  ## (default behaviour by sqlite).
+  if bind_null(ps.PStmt, paramIdx.int32) != SQLITE_OK:
+    dbBindParamError(paramIdx, val)
+
+proc bindParam*(ps: SqlPrepared, paramIdx: int, val: string) {.since: (1, 3).} =
+  ## Binds a string to the specified paramIndex.
+  if bind_text(ps.PStmt, paramIdx.int32, val.cstring, -1.int32, SQLITE_STATIC) != SQLITE_OK:
```
alaviss: Our strings may be NUL-terminated, but that doesn't stop anyone from adding NUL to their strings. Instead of making SQLite guess the length, pass them the length that we already know.

```suggestion
  if bind_text(ps.PStmt, paramIdx.int32, val.cstring, val.len.int32, SQLITE_STATIC) != SQLITE_OK:
```

I still think this should be `bind_blob` however.

Since we are using `SQLITE_STATIC` here, a warning to potential users would be nice.

Alternatively, maybe introduce a `copy` parameter at the end which dictates whether the data should be copied by SQLite? We can give it a default value of `true` to be on the safe side.

lib/impure/db_sqlite.nim
bung87: `SqlPrepared` is last PR I added

lib/impure/db_sqlite.nim
```diff
@@ -728,22 +723,93 @@ proc setEncoding*(connection: DbConn, encoding: string): bool {.
...
+  if n.kind == k: error("UnExpected a node of kind " & $k ,n)
+
+macro bindParams*(ps: SqlPrepared, params: varargs[untyped]): untyped {.since: (1, 3).} =
+  let bindParam = bindSym("bindParam", brOpen)
+  let preparedStatement = genSym()
+  result = newStmtList()
+  # Store `ps` in a temporary variable. This prevents `ps` from being evaluated every call.
+  result.add newNimNode(nnkLetSection).add(newIdentDefs(preparedStatement, newEmptyNode(), ps))
+  for idx, param in params:
+    unExpectKind param, nnkBracket
```
alaviss: Why is this?

alaviss:
We should offer an API to reset a prepared statement so that it could be reused too.

lib/impure/db_sqlite.nim
bung87: avoiding developer pass empty array, that bind typEmpty to param. which is a runtime error, as prevous api test pass `[]`. that may cause miss leading.

bung87:
> We should offer an API to reset a prepared statement so that it could be reused too.

that's in `sqlite3` lib ,should make a wrap version in `db_sqlite` ? if so I'll do that.

lib/impure/db_sqlite.nim
alaviss: Well then this wouldn't work. Then can use a long expression which result in an empty array and we wouldn't know. It's best not to meddle with this, the devs are supposed to understand nim's parameter passing semantics.

We might also want to map `nil` to `bindNull`. We can make it so that only `nil` literals can be bound like this.

bung87: ok , about bind in next PR or this PR , what do you think ?

Araq: This PR.

Issue #9514, opened 2018-10-25
[Perf] Max/min are not inline and slow (7x slower than memory-bound implementation)


mratsim:
## Context

While implementing high-performance computing kernels and checking [kadro library](https://github.com/bluenote10/kadro) by @bluenote10, I noticed that he tried to [optimize Nim `max` implementation](https://github.com/bluenote10/kadro/blob/f785c6863028820c240180ae391d8150f08a93d4/src/kadro/backend_standard/p_columns_ops_typed.nim#L124-L151). I also benchmarked it and there a huge function call overhead in the current implementation in system.nim. See also https://github.com/bluenote10/kadro/issues/2#issuecomment-433047530. This also part of the cause of Arraymancer perf issues mentioned by @andreaferretti in https://github.com/mratsim/Arraymancer/issues/265.

The `relu` operation is basically max(0, x) and is a staple in Neural Networks and can be applied repeatedly on tensors of hundreds of millions of parameters:

```Nim
# This is a simple image recognition neural net
class Net(nn.Module):
    def __init__(self):
        super(Net, self).__init__()
        self.conv1 = nn.Conv2d(1, 20, 5)
        self.conv2 = nn.Conv2d(20, 50, 5)
        self.linear1 = nn.Linear(800, 500)
        self.linear2 = nn.Linear(500, 10)

    def forward(self, x):
        batch_size, _, _, _ = x.size()
        x = self.conv1(x)
        x = F.relu(F.max_pool2d(x, 2)) # <------ max(0, x)
        x = self.conv2(x)
        x = F.relu(F.max_pool2d(x, 2)) # <------ max(0, x)
        x = x.view(batch_size, 800)
        x = F.relu(self.linear1(x)) # <------ max(0, x)
        x = self.linear2(x)
        return x
```

## Benchmark results

Benchmark is done on i5-5257U 2.7Ghz turbo 3.1 of Broadwell generation from a macBook Pro 2015

I benchmarked Nim system, an "inline max" which is just a copy paste with inline, and a SSE3 implementation which bottlenecks on how fast you can retrieve data from memory for the theoretical maximum throughput.

Let's start with the results

### Only `-d:release`
```bash
Warmup: 1.2016 s, result 224 (displayed to avoid compiler optimizing warmup away)

Reduction - system.nim - max - float32
Collected 1000 samples in 20.024 seconds
Average time: 20.020 ms
Stddev  time: 0.605 ms
Min     time: 19.646 ms
[...]

PR #14417, opened 2020-05-21
The whole options module should be inline

mratsim:
`none`  shows up as a bottleneck in benchmark :/

I'm not going to turn this into another blog post like https://github.com/nim-lang/Nim/issues/9514 but this takes 45s to raytrace in C++ and the Nim code is almost equivalent besides choosing to use Option return values.

For example the `none` function in the benchmark below

image

Ideally this is a small enough change to be backported in 1.2.2 because we do use Option in Nimbus.

Furthermore the values are quite huge to wrap/unwrap (256-bit hash) so inline should help significantly.


Araq:
I prefer explict ``{.inline.}`` on the procs you need to be inline. Also, ``lent`` annotations might help.

mratsim:
`lent` doesn't seem to work at compile-time.

The tests have a lot of
```
../../lib/pure/options.nim(171, 3) Error: limited VM support for 'addr'
```

arnetheduck:
@mratsim did you try to simply add `-flto` to the compiler flags?

PR #14418, opened 2020-05-21
change the [Processing] messages into dots

Araq:


mratsim:
Is there a reason for this?

I've had the compiler suddenly `quit 0` on me with no error message and the hint was helpful to narrow down a minimal reproducing example.

Araq:
> I've had the compiler suddenly quit 0 on me with no error message and the hint was helpful to narrow down a minimal reproducing example.

Well we have ``--verbosity:2`` or similar for that.

ghost:
A small bug - if there are no warnings when compiling, there's no newline before the first "CC" message:

timotheecour:
@Yardanico did https://github.com/nim-lang/Nim/pull/14415 fix your issue? I'm not seeing it after that PR

```
nim c --gc:arc -d:danger -r --hint:cc --listcmd:off --hint:conf -f $timn_test_hello_F
Hint: used config file '/Users/timothee/git_clone/nim/Nim_devel/config/nim.cfg' [Conf]
Hint: used config file '/Users/timothee/git_clone/nim/Nim_devel/config/config.nims' [Conf]
Hint: used config file '/Users/timothee/git_clone/nim/timn/nim.cfg' [Conf]
Hint: used config file '/Users/timothee/git_clone/nim/timn/config.nims' [Conf]
compiling /Users/timothee/git_clone/nim/timn/src/timn/examples/thello.nim
CC: stdlib_io.nim
CC: stdlib_system.nim
CC: thello.nim
Hint: 82917 lines; 0.482s; 95.363MiB peakmem; Dangerous Release build; proj: /Users/timothee/git_clone/nim/timn/src/timn/examples/thello.nim; out: /Users/timothee/git_clone/nim/timn/src/timn/examples/thello [SuccessX]
hello world

```

using nim c61f7466292cb19c910f4a2f2976f738796e6bdf 1.3.5

@mratsim 
> I've had the compiler suddenly quit 0 on me with no error message and the hint was helpful to narrow down a minimal reproducing example.

`--hint:msgorigin` should help (should work with all commands as of https://github.com/nim-lang/Nim/pull/14415)

that being said, I recently had cases where I wanted to see full path of files as they were being processed to debug which file was actually picked up (common case when you're debugging --`path` issues); 

instead of using a `.`, for systems that support ansii escape sequences (posix + windows 10), there's a better alternative: write the full path, then erase the line using exact same logic as `flushDot`

this would be a strict improvement; however it still doesn't help if you need to see prior paths, not just last one; I'm scratching my head real hard for how to allow that without introducing a new flag (maybe `-d:nimHintProcessingFullPath` doesn't count as a new flag?)

ghost:
@timotheecour seems like it didn't help (no newline before first CC with no arguments, the nim.cfg just contains threads:on)

PR #14446, opened 2020-05-25
Document NVRO

Araq:
WIP; refs https://github.com/nim-lang/RFCs/issues/230

### DIFF
--- a/doc/manual.rst
+++ b/doc/manual.rst
@@ -1802,7 +1802,7 @@ Mixing GC'ed memory with ``ptr``
 
 Special care has to be taken if an untraced object contains traced objects like
 traced references, strings or sequences: in order to free everything properly,
-the built-in procedure ``GCunref`` has to be called before freeing the untraced
+the built-in procedure ``reset`` has to be called before freeing the untraced
 memory manually:
 
 .. code-block:: nim
@@ -1816,12 +1816,12 @@ memory manually:
   d.s = "abc"
 
   # tell the GC that the string is not needed anymore:
-  GCunref(d.s)
+  reset(d.s)
 
   # free the memory:
   dealloc(d)
 
-Without the ``GCunref`` call the memory allocated for the ``d.s`` string would
+Without the ``reset`` call the memory allocated for the ``d.s`` string would
 never be freed. The example also demonstrates two important features for low
 level programming: the ``sizeof`` proc returns the size of a type or value
 in bytes. The ``cast`` operator can circumvent the type system: the compiler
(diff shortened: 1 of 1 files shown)


doc/manual.rst
```diff
@@ -3707,6 +3707,71 @@ location is derived from the second parameter (called
...
+
+  var x; p(x)
+  p(x)
+
+
+Let ``T``'s be ``p``'s return type. NRVO applies for ``T``
+if ``sizeof(T) >= N`` (where ``N`` is implementation dependent),
+in other words, it applies for "big" structures.
+
+If ``p`` can raise an exception, NRVO applies regardless. This can produce
```
arnetheduck, line 3742: is this really going to be documented as a feature?

It's extremely surprising behaviour that breaks with the most basic foundations of what a function invocation is, sacrificing basic correctness in the name of a minor optimization - if this behaviour is wanted for whatever reason (in a case where the performance would matter), it's trivial to rewrite the function to take a `var` parameter to make the destructive behaviour clearly visible, but the bugs that will follow from this kind of approach will be hard to find in any larger codebase, and will worsen as more deterministic resource management is baked into the language - ie consider when a destructor has to roll back a partially updated instance due to a gotcha like this.

the compiler (in theory) can already remove most cases of the copying when it can prove that the result is not observable - ie it can apply to optimization to where the instance is "fresh" - `let x = xxx()` etc.

If anything, one could imagine a hint for when `NRVO` is _not_ applied could be useful - ie "missed optimization" warning.

doc/manual.rst
```diff
@@ -3707,6 +3707,71 @@ location is derived from the second parameter (called
...
+    except ValueError:
+      doAssert x == [0, 1, 2, 3, 4, 5, 6, 7, 0, 0, 0, 0, 0, 0, 0, 0]
+
+  main()
+
+
+However, the current implementation produces a warning in these cases.
+There are different ways to deal with this warning:
+
+1. Disable the warning via ``{.push warning[ObservableStores]: off.}`` ... ``{.pop.}``.
```
arnetheduck, line 3768: it's pretty odd if the compiler identifies a language behaviour as undesireable enough to produce a warning, and next the manual recommends disabling it - is the manual writer fighting the compiler writer over this?

it seems that point `2.` would be enough as far as recommendations go - it's a workaround for now before the compiler / language can be improved to avoid the behaviour altogether - that warnings can be disabled is documented elsewhere already.

doc/manual.rst
Araq: I think the place for discussion is here https://github.com/nim-lang/RFCs/issues/230

doc/manual.rst
Araq: Again, please argue in https://github.com/nim-lang/RFCs/issues/230 

I'm very willing to do it differently, I don't particularly like this solution, it merely seemed the least problematic since I really want to have inplace construction.

PR #14457, opened 2020-05-26
Change `UnpackError` with `UnpackDefect`

bobeff:
The deprecation warning for `UnpackError` exception raised by some `inline` procedures in the Nim standard library propagates to the user code. If the user code has a requirement for building without warnings this is a problem for the successful execution of the tests. In order to resolve this, all occurrences of `UnpackError` in the Nim code base are changed to `UnpackDefect`. Only the type alias is retained to not break other people's user code since `UnpackError` is exported type.

### DIFF
--- a/lib/pure/nativesockets.nim
+++ b/lib/pure/nativesockets.nim
@@ -445,9 +445,10 @@ proc getSockDomain*(socket: SocketHandle): Domain =
   if getsockname(socket, cast[ptr SockAddr](addr(name)),
                  addr(namelen)) == -1'i32:
     raiseOSError(osLastError())
-  try:
-    result = toKnownDomain(name.sin6_family.cint).get()
-  except UnpackDefect:
+  let knownDomain = toKnownDomain(name.sin6_family.cint)
+  if knownDomain.isSome:
+    result = knownDomain.get()
+  else:
     raise newException(IOError, "Unknown socket family in getSockDomain")
 
 proc getAddrString*(sockAddr: ptr SockAddr): string =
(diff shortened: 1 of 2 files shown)


ghost:
There might be a problem with backwards compatibility though

bobeff:
> There might be a problem with backwards compatibility though

I can't imagine why. I retained the type alias and both `UnpackError` and `UnpackDefect` should be valid in the user code. Can you give an example?

Araq:
Your patch is very good but the code was wrong to begin with, one cannot catch an `UnpackDefect`.

lib/pure/nativesockets.nim
```diff
@@ -447,7 +447,7 @@ proc getSockDomain*(socket: SocketHandle): Domain =
     raiseOSError(osLastError())
   try:
     result = toKnownDomain(name.sin6_family.cint).get()
-  except UnpackError:
+  except UnpackDefect:
```
ghost: As @Araq said, I guess you should change this code to something like:
```nim
let res = toKnownDomain(name.sin6_family.cint)
if res.isSome():
  result = res.get()
else:
  raise newException(IOError, "Unknown socket family in getSockDomain")
```

bobeff: It's fixed.

lib/pure/nativesockets.nim
```diff
@@ -445,9 +445,10 @@ proc getSockDomain*(socket: SocketHandle): Domain =
   if getsockname(socket, cast[ptr SockAddr](addr(name)),
                  addr(namelen)) == -1'i32:
     raiseOSError(osLastError())
-  try:
+  let knownDomain = toKnownDomain(name.sin6_family.cint)
+  if knownDomain.isSome:
     result = toKnownDomain(name.sin6_family.cint).get()
```
ghost: uhh, this is not correct since you call it again, you should have "knownDomain.get()" instead :P

bobeff: Yes. I'm very careless. :(

bobeff: It's fixed.

PR #14462, opened 2020-05-26
Revert "niminst: use threaded compression when supported"

alaviss:
While compression speed is greatly increased with threaded xz, the compression ratio suffers heavily.

The generated tarball went from 4.9MB to 10MB, which is 2x the size and doesn't justify the speed up.

I may look into the potential to employ zstd here instead and see if it provides us with better speed/compression ratio tradeoff.

Reverts nim-lang/Nim#14455

### DIFF
--- a/tools/niminst/niminst.nim
+++ b/tools/niminst/niminst.nim
@@ -700,10 +700,8 @@ RunProgram="tools\downloader.exe"
           if execShellCmd("tar cf $1.tar $1" % proj) != 0:
             echo("External program failed")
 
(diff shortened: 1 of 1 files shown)


juancarlospaco:
Maybe use a `when defined(threadedCompression)`, let user choose Speed or Compression (?)

Whatsoever both can not be used because https://github.com/nim-lang/Nim/issues/11433

alaviss:
Nah, tools like these are better off having more customizable command line, not compile flags.

Issue #13967, opened 2020-04-13
dup/with cannot be used to modify fields?


bluenote10:
### Summary

I was trying to upgrade a DSL of mine to modern Nim using `dup`/`with` (relying on a builder pattern so far), but I cannot figure out how to write one-line-expressions that modify fields.

I thought I can use them to turn an immutable instance of an object into a new immutable instance with some fields modified (the equivalent of case class copies like `original.copy(field="modified")` in Scala).

### Description

```nim
import sugar, with

type
  User = object
    id: int
    name: string
    age: int
  
let user = User(id: 42, name: "original", age: 42)
echo user
echo user.dup(name="modified")
echo user.with(name="modified")
echo user.dup.with(name="modified")
echo user.shallowCopy().with(name="modified")
```

First line errors with:
```
Error: invalid expression: name = "modified"
```

Second / third lines have similar errors:
```
Error: type mismatch: got <User, name: string>
but expected one of: 
macro with(x: typed; body: untyped): untyped
  first type mismatch at position: 2
  unknown named parameter: name

expression: with(
  var dupResult_3786220 = user
  dupResult_3786220, name = "modified")
```

Forth line errors with:
```
Error: type mismatch: got <User>
but expected one of: 
proc shallowCopy[T](x: var T; y: T)
  first type mismatch at position: 1
  required type for x: var T
  but expression 'user' is immutable, not 'var'

expression: shallowCopy(user)
```

The only expression that I got to compile so far was
```nim
echo (block:
  var tmpUser = user
  with tmpUser:
    name = "modified"
  tmpUser)
```
which isn't a real help to replace a builder pattern.

Am I missing something? If not, is field modification a feature that can be added to `with`/`dup`?

Araq:
Field modifications are not part of the design. Can they be added? I don't know, feel free to tinker with the implementation, it's rather simple IMO.

Clyybber:
Fixed by #14484

PR #14484, opened 2020-05-28
std/with support field assign

slangmgh:
std/with support field assign like this:
```nim
   type Foo = object
      x, y: int

   proc addx(v: var Foo, n: int) =
      v.x += n

   var v = Foo()
   with v:
      x = 10
      y = 20
      addx 5

   echo v
```


Araq:
Also needs a changelog entry.

lib/std/private/underscored_calls.nim
```diff
@@ -26,6 +26,8 @@ proc underscoredCall(n, arg0: NimNode): NimNode =
     for i in 1..u-1: result.add n[i]
     result.add arg0
     for i in u+1..n.len-1: result.add n[i]
+  elif n.kind == nnkAsgn:
+    result = newDotExpr(arg0, n[0]).newAssignment n[1]
```
alaviss: Can we have support for the `_.field` syntax too?

alaviss (commented):
I think this should applies to `sugar.dup` as well, right? Then maybe a test or two for that would be nice too.

alaviss:
Possibly solves #13967.

lib/std/private/underscored_calls.nim
slangmgh: Support ``_.field = ...`` have been added, and it can work with sugar.dup now.

lib/std/with.nim
```diff
@@ -49,3 +52,14 @@ when isMainModule:
...
+  f = Foo()
+  with f:
+    col = $(2, 3, 4)
+    pos = $(0.0, 1.0)
+    name = "bar"
+  echo f
+
+  let f2 = f.dup(setPosition(2.0, 3.0)).with:
+    col = $(3, 4, 5)
+  echo f2
```
alaviss: I think you misunderstood. What I meant was that since `sugar.dup` also uses `underscoredCall`, your changes to it should enable it to have syntax like this too:

```nim
let f2 = f.dup:
  setPosition(2.0, 3.0)
  col = $(3, 4, 5)
```

I asked if you could verify this to be the case and add tests for `dup()` to ensure it will keep working, not adding special casing for `.dup.with` :P

lib/std/private/underscored_calls.nim
```diff
@@ -26,6 +26,12 @@ proc underscoredCall(n, arg0: NimNode): NimNode =
     for i in 1..u-1: result.add n[i]
     result.add arg0
     for i in u+1..n.len-1: result.add n[i]
+  elif n.kind == nnkAsgn:
+    var field = n[0]
+    if n[0].kind == nnkDotExpr and n[0][0].eqIdent("_"):
+      # handle _.field = ...
+      field = n[0][1]
```
alaviss, line 33: Can we have tests for this too? It's important to keep things like this working.

alaviss (changes requested):

lib/std/with.nim
slangmgh: The following code  already works:
```nim
let f2 = f.dup:
  setPosition(2.0, 3.0)
  col = $(3, 4, 5)
```

alaviss: Then please add a test for it (in sugar module I persume), and revert the `.dup.with` change.

lib/std/private/underscored_calls.nim
```diff
@@ -41,9 +47,22 @@ proc underscoredCalls*(result, calls, arg0: NimNode) =
         result.add underscoredCall(a, arg0)
 
   expectKind calls, nnkArgList
+
+  # handle the case when arg0 is x.dup
```
Araq: I don't understand why this case has to be handled with special logic. I don't understand how ``arg0[^1].kind == nnkSym`` indicates ``isDup``.

slangmgh: Because the code ``x.dup(sort).with`` will make the arg0 like this:
```nim
var dupResult = x
sort(dupResult)
dupResult
```
Of cause, it is tricky, not understandable and depends on implementation of ``dup``, I will remove it later.

Clyybber (approved):

changelog.md
```diff
@@ -99,6 +99,21 @@
 - `osproc.execCmdEx` now takes an optional `input` for stdin, `workingDir` and `env`
   parameters.
 
+- `std/with` now support object field assignment expression:
```
alaviss: Can you expand this changelog to include `sugar.dup` as well?

alaviss (approved):
LGTM, just one small point.

changelog.md
```diff
@@ -99,7 +99,7 @@
 - `osproc.execCmdEx` now takes an optional `input` for stdin, `workingDir` and `env`
   parameters.
 
-- `std/with` now support object field assignment expression:
+- `std/with`, `sugar/dup` now support object field assignment expression:
```
```suggestion
- `std/with`, `sugar.dup` now support object field assignment expression:
```

lib/std/private/underscored_calls.nim
```diff
@@ -32,6 +32,8 @@ proc underscoredCall(n, arg0: NimNode): NimNode =
       # handle _.field = ...
       field = n[0][1]
     result = newDotExpr(arg0, field).newAssignment n[1]
+  elif n.kind in {nnkExprEqExpr, nnkExprColonExpr}:
+    result = newDotExpr(arg0, n[0]).newAssignment n[1]
```
Clyybber, line 36: Is this for `test(arg = _)` ?
Would be good to have a test for this I think.

lib/std/private/underscored_calls.nim
```diff
@@ -26,24 +26,25 @@ proc underscoredCall(n, arg0: NimNode): NimNode =
...
     result.add arg0
     for i in u+1..n.len-1: result.add n[i]
+  elif n.kind == nnkAsgn:
+    var field = n[0]
+    if n[0].kind == nnkDotExpr and n[0][0].eqIdent("_"):
+      # handle _.field = ...
+      field = n[0][1]
+    result = newDotExpr(arg0, field).newAssignment n[1]
+  elif n.kind in {nnkExprEqExpr, nnkExprColonExpr}:
+    result = newDotExpr(arg0, n[0]).newAssignment n[1]
```
Clyybber:
```suggestion
  elif n.kind in {nnkAsgn, nnkExprEqExpr}:
    var field = n[0]
    if n[0].kind == nnkDotExpr and n[0][0].eqIdent("_"):
      # handle _.field = ...
      field = n[0][1]
    result = newDotExpr(arg0, field).newAssignment n[1]
```
I don't think we should support this weird colon syntax.

lib/pure/sugar.nim
```diff
@@ -337,6 +337,32 @@ macro collect*(init, body: untyped): untyped {.since: (1, 1).} =
...
+      block:
+        let b = a.dup(inc_col, inc_pos):
+          _.pos = 3
+          name_append("bar")
+          inc_pos
+
+        doAssert(b == Foo(col: 2, pos: 4, name: "foobar"))
+
+      block:
+        let b = a.dup(inc_col, pos = 3, name: "bar"):
```
Clyybber:
```suggestion
        let b = a.dup(inc_col, pos = 3, name = "bar"):
```

PR #14502, opened 2020-05-30
Add generic procs to endians module

demotomohiro:
Existsing `littleEndian*` and `bigEndian` are not easy to use as they take addresses of arguments.
So I added 2 generic procedures.

### DIFF
--- a/lib/pure/endians.nim
+++ b/lib/pure/endians.nim
@@ -111,3 +111,29 @@ else:
   proc bigEndian64*(outp, inp: pointer) {.inline.} = swapEndian64(outp, inp)
   proc bigEndian32*(outp, inp: pointer) {.inline.} = swapEndian32(outp, inp)
   proc bigEndian16*(outp, inp: pointer) {.inline.} = swapEndian16(outp, inp)
+
+proc getLittleEndian*[T](val: T): T {.noSideEffect, inline.} =
+  var src = val
+  when sizeof(T) == 1:
+    result = src
+  elif sizeof(T) == 2:
+    littleEndian16(addr result, addr src)
+  elif sizeof(T) == 4:
+    littleEndian32(addr result, addr src)
+  elif sizeof(T) == 8:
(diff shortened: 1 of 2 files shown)


alaviss:
You might want to consider porting [the work done by Status](https://github.com/status-im/nim-stew/blob/master/stew/endians2.nim) to our endians module.

They have a pretty mature API, and not to mention that their version works at compile-time :)

demotomohiro:
@alaviss Thx!
I didn't know about that module and that looks better.
But I'm not sure it should be merged to Nim standard library.
Maybe just adding a link to stew/endian2 in the documentation comment in std/endians module would be enough?

alaviss:
> But I'm not sure it should be merged to Nim standard library.

The end goal of modules in `nim-stew` is to either spin off into their own projects, or become a part of the stdlib.

For `endians2`, it's better for it to be a part of the stdlib, esp. since the stdlib version is terrible compared to it.

demotomohiro:
I found a PR that tried to merge stew/endian2 to std/endian module but didn't get merged.
https://github.com/nim-lang/Nim/pull/13463

alaviss:
https://github.com/nim-lang/Nim/pull/13463#issuecomment-595123705

Apparently that PR introduced a heavily modified version of `endians2`. By the end of the conversation they seem to agree that `endians2` should be merged as-is. So I think the way to go here is to copy `endians2` in-place of `endians`, then implement old APIs on top with `{.deprecated.}` on them.

/cc @Araq @arnetheduck

arnetheduck:
we're happy if stew/endians2 lands in the standard library indeed - we're using it in a codebase that has gone through security audits and significant testing - like anything with `ops` in it, it covers the most basic building blocks / hardware level operations upon which one might build more abstract tools

arnetheduck:
no particular opinion on this PR - we continue to enjoy using https://github.com/status-im/nim-stew/blob/master/stew/endians2.nim

PR #14507, opened 2020-05-30
Always semOperand params

Clyybber:
This is most definitely wrong @zah @Araq .

It fixes this snippet infinitely recursing:
```nim
import macros

var a {.compileTime.} = 0

macro test(n: typed): bool =
  n.expectKind nnkDotExpr
  #if a > 3: quit 1 #Uncomment this to limit the recursion
  #inc a
  if n[1].kind == nnkIdent:
    var n = copyNimTree(n)
    n[0] = newCall("default", n[0])
    result = newCall("test", n)
    echo treeRepr result
  else:
    # This is never reached..
    result = newLit(true)

type Foo = object
  x: int

echo test(Foo.x)
```
because the `x` in the macro generated AST `test(default(Foo).x)` is never turned into a sym, while it should as it does in this example:
```nim
import macros

macro test(n: typed): bool =
  n.expectKind nnkDotExpr
  echo treeRepr n
  result = newLit(true)

type Foo = object
  x: int

echo test(default(Foo).x)
```

### DIFF
--- a/compiler/semexprs.nim
+++ b/compiler/semexprs.nim
@@ -47,7 +47,7 @@ template rejectEmptyNode(n: PNode) =
   if n.kind == nkEmpty: illFormedAst(n, c.config)
 
 proc semOperand(c: PContext, n: PNode, flags: TExprFlags = {}): PNode =
-  rejectEmptyNode(n)
+  if n.kind == nkEmpty: return n #rejectEmptyNode(n)
   # same as 'semExprWithType' but doesn't check for proc vars
   result = semExpr(c, n, flags + {efOperand})
   if result.typ != nil:
--- a/compiler/sigmatch.nim
+++ b/compiler/sigmatch.nim
@@ -2271,7 +2271,7 @@ proc prepareOperand(c: PContext; formal: PType; a: PNode): PNode =
                 #else: {efDetermineType}
     result = c.semOperand(c, a, flags)
   else:
-    result = a
+    result = c.semOperand(c, a, {efDetermineType})
     considerGenSyms(c, result)
 
 proc prepareOperand(c: PContext; a: PNode): PNode =
--- a/compiler/sigmatch.nim
+++ b/compiler/sigmatch.nim
@@ -2271,7 +2271,10 @@ proc prepareOperand(c: PContext; formal: PType; a: PNode): PNode =
                 #else: {efDetermineType}
     result = c.semOperand(c, a, flags)
   else:
-    result = c.semOperand(c, a, {efDetermineType})
+    if a.kind in nkCallKinds and a.len > 0 and a[0].kind == nkSym and a[0].sym.magic == mNBindSym:
+      result = a
+    else:
+      result = c.semOperand(c, a, {efDetermineType})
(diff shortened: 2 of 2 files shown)


zah:
The CI tells you why it's wrong, but I can comment on the type system aspects of this code.

`Foo.x` should internally be typed with the special type `tyFieldAccessor`. Semantically, this type is somewhat similar to the function `proc (x: Foo): type(x)`, because you can imagine that the following should be possible:

```nim
type
  Foo = object
    x: int
    y: float

var fooSeq = @[Foo(x: 10, y: 1.0), Foo(x: 20, y: 2.0)]
var intSeq = map(fooSeq, Foo.x)
```

`tyFieldAccessor` is a strange type because these is no way to express it with regular Nim code and currently no way to create a function that accepts such a parameter. I guess there is nothing wrong if the macro system sees an `nkDotExpr` with two `nkSym` children (one for the type and one for the regular field), but the language has to decide whether the `tyFieldAccessor` type will be visible to the user code when `getType` is used. In some ways, Nim has already made the decision to hide the type, because the `typeof` operator or the `is` operator transparently skip `tyFieldAccessor`.

Right now, I consider the `getType` treatment broken. Consider the following simple program:

```nim
import
  macros

type
  Foo = object
    x: int

macro inspect(x: typed) =
  echo getType(x).treeRepr

var x: int

inspect x
inspect int
inspect Foo.x
```

The output is:

```
Sym "int"
BracketExpr
  Sym "typeDesc"
  Sym "int"
BracketExpr
  Sym "typeDesc"
  Sym "int"
```

In other words, `Foo.x` and `int` are considered to be expressions with the same type, namely `typedesc[int]` (aka the [int kind][1] - the type of the int type).

Arguably, the type of `Foo.x` should either be `int` or `FieldAccessor[int]`.

[1]: https://en.wikipedia.org/wiki/Kind_(type_theory)

Clyybber:
```nim
import macros

macro bar(i: int, n: typed): untyped =
  result = n
  echo i.intVal

macro foo(i: int, n: typed): untyped =
  result = copyNimTree n
  echo i.intVal
  
  result.body = newStmtList:
    newCall(bindSym"bar", newLit 2):
      result.body

macro addFoo(n: typed): untyped =
  result = copyNimTree n
  
  result.addPragma:
    nnkExprColonExpr.newTree(bindSym"foo"):
      newLit 1
  result.addPragma:
    nnkExprColonExpr.newTree(bindSym"bar"):
      newLit 3
      
proc test() {.foo(1), bar(3).} =
  discard

proc test1() {.addFoo.} =
  discard

#proc test2() {.bar(3), foo(1).} =
#  discard
```
demonstrates another consequence of I think the same issue as described here.
The expected output be `3 1 2 3 1 2`, but the actual output is `3 1 2 1 3 2` because `addFoo` returns AST that has it's `.typ` field already set, which causes `prepareOperand` to skip semming it again, which in turn causes `foo(1)` to be evaluated before `bar(3)`.
I think we may need to strip `.typ` from macro generated AST generally, instead of what this PR does.

PR #14536, opened 2020-06-01
manual: removed subjective phrase from 'macros' section about Nims syntax being "flexible enough anyway"

zevv:
Removed phrase "However, this is no real restriction because Nim's syntax is flexible enough anyway." from the manual - this is subjective, and I sometimes *do* find myself restricted by Nim's syntax when writing DSLs


zevv:
Sorry for the nitpicking, but this one always annoys me a bit. It's even on the nim-lang.org front page.

PR #14541, opened 2020-06-01
Add the ability to pass title case headers to an HTTP server

wonderix:
My [original PR](https://github.com/nim-lang/Nim/pull/13968) is working when the headers are passed in the constructor to `newHttpClient` 

```
client = newHttpClient(headers=newHttpHeader(...,titleCase=true))
```
but it is not working in case the headers are passed in each request. 

```
client.request(url,HttpGet,headers=newHttpHeader(...,titleCase=true))
```

This is caused by the fact that a new `HttpHeader`object is created in [`override`](https://github.com/nim-lang/Nim/blob/fbc97e712a67f30076f7633880383c8c4ae7866f/lib/pure/httpclient.nim#L963).

Therefore, I would like to revert my [original PR](https://github.com/nim-lang/Nim/pull/13968) and propose a new solution.

### DIFF
--- a/lib/pure/httpclient.nim
+++ b/lib/pure/httpclient.nim
@@ -474,8 +474,15 @@ proc getNewLocation(lastURL: string, headers: HttpHeaders): string =
     parsed.anchor = r.anchor
     result = $parsed
 
+proc toTitleCase(s: string): string =
+  result = newString(len(s))
+  var upper = true
+  for i in 0..len(s) - 1:
+    result[i] = if upper: toUpperAscii(s[i]) else: toLowerAscii(s[i])
+    upper = s[i] == '-'
+
 proc generateHeaders(requestUrl: Uri, httpMethod: string, headers: HttpHeaders,
-                     proxy: Proxy): string =
+                     proxy: Proxy, titleCaseHeaders: bool): string =
   # GET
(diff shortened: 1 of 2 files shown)


dom96:
Did your last implementation make it into a release? If so we will be breaking an API that was in a released version... I guess we can be pragmatic here, but ugh. Any chance you could fix the problem instead of reimplementing your original solution?

wonderix:
The last PR was merged after Release 1.2.0. Therefore, I suggested a new implementation. It would also be possible to fix the problem. But I'm not sure how to merge two `HttpHeaders` with different `titleCase` setting in [`override`](https://github.com/nim-lang/Nim/blob/fbc97e712a67f30076f7633880383c8c4ae7866f/lib/pure/httpclient.nim#L963)

wonderix:
A solution without breaking change might look like [this](https://github.com/wonderix/Nim/compare/devel...wonderix:title-case-headers-2)

Araq:
Since it happened after 1.2, let's break it.

dom96:
Sure, let's keep breaking things, even when the alternative is much better. Seriously, look at @wonderix's link, that's what you should PR instead. It makes sense to have this option as part of the HttpHeaders anyway.

PR #14565, opened 2020-06-05
Fix asynchttpserver newline breaking content-length

zedeus:
#13846 made it possible to manually specify a content-length, but the newline for the default header value wasn't made conditional. Without this change, the output includes an extra newline which is incorrect and breaks all clients I've tested (curl, chrome, firefox, etc.)

Backport?


dom96 (approved):
Looks good to me. And yeah, this should be backported. I'll leave @narimiran to merge and add the necessary tags.

PR #14597, opened 2020-06-07
openssl: use explicit result for SSL_in_init

### LINKED ISSUE #14591: compile issue on latest devel branch
/root/Nim/lib/wrappers/openssl.nim(421, 14) Error: expression 'theProc(ssl)' is of type 'cint' and has to be used (or discarded)

```
$ nim -v
Compiled at 2020-06-07

git hash: 51b71e35f2c0000def4f35d65ceb28ecdeb7ee10
```

alaviss:
Sometimes the analysis might get funky and decide that the if-expression
below is not an if-expression. Details of how this can happen is largely
unknown, but it's easy to prevent so we will just go forward with that.

Fix #14591


alaviss:
Unrelated CI failures, one due to nitter and one due to the flaky async test on bsd

PR #14632, opened 2020-06-10
porting name mangling from ic effort

disruptek:
You can see that this drags in the literal labeling and reconstitution crap.

I will add some better docs before review.

### DIFF
--- a/compiler/ccgexprs.nim
+++ b/compiler/ccgexprs.nim
@@ -1394,13 +1394,10 @@ proc rawConstExpr(p: BProc, n: PNode; d: var TLoc) =
     fillLoc(d, locData, n, name, OnStatic)
     # expression not found in the cache:
     p.module.s[cfsData].addf("static NIM_CONST $1 $2 = $3;$n",
-<<<<<<< HEAD
-          [getTypeDesc(p.module, t), d.r, genBracedInit(p, n, isConst = true, t)])
-=======
                              [getTypeDesc(p.module, t), name,
-                              genBracedInit(p, n, isConst = true)])
-  fillLoc(d, locData, n, name, OnStatic)
->>>>>>> e4d39c9cf... more literals
+                              genBracedInit(p, n, isConst = true, t)])
+  else:
+    fillLoc(d, locData, n, name, OnStatic)
 
 proc handleConstExpr(p: BProc, n: PNode, d: var TLoc): bool =
   if d.k == locNone and n.len > ord(n.kind == nkObjConstr) and n.isDeepConstExpr:
--- a/compiler/ccgexprs.nim
+++ b/compiler/ccgexprs.nim
@@ -1390,14 +1390,11 @@ proc rawConstExpr(p: BProc, n: PNode; d: var TLoc) =
   discard getTypeDesc(p.module, t) # so that any fields are initialized
   var name: Rope
   if getTempName(p.module, n, name):
-    # duplication, but we cannot know if it's unnecessary
-    fillLoc(d, locData, n, name, OnStatic)
     # expression not found in the cache:
     p.module.s[cfsData].addf("static NIM_CONST $1 $2 = $3;$n",
                              [getTypeDesc(p.module, t), name,
                               genBracedInit(p, n, isConst = true, t)])
-  else:
-    fillLoc(d, locData, n, name, OnStatic)
+  fillLoc(d, locData, n, name, OnStatic)
 
 proc handleConstExpr(p: BProc, n: PNode, d: var TLoc): bool =
   if d.k == locNone and n.len > ord(n.kind == nkObjConstr) and n.isDeepConstExpr:
--- a/compiler/ccgexprs.nim
+++ b/compiler/ccgexprs.nim
@@ -2646,7 +2646,7 @@ proc expr(p: BProc, n: PNode, d: var TLoc) =
       if {sfDispatcher, sfForward} * sym.flags != {}:
         # we cannot produce code for the dispatcher yet:
         fillProcLoc(p.module, n)
-        genProcPrototype(p.module, sym)
+        genProcPrototype(p, sym)
       else:
         genProc(p.module, sym)
       putLocIntoDest(p, d, sym.loc)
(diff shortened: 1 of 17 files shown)


treeform:
Why did the work on this stop? I would love to have better generated names so its easier to debug.

disruptek:
The work stopped because I ran out of money.  You are welcome to pick up the PR.

disruptek:
I think this is the list of remaining _known_ defects:
- do something for js; I think `idOrSig` pukes or something, I forget.  should be a line or two to fix.
- do something for HCR because I sorta clobbered it relatively carelessly.  at least I kept the comments.  in theory, we should have improved the situation for HCR but I have never tested it.
- fixup proc param type names changing (generics, proc ptrs?) which changes the symbol under C++ and causes link errors.  I don't know what the issue is exactly, but I have a few things I want to check...

disruptek:
`jsonutils` is the only test that fails for me.  I had to go back to signature hashes for proc type names (and, preemptively, tuple type names) because I could not come up with any other way to match argument types from foreign procs when the proc type is defined locally.  I spent a couple days trying to figure this out, so a solution here would be very much appreciated.

It's likely that `jsonutils` reveals a similar problem, but I haven't looked at the test yet.

disruptek:
I should say that one solution I considered would be to actually scan all procs and modules for types with matching signatures, but I feel that this is a pretty expensive approach...

juancarlospaco:
Nice to see the progress!, keep it up. :+1:

disruptek:
It seems that it's just Arraymancer that doesn't work, but I cannot seem to reproduce the problem on my Linux box, either via #head or the current v0.6.0 release.

Oh, and jsonutils still doesn't work, but who cares, amiright?

compiler/ccgexprs.nim
```diff
@@ -1389,13 +1383,15 @@ proc genNewSeqOfCap(p: BProc; e: PNode; d: var TLoc) =
 proc rawConstExpr(p: BProc, n: PNode; d: var TLoc) =
   let t = n.typ
   discard getTypeDesc(p.module, t) # so that any fields are initialized
-  let id = nodeTableTestOrSet(p.module.dataCache, n, p.module.labels)
-  fillLoc(d, locData, n, p.module.tmpBase & rope(id), OnStatic)
-  if id == p.module.labels:
+  var name: Rope
+  block:
+    if not getTempName(p.module, n, name):
+      break
```
Clyybber:
```suggestion
  if getTempName(p.module, n, name):
```

juancarlospaco:
new_manging

:sunglasses: :+1:

disruptek:
The current issue is that the type cache(s) don't discriminate between `TaintedString` and `string`, so during different phases of proc generation, we're resetting the cache with the wrong name.
  
This causes a proc like `add(seq[string])` to retain a name of `add_seqString`, which was serialized earlier, but subsequently lose the correct type name of the first parameter; the prototype will be `add_seqString(seqString)` but the later definition turns into `add_seqString(seqTaintedString)`, which causes a linking error on C++.

I'm going to backtrack and put a little effort into a `hashType` that is a bit more sensitive, once I figure out how this is even a thing... :wink:

disruptek:
@cooldome PTAL :smile:

compiler/ccgtypes.nim
```diff
@@ -700,9 +570,9 @@ proc getOpenArrayDesc(m: BModule, t: PType, check: var IntSet; kind: TSymKind):
       m.s[cfsTypes].addf("typedef struct {$n$2* Field0;$nNI Field1;$n} $1;$n",
                          [result, elemType])
 
-proc getTypeDescAux(m: BModule, origTyp: PType, check: var IntSet; kind: TSymKind): Rope =
+proc getTypeDescAux(p: ModuleOrProc, origTyp: PType, check: var IntSet; kind: TSymKind): Rope =
```
Araq: I don't like this ``ModuleOrProc`` type at all. Please get rid of it.

compiler/mangler.nim
```diff
@@ -0,0 +1,704 @@
...
+  irrelevantForNaming = irrelevantForBackend + {tyVar} - {tyRange}
+  # these types may prefix the types they wrap, eg. `seqStr` or `setInt`
+  abbrevTypeArg = {tySet, tySequence, tyTypeDesc, tyArray, tyRange}
+  # these types get unwrapped and discarded from mangling
+  unwrapTypeArg = irrelevantForNaming - {tyRange} + {tyRef, tyPtr,
+                                         tyUserTypeClass, tyUserTypeClassInst}
+  unIdentity = ItemId(module: -1, item: -1)
+
+type
+  ModuleOrProc* = BProc or BModule
```
Araq, line 60: This means all this stuff just got generic with two different instantations. For no good reason. Please remove this and keep the old structure, generating a C type declaration doesn't need a BProc context, so don't pretend that it does for your hacking conveniences.

disruptek: Fair enough, but I still need the typeclass for `BProc|BModule` shenanigans.  Would you rather it was explicit in each declaration, instead?

Araq: I didn't need this typeclass for 10 years. What do you need it for?

disruptek: I know, I didn't need it for 10 years, either.  Then I discovered this new mangling code and it came with functionality that is generic to `BProc` and `BModule`.  The way the author solved this is by creating a typeclass to allow the same mechanisms to update the `sigConflicts` structure that is duplicated in each of these types.

If you're happy with the backend from 10 years ago, we can go there.  But I'm personally annoyed that I have to play games because the proc signature is generated with the module instead of the proc, for example.  I would like to step away from the old code and embrace abstraction in the interests of evolution.  To me, it's worth a second copy of the proc in the nim compiler binary.

New code is new.  That's why we call it new.  It's due to the newness of the thing.

Araq: New code can be better code. But this is so hacky that it's unacceptable. We all want better name mangling but I look at this Nim code that you wrote far more often than I look at the produced C code so it's a bad trade-off.

disruptek: Okay, well, if you come up with a better solution, I'll take another crack at it.

PR #14639, opened 2020-06-11
httpcore: Add http code 308 Permanent Redirect

daniel-j:
https://tools.ietf.org/html/rfc7538

> This status code is similar to 301 (Moved Permanently) ([RFC7231], Section 6.4.2), except that it does not allow changing the request method from POST to GET.

https://evertpot.com/http/308-permanent-redirect

> The difference between 301 and 308 is that a client that sees a 308 redirect MUST do the exact same request on the target location. If the request was a POST and and had a body, then the client must do a POST request with a body on the new location.


Araq:
Thank you!

PR #14644, opened 2020-06-11
Add browsers.findInstalledBrowsers, browsers.openBrowsers

juancarlospaco:
- Add `browsers.findInstalledBrowsers` to find *all* installed web browsers, not just the default one.
- Add `browsers.openBrowsers` to open an `url` with one or more web browsers with arbitrary `params`.
- `since`, changelog, examples, documentation, etc.

You can open a page you are coding locally with Firefox and Chrome, with DevTools open, all at once.
:)

### DIFF
--- a/lib/pure/browsers.nim
+++ b/lib/pure/browsers.nim
@@ -126,21 +126,19 @@ proc findInstalledBrowsers*(limit = 3.Positive, followSymlinks = true): set[WebB
       inc i
       incl result, x
 
-proc openBrowsers*(url: string, browsers: openArray[tuple[browser: WebBrowsers, params: string]]) {.since: (1, 3).} =
-  ## Open an `url` with one or more web browsers with arbitrary `params`.
+proc openBrowser*(url: string, browser: WebBrowsers, params: seq[string]) {.since: (1, 3).} =
+  ## Open an `url` with given web browser with arbitrary `params`.
   ## This does not block. The URL must not be empty string.
-  ## This proc does not raise an exception on error, beware.
   ## To check for installed web browsers see `findInstalledBrowsers()`.
   ##
   ## .. code-block:: nim
-  ##   openBrowsers("https://nim-lang.org", [(wbFirefox, "-new-tab"), (wbChromium, "--new-window")])
-  ##   openBrowsers("http://localhost/karax-spa/index.html", [(wbFirefox, "--devtools")])
+  ##   openBrowser("https://nim-lang.org", wbFirefox, "-new-tab")
+  ##   openBrowser("https://nim-lang.org", wbChromium, "--new-window")
+  ##   openBrowser("http://localhost/karax-spa/index.html", wbFirefox, "--devtools")
   doAssert url.len > 0, "URL must not be empty string"
-  assert browsers.len > 0
-  for x in browsers:
-    when defined(windows):
-      var o = newWideCString($x.browser & " " & x.params)
-      var u = newWideCString(url)
-      discard shellExecuteW(0'i32, o, u, nil, nil, SW_SHOWNORMAL)
-    else:
-      discard execShellCmd($x.browser & " " & quoteShell(x.params) & " " & quoteShell(url))
+  when defined(windows):
+    var o = newWideCString($browser & " " & params.join(" "))
+    var u = newWideCString(url)
+    discard shellExecuteW(0'i32, o, u, nil, nil, SW_SHOWNORMAL)
+  else:
+    discard execShellCmd($browser & " " & quoteShell(params.join(" ")) & " " & quoteShell(url))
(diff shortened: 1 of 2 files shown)


lib/pure/browsers.nim
```diff
@@ -79,3 +111,36 @@ proc openDefaultBrowser*() {.since: (1, 1).} =
   ## .. code-block:: nim
   ##   block: openDefaultBrowser()
   openDefaultBrowserImpl("http:about:blank")  # See IETF RFC-6694 Section 3.
+
+proc findInstalledBrowsers*(limit = 3.Positive, followSymlinks = true): set[WebBrowsers] {.inline, since: (1, 3).} =
+  ## Try to find *all* installed web browsers, not just the default one.
```
alaviss:
```suggestion
  ## Try to find *all* installed web browsers within `PATH`, not just the default one.
```

It's better to clarify this, since Windows and certain package managers (ie. `flatpak`) do not export their browsers to `PATH`. Also this may not work for macOS, which IIRC doesn't add installed applications to `PATH`.

So effectively this proc will only works on Linux...

lib/pure/browsers.nim
```diff
@@ -79,3 +111,36 @@ proc openDefaultBrowser*() {.since: (1, 1).} =
...
+  for x in WebBrowsers:
+    if i > limit: break
+    if findExe($x, followSymlinks).len > 0:
+      inc i
+      incl result, x
+
+proc openBrowsers*(url: string, browsers: openArray[tuple[browser: WebBrowsers, params: string]]) {.since: (1, 3).} =
+  ## Open an `url` with one or more web browsers with arbitrary `params`.
+  ## This does not block. The URL must not be empty string.
+  ## This proc does not raise an exception on error, beware.
```
alaviss: Why wouldn't this raise exceptions for when the browser couldn't be found?

I'm also finding this to be a bit weird to allow opening multiple instances of a web page in different browsers. Wouldn't an `openBrowser` proc that just open exactly one browser works well enough? If users really want to open a page in multiple browsers, they can just do a for-loop over an array, there's no need for fancy things like this.

Actually, wouldn't it be simpler if they just `import osproc` and uses `execShellCmd()`? I don't quite see the point of this proc.

lib/pure/browsers.nim
juancarlospaco: I agree, if you know how to improve it, please send your Pull Request to `os.findExe()`.
:)

lib/pure/browsers.nim
juancarlospaco: The cost of a 1 loop `for` loop is not relevant compared to the time it takes to actually open up the browser with the rendered page,
so basically it works just Ok for 1 url and 1 browser with no options too.

alaviss: I'm not questioning the runtime cost of this proc. I'm questioning whether it's any useful.

This proc can be easily replicated with `execCmdShell()` and a for-loop over the browsers you want to open in, if you ever need such an use case. Not to mention that as with `findInstalledBrowsers`, effectively this proc can only work on Linux.

alaviss: Also, on Linux, from what I understand here, the browsers will be opened one-by-one, and it can only continues after the previously launched browser is closed, making it even less useful.

lib/pure/browsers.nim
```diff
@@ -13,14 +13,46 @@
...
+from os import findExe
 import strutils
 
 when defined(windows):
   import winlean
 else:
   import os, osproc
 
+since (1, 3):
+  type WebBrowsers* = enum          ## Web Browsers
```
kaushalmodi, line 25: I think that such hard coded list of browsers is not very future proof, so not very stdlib-apt. Wouldn't xdg-open or alike be more appropriate?

juancarlospaco: `openDefaultBrowser()` already uses `xdg-open`.
This is to open with browsers that are not the default one.

juancarlospaco: @kaushalmodi 
What Type do you suggest for Enumeration better than `enum` ?.
It can be a big `seq[string]` as default argument, but I dont think thats better neither.

This is not to have all browsers, just the most *well known* popular ones,
because browsers are open source you can theoretically create your own, 
so is impossible to known them all. 

[Looking at other langs it does not looks like a bad implementation in general.](https://github.com/python/cpython/tree/3.8/Lib/webbrowser.py)

kaushalmodi: IMO, I was thinking of an stdlib as something "solid". Here, it's relying on a lot of 3rd party projects. New open source projects get added and removed, and that would affect the quality of this stdlib. 

Also, why does this need to be an enum? Why not allow user to do `openBrowser("https://nim-lang.org", "firefox", "-new-tab")`?

The list of browsers looks over-specified. If a user is using a non-popular browser, they cannot use this proc. But if they have that non-popular browser named "foo" in their PATH, they can just do `openBrowser("https://nim-lang.org", "foo")`.

PR #14659, opened 2020-06-14
Fix constant definitions in lib/wrappers/iup.nim

Prince213:


Araq:
The old names need a deprecation period.

ringabout:
`iup` has been removed from devel.

Please contribute to this repo instead:
https://github.com/nim-lang/iup

PR #14662, opened 2020-06-14
Update newLit set[T] to work when set is empty.

solo989:


lib/core/macros.nim
```diff
@@ -793,6 +793,10 @@ proc newLit*[T](s: set[T]): NimNode {.compileTime.} =
   result = nnkCurly.newTree
   for x in s:
     result.add newLit(x)
+   if result.len == 0:
```
Clyybber:
```suggestion
  if result.len == 0:
```

PR #14681, opened 2020-06-16
Clarify the use of the backwards index operator (^N) in tut1

kaushalmodi:
For consistency:
- Do `[a .. ^b]` (notice spaces on both sides of `..`)
- Do `[c ..< d]` (notice spaces on both sides of `..<`)

Ref: https://github.com/nim-lang/Nim/issues/14671.

### DIFF
--- a/doc/tut1.rst
+++ b/doc/tut1.rst
@@ -1444,8 +1444,12 @@ string that is "useless" and replace it with "useful".
 ``b[11 .. ^2]`` is the portion "useless", and ``b[11 .. ^2] = "useful"`` replaces the
 "useless" portion with "useful", giving the result "Slices are useful."
 
-Note: alternate ways of writing this are ``b[^8 .. ^2] = "useful"`` or
+Note 1: alternate ways of writing this are ``b[^8 .. ^2] = "useful"`` or
 as ``b[11 .. b.len-2] = "useful"`` or as ``b[11 ..< b.len-1] = "useful"``.
+
+Note 2: As the ``^`` template returns a `distinct int <manual.html#types-distinct-type>`_
+of type ``BackwardsIndex``, we can have a ``lastIndex`` constant defined as ``const lastIndex = ^1``,
+and later used as ``b[0 .. lastIndex]``.
 
 Objects
 -------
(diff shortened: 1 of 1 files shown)


doc/tut1.rst
```diff
@@ -395,20 +395,29 @@ Since counting up occurs so often in programs, Nim also has a `..
...
   var s = "some string"
-  for i in 0..<s.len:
+  for i in 0 ..< s.len:
+    ...
+
+or
+
+.. code-block:: nim
+  var s = "some string"
+  for i in 0 .. ^1:
```
ghost: What is that supposed to do?

kaushalmodi: ugh.. a typo.. will fix it soon. Thanks.

juancarlospaco:
@kaushalmodi 
I think that would be nice to mention that one of the benefits of Nim is that it is a Type `ReverseIndex`,
and that it can be "prepared" at compile time like `const lastOne = ^1`.
:)

doc/tut1.rst
```diff
@@ -395,20 +395,29 @@ Since counting up occurs so often in programs, Nim also has a `..
 <system.html#...i,T,T>`_ iterator that does the same:
 
 .. code-block:: nim
-  for i in 1..10:
+  for i in 1 .. 10:
     ...
 
-Zero-indexed counting have two shortcuts ``..<`` and ``..^`` to simplify counting to one less than the higher index:
+Zero-indexed counting have two shortcuts ``..<`` and ``.. ^1``
```
Varriount: Should be "has two"

Varriount (changes requested):

kaushalmodi:
> 
> 
> @kaushalmodi
> I think that would be nice to mention that one of the benefits of Nim is that it is a Type `ReverseIndex`,
> and that it can be "prepared" at compile time like `const lastOne = ^1`.
> :)

Where will be a good place to mention that? It feels quite arbitrary to add a note about this specific type and not any other.

juancarlospaco:
@kaushalmodi 
Just a line with `Note:`  at bottom.
:)

kaushalmodi:
> 
> 
> @kaushalmodi
> Just a line with `Note:` at bottom.
> :)

Done in https://github.com/nim-lang/Nim/pull/14681/commits/c20786d9339772811badcb16be4108029e016216

Varriount (approved):

PR #14688, opened 2020-06-16
Update tuple newLit

solo989:
Supports unnamed tuples and empty tuples.

### DIFF
--- a/lib/core/macros.nim
+++ b/lib/core/macros.nim
@@ -805,7 +805,7 @@ proc newLit*[T: tuple](arg: T): NimNode {.compileTime.} =
   when isNamedTuple(T):
     result = nnkTupleConstr.newTree
     for a,b in arg.fieldPairs:
-      result.add nnkExprColonExpr.newTree(newIdentNode(a), newLit(b))
+      result.add nnkExprColonExpr.newTree(ident(a), newLit(b))
   else:
     result = nnkTupleConstr.newTree
     for b in arg.fields:
--- a/lib/core/macros.nim
+++ b/lib/core/macros.nim
@@ -801,7 +801,7 @@ proc newLit*[T](s: set[T]): NimNode {.compileTime.} =
 proc isNamedTuple(T: typedesc): bool {.magic: "TypeTrait".}
   ## Return true for named tuples, false for any other type.
 
-proc newLit*[T:tuple](arg: T): NimNode {.compileTime.} =
+proc newLit*[T: tuple](arg: T): NimNode {.compileTime.} =
   when isNamedTuple(T):
     result = nnkTupleConstr.newTree
     for a,b in arg.fieldPairs:
--- a/lib/core/macros.nim
+++ b/lib/core/macros.nim
@@ -758,7 +758,7 @@ proc newLit*(arg: enum): NimNode {.compileTime.} =
 proc newLit*[N,T](arg: array[N,T]): NimNode {.compileTime.}
 proc newLit*[T](arg: seq[T]): NimNode {.compileTime.}
 proc newLit*[T](s: set[T]): NimNode {.compileTime.}
-proc newLit*(arg: tuple): NimNode {.compileTime.}
+proc newLit*[T:tuple](arg: T): NimNode {.compileTime.}
 
 proc newLit*(arg: object): NimNode {.compileTime.} =
   result = nnkObjConstr.newTree(arg.type.getTypeInst[1])
@@ -801,8 +801,8 @@ proc newLit*[T](s: set[T]): NimNode {.compileTime.} =
 proc isNamedTuple(T: typedesc): bool {.magic: "TypeTrait".}
   ## Return true for named tuples, false for any other type.
 
-proc newLit*(arg: tuple): NimNode {.compileTime.} =
-  if isNamedTuple(typeof(arg)):
+proc newLit*[T:tuple](arg: T): NimNode {.compileTime.} =
+  if isNamedTuple(T):
     result = nnkTupleConstr.newTree
     for a,b in arg.fieldPairs:
       result.add nnkExprColonExpr.newTree(newIdentNode(a), newLit(b))
--- a/lib/core/macros.nim
+++ b/lib/core/macros.nim
@@ -758,7 +758,7 @@ proc newLit*(arg: enum): NimNode {.compileTime.} =
 proc newLit*[N,T](arg: array[N,T]): NimNode {.compileTime.}
 proc newLit*[T](arg: seq[T]): NimNode {.compileTime.}
 proc newLit*[T](s: set[T]): NimNode {.compileTime.}
-proc newLit*[T:tuple](arg: T): NimNode {.compileTime.}
+proc newLit*[T: tuple](arg: T): NimNode {.compileTime.}
 
 proc newLit*(arg: object): NimNode {.compileTime.} =
   result = nnkObjConstr.newTree(arg.type.getTypeInst[1])
--- a/lib/core/macros.nim
+++ b/lib/core/macros.nim
@@ -802,12 +802,11 @@ proc isNamedTuple(T: typedesc): bool {.magic: "TypeTrait".}
   ## Return true for named tuples, false for any other type.
 
 proc newLit*[T: tuple](arg: T): NimNode {.compileTime.} =
+  result = nnkTupleConstr.newTree
   when isNamedTuple(T):
-    result = nnkTupleConstr.newTree
     for a,b in arg.fieldPairs:
       result.add nnkExprColonExpr.newTree(ident(a), newLit(b))
   else:
-    result = nnkTupleConstr.newTree
     for b in arg.fields:
       result.add newLit(b)
 
--- a/lib/core/macros.nim
+++ b/lib/core/macros.nim
@@ -805,7 +805,7 @@ proc newLit*[T: tuple](arg: T): NimNode {.compileTime.} =
   result = nnkTupleConstr.newTree
   when isNamedTuple(T):
     for a,b in arg.fieldPairs:
-      result.add nnkExprColonExpr.newTree(newIdentNode(a), newLit(b))
+      result.add nnkExprColonExpr.newTree(ident(a), newLit(b))
   else:
     for b in arg.fields:
       result.add newLit(b)
--- a/lib/core/macros.nim
+++ b/lib/core/macros.nim
@@ -799,7 +799,7 @@ proc newLit*[T](s: set[T]): NimNode {.compileTime.} =
     result = newCall(typ,result)
 
 proc isNamedTuple(T: typedesc): bool {.magic: "TypeTrait".}
-  ## Return true for named tuples, false for any other type.
+  ## See typetraits.isNamedTuple
 
 proc newLit*[T: tuple](arg: T): NimNode {.compileTime.} =
   result = nnkTupleConstr.newTree
--- a/lib/core/macros.nim
+++ b/lib/core/macros.nim
@@ -798,10 +798,18 @@ proc newLit*[T](s: set[T]): NimNode {.compileTime.} =
     var typ = getTypeInst(typeof(s))[1]
     result = newCall(typ,result)
 
+proc isNamedTuple(T: typedesc): bool {.magic: "TypeTrait".}
+  ## Return true for named tuples, false for any other type.
+
 proc newLit*(arg: tuple): NimNode {.compileTime.} =
-  result = nnkPar.newTree
-  for a,b in arg.fieldPairs:
-    result.add nnkExprColonExpr.newTree(newIdentNode(a), newLit(b))
+  if isNamedTuple(typeof(arg)):
+    result = nnkTupleConstr.newTree
+    for a,b in arg.fieldPairs:
+      result.add nnkExprColonExpr.newTree(newIdentNode(a), newLit(b))
+  else:
+    result = nnkTupleConstr.newTree
+    for b in arg.fields:
+      result.add newLit(b)
 
 proc nestList*(op: NimNode; pack: NimNode): NimNode {.compileTime.} =
   ## Nests the list `pack` into a tree of call expressions:
--- a/lib/core/macros.nim
+++ b/lib/core/macros.nim
@@ -805,7 +805,7 @@ proc newLit*[T: tuple](arg: T): NimNode {.compileTime.} =
   result = nnkTupleConstr.newTree
   when isNamedTuple(T):
     for a,b in arg.fieldPairs:
-      result.add nnkExprColonExpr.newTree(ident(a), newLit(b))
+      result.add nnkExprColonExpr.newTree(newIdentNode(a), newLit(b))
   else:
     for b in arg.fields:
       result.add newLit(b)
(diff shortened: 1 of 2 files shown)


lib/core/macros.nim
```diff
@@ -758,7 +758,7 @@ proc newLit*(arg: enum): NimNode {.compileTime.} =
 proc newLit*[N,T](arg: array[N,T]): NimNode {.compileTime.}
 proc newLit*[T](arg: seq[T]): NimNode {.compileTime.}
 proc newLit*[T](s: set[T]): NimNode {.compileTime.}
-proc newLit*(arg: tuple): NimNode {.compileTime.}
+proc newLit*[T:tuple](arg: T): NimNode {.compileTime.}
```
ghost: nit: 
```suggestion
proc newLit*[T: tuple](arg: T): NimNode {.compileTime.}
```

lib/core/macros.nim
```diff
@@ -798,10 +798,18 @@ proc newLit*[T](s: set[T]): NimNode {.compileTime.} =
...
     result = newCall(typ,result)
 
-proc newLit*(arg: tuple): NimNode {.compileTime.} =
-  result = nnkPar.newTree
-  for a,b in arg.fieldPairs:
-    result.add nnkExprColonExpr.newTree(newIdentNode(a), newLit(b))
+proc isNamedTuple(T: typedesc): bool {.magic: "TypeTrait".}
+  ## Return true for named tuples, false for any other type.
+
+proc newLit*[T:tuple](arg: T): NimNode {.compileTime.} =
```
ghost: really sorry, but same here ;)

timotheecour:
can you add tests to tests/macros/tmacros_various.nim ?

tests/macros/tmacros_various.nim
```diff
@@ -194,3 +194,66 @@ static:
     doAssert(v == a)
 
   echo "macrocache ok"
+  
+block tupleNewLitTests:
+  echo ""
```
timotheecour: remove all echo's, doAssert is all that's needed

tests/macros/tmacros_various.nim
```diff
@@ -194,3 +194,66 @@ static:
...
+    x3 = t3()
+    x4 = t4()
+    x5 = t5()
+    x6 = t6()
+    x7 = t7()
+    x8 = t8()
+    x9 = t9()
+  doAssert(x0 == eT)
+  echo x0
+  doAssert(x1 == (5,))
```
timotheecour: optional nit: `doAssert(x1 == (5,))` => `doAssert x1 == (5,)` is more idiomatic and slightly easier to read

(plus elsewhere)

tests/macros/tmacros_various.nim
```diff
@@ -194,3 +194,66 @@ static:
...
+  echo x1
+  doAssert(x2 == (a:5))
+  echo x2
+  doAssert(x3 == (5,"5"))
```
timotheecour: * no need for intermediate `const`, all you need is to put it in a static block
* put the corresponding assert right below the definition, it's much easier to follow what's going on, eg:
```nim
macro t8() : untyped =
  result = newLit(@[(a:5,b:"5")])
static: doAssert t8 == @[(a:5, b:"5")]
```

tests/macros/tmacros_various.nim
```diff
@@ -194,3 +194,54 @@ static:
     doAssert(v == a)
 
   echo "macrocache ok"
+  
+block tupleNewLitTests:
+  echo "tupleNewLitTests"
+  type
+    EmptyT = tuple
```
timotheecour: I didn't know that was even legal... not sure whether it's a bug; @araq what do you think?

tests/macros/tmacros_various.nim
```diff
@@ -194,3 +194,54 @@ static:
...
+  proc emptyT() : EmptyT =
+    discard
+  macro t0() : untyped =
+    result = newLit(emptyT())
```
timotheecour: result = newLit(())

tests/macros/tmacros_various.nim
```diff
@@ -194,3 +194,41 @@ static:
...
+  doAssert t0 == emptyT()
+  macro t1() : untyped =
```
timotheecour: no space before `:`
`macro t1() : untyped =`
=>
`macro t1(): untyped =`


and I'm pretty sure other reviewers will require you to add space AFTER `:` and `,`:
```
doAssert t4 == (a:5,b:"5")
=>
doAssert t4 == (a: 5, b: "5")
```

lib/core/macros.nim
```diff
@@ -798,10 +798,18 @@ proc newLit*[T](s: set[T]): NimNode {.compileTime.} =
...
-proc newLit*(arg: tuple): NimNode {.compileTime.} =
-  result = nnkPar.newTree
-  for a,b in arg.fieldPairs:
-    result.add nnkExprColonExpr.newTree(newIdentNode(a), newLit(b))
+proc isNamedTuple(T: typedesc): bool {.magic: "TypeTrait".}
+  ## Return true for named tuples, false for any other type.
+
+proc newLit*[T: tuple](arg: T): NimNode {.compileTime.} =
+  when isNamedTuple(T):
+    result = nnkTupleConstr.newTree
```
timotheecour: this is in common for both branches; move it before `when`

lib/core/macros.nim
```diff
@@ -798,10 +798,18 @@ proc newLit*[T](s: set[T]): NimNode {.compileTime.} =
...
+    for a,b in arg.fieldPairs:
+      result.add nnkExprColonExpr.newTree(newIdentNode(a), newLit(b))
```
timotheecour, line 808: optional nit: docs recommend `ident(a)` instead of `newIdentNode(a)`

tests/macros/tmacros_various.nim
```diff
@@ -194,3 +194,41 @@ static:
...
+  
+block tupleNewLitTests:
+  echo "tupleNewLitTests"
+  type
+    EmptyT = tuple
+  proc emptyT(): EmptyT =
+    discard
+  macro t0(): untyped =
+    result = newLit(())
+  doAssert t0 == emptyT()
```
timotheecour: i'd just use `()` everywhere and remove `emptyT` altogether

timotheecour:
@solo989 thanks for your patience, almost there :-)

tests/macros/tmacros_various.nim
```diff
@@ -194,3 +194,41 @@ static:
     doAssert(v == a)
 
   echo "macrocache ok"
+  
+block tupleNewLitTests:
+  echo "tupleNewLitTests"
```
timotheecour: @solo989 the tests fail because you're adding `echo` and you haven't updated the `nimout` section on top; I recommend simply removing the `echo` statements you've added; it serves no purpose (doAssert would show the context on error)

solo989:
```nim
proc emptyT(): () =
    discard
```
didn't work so I did
```nim
type
    EmptyT = tuple
```
instead
though
```nim
proc emptyT(): tuple =
    discard
```
works
which I suppose is consistent
with
```nim
type
  EmptyT = tuple
```
working
would you prefer that?

tests/macros/tmacros_various.nim
ghost: @timotheecour an empty object is allowed too

solo989:
I changed ident back to newIdentNode since all the other macro procs seem to use that.
Also I'm getting mysterious check failures with no errors and no warnings associated with them.
So maybe this will fix that.

solo989:
I dropped emptyT.

tests/macros/tmacros_various.nim
timotheecour: nevermind i was thinking of the `tuple[]` syntax but, i guess it's legal as this syntax w no fields:
```nim
type Foo = tuple
  f1: int
```

tests/macros/tmacros_various.nim
timotheecour: resolved; (@solo989 you can click on "resolve conversation"  when resolved)

solo989:
newIdentNode is defined as
```nim
proc newIdentNode*(i: string): NimNode {.magic: "StrToIdent", noSideEffect, compilerproc.}
```
while ident is defined as
```nim
proc ident*(name: string): NimNode {.magic: "StrToIdent", noSideEffect.}
```
I don't know if the addition of the compilerproc pragma makes a difference.

tests/macros/tmacros_various.nim
timotheecour: yay! last commit resolved this

timotheecour:
@solo989 yes, my bad regarding `proc emptyT(): tuple = discard`, it is correct indeed (but, as I said, was not needed, and you've removed it so all good)
note that these would've also worked:
```nim
proc emptyT(): tuple[] = ()
proc emptyT(): auto = ()
```
(but again, not needed)

> I changed ident back to newIdentNode since all the other macro procs seem to use that.

that's fine

> Also I'm getting mysterious check failures with no errors and no warnings associated with them.

hmm which ones?

> I don't know if the addition of the compilerproc pragma makes a difference.

this seems like a bug in `newIdentNode`, https://nim-lang.github.io/Nim/intern.html
> A magic proc is not the same as a compilerproc: A magic is a proc that needs compiler magic for its semantic checking, a compilerproc is a proc that is used by the code generator.

I'll send a PR to fix `newIdentNode` (EDIT: https://github.com/nim-lang/Nim/pull/14692)


the only thing left is to check whether CI failures are related to this PR or not

solo989:
I think I was getting mysterious failures with one of the azure pipeline tests but maybe those never give good error messages.

lib/core/macros.nim
```diff
@@ -798,10 +798,17 @@ proc newLit*[T](s: set[T]): NimNode {.compileTime.} =
     var typ = getTypeInst(typeof(s))[1]
     result = newCall(typ,result)
 
-proc newLit*(arg: tuple): NimNode {.compileTime.} =
-  result = nnkPar.newTree
-  for a,b in arg.fieldPairs:
-    result.add nnkExprColonExpr.newTree(newIdentNode(a), newLit(b))
+proc isNamedTuple(T: typedesc): bool {.magic: "TypeTrait".}
+  ## Return true for named tuples, false for any other type.
```
timotheecour: nit: i'd rather have a comment like: `see typetraits.isNamedTuple` to keep things DRY (in case doc comment changes in typetraits.isNamedTuple)

timotheecour (approved):

solo989:
I'll change it back to ident instead of newIdentNode after I see if these checks go through.

timotheecour:
> I'll change it back to ident instead of newIdentNode after I see if these checks go through.

@solo989 I don't think that's the problem, here I can reproduce locally as follows:

```
git clone https://github.com/Vindaar/ggplotnim
cd ggplotnim
NIMBLE_DIR=$HOME/.nimble_fake29 nimble develop
NIMBLE_DIR=$HOME/.nimble_fake29 nim c -d:noCairo -r -d:nimWorkaround14447 tests/tests.nim
before PR: works
after PR:
/Users/timothee/git_clone/nim/temp/ggplotnim/src/ggplotnim/dataframe/arraymancer_backend.nim(1105, 16) Error: cannot get child of node kind: nnkSym
```

the next question is what to do about this :)

solo989:
Someone could open another pr after this one to change all those newIdentNode calls to ident calls and see if anything breaks.

timotheecour:
> Someone could open another pr after this one to change all those newIdentNode calls to ident calls and see if anything breaks.

that's not the problem I believe, see https://github.com/nim-lang/Nim/pull/14688#issuecomment-645100429
what i suspect is that the bug you've fixed here causes some packages to fail; and the question is what to do about it; there are a few options:
* let them break
* allow old behavior via `when defined(nimHasWorkaround14688)` so other packages can migrate smoothly (probably best option)

EDIT: indeed, in your patch, if I change:
```nim
when isNamedTuple(T):
=>
when defined(nimHasWorkaround14688) or isNamedTuple(T):
```
then this works:
```
NIMBLE_DIR=$HOME/.nimble_fake29 $nim_bisect_D/bin/nim.detached.6597beb75 c -d:noCairo -r -d:nimWorkaround14447 -d:nimHasWorkaround14688 tests/tests.nim
```


so you can also change this in `testament/important_packages.nim`:
```
pkg1 "ggplotnim", true, "nim c -d:noCairo -r -d:nimWorkaround14447 tests/tests.nim"
=>
pkg1 "ggplotnim", true, "nim c -d:noCairo -r -d:nimWorkaround14447 -d:nimHasWorkaround14688 tests/tests.nim"
```

(likewise if other packages fail)
yes, this is the 2nd time a workaround flag is used for ggplotnim...

please also send an issue to https://github.com/Vindaar/ggplotnim so that @Vindaar can track this

solo989:
@timotheecour
While I have your attention I was wondering if
```nim
proc `[]`*(n: NimNode, i: int): NimNode {.magic: "NChild", noSideEffect.}
```
being unable to return a var NimNode is a fundamental restriction of the language.
It would be very convenient to pass it to
```nim
proc t(a : var NimNode)
```
without having to do
```nim
var x = a[i]
t(x)
x = a[i]
```

solo989:
Actually in some of my own projects I've wrapped NimNodes in ref object wrappers
to get around this and add other features though ultimately I decided it was too much of a hassle recreating all the functionality of a NimNode on another object.

timotheecour:
> While I have your attention I was wondering if

let's discuss this in https://github.com/nim-lang/Nim/issues/14670 though

solo989:
Ok these were the only two to fail.
Packages CI / ubuntu-18.04 (pkg: 1) (pull_request)
Packages CI / macos-10.15 (pkg: 1) (pull_request)
I will switch newIdentNode with ident and see if the results are different.

timotheecour:
> I will switch newIdentNode with ident and see if the results are different.

you don't have to try and guess on CI, you can test locally as I showed here https://github.com/nim-lang/Nim/pull/14688#issuecomment-645102392
and as I wrote there, I don't think it's related to newIdentNode vs ident; see https://github.com/nim-lang/Nim/pull/14688#issuecomment-645102392

solo989:
nim-lang.Nim (packages OSX_amd64)
failing
nim-lang.Nim (packages OSX_amd64_cpp)
failing
thats new
I think I may have stumbled on a worse bug.
The only change I made was to replace newIdentNode with ident.

solo989:
The problem is the only overload that newLit can find of ident is
```nim
proc ident(n: NimNode): NimIdent
```
The magic 
```nim
proc ident*(name: string): NimNode {.magic: "StrToIdent", noSideEffect.}
```
isn't being properly picked up while inside macros.nim

I should have tested this version earlier on my local copy.

solo989:
I'm switching it back to newIdentNode, this seems to be an entirely seperate issue.

solo989:
I figured it out : ident is defined near the bottom of macros.nim
Moving ident to the top of macros.nim can be a seperate PR.

timotheecour:
> I figured it out : ident is defined near the bottom of macros.nim

this is unrelated; I've explained here the issue + solution:
https://github.com/nim-lang/Nim/pull/14688#issuecomment-645102392

solo989:
It's unrelated to those two particular failures but when I changed newIdentNode to ident everything failed. Not just those two. Regardless I did a seperate pr to put ident at the top of macros.nim so calls to ident(string) should work for any future prs. It has already been merged.

Those two failures have to do with using newLit to generate tuples and then trying to extract the values from the resulting NimNodes without checking to see if they were named tuples or not. At least in Vindaars case.

Issue #13935, opened 2020-04-09
SIGSEV encountered when creating threads in a loop w/ --gc:arc


zacharycarter:
SIGSEV encountered when creating threads in a loop w/ `--gc:arc`.

### Example
```nim
import osproc

var threads: seq[Thread[int]]

proc threadFn(a: int) {.thread.} =
  echo a

let numThreads = countProcessors() - 1

threads.setLen(numThreads)

for i in 0 ..< numThreads:
  createThread(threads[i], threadFn, i)
joinThreads(threads)
```

### Current output
```
➜  junkers git:(master) ✗ nim c --gc:arc --threads:on --debugger:native -r thread_error.nim
Hint: used config file '/Users/zacharycarter/.choosenim/toolchains/nim-#devel/config/nim.cfg' [Conf]
Hint: system [Processing]
Hint: repr_v2 [Processing]
Hint: widestrs [Processing]
Hint: io [Processing]
Hint: thread_error [Processing]
Hint: osproc [Processing]
Hint: strutils [Processing]
Hint: parseutils [Processing]
Hint: math [Processing]
Hint: bitops [Processing]
Hint: macros [Processing]
Hint: algorithm [Processing]
Hint: unicode [Processing]
Hint: os [Processing]
Hint: pathnorm [Processing]
Hint: osseps [Processing]
Hint: posix [Processing]
Hint: times [Processing]
Hint: options [Processing]
Hint: typetraits [Processing]
Hint: strtabs [Processing]
Hint: hashes [Processing]
Hint: streams [Processing]
Hint: cpuinfo [Processing]
Hint: kqueue [Processing]
CC: stdlib_system.nim
CC: thread_error.nim
Hint:  [Link]
Hint: dsymutil /Users/zacharycarter/dev/junkers/thread_error [Exec]
Hint: 39581 LOC; 0.993 sec; 52.48MiB peakmem; Debug build; proj: /Users/zacharycarter/dev/junkers/thread_error.nim; out: /Users/zacharycarter/dev/junkers/thread_error [SuccessX]
Hint: /Users/zacharycarter/dev/junkers/thread_error  [Exec]
0
3
2
No stack traceback available
SIGSEGV: Illegal storage access. (Attempt to read from nil?)
1
5
No stack traceback available
SIGSEGV: Illegal storage access. (Attempt to read from nil?)
4
```

### Additional Information

https://gist.github.com/zacharycarter/dbee66c50a6c26e7f187512235c4ed1f

```
[...]

PR #14690, opened 2020-06-17
Add an ARC test with threads in a loop

ghost:
Just to test where it works and where it fails. Probably don't merge unless it works on all platforms :P
See https://github.com/nim-lang/Nim/issues/13935

### DIFF
--- a/tests/arc/tthread_loop.nim
+++ b/tests/arc/tthread_loop.nim
@@ -1,0 +1,17 @@
+discard """
+  output: "test ok"
+  cmd: "nim c --gc:arc --threads:on $file"
+"""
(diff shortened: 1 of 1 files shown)


ghost:
Stacktrace is off because otherwise it was hanging locally for me, with -d:danger it SIGSEGVs instead.

ghost:
So it fails for Linux (both i386 and amd64), FreeBSD, works on Windows, macOS and OpenBSD. Weird stuff.

Araq:
Problem is that valgrind does complain but points me into the guts of the OS, so it's not very helpful.

PR #14700, opened 2020-06-17
make nim-gdb compatible with BSD systems

BarrOff:
Hello,

I made some small changes to the nim-gdb bash script:

- unified test if readlink executable is available
- also check if os is some *BSD
- MacOS uses the bsd readlink version, where the `-f` flag is equal to the coreutils readlink `-e` flag

I have only been able to test this on FreeBSD, but it should work the same on all the other *BSDs.

### DIFF
--- a/bin/nim-gdb
+++ b/bin/nim-gdb
@@ -5,17 +5,15 @@ set -e
 
 which nim > /dev/null || (echo "nim not in PATH"; exit 1)
 which gdb > /dev/null || (echo "gdb not in PATH"; exit 1)
+which readlink > /dev/null || (echo "readlink not in PATH."; exit 1)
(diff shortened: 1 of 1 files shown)


euantorano:
OpenBSD failure is unrelated to this PR, it looks like builds.sr.ht had a bug - it was mentioned in IRC earlier today for another project using FreeBSD and ddevault responded:

> 00:22:23 this error is a fluke
> 00:22:47 it happens intermittedly
> 00:22:52 race condition I haven't gotten around to fixing
> 00:22:58 if you resubmit it'll likely work

The build should be able to be resubmitted in the builds.sr.ht UI or by pushing an empty commit.

PR #14722, opened 2020-06-18
Fix allocating 0-sized blocks

k0zmo:
This fixes a nasty bug with `--gc:arc`. With other `--gc`'s when you request 0-sized block using `allocShared` an allocator adds also a `FreeCell` object for bookkeeping which makes `rawAlloc` function to never receive a `requestedSize == 0`. 

ARC does not add such overhead and calls `rawAlloc` with whatever it gets from `allocShared`. This makes it possible for `requestedSize` to be 0. If so, it allocates a block from OS and adds it to the free list. However. because the `requestedSize` is 0, the next allocation request will return the same address as before using the chunk of size=0 from the free list. In the end when we need to deallocate we end up with what is basically an double free operation.

Also, fixes crash in #14690 when on thread creation we allocate a `sizeof(GcThread)`-sized block which is 0.

### DIFF
--- a/lib/system/alloc.nim
+++ b/lib/system/alloc.nim
@@ -749,7 +749,7 @@ proc rawAlloc(a: var MemRegion, requestedSize: int): pointer =
   sysAssert(allocInv(a), "rawAlloc: begin")
   sysAssert(roundup(65, 8) == 72, "rawAlloc: roundup broken")
   sysAssert(requestedSize >= sizeof(FreeCell), "rawAlloc: requested size too small")
-  var size = roundup(requestedSize, MemAlign)
+  var size = roundup(max(1, requestedSize), MemAlign)
   sysAssert(size >= requestedSize, "insufficient allocated size!")
   #c_fprintf(stdout, "alloc; size: %ld; %ld\n", requestedSize, size)
   if size <= SmallChunkSize-smallChunkOverhead():
--- a/tests/system/talloc3.nim
+++ b/tests/system/talloc3.nim
@@ -1,0 +1,10 @@
+discard """
+   cmd: "nim c --gc:arc --stacktrace:off $file"
+"""
+
+var x = allocShared0(0)
+var y = allocShared0(0)
+doAssert x != y
+
+x.deallocShared()
+y.deallocShared()


ghost:
Can confirm that `--gc:arc` with threads now works correctly on Linux:
```
==32375== Memcheck, a memory error detector
==32375== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==32375== Using Valgrind-3.16.0.GIT and LibVEX; rerun with -h for copyright info
==32375== Command: ./tt
==32375== 
test ok
==32375== 
==32375== HEAP SUMMARY:
==32375==     in use at exit: 0 bytes in 0 blocks
==32375==   total heap usage: 11 allocs, 11 frees, 2,464 bytes allocated
==32375== 
==32375== All heap blocks were freed -- no leaks are possible
==32375== 
==32375== For lists of detected and suppressed errors, rerun with: -s
==32375== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

ghost:
Wonder why this worked for Windows before though :P Because of different system allocators?

Araq:
With ``-d:useMalloc`` we don't use our allocator though. The fix is great but looks incomplete. :-)

k0zmo:
Windows was also affected but it didn't show the symptoms. The faulty code is the same on both platforms. It didn't crash in your test probably because of different threading timings. To get a crash another thread needs to start before the previous one is finished. In other words there must be a call to `allocShared` in `createThread` for next thread before previous thread's `threadProcWrapper` finishes:

No crash:
```nim
var x = allocShared(0)
deallocShared(x)
var y = allocShared(0)
deallocShared(y)
```

Crash:
```nim
var x = allocShared(0)
var y = allocShared(0)
deallocShared(x) # This invalidates y
deallocShared(y)
```

@Araq What do you have in mind?

ghost:
@Araq seems like with `--gc:arc -d:useMalloc` it already works (simple examples from here, the threads test, the channels test). 

I searched a bit and it seems like most compilers would actually allocate at least 1 byte so it can be freed later.
From https://pubs.opengroup.org/onlinepubs/009695399/functions/malloc.html:
> If the size of the space requested is 0, the behavior is implementation-defined: the value returned shall be either a null pointer or a unique pointer.

And even if `malloc(0)` returns `NULL`, `free(NULL)` will work just fine.
https://pubs.opengroup.org/onlinepubs/009695399/functions/free.html
> If ptr is a null pointer, no action shall occur.

ghost:
For example when running this code by @k0zmo:
```nim
var x = allocShared(0)
var y = allocShared(0)
deallocShared(x) # This invalidates y
deallocShared(y)
```
with `-d:useMalloc --gc:arc -d:danger` with valgrind I get:
```
==20363== Memcheck, a memory error detector
==20363== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==20363== Using Valgrind-3.16.0.GIT and LibVEX; rerun with -h for copyright info
==20363== Command: ./tt
==20363== 
==20363== 
==20363== HEAP SUMMARY:
==20363==     in use at exit: 0 bytes in 0 blocks
==20363==   total heap usage: 2 allocs, 2 frees, 0 bytes allocated
==20363== 
==20363== All heap blocks were freed -- no leaks are possible
==20363== 
==20363== For lists of detected and suppressed errors, rerun with: -s
==20363== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```
So it actually allocates 0 bytes, but still works just fine

k0zmo:
I wouldn't equate valgrind's summary with him not allocating anything with `malloc(0)`. In both cases -  natively and under valgrind all `malloc(0)` returned pointers have unique addresses. This is also evident when you don't `free` these pointers:

```
==9667== HEAP SUMMARY:
==9667==     in use at exit: 0 bytes in 1,000 blocks
==9667==   total heap usage: 1,000 allocs, 0 frees, 0 bytes allocated
==9667==
==9667== 0 bytes in 1,000 blocks are definitely lost in loss record 1 of 1
==9667==    at 0x4C31B25: calloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==9667==    by 0x10A576: main__Fjz3xT8gB6ZowWk5Tic1HQ (in /mnt/d/Programowanie/Projects/nim/sandbox/alloc0)
==9667==    by 0x10A588: NimMainInner (in /mnt/d/Programowanie/Projects/nim/sandbox/alloc0)
==9667==    by 0x10891C: main (in /mnt/d/Programowanie/Projects/nim/sandbox/alloc0)
==9667==
==9667== LEAK SUMMARY:
==9667==    definitely lost: 0 bytes in 1,000 blocks
==9667==    indirectly lost: 0 bytes in 0 blocks
==9667==      possibly lost: 0 bytes in 0 blocks
==9667==    still reachable: 0 bytes in 0 blocks
==9667==         suppressed: 0 bytes in 0 blocks
```

Araq:
I fixed it differently, see https://github.com/nim-lang/Nim/pull/14751

k0zmo:
In terms of threads crashing under `--gc:arc` - yes, it's fixed now. 

However, I wouldn't say you fixed "allocating 0-sized blocks"-part. You'd still corrupt your heap with what even C considers a valid code. I don't see an `allocShared` contract, saying calling this function with 0 is undefined/invalid behavior. What's more - its argument is `Natural`, not `Positive`.

Araq:
It's argument should be ``Positive`` indeed. I don't care about C's broken behaviours.

PR #14733, opened 2020-06-19
Lower the not operators precedence

Clyybber:
This makes the `not` operator not bind more tightly than binary operators with a higher precedence.


Clyybber:
Probably a bad idea, while more inline with python we also use `not` as a binary operator and this change changes behaviour in `not a shr b`.

PR #14742, opened 2020-06-20
Enable fidget in important_packages.

treeform:
Fidget now works with `nimble test` and there is a CI: https://github.com/treeform/fidget/commits/master

### DIFF
--- a/.github/workflows/ci_packages.yml
+++ b/.github/workflows/ci_packages.yml
@@ -33,7 +33,7 @@ jobs:
           DEBIAN_FRONTEND='noninteractive' \
             sudo apt-fast install --no-install-recommends -yq \
               libcurl4-openssl-dev libgc-dev libsdl1.2-dev libsfml-dev \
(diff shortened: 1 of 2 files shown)


treeform:
Hmm fidget requires some stuff to be installed on linux ... mainly the xorg-dev stuff. Not sure how to specify it as part of important packages? 

It seems like the other ubuntu's pass. So maybe that particular 1 test needs stuff added?

This is what I use in my CI
`sudo apt install -y  build-essential libalut-dev libasound2-dev libc6-dev libgl1-mesa-dev libglu1-mesa-dev libxcursor-dev libxi-dev libxinerama-dev libxrandr-dev libxxf86vm-dev mesa-utils pkg-config xorg-dev xvfb`

alaviss:
You can add them here:

https://github.com/nim-lang/Nim/blob/1ffe1134501210e23e71e0f23099c2005edd98c9/.github/workflows/ci_packages.yml#L29-L36

PR #14749, opened 2020-06-21
Docs: Fix a broken `code-block`

ee7:
This PR fixes a broken `code-block` at the top of the [devel httpclient docs](https://nim-lang.github.io/Nim/httpclient.html). The bug was introduced by https://github.com/nim-lang/Nim/pull/14037 (it is not included in the [1.2.2 httpclient docs](https://nim-lang.org/docs/httpclient.html))

I did a quick search for other `code-block`s that are broken in the same way, but the only other one I found (in `pegs.nim`) is not included in the generated documentation.

It would be nice if we could detect such problems during CI.

### DIFF
--- a/lib/pure/httpclient.nim
+++ b/lib/pure/httpclient.nim
@@ -25,13 +25,13 @@
 ## ``AsyncHttpClient``:
 ##
 ## .. code-block:: Nim
-## import asyncdispatch, httpclient
+##   import asyncdispatch, httpclient
 ##
-## proc asyncProc(): Future[string] {.async.} =
-##   var client = newAsyncHttpClient()
-##   return await client.getContent("http://example.com")
+##   proc asyncProc(): Future[string] {.async.} =
+##     var client = newAsyncHttpClient()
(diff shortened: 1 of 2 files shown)


planetis-m:
This PR needs a tool to detect broken code blocks. It needs to be integrated to nim doc, so it is run during CI.

planetis-m:
I also suggest a SWOT analysis of why such tool would bee useful.

ee7:
In retrospect, this PR's title and the top line of the commit message is misleading.

My intent with this PR was just to:
- Fix the broken `code-block` at the top of the `httpclient.nim` docs.
- Say "I looked for others and fixed the only other occurrence I found, which wasn't important".

I don't intend to add a new check to CI; I don't have time right now to fully test such code. Somebody else is free to do add a check.

I can change the PR title, but I don't know if it's considered wasting CI resources if I force-push for a commit title change.

If this PR was deemed acceptable, the person merging could edit the commit title when squash-merging to be

```
Docs: Fix broken `code-block` (#14749)
```

or similar.

PR #14753, opened 2020-06-21
posix_other: add define to force time_t to 64 bit [backport]

alaviss:
This is a temporary remedy for time_t transition to 64 bit from 32 bit in 
newer Linux libc (musl >= 1.2.0, glibc >= 2.32).

In the future we might want to move away from libc, or figure out a way
to reliably detect the real size of C types at compile time, both of
which are difficult.

### DIFF
--- a/lib/posix/posix_other.nim
+++ b/lib/posix/posix_other.nim
@@ -32,7 +32,12 @@ type
   SocketHandle* = distinct cint # The type used to represent socket descriptors
 
 type
-  Time* {.importc: "time_t", header: "<time.h>".} = distinct clong
+  Time* {.importc: "time_t", header: "<time.h>".} = distinct (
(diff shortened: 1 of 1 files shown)


ghost:
Don't we have clonglong? I suppose it doesn't matter but I thought that using c-prefixed types for C interop was preferred

alaviss:
> Don't we have clonglong?

Yes but why use that when we can use something concrete?

ghost:
@alaviss well as I said I just don't understand why we use c-prefixed types at all then

alaviss:
We use them when we want to interact with code that make explicit use of them (ie. `int foo(long a);`). However `time_t` is considered opaque, all we know is that they resolve to an integer type, we just don't know which.

PR #14754, opened 2020-06-21
Add sugar.c

juancarlospaco:
- Add `sugar.c` convenience template for `cstring("some string")`, only for Literal `string`, still allows functions named `c(s: string)`.
- Documentation, `runnableExamples` with `doAssert` and `proc c()` to prove it wont match, `since`, changelog, etc.
- Meant for working with JavaScript or C or C++ interoperability with tons of `cstring("a"), cstring("b"), cstring("c"), ...`

### DIFF
--- a/lib/pure/sugar.nim
+++ b/lib/pure/sugar.nim
@@ -346,7 +346,8 @@ since (1, 3):
         doAssert c"Nim" == cstring("Nim")
         doAssert c" " == cstring(" ")
         doAssert c"" == cstring("")
-    system.cstring(s)
+    bind cstring
+    cstring(s)
 
 
 when isMainModule:
--- a/lib/pure/sugar.nim
+++ b/lib/pure/sugar.nim
@@ -335,6 +335,19 @@ macro collect*(init, body: untyped): untyped {.since: (1, 1).} =
       call.add init[i]
   result = newTree(nnkStmtListExpr, newVarStmt(res, call), resBody, res)
 
+template c*(s: string{nkRStrLit}): cstring {.since: (1, 3).} =
+  ## Convenience template for `cstring("some string")`, *only for Literal* `string`.
+  ## Wont match `"str".c` nor `c("str")`, just `c"str"`, to allow functions named `c()`.
+  ## Meant for working with JavaScript or C or C++ interoperability with lots of `cstring`.
+  runnableExamples:
+    static:
+      proc c(s: string) = doAssert false, "Must not match this"
+      doAssert c"Nim" == cstring("Nim")
+      doAssert c" " == cstring(" ")
+      doAssert c"" == cstring("")
+  system.cstring(s)
+
+
 when isMainModule:
   since (1, 1):
     import algorithm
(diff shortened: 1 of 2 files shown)


ghost:
I think `cstring` is the way it is because it shouldn't really be used in normal code, only in bindings as you said. But generally I'm neutral to the PR :)

juancarlospaco:
I agree, but when you have tons of `cstring("")` is too long for maybe few letter strings, other langs use `str` thats shorter,
thats why is designed to not match stuff named `c()` so you can still use templates or func named `c()` if any.
:)

lib/pure/sugar.nim
```diff
@@ -335,6 +335,20 @@ macro collect*(init, body: untyped): untyped {.since: (1, 1).} =
...
+    ## Convenience template for `cstring("some string")`, *only for Literal* `string`.
+    ## Wont match `"str".c` nor `c("str")`, just `c"str"`, to allow functions named `c()`.
+    ## Meant for working with JavaScript or C or C++ interoperability with lots of `cstring`.
+    runnableExamples:
+      static:
+        proc c(s: string) = doAssert false, "Must not match this"
+        doAssert c"Nim" == cstring("Nim")
+        doAssert c" " == cstring(" ")
+        doAssert c"" == cstring("")
+    system.cstring(s)
```
alaviss:
```suggestion
    bind cstring
    cstring(s)
```

Explicit `bind()` is usually better as the symbol will be resolved directly at definition site.

lib/pure/sugar.nim
```diff
@@ -335,6 +335,20 @@ macro collect*(init, body: untyped): untyped {.since: (1, 1).} =
       call.add init[i]
   result = newTree(nnkStmtListExpr, newVarStmt(res, call), resBody, res)
 
+since (1, 3):
+  template c*(s: string{nkRStrLit}): cstring =
+    ## Convenience template for `cstring("some string")`, *only for Literal* `string`.
+    ## Wont match `"str".c` nor `c("str")`, just `c"str"`, to allow functions named `c()`.
+    ## Meant for working with JavaScript or C or C++ interoperability with lots of `cstring`.
+    runnableExamples:
+      static:
```
alaviss: I don't think `static` is necessary here.

lib/pure/sugar.nim
```diff
@@ -335,6 +335,20 @@ macro collect*(init, body: untyped): untyped {.since: (1, 1).} =
       call.add init[i]
   result = newTree(nnkStmtListExpr, newVarStmt(res, call), resBody, res)
 
+since (1, 3):
```
alaviss, line 338: This should be applied as a pragma to the template instead.

ghost:
@juancarlospaco but if you have a lot of cstring in your project - you can just make this 1-line template yourself, no? Genuinely curious

alaviss:
I'm not sure whether this is working as expected. CI is complaining about `c` being redefined.

Araq:
Nah, sorry. Just use a template in your project, it's allowed.

lib/pure/sugar.nim
Araq: Alas, it currently cannot because we don't have custom pragmas on template declarations.

Issue #11795, opened 2019-07-20
docgen: broken sidebar links for generic procs with type constraints and Object[T, T] params


Jjp137:
If a generic proc has a type constraint, and one of the proc's parameters is a generic object with at least two type parameters (such as `Foo[T, T]`, and this includes `HSlice[T, T]`), then the generated sidebar link for that proc does not work.

### Example

```nim
## This is a module.
##
## Stuff
## -----
## This is filler meant to make the page long enough so that the problem
## is easy to see.
##
## More stuff
## ----------
## ...and more filler. If this page still fits within a single screen,
## make the browser window shorter or zoom in.

type Foo*[T; U] = object
  ## Example object.
  a*: T
  b*: U

proc working*[T](x: Foo[T, T]) =
  ## This example has a working sidebar link.
  discard

proc broken*[T: int](x: Foo[T, T]) =
  ## This is an example that results in a broken link. The link from the
  ## sidebar does not result in the browser jumping to this proc.
  discard

proc brokenTwo*[T: Ordinal or SomeFloat](x: Foo[T, T]) =
  ## This example, modeled after a similar proc in the random module, also
  ## results in a broken link. (The resulting link is also rather long.)
  discard
```

Generate the documentation with `nim doc`.

### Current Behavior

The sidebar link for the `working` proc works, but the sidebar links for the `broken` and `brokenTwo` procs do not work.

Also, within the documentation body, the name of each type and proc at the location it is defined is a link to itself, and usually those links work. While the `working` proc's link to itself is not broken, the `broken` and `brokenTwo` links do not work.

For reference, this is the generated HTML of the procs section within the sidebar:
```HTML
<li>
  <a class="reference reference-toplevel" href="#12" id="62">Procs</a>
  <ul class="simple simple-toc-section">
      <li><a class="reference" href="#working%2CFoo%5BT%2CT%5D"
    title="working[T](x: Foo[T, T])"><wbr />working<span class="attachedType">Foo</span></a></li>
  <li><a class="reference" href="#broken%2CFoo%5BT%3A+int%2CT%3A+int%5D"
    title="broken[T: int](x: Foo[T, T])"><wbr />broken<span class="attachedType">Foo</span></a></li>
  <li><a class="reference" href="#brokenTwo%2CFoo%5BT%3A+Ordinal+or+float+or+float32+or+float64%2CT%3A+Ordinal+or+float+or+float32+or+float64%5D"
[...]

PR #14755, opened 2020-06-21
Refactor random

planetis-m:
With this refactor we can get rid duplicate procs in this module. It adds ``Rand`` as last parameter, which defaults to ``state`` and deprecates procs that accepts ``r`` as first parameter.

### DIFF
--- a/lib/pure/random.nim
+++ b/lib/pure/random.nim
@@ -123,6 +123,7 @@ proc rotl(x, k: Ui): Ui =
 
 template randState*(): untyped =
   ## Makes the default Rand state accessible from other modules.
+  ## Useful for module authors.
   state
 
 proc next*(r: var Rand): uint64 =
@@ -219,7 +220,7 @@ proc rand*(max: Natural; r: var Rand = state): int {.benign.} =
   ## numbers returned from this proc will always be the same.
   ##
   ## See also:
-  ## * `rand proc<#rand,float,Rand>`_ that returns a float
+  ## * `rand proc<#rand,range[],Rand>`_ that returns a float
   ## * `rand proc<#rand,HSlice[T,T],Rand>`_ that accepts a slice
   ## * `rand proc<#rand,typedesc[T],Rand>`_ that accepts an integer or range type
   runnableExamples:
@@ -243,7 +244,7 @@ proc rand*(r: var Rand; max: int): int {.benign, deprecated.} =
   ## Returns a random integer in the range `0..max` using the given state.
   ##
   ## **Deprecated since version 1.3**:
-  ## Use the `rand proc<#rand,int,Rand>`_ instead.
+  ## Use the `rand proc<#rand,Natural,Rand>`_ instead.
   rand(max, r)
 
 proc rand*(max: range[0.0 .. high(float)]; r: var Rand = state): float {.benign.} =
@@ -255,18 +256,16 @@ proc rand*(max: range[0.0 .. high(float)]; r: var Rand = state): float {.benign.
   ## numbers returned from this proc will always be the same.
   ##
   ## See also:
-  ## * `rand proc<#rand,int,Rand>`_ that returns an integer
+  ## * `rand proc<#rand,Natural,Rand>`_ that returns an integer
   ## * `rand proc<#rand,HSlice[T,T],Rand>`_ that accepts a slice
   ## * `rand proc<#rand,typedesc[T],Rand>`_ that accepts an integer or range type
   runnableExamples:
     randomize(234)
     let f1 = rand(1.0)
-    # f1 = 8.717181376738381e-07
-
     # using the provided state
     var r = initRand(234)
     let f2 = rand(1.0, r)
-    # f2 = 8.717181376738381e-07
+    doAssert f1 == f2
   let x = next(r)
   when defined(js):
     result = (float(x) / float(high(uint32))) * max
@@ -279,7 +278,7 @@ proc rand*(r: var Rand; max: float): float {.benign, deprecated.} =
   ## using the given state.
   ##
   ## **Deprecated since version 1.3**:
-  ## Use the `rand proc<#rand,float,Rand>`_ instead.
+  ## Use the `rand proc<#rand,range[],Rand>`_ instead.
   rand(max, r)
 
 proc rand*[T: Ordinal or SomeFloat](x: HSlice[T, T]; r: var Rand = state): T =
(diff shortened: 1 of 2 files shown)


planetis-m:
Also if the tests pass I will export ``state``

lib/pure/random.nim
```diff
@@ -363,11 +356,36 @@ proc rand*[T: SomeInteger](t: typedesc[T]): T =
...
+  ## * `sample proc<#sample,openArray[T],openArray[U]>`_ that uses a
+  ##   cumulative distribution function
+  runnableExamples:
+    randomize(987)
+    let s = {1, 3, 5, 7, 9}
+    doAssert sample(s) == 5
+    doAssert sample(s) == 7
+    doAssert sample(s) == 1
+  assert card(s) != 0
+  var i = rand(r, card(s) - 1)
```
planetis-m: missed one

lib/pure/random.nim
```diff
@@ -425,31 +438,45 @@ proc sample*[T](r: var Rand; a: openArray[T]): T =
...
+    let cdf = count.cumsummed
+    randomize(789)
+    doAssert sample(marbles, cdf) == "red"
+    doAssert sample(marbles, cdf) == "green"
+    doAssert sample(marbles, cdf) == "blue"
+  assert(cdf.len == a.len) # Two basic sanity checks.
+  assert(float(cdf[^1]) > 0.0)
+  #While we could check cdf[i-1] <= cdf[i] for i in 1..cdf.len, that could get
+  #awfully expensive even in debugging modes.
+  let u = r.rand(float(cdf[^1]))
```
planetis-m: and here

lib/pure/random.nim
```diff
@@ -477,44 +504,39 @@ proc sample*[T, U](r: var Rand; a: openArray[T]; cdf: openArray[U]): T =
...
-  ## * `sample proc<#sample,set[T]>`_ for sets
+  ## * `shuffle proc<#shuffle,Rand,openArray[T]>`_ that uses a provided state
   runnableExamples:
-    from math import cumsummed
+    var cards = ["Ace", "King", "Queen", "Jack", "Ten"]
+    randomize(678)
+    shuffle(cards)
+    doAssert cards == ["King", "Ace", "Queen", "Ten", "Jack"]
+  for i in countdown(x.high, 1):
+    let j = r.rand(i)
```
planetis-m: here too

lib/pure/random.nim
```diff
@@ -656,5 +645,7 @@ when isMainModule:
 
     # don't use causes integer overflow
     doAssert compiles(rand[int](low(int) .. high(int)))
+    var d = initRand(123)
```
planetis-m: delete

planetis-m:
Ready

lib/pure/random.nim
```diff
@@ -121,15 +121,19 @@ else:
...
+template randState*(): untyped =
+  ## Makes the default Rand state accessible from other modules.
+  state
+
 proc next*(r: var Rand): uint64 =
   ## Computes a random ``uint64`` number using the given state.
   ##
   ## See also:
-  ## * `rand proc<#rand,Rand,Natural>`_ that returns an integer between zero and
+  ## * `rand proc<#rand,Natural,Rand>`_ that returns an integer between zero and
```
planetis-m, line 133: Someone who knows, is this link correct?

ghost: FYI: don't do co skip for doc changes in the future :)

planetis-m: [ci skip] doesn't work anyway.

lib/pure/random.nim
```diff
@@ -121,15 +121,19 @@ else:
...
   ##   a given upper bound
-  ## * `rand proc<#rand,Rand,range[]>`_ that returns a float
-  ## * `rand proc<#rand,Rand,HSlice[T,T]>`_ that accepts a slice
-  ## * `rand proc<#rand,typedesc[T]>`_ that accepts an integer or range type
+  ## * `rand proc<#rand,range[],Rand>`_ that returns a float
+  ## * `rand proc<#rand,HSlice[T,T],Rand>`_ that accepts a slice
```
planetis-m, line 136: All links to slice don't work. need help.

Jjp137: It's probably related to #11795

planetis-m: Ok nothing i can do in this pr

Araq:
big -1 from me. The state accepting procs should not be deprecated, they were introduced to fix the old procs not taking the state. In the future we avoid the factor of 2 in the API surface by not introducing procs that do not take in the state explicitly.

planetis-m:
So which proc prototype is recommended? Is it still okay to add default state parameter in the `rand(1)`procs?

Araq:
It's ok to export the global state. Nothing more should be done.

planetis-m:
will start from clean in another Pr

c-blake:
Not sure of all your goals in your new PR, but in the past for unit normal deviates the [Ziggurat method](https://en.wikipedia.org/wiki/Ziggurat_algorithm) been fastest.  In general "batch/vector" APIs can often be convenient in this space.  Might be something to keep in mind and @mratsim may well want to comment.

mratsim:
@b3liever timed the Ziggurat vs the Polar and the difference was minutes: https://gist.github.com/b3liever/23505f9b47124bc9cc40e60d7279b067#gistcomment-3346707

In terms of API, I prefer having the Rand as first parameter because I often use the idiom:
```Nim
rng.random(MyType)
```
MyType being an elliptic curve, BigInt, Modular Arithmetic integer: https://github.com/mratsim/constantine/blob/0400187f/helpers/prng_unsafe.nim#L313-L357
or Vec3 or Colors: https://github.com/mratsim/trace-of-radiance/blob/99f7d85d/trace_of_radiance/sampling.nim#L30-L86

That said `random(MyType, rng)` is possible as well.

c-blake:
@mratsim - that link did not include the Ziggurat method (which usually has three 256 entry pre-computed tables).  I am not sure the size of test leading to your "minutes", but I agree it is much faster.  To better motivate things in a tight/batch loop I have found it to be approximately 4-6x faster than competing alternatives (for 2009 Nehalem).  In a non-batch context, the 3 KiB of table space can cause more variation from L1 cache competition.

c-blake:
It seems @krux02 has a Nim impl in https://github.com/krux02/opengl-sandbox, specifically under `fancygl/ziggurat_normal_dist.nim`, although that impl does not "show off" Nim's compile-time compute ability very nicely (the tables should be computed as some `const` section!)

The charm of the method is that 98.5% of the time a sample reduces to just int uniform RNG + an int mask, int compare, and 1 float product.  For the rare case you do something a bit more expensive like a typical Normal deviate generator.  I just timed my old C code on an AMD2950x and saw it being over 2.5x faster than a Box-Muller algo { one careful to leverage that you get a pair of normal deviates per inner calculation..Would be over 5x without that leveraging }.   Anyway, I think it's faster-enough to be worth learning/doing. Of course, it's rare that the time to generate numbers is a big fraction of whatever else one is doing.

PR #14782, opened 2020-06-23
readLine: Unicode support for Windows console

ktamp:
When input is read from the Windows console, input encoding is UTF16. This is translated internally to UTF8.

### DIFF
--- a/lib/system/io.nim
+++ b/lib/system/io.nim
@@ -14,6 +14,11 @@ include inclrtl
 import std/private/since
 import formatfloat
 
+when defined(windows) and not defined(useWinAnsi):
+  import os
+  import system/widestrs
+  import terminal
+
 # ----------------- IO Part ------------------------------------------------
 type
   CFile {.importc: "FILE", header: "<stdio.h>",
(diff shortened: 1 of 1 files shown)


lib/system/io.nim
```diff
@@ -14,6 +14,11 @@ include inclrtl
 import std/private/since
 import formatfloat
 
+when defined(windows) and not defined(useWinAnsi):
+  import os
+  import system/widestrs
+  import terminal
```
alaviss: This is bad, as all modules have `system/io` implicitly imported. This will cause recursive imports, which we don't support (but we should) atm.

ktamp: We can work around **os** by calling **GetLastError**.
We can also work around **terminal** by accessing **File** struct contents (although this may not be pretty).
But working around **widestrs** is another matter...

This is what makes code work across compilers.
GCC has this functionality built-in, but what about other compilers?

alaviss: IIRC console have special HANDLE values that we can just compare to instead.

As for widestrs, you might not even have to import it.

alaviss: https://github.com/nim-lang/Nim/blob/devel/lib/pure/terminal.nim#L669-L672

`isatty()` is simple enough to just duplicate into `io` for this.

PR #14716, opened 2020-06-18
Fix #14715 - detect tool fails on FreeBSD

euantorano:
Fixes #14715

FreeBSD also requires linking the math library (see [`fenv`](https://www.freebsd.org/cgi/man.cgi?query=fenv)). It [seems NetBSD does too](https://netbsd.gw.com/cgi-bin/man-cgi?fenv+3+NetBSD-current), though I haven't tested there.

Tested locally on FreeBSD and it appears to work - I now get a `freebsd_amd64_consts.nim` being generated at least.

PR #14792, opened 2020-06-24
Patch #14716 - add missing `when`

euantorano:
Patch for #14716 - seems a `when` was lost in the PR and was missed before merging.

### DIFF
--- a/tools/detect/detect.nim
+++ b/tools/detect/detect.nim
@@ -98,8 +98,8 @@ proc main =
     f.write(nimfile % [other])
     close(f)
 
-  let cCompile = defined(openbsd) or defined(freebsd) or defined(netbsd): ccLinkMath else: cc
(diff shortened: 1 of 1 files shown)


timotheecour:
@euantorano this means CI doesn't catch this and should be fixed, at very least by trying to compile detect.nim; do you mind adding some test?

euantorano:
I’m not sure if any of the tools are built or ran under CI at all at the moment.

It would be easy enough to add another CI step to try build and run them and just ensure they run successfully (e.g. have a 0 exit code) which is probably enough?

timotheecour:
any of the steps below is better than nothing:
* step 0: ensure tools compiles
* step 1: ensure they run
* step 2: ensure they give correct results

so long it doesn't add significant time to CI, there's no drawbacks in doing this
but please don't add a new CI pipeline, just add code under koch.runCI

euantorano:
I'll try find a chance to look into it. Checking for correct results may be difficult due to the results being platform dependent. I'll open an issue.

PR #14793, opened 2020-06-24
Fix #14396

### LINKED ISSUE #14396: [ARC] destructive move destroys the object too early
This code works with the default GC, but with ARC it crashes with a SIGSEGV because ``spinny`` is getting sinked into ``createThread`` but it's being used after that.
### Example
```nim
# Compile with --threads:on
type
  Spinny = ref object
    t: Thread[Spinny]
    text: string

proc newSpinny*(): Spinny =
  Spinny(text: "hello")

proc spinnyLoop(spinny: Spinny) {.thread.} =
  echo spinny.text
# Adding {.nosinks.} either to this proc or to createThread solves the issue
proc start*(spinny: Spinny) =
  createThread(spinny.t, spinnyLoop, spinny)

proc stop*(spinny: Spinny) =
  joinThread(spinny.t)

var spinner1 = newSpinny()
spinner1.start()
spinner1.stop()
```

### Current Output
```
/home/dian/Projects/spinny/spinny.nim(19) spinny
/home/dian/Projects/spinny/spinny.nim(13) start
```

### Expected Output
```
hello
```

```
$ nim -v
Compiled at 2020-05-17

git hash: 40ac19572a86b5bfa7b57cf0482ae3a30432176a
```

Clyybber:


compiler/ccgcalls.nim
```diff
@@ -246,26 +255,72 @@ proc genArg(p: BProc, n: PNode, param: PSym; call: PNode): Rope =
...
+  # is a bit pessimistic currently
+  var needTmp = newSeq[bool](ri.len - 1)
+  var potentialWrites: seq[PNode]
+  for i in countdown(ri.len - 1, 1):
+    if ri[i].skipTrivialIndirections.kind == nkSym:
+      needTmp[i - 1] = potentialAlias(ri[i], potentialWrites)
+    else:
+      let cfg = constructCfg(ri[i])
+      var newPotentialWrites: seq[PNode]
+      # Simple linear pass, ignoring forks and gotos
```
Araq: Er, why do you need the DFA then? There is only ``nkAsgn, nkFastAsgn`` and addr(x) that is potentially destructive. Easy enough to check for without the DFA.

Clyybber: Done, though for some reason there are cases like `var f = 0; f += 1` where `f` is not an nkHiddenAddr, but an nkSym directly..

PR #14795, opened 2020-06-25
Add test-cases to some fixed issues to close them

ghost:


tests/async/tasync_misc.nim
```diff
@@ -74,3 +74,10 @@ block: # issue #13899
       result = true
   doAssert waitFor testFooSucceed()
   doAssert waitFor testFooFails()
+
+block: # 9313
```
timotheecour: `s/9313/issue #9313` so that it's easier to search than a raw number, which would give tons of false positives.

ditto elsewhere.

(=> https://github.com/nim-lang/Nim/pull/14796)

ghost: @timotheecour I tried to follow whatever namings are already there in files but okay

timotheecour: thanks for the change; as noted in  #14796 following surrounding code leads to dialects that survive ;-)

timotheecour (approved):

timotheecour:
@araq I'm merging this

timotheecour:
@Yardanico you need to add the closing string inside top PR message otherwise issues don't auto-close; I'll close them manually

EDIT: ok, it's partially my fault too, when I squashed, I should've kept the individual PR commit messages in this case; will remember for next time ;-) still though it's preferable for PR top message to mention those.

PR #14798, opened 2020-06-25
Clean out Deprecated toUint16,toUint32,toUint64,etc

juancarlospaco:
- Remove and/or clean out Deprecated `toUint16`, `toUint32`, `toUint64`, toUsomething, etc.  Deprecated since `0.19`.


Araq:
Why remove them? Still too much code that uses them?

PR #14801, opened 2020-06-25
New freebsd platforms

lbartoletti:
This PR proposes to add all platforms supported by FreeBSD. 

image

cc @euantorano

### DIFF
--- a/compiler/installer.ini
+++ b/compiler/installer.ini
@@ -9,7 +9,7 @@ Platforms: """
   linux: i386;hppa;ia64;alpha;amd64;powerpc64;arm;sparc;sparc64;m68k;mips;mipsel;mips64;mips64el;powerpc;powerpc64el;arm64;riscv64
   macosx: i386;amd64;powerpc64
   solaris: i386;amd64;sparc;sparc64
-  freebsd: i386;amd64;powerpc64;arm;arm64
+  freebsd: i386;amd64;powerpc64;arm;arm64;riscv64;sparc64;mips;mipsel;mips64;mips64el;powerpc
   netbsd: i386;amd64
   openbsd: i386;amd64;arm;arm64
   dragonfly: i386;amd64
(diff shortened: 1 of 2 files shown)


euantorano:
Unfortunately I only have access to an AMD64 based FreeBSD system at the moment, [as does builds.sr.ht](https://man.sr.ht/builds.sr.ht/compatibility.md#freebsd). This doesn't look like it'd cause any problems though as far as I can tell.

euantorano:
PS: note that the FreeBSD build failure is unrelated to this PR - sr.ht seems to have messed up due to a race condition that they're aware of. Pushing an empty commit should re-trigger it.

timotheecour:
Pushing an empty commit restarts all CI, instead you should only restart the CI that needs to be restarted

euantorano:
The only person who can do that is Araq as far as I know, and I prefer not to annoy him if possible.

timotheecour:
Just create a sourcehut account and you can

euantorano:
Yes, but then it gets re-submitted under my name instead: https://builds.sr.ht/~euantorano/job/241766

Thinking about it, I'm not sure if re-submitting even re-triggers the GitHub webhook to update the status.

timotheecour:
as you can see, it still works (it's now green for builds.sr.ht), but until it finishes, the badge isn't updated; I've reported it here: https://todo.sr.ht/~sircmpwn/builds.sr.ht/271 feel free to upvode

euantorano:
Ah, awesome. I didn’t think that would work since it would be under my name instead but it’s good to know it does.

timotheecour:
We need to add a section under contributing on how to restart CI in a resource effective way, it differs for azure vs github vs buildshrt

lbartoletti:
> Unfortunately I only have access to an AMD64 based FreeBSD system at the moment

No worries, I tested FreeBSD / RiscV with Qemu. PPC works already, Mips should work too. Like my previous PR for arm, with this PR our Nim package will be compiled on supported platforms (if it doesn't work in our package infrastructure, I can fix it ;) )

PR #14808, opened 2020-06-25
add debug fmt string like python's

ringabout:
Add `=` specifier to fmt-strings. `fmt"{expr=}"` expands to `fmt"expr={expr}"`.

## Examples
```nim
import strformat


let y = 3.1415926
doAssert fmt"{y=:.2f}" == "y=3.14"
```

[EDITED]
Print the error msgs：

```nim
import strformat


let age = 15
discard fmt"{age = = 12}"

# Error: could not parse `name = = 12`.
# Error: expression expected, but found '='
```
**before**
```nim
import strformat


let age = 15
discard fmt"{name != }"

# Error: could not parse `name != `.
# Error: expression expected, but found '[EOF]'
```

**after**
```nim
import strformat


let age = 15
discard fmt"{name != }"
# Error: could not parse `name !`.
# Error: expression expected, but found '[EOF]'
```

```nim
import strformat


let age = 15
discard fmt"{name ! = 12}"

# Error: could not parse `name ! = 12`.
# Error: expression expected, but found '='
```

```nim
import strformat


let name = 4
echo fmt"{name="
# Error: index 6 not in 0 .. 5
```


lib/pure/strformat.nim
```diff
@@ -552,8 +552,22 @@ proc strformatImpl(pattern: NimNode; openChar, closeChar: char): NimNode =
...
-          inc i
+          if f[i] == '=':
+            let start = i
+            inc i
+            i += f.skipWhitespace(i)
+            if f[i] == closeChar or f[i] == ':':
+              try:
+                discard parseExpr(subexpr)
+                result.add newCall(bindSym"add", res, newLit(subexpr & f[start ..< i]))
+              except ValueError:
```
Araq: That looks dangerous. Secretly misinterpreting what somebody might have intented.

Araq:
Nice idea!

lib/pure/strformat.nim
ringabout: I think fmt's safety is not much difference with the previous? The previous code used `parseExpr` too.

Araq: Can you remove the ``except ValueError: ...`` section please?

ghost:
Need to update the docs too I guess? But looks nice :)

ringabout:
Thanks for merging! Should I open a separate pr to add docs?

PR #14825, opened 2020-06-26
Add all missing css properties to dom.nim.

treeform:
* Add a ton of new properties.
* `zIndex` needs to be `cstring` as it can be "auto"
* Many of the `scrollbar*` are not part of the CSS standard, but I left them here for backward compatibility.
* Some things moved around due to alphabetical order.


dom96 (approved):
looks good

PR #14834, opened 2020-06-27
Fix nimdoc invalid css on theme switch class

KingDarBoja:
For some reason, there was an invalid css syntax on the `nimdoc.css` file. This PR fixes that 😆


Clyybber:
Interesting but unrelated test failure.

Issue #14820, opened 2020-06-26
timeout of asyncdispatch.drain is not respected and behaviour doesn't match poll


D-Nice:
The timeout behaviour appears buggy to me or at least non-intuitive, in that, if I have something async that takes 500ms, I can `poll` 5 times with 100ms timeouts, for it to resolve, whilst `drain` resolves on the 3rd call, waiting 200ms instead of the requested 100ms. The timeout will differ further, the more async events you may have in the pipeline, because of what appears to be a regression to me.

Simply said, I expected `drain` to allow events up to the timeout to complete and move on either after the timeout exceeds, or early if all pending events are complete.

My understanding was that it was meant as a replacement to `poll`, for when you'd like to allow all events up to a certain amount of timeout to complete. The commit message reads as: `introduce asyncdispatch.drain that completes all operations that can be completed immediately;` although I read `immediately` as within the timeout. The current behaviour https://github.com/nim-lang/Nim/commit/4e6f2b5313418c5f070611adb3b29a70c621782b doesn't match this at all and creates a dangerous, varying timeout, and the previous, pre-regression behaviour https://github.com/nim-lang/Nim/commit/3073f08e4857b14a6bd1b0022fe45f461bf88171 still doesn't match this and appears to behave as a carbon copy of poll.

### Example
https://play.nim-lang.org/#ix=2qch
```nim
import asyncdispatch, times

var done = false
proc somethingAsync() {.async.} =
  yield sleepAsync 5000
  echo "async done"
  done = true
  
discard somethingAsync()
while not done:
  let s0 = now()
  drain 1000
  echo "ms: ", (now() - s0).inMilliseconds
```

### Current Output
```
ms: 2002
ms: 2002
async done
ms: 997
```

### Expected Output
replacing drain with poll, one gets:
```
ms: 1001
ms: 1001
ms: 1001
ms: 1001
async done
ms: 997
```

and as it currently stands, `waitFor sleepAsync 1000` more closely matches poll, than drain does, except that the last line is 1001 as well, since it waits for this additional sleep, which is why drain is useful over just the sleepAsync hack.

### Possible Solution

* The current regression appears to have been introduced here: https://github.com/nim-lang/Nim/commit/4e6f2b5313418c5f070611adb3b29a70c621782b I think this simply needs to be reverted as a first order of business.

[...]

PR #14838, opened 2020-06-28
Fix asyncdispatch drain behavior (#14820)

rayman22201:
Drain was totally broken. Thanks to @D-Nice for pointing it out. This should be a better implementation.
I also made a tweak to the docs to make the behavior more clear, and added a test case based on the original #14820 report.
It should now have much more intuitive behavior.

@dom96 If you could please give me a code review I would very much appreciate it! Thanks! 😄 ❤️

### DIFF
--- a/tests/async/t14820.nim
+++ b/tests/async/t14820.nim
@@ -19,7 +19,9 @@ proc somethingAsync() {.async.} =
   
 asyncCheck somethingAsync()
 var count = 0
+let s0 = now()
 while not done:
   count += 1
   drain 100
   echo "iteration: ", count 
+  echo "ms: ", (now() - s0).inMilliseconds
--- a/tests/async/t14820.nim
+++ b/tests/async/t14820.nim
@@ -17,7 +17,7 @@ proc somethingAsync() {.async.} =
   echo "async done"
   done = true
   
-discard somethingAsync()
+asyncCheck somethingAsync()
 var count = 0
 while not done:
   count += 1
--- a/tests/async/t14820.nim
+++ b/tests/async/t14820.nim
@@ -13,15 +13,13 @@ import asyncdispatch, times
 
 var done = false
 proc somethingAsync() {.async.} =
-  yield sleepAsync 500
+  yield sleepAsync 5000
   echo "async done"
   done = true
   
 asyncCheck somethingAsync()
 var count = 0
-let s0 = now()
 while not done:
   count += 1
-  drain 100
+  drain 1000
   echo "iteration: ", count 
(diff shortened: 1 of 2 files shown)


rayman22201:
I'm investigating the Mac failures. It seems drain is ending one iteration early on Mac... because freaking Mac 😢

tests/async/t14820.nim
```diff
@@ -0,0 +1,25 @@
...
+
+import asyncdispatch, times
+
+var done = false
+proc somethingAsync() {.async.} =
+  yield sleepAsync 500
+  echo "async done"
+  done = true
+  
+discard somethingAsync()
```
D-Nice: I know this line is my fault, but just a heads up and maybe @dom96 would chime in that there's talk about forbidding discarding of futures (discardable pragma on async is already an error in devel), so maybe asyncCheck here instead would alleviate this becoming a blocker for such a PR in the future.

dom96: yeah, asyncCheck would be wise here

dom96 (approved):
Makes sense to me, I think we can merge this.

D-Nice:
Test passed fine on a macbook air:
https://i.imgur.com/N5VGSMw.png
https://i.imgur.com/lZn8lyW.png

rayman22201:
> 
> 
> Test passed fine on a macbook air:
> https://i.imgur.com/N5VGSMw.png
> https://i.imgur.com/lZn8lyW.png

Really strange. I wonder why the Azure CI is failing the Mac build...
https://dev.azure.com/nim-lang/Nim/_build/results?buildId=6468&view=logs&j=30931762-47c4-53b3-6a83-316eb5a6b9d7&t=b1c7d701-c448-5ecb-905d-689d8a5921e0&l=1446

It looks like Azure is using MacOS 10.15.
It's also only failing on the C back end. <s>The C++ backend succeeds.</s>
edit: nope, latest re-run fails with the C++ backend also 😞

rayman22201:
I tried to reproduce on my own Macbook air with MacOS 10.15 and I can't get the test to fail.
I even tried running the entire test suite to see if it was some kind of interference from another test. It just works...

At this point I'm not sure what to do other than push a bunch of debug logging to the PR in order to get more details into what is going on inside Azure...

rayman22201:
After some help from Leorize on irc, it's because the Azure VM's are slow and can easily stall async. Slowing down the test by increasing the timeouts solved the problem.

I'm not sure I'm completely satisfied by this solution, but it's good enough for now, and hopefully good enough to merge.

PR #14856, opened 2020-06-30
Clean out sets

juancarlospaco:
- Remove `sets.isValid`, `sets.toSet`, `sets.initSet`, was deprecated since `0.20`, Sets are initialized by default.
- Unrelated CI fail, failed to read `arraymancer` package for some reason.


narimiran:
I'm against this one.

I had all those removed in my initial version of https://github.com/nim-lang/Nim/pull/14699 but soon realized that lots of packages still use those. (I've send PRs to the packages I've noticed, but who knows how many are still there)

Araq:
Ok, rejected for the time being. But thanks for the effort!

PR #14861, opened 2020-06-30
add docs and more tests for debug format strings

ringabout:
ref https://github.com/nim-lang/Nim/pull/14808


lib/pure/strformat.nim
```diff
@@ -81,6 +80,44 @@ Formatting floats
     doAssert fmt"{123.456:13e}" == " 1.234560e+02"
 
 
+Debugging strings
+=================
+
+``fmt"{expr=}"`` expands to ``fmt"expr={expr}"`` namely the text of the exprssion,
```
juancarlospaco:
```suggestion
``fmt"{expr=}"`` expands to ``fmt"expr={expr}"`` namely the text of the expression, 
```

Varriount:
I'm wondering if we should put a disclaimer stating that the debug output is subject to change...

PR #14882, opened 2020-07-02
Create SECURITY.md

FedericoCeratto:
A basic security policy.
It could be improved with PGP encrypted email

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,15 @@
+# Security Policy
+
+## Supported Versions
+
+Security fixes are provided in new releases and bugfix releases.
+
+We do not backport security fixes to older releases.
+
+(Yet, Linux distributions might backport security fixes for their packages)
+
+## Reporting a Vulnerability
+
+Please do not report vulnerabilities on GitHub issues.
+
+Send an email to security@nim-lang.org or contact @d0m96 on Telegram using Secret Chat


ee7:
Thanks for making this PR. I think it's important to add a visible security policy.

My main criticism: Telegram has a poor reputation in the security community, and mentioning it in `SECURITY.md` would imply a certain lack of seriousness in Nim's approach to security.

I suggest:
- For this PR: remove the Telegram mention.
- Later: consider adding a PGP key if somebody is happy to deal with it. Note that even https://golang.org/security contains "We accept PGP-encrypted email, but the majority of the security team are not regular PGP users so it's somewhat inconvenient. Please only use PGP for critical security reports."
- More long-term: add a security page to the website (e.g. https://nim-lang.org/security) and add a link to that page in `SECURITY.md`.

Here are some example security files/policies for some well-known projects:
- https://github.com/golang/go/blob/master/SECURITY.md
- https://github.com/rust-lang/rust/security/policy
- https://github.com/kubernetes/kubernetes/security/policy
- https://github.com/django/django/security/policy
- https://github.com/rails/rails/security/policy
- https://github.com/tensorflow/tensorflow/blob/master/SECURITY.md#reporting-vulnerabilities
- https://github.com/electron/electron/blob/master/SECURITY.md
- https://github.com/facebook/react/blob/master/SECURITY.md
- https://github.com/twbs/bootstrap/blob/main/SECURITY.md

There's some good ideas there about other things to mention.

If you really wanted something secure with better usability than PGP then use Signal, not Telegram. Get a dedicated SIM for it so you don't need to put a personal phone number out there. This is a common practice for things like whistleblowing (where the conversation typically includes one or two non-technical people), but is less common for disclosing vulnerabilities.

This entire comment is non-blocking: writing `security@nim-lang.org` anywhere at all is a significant improvement on the current situation - I didn't find any mention of this email address when searching in this repo, the website repo, or Google.

Araq:
Should this instead be part of ``readme.md``, right before ``## Community`` ?

Jjp137:
GitHub recognizes [any file named `security.md` in the repository's root as the security policy](https://docs.github.com/en/github/managing-security-vulnerabilities/adding-a-security-policy-to-your-repository), but referring to it in `readme.md` should also be done.

SECURITY.md
```diff
@@ -0,0 +1,15 @@
...
+
+We do not backport security fixes to older releases.
+
+(Yet, Linux distributions might backport security fixes for their packages)
+
+## Reporting a Vulnerability
+
+Please do not report vulnerabilities on GitHub issues.
+
+Send an email to security@nim-lang.org or contact @d0m96 on Telegram using Secret Chat
```
dom96, line 15: Probably better not to have me as the PoC for this.

FedericoCeratto:
1. This is the initial step.
1. It can be merged and people can improve it in future if/once an official PGP key is provided, as I suggested
1. SECURITY.md is the default filename and location recommended by github and immediately recognized by security researchers
1. the telegram contact was given to me by @dom96 . I can remove it but an alternative PoC other than cleartext email would be nice. Understandably, not many people want to publish their phone numbers as Signal contacts. Is there any other option? @Araq ?

Araq:
I took over.

PR #14888, opened 2020-07-03
Add Deprecated pragma to Deprecated module

juancarlospaco:
- Add `{.deprecated.}` pragma to Deprecated module.
- Documentation explicitly says its Deprecated since `0.19` but it has no pragma.

### DIFF
--- a/lib/core/macrocache.nim
+++ b/lib/core/macrocache.nim
@@ -13,6 +13,8 @@
 ## as it breaks incremental compilations.
 ## Instead the API here needs to be used.
 
+{.deprecated: "Deprecated since 0.19; Not supported anymore as it breaks incremental compilation.".}
+
 type
   CacheSeq* = distinct string
   CacheTable* = distinct string


lib/core/macrocache.nim
```diff
@@ -13,6 +13,8 @@
 ## as it breaks incremental compilations.
 ## Instead the API here needs to be used.
 
+{.deprecated: "Deprecated since 0.19; Not supported anymore as it breaks incremental compilation.".}
```
alaviss, line 16: Huh? This is not deprecated.

According to the doc comment:
- Collecting compile time information across module boundaries in global variables is deprecated
- This module provides an alternative way to achieve that.

I can see why the documentation is confusing. If you notice the `git blame` then this module was added with those exact lines, so that rules out the "this module is deprecated" part.

Instead you should clarify the documentation comment above to separate the points correctly.

alaviss (changes requested):

PR #14896, opened 2020-07-04
Minor improvements to typecast section of manual

awr1:


doc/manual.rst
```diff
@@ -3265,15 +3265,18 @@ conversions too are now *always unchecked*.
...
+programming and are inherently unsafe.
 
 .. code-block:: nim
   cast[int](x)
 
-Type casts are a crude mechanism to interpret the bit pattern of
-an expression as if it would be of another type. Type casts are
-only needed for low-level programming and are inherently unsafe.
-
+Type casts should not be confused with type conversions, as the underlying bit
```
Varriount: I would rewrite the first couple of sentences to something like this, to reduce grammatical ambiguity:
> Type casts should not be confused with type conversions. Unlike type conversations, type casts do not change the underlying bit pattern of the data being casted (aside from possibly changing the size of the target type) [...]

Varriount (changes requested):

doc/manual.rst
awr1: Technically a type conversion can leave a bit pattern unaltered too (some, certainly not all). I will rephrase this however

PR #14903, opened 2020-07-04
Fix #13608 - discard staticExec

### LINKED ISSUE #13608: Cannot discard result of staticExec
I cannot discard the result of staticExec.

### Example
```nim
discard staticExec(r"echo ""hello""")
```

### Current Output
```
test.nim(1, 19) Error: 'staticExec' can only be used in compile-time context
```

### Expected Output
```
Successful compilation
```

### Possible Solution

It works if i write: 

```nim
const s = staticExec(r"echo ""hello""")
```

### Additional Information

I'm just trying to learn the language and find this behavior unexpected. 
Why would the program stop compiling if i discard the return value?

```
$ nim -v
```

genotrance:
Fixed as suggested by @krux02 in #13608.


genotrance:
This fix does not work since `gorge()` now runs in `$nim/lib` since that's where `system.nim` is. Any calls to `gorge()` that use relative path break as the PR did.

PR #14916, opened 2020-07-05
Allow overwritting of values in B-Tree

decanus:


Varriount:
What was the previous behavior? Isn't the point of insert not to replace things?

decanus:
@Varriount Prior to this it didn’t let you rewrite values. I’m not sure that is the point, I can’t find anything that states b-tree values should be immutable. What currently happens is it readds the key but the newer one is simply unaccessible, due to the get algorithm taking the first occurrence.

PR #14927, opened 2020-07-07
lexer speedups

Araq:


zah:
Is this measurable in practice? I'm asking because I have few simple ideas for how to speed up the lexer further.

Araq:
It "feels" faster for the Nim compiler but I have a different benchmark where it makes a difference of a factor of 3.

PR #14938, opened 2020-07-08
Move wasMoved out of destructor

Clyybber:
This should pave the road for some optimizations.

### DIFF
--- a/compiler/injectdestructors.nim
+++ b/compiler/injectdestructors.nim
@@ -93,7 +93,7 @@ proc optimize(s: var Scope) =
   proc findCorrespondingDestroy(final: seq[PNode]; moved: PNode): int =
     # remember that it's destroy(addr(x))
     for i in 0 ..< final.len:
-      if final[i] != nil and exprStructuralEquivalent(final[i][1].skipAddr, moved, strictSymEquality = true):
+      if final[i] != nil and exprStructuralEquivalent(final[i][0][1].skipAddr, moved, strictSymEquality = true):
         return i
     return -1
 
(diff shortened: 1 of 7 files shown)


Araq:
Needs a changelog entry and a documentation update.

Varriount:
Could there be situations where this new behavior is unwanted?

Clyybber:
@Varriount Nope. This design is better since we can then optimize those wasMoved calls away and we can elide the pointer indirection in `=destroy(a: var T)` if its not actually needed.

compiler/injectdestructors.nim
```diff
@@ -93,7 +93,7 @@ proc optimize(s: var Scope) =
   proc findCorrespondingDestroy(final: seq[PNode]; moved: PNode): int =
     # remember that it's destroy(addr(x))
     for i in 0 ..< final.len:
-      if final[i] != nil and exprStructuralEquivalent(final[i][1].skipAddr, moved, strictSymEquality = true):
+      if final[i] != nil and exprStructuralEquivalent(final[i][0][1].skipAddr, moved, strictSymEquality = true):
```
Araq, line 96: I don't understand this change.

compiler/injectdestructors.nim
```diff
@@ -547,15 +546,15 @@ proc ensureDestruction(arg: PNode; c: var Con; s: var Scope): PNode =
     result = newNodeIT(nkStmtListExpr, arg.info, arg.typ)
 
     if s.parent != nil:
-      let tmp = getTemp(c, s.parent[], arg.typ, arg.info)
-      result.add genSink(c, s, tmp, arg, isDecl = true)
+      let tmp = c.getTemp(s.parent[], arg.typ, arg.info)
+      result.add c.genSink(s, tmp, arg, isDecl = true)
       result.add tmp
-      s.parent[].final.add genDestroy(c, tmp)
+      s.parent[].final.add newTree(nkStmtList, c.genDestroy(tmp), c.genWasMoved(tmp))
```
Araq, line 552: Well the variable is to disappear. We can avoid the ``wasMoved(x)`` call unless it's in a loop. In fact, even if in a loop if we use nkFastAssign instead of =sink or '='.

lib/system/seqs_v2.nim
```diff
@@ -73,11 +73,10 @@ proc shrink*[T](x: var seq[T]; newLen: Natural) =
   when nimvm:
     setLen(x, newLen)
   else:
-    mixin `=destroy`
     #sysAssert newLen <= x.len, "invalid newLen parameter for 'shrink'"
     when not supportsCopyMem(T):
       for i in countdown(x.len - 1, newLen):
-        `=destroy`(x[i])
+        reset x[i]
```
Araq, line 79: This is quite useful and should make it into the manual. Custom destructors should call ``reset`` rather than ``=destroy`` if in doubt.

Araq:
I'm taking over...

PR #14949, opened 2020-07-10
Fix #14906

### LINKED ISSUE #14906: Calling peekChar with Stream returned from osproc.outputStream generate runtime error
On Windows, `peekDataImpl` for pipe is not implemented.

On Linux, `ftell` is called with pipe but it doesn't work with pipe.
And you can't seek backwards from a pipe.
https://stackoverflow.com/questions/19187131/how-to-fseek-on-a-pipe

Following code generate runtime error on both Linux and Windows.

```nim
import osproc, streams

var
  p = startProcess("echo", args = ["foobar"], options = {poUsePath})
  outStrm = p.outputStream

echo outStrm.peekChar
p.close
```

Output on windows:
```
f:\temp>peekchar.exe
f:\temp\peekchar.nim(7)  peekchar
f:\project\demotomohiro\Nim\lib\pure\streams.nim(469) peekChar
f:\project\demotomohiro\Nim\lib\pure\streams.nim(307) peekData
```

Output on Linux:
```
colab@:~$ ./peekchar
/home/colab/peekchar.nim(7) peekchar
/home/colab/Nim/lib/pure/streams.nim(469) peekChar
/home/colab/Nim/lib/pure/streams.nim(307) peekData
/home/colab/Nim/lib/pure/streams.nim(1328) fsPeekData
/home/colab/Nim/lib/pure/streams.nim(1319) fsGetPosition
/home/colab/Nim/lib/system/io.nim(677) getFilePos
/home/colab/Nim/lib/system/io.nim(139) raiseEIO
Error: unhandled exception: cannot retrieve file position [IOError]
```

Nim version on Windows:
```
f:\temp>nim -v
Compiled at 2020-07-05

git hash: af27e6bdea63bbf66718193ec44bc61e745ded38
```

Nim version on Linux:
```
colab@:~$ nim -v
Compiled at 2020-07-05

git hash: af27e6bdea63bbf66718193ec44bc61e745ded38
```

demotomohiro:
There is a windows API [PeekNamedPipe](https://docs.microsoft.com/en-us/windows/win32/api/namedpipeapi/nf-namedpipeapi-peeknamedpipe) that can peek anonymous pipe but it returns without waiting for data to read.

I couldn't find how to peek a pipe on Linux.
[It is not possible to apply lseek to a pipe.](https://man7.org/linux/man-pages/man7/pipe.7.html)

So I have implemented a `PipeOutStream` ref object type that wrap exising stream type used for outputStream.
When it peeks, it reads data from stream and store data to buffer and also return it.
Next time it reads, it return data in buffer so that previus peek call looks like didn't changed read position.

### DIFF
--- a/lib/pure/osproc.nim
+++ b/lib/pure/osproc.nim
@@ -438,6 +438,13 @@ template streamAccess(p) =
 # when setPosition/getPosition is called or write operation is performed.
 type
   PipeOutStream[T] = ref object of T
+    # When stream peek operation is called, it reads from base stream
+    # type using `baseReadDataImpl` and stores the content to this buffer.
+    # Next stream read operation returns data in the buffer so that previus peek
+    # operation looks like didn't changed read positon.
+    # When stream read operation that returns N byte data is called and the size is smaller than buffer size,
+    # first N elements are removed from buffer.
+    # Deque type can do such operation more efficiently than seq type.
     buffer: Deque[char]
     baseReadLineImpl: typeof(StreamObj.readLineImpl)
     baseReadDataImpl: typeof(StreamObj.readDataImpl)
--- a/lib/pure/osproc.nim
+++ b/lib/pure/osproc.nim
@@ -482,17 +482,21 @@ proc posPeekData[T](s: Stream, buffer: pointer, bufLen: int): int =
   var s = PipeOutStream[T](s)
   assert s.baseReadDataImpl != nil
 
-  let n = min(s.buffer.len, bufLen)
+  let
+    dest = cast[ptr UncheckedArray[char]](buffer)
+    n = min(s.buffer.len, bufLen)
+
   result = n
-  if bufLen > n:
-    let m = bufLen - n
-    var buf = newSeq[char](m)
-    result += s.baseReadDataImpl(s, addr buf[0], m)
-    for i in 0..<m:
-      s.buffer.addLast buf[i]
-  let dest = cast[ptr UncheckedArray[char]](buffer)
-  for i in 0..<bufLen:
+  for i in 0..<n:
     dest[i] = s.buffer[i]
+
+  if bufLen > n:
+    let
+      newDataNeeded = bufLen - n
+      numRead = s.baseReadDataImpl(s, addr dest[n], newDataNeeded)
+    result += numRead
+    for i in 0..<numRead:
+      s.buffer.addLast dest[n + i]
 
 proc newPipeOutStream[T](s: sink (ref T)): owned PipeOutStream[T] =
   assert s.readDataImpl != nil
--- a/lib/pure/osproc.nim
+++ b/lib/pure/osproc.nim
@@ -18,7 +18,7 @@
 include "system/inclrtl"
 
 import
-  strutils, os, strtabs, streams, cpuinfo
+  strutils, os, strtabs, streams, cpuinfo, deques
 
 export quoteShell, quoteShellWindows, quoteShellPosix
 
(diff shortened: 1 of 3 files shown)


lib/pure/osproc.nim
```diff
@@ -434,6 +434,93 @@ when not defined(useNimRtl):
...
+proc posReadData[T](s: Stream, buffer: pointer, bufLen: int): int =
+  var s = PipeOutStream[T](s)
+  assert s.baseReadDataImpl != nil
+
+  let
+    pbuf = cast[ByteAddress](buffer)
+    n = min(s.buffer.len, bufLen)
+  result = n
+  for i in 0..<n:
+    cast[ptr char](pbuf + i)[] = s.buffer.popFirst
```
Araq: Please use a ``ptr UncheckedArray`` instead for ``pbuf``.

lib/pure/osproc.nim
```diff
@@ -434,6 +434,93 @@ when not defined(useNimRtl):
...
+  func head(x: ref T): pointer =
+    cast[pointer](cast[ByteAddress](addr x[]) + sizeof(pointer))
+
+  copyMem(dest.head, source.head, sizeof(T) - sizeof(pointer))
+
+proc newPipeOutStream[T](s: sink (ref T)): owned PipeOutStream[T] =
+  assert s.readDataImpl != nil
+
+  new(result)
+  # Please tell me how to done this correclty.
```
Araq: Maybe like this:

```nim

for dest, src in fields(result[], s[]): dest = src
```

Varriount:
I'd like to point out this note in the [PeekNamedPipe documentation](https://docs.microsoft.com/en-us/windows/win32/api/namedpipeapi/nf-namedpipeapi-peeknamedpipe):

> The PeekNamedPipe function can block thread execution the same way any I/O function can when called on a synchronous handle in a multi-threaded application. To avoid this condition, use a pipe handle created for asynchronous I/O.

I'm not sure this addition will always return immediately.

lib/pure/osproc.nim
```diff
@@ -434,6 +434,87 @@ when not defined(useNimRtl):
 template streamAccess(p) =
   assert poParentStreams notin p.options, "API usage error: stream access not allowed when you use poParentStreams"
 
+# Wrap pipe for reading with PipeOutStream so that you can use peek* and generate runtime error
+# when setPosition/getPosition is called or write operation is performed.
+type
+  PipeOutStream[T] = ref object of T
+    buffer: Deque[char]
```
Varriount: A comment should be added explaining why a deque is used.

lib/pure/osproc.nim
```diff
@@ -434,6 +434,87 @@ when not defined(useNimRtl):
...
+proc posReadLine[T](s: Stream, line: var TaintedString): bool =
+  var s = PipeOutStream[T](s)
+  assert s.baseReadLineImpl != nil
+
+  let n = s.buffer.len
+  line.string.setLen(0)
+  for i in 0..<n:
+    var c = s.buffer.popFirst
+    if c == '\c':
+      c = readChar(s)
```
Varriount: Why is this reading a character? Won't that data be lost?

lib/pure/osproc.nim
```diff
@@ -434,6 +434,87 @@ when not defined(useNimRtl):
...
+
+proc posPeekData[T](s: Stream, buffer: pointer, bufLen: int): int =
+  var s = PipeOutStream[T](s)
+  assert s.baseReadDataImpl != nil
+
+  let n = min(s.buffer.len, bufLen)
+  result = n
+  if bufLen > n:
+    let m = bufLen - n
+    var buf = newSeq[char](m)
```
Varriount: Rather than create a sequence here, is it possible to extend the existing buffer, and read data directly into it?
Something like:
```
  let dest = cast[ptr UncheckedArray[char]](buffer)

  let newDataNeeded = bufLen - s.buffer.len
  if newDataNeeded > 0:
    # We need more data than the internal buffer can supply, so
    # use the user's buffer to retrieve the new data
    result = bufLen
    result += s.baseReadDataImpl(s, addr dest[s.buffer.len], newDataNeeded)

    # Copy the user buffer's new data into the stream buffer
    for i in 0 ..< newDataNeeded:
      s.buffer.addLast dest[i + s.buffer.len]
    # Copy the stream buffer's old data into the user buffer
    for i in 0 .. s.buffer.len:
      dest[i] = s.buffer[i]

  else:
    # Copy some of the stream buffer data into the user buffer
    for i in 0 ..< bufLen:
      dest[i] = s.buffer[i]
```

lib/pure/osproc.nim
```diff
@@ -434,6 +434,87 @@ when not defined(useNimRtl):
...
+
+  let n = s.buffer.len
+  line.string.setLen(0)
+  for i in 0..<n:
+    var c = s.buffer.popFirst
+    if c == '\c':
+      c = readChar(s)
+      return true
+    elif c == '\L': return true
+    elif c == '\0':
```
Varriount: Why is a null checked for here? What if the stream contains null characters?

Varriount (commented):
This looks good, although I posted a possible improvement to the `posPeeekData` code.
I also have some questions about certain parts of the implementation

lib/pure/osproc.nim
demotomohiro: I copied code from `readLine` proc in `streams` module so that `posReadLine` take new line from the buffer in the same way as `readLine` in `streams` module.
On windows and dos, a line of text is delimited ``CRLF``.
The newline character(s) are not part of the returned string.

lib/pure/osproc.nim
demotomohiro: This code was also copied from `readLine` proc in `streams` module.

Araq:
The code is good now but this new stream type should be moved into its own module as a stream wrapper. And then people who need ``peek`` on osproc's streams can wrap the stream themselves in a PeekableStream. Otherwise we complexify the codebase more and more for niché use cases.

demotomohiro:
I have moved implementation of `PipeOutStream` to new module but added `peekableOutputStream` proc and `peekableErrorStream` proc to osproc module.

`PipeOutStream` is a generic ref objec type that wrap pipe stream by inheriting it.
But if `PipeOutStream` was moved to other file, `newPipeOutStream` cannot get actual type of pipe stream because `outputStream` return `Stream` type.
On windows, `outputStream` returns `FileHandleStream` type ref object as `Stream` type but it is not public.

If you don't like to add any code to osproc module, `PipeOutStream` type need to have `Stream` type instead of inheriting the type that inherits `Stream` type.
For example:
```nim
type
  PipeOutStream = ref object of StreamObj
    baseStream: Stream
```

But it become a type that has 10 unused procedural type because `PipeOutStream` also need to inherit `Stream` type.
`StreamObj` type has 10 procedural types.

Araq:
Thank you, superb work.

PR #14960, opened 2020-07-10
Add testcase for some old fixed issues

### LINKED ISSUE #2586: Integer literals' types not inferred uniformly
``` nimrod
import unsigned
var x = 10'u8
echo x < 9 # Works
echo x > 9 # Error: type mismatch
```

It is unexpected that `x < 9` works but `x > 9` doesn't. Of course this happens because `>` is implemented as a template rewriting it to `9 < x`. Because 9 occurs before x now, 9 is automatically an int. Would be nicer if the integer type could be inferred here as well.

### LINKED ISSUE #2703: Generic non-ref type conversion causes error at C compiler
``` nimrod
type
  fooObj[T] = object of RootObj
  bazObj[T] = object of fooObj[T]
    x: T

var troz: fooObj[string]
echo bazObj[string](troz).x
```

Fails with:

```
nimcache/x.c: In function ‘xInit’:
nimcache/x.c:144:2: error: conversion to non-scalar type requested
  printf("%s\012", ((bazobj88034) (troz_88025)).x? (((bazobj88034) (troz_88025)).x)->data:"nil");
  ^
nimcache/x.c:144:2: error: conversion to non-scalar type requested
```

At least it should fail with an error by the Nim compiler.

Clyybber:
Closes #2703
Closes #2586

### DIFF
--- a/tests/template/template_issues.nim
+++ b/tests/template/template_issues.nim
@@ -271,6 +271,6 @@ ggg(hello)
 
 
 # bug #2586
-var x = 10'u8
-echo x < 9 # Works
-echo x > 9 # Error: type mismatch
+var z = 10'u8
+echo z < 9 # Works
+echo z > 9 # Error: type mismatch
--- a/tests/template/template_issues.nim
+++ b/tests/template/template_issues.nim
@@ -11,6 +11,8 @@ hey
 foo
 foo
 foo
+false
+true
 '''
 """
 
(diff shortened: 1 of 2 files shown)


ee7:
@Clyybber This PR moved [tissues.nim](https://github.com/nim-lang/Nim/blob/21772b6419b6cb3b59b436fbdcd8975a9da62349/tissues.nim) to the root directory of the repo. Did you mean a different location?

Clyybber:
Thanks, fixed! (meant to put it in `tests/ccgbugs/`)

PR #14972, opened 2020-07-13
fix #14082 by raising a ValueError for illegal hex value in uri.nim

### LINKED ISSUE #14082: uri.decodeUrl crashes on incorrectly formatted input
Function `uri.decodeUrl` crashes when encoding sequences are incorrect. In server-side contexts, this might cause denial of service if not handled properly.

### Example
```nim
import uri
echo decodeUrl("abc%xx")
```

```nim
import uri
echo decodeUrl("abc%")
```

### Current Output

```
/usercode/in.nim(3)      in
/playground/nim/lib/pure/uri.nim(109) decodeUrl
/playground/nim/lib/pure/uri.nim(100) handleHexChar
/playground/nim/lib/system/assertions.nim(29) failedAssertImpl
/playground/nim/lib/system/assertions.nim(22) raiseAssert
Error: unhandled exception: /playground/nim/lib/pure/uri.nim(100, 17) `false`  [AssertionError]
```

or

```
/usercode/in.nim(3)      in
/playground/nim/lib/pure/uri.nim(109) decodeUrl
Error: unhandled exception: index 4 not in 0 .. 3 [IndexError]
```

### Expected Output

Either some correct way to handle the error (I don't think there is, in go, the error is returned as an error object) or the situation should be properly checked to avoid runtime errors.

In any case, the documentation should mention the possibility of errors.

### Possible Solution

Add an assert for the first case, or detect both cases properly and raise a specific error

### Additional Information

I found this while I was using the following code: https://github.com/nim-lang/Nim/pull/7761#issuecomment-616810243

Nim version 1.2.0 from the playground

narimiran:
Another week, another opportunity for some ~~bikeshedding~~ discussion :)

### DIFF
--- a/lib/pure/cgi.nim
+++ b/lib/pure/cgi.nim
@@ -37,7 +37,7 @@ proc handleHexChar(c: char, x: var int) {.inline.} =
   of '0'..'9': x = (x shl 4) or (ord(c) - ord('0'))
   of 'a'..'f': x = (x shl 4) or (ord(c) - ord('a') + 10)
   of 'A'..'F': x = (x shl 4) or (ord(c) - ord('A') + 10)
-  else: assert(false)
+  else: raise new ValueError
 
 proc addXmlChar(dest: var string, c: char) {.inline.} =
   case c
@@ -102,11 +102,16 @@ iterator decodeData*(data: string): tuple[key, value: TaintedString] =
     while i < data.len:
       case data[i]
       of '%':
-        var x = 0
-        handleHexChar(data[i+1], x)
-        handleHexChar(data[i+2], x)
-        inc(i, 2)
-        add(name, chr(x))
+        try:
+          var x = 0
+          handleHexChar(data[i+1], x)
+          handleHexChar(data[i+2], x)
+          inc(i, 2)
+          add(name, chr(x))
+        except ValueError:
+          raise newException(ValueError, data[i .. i+2] & " is not a valid hex value")
+        except IndexDefect:
+          raise newException(ValueError, "not a two-digit hex value after '%'")
       of '+': add(name, ' ')
       of '=', '&': break
       else: add(name, data[i])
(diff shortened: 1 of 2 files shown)


lib/pure/cgi.nim
```diff
@@ -37,7 +37,7 @@ proc handleHexChar(c: char, x: var int) {.inline.} =
   of '0'..'9': x = (x shl 4) or (ord(c) - ord('0'))
   of 'a'..'f': x = (x shl 4) or (ord(c) - ord('a') + 10)
   of 'A'..'F': x = (x shl 4) or (ord(c) - ord('A') + 10)
-  else: assert(false)
+  else: raise new ValueError
```
Araq, line 40: Make it return  a bool instead, maybe.

lib/pure/cgi.nim
```diff
@@ -102,11 +102,16 @@ iterator decodeData*(data: string): tuple[key, value: TaintedString] =
...
-        add(name, chr(x))
+        try:
+          var x = 0
+          handleHexChar(data[i+1], x)
+          handleHexChar(data[i+2], x)
+          inc(i, 2)
+          add(name, chr(x))
+        except ValueError:
+          raise newException(ValueError, data[i .. i+2] & " is not a valid hex value")
+        except IndexDefect:
```
Araq, line 113: Not acceptable, do the ``i + 2 < data.len`` check instead.

lib/pure/uri.nim
```diff
@@ -101,24 +103,34 @@ proc decodeUrl*(s: string, decodePlus = true): string =
     assert decodeUrl("https%3A%2F%2Fnim-lang.org%2Fthis+is+a+test") == "https://nim-lang.org/this is a test"
     assert decodeUrl("https%3A%2F%2Fnim-lang.org%2Fthis%20is%20a%20test",
         false) == "https://nim-lang.org/this is a test"
+    doAssertRaises(ValueError):
+      discard decodeUrl("abc%xx")
+    doAssertRaises(ValueError):
+      discard decodeUrl("abc%")
+
   proc handleHexChar(c: char, x: var int) {.inline.} =
```
Araq, line 111: While we're at it, remove the code duplication.

narimiran:
@Araq the review comments addressed in https://github.com/nim-lang/Nim/pull/14977

PR #14974, opened 2020-07-13
fix index error in logging (substituteLog)

hendrikgit:
Calling substituteLog with a frmt string that only has a $variable (or only $) and no space or other character after that leads to an IndexError at runtime.

Output of
```nim
import logging
echo substituteLog("$levelid", lvlDebug, "error")
```
is
```
...
.../lib/pure/logging.nim(305) substituteLog
.../lib/system/fatal.nim(49) sysFatal
Error: unhandled exception: index 8 not in 0 .. 7 [IndexError]
```

After seeing the output everyone will probably want to add a space anyway but I think it still should not lead to an IndexError to not do so. There is a runnableExamples test and a fix for the IndexError in this PR. The length check solves this error but I don't know if it is the most fitting solution.


Varriount:
Could you add a unit test for this (or add to the module tests, if they exist)?

Issue #9535, opened 2018-10-27
Shadow DOM support in libs/js/dom.nim [Feature Request]


Epictek:
### Summary
We should be able to create and attach Shadow DOMs now that most browsers support them.
Firefox is to support them in the latest update/already does in nightly and chromium has support for them already.

### Additional Information
Dom Spec: https://www.w3.org/TR/dom41/#interface-shadowroot

ringabout:
This can be closed now?

PR #14979, opened 2020-07-14
Shadow Dom apis

bung87:
maybe there's better name for proc parameter's type

### DIFF
--- a/lib/js/dom.nim
+++ b/lib/js/dom.nim
@@ -1340,7 +1340,7 @@ since (1, 3):
       delegatesFocus*: bool
 
     HTMLSlotElement* = ref HTMLSlotElementObj
-    HTMLSlotElementObj {.importc.} = object of ElemeRootObjnRootObjtObj
+    HTMLSlotElementObj {.importc.} = object of RootObj
       name*: cstring
     SlotOptions* = object of RootObj
       flatten*: bool
--- a/lib/js/dom.nim
+++ b/lib/js/dom.nim
@@ -1528,8 +1528,8 @@ since (1, 3):
 
   # DocumentOrShadowRoot
   proc getSelection*(n: DocumentOrShadowRoot): Selection
-  proc elementFromPoint*(n: DocumentOrShadowRoot; x,y: float): Element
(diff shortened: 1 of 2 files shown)


lib/js/dom.nim
```diff
@@ -1503,6 +1523,19 @@ proc contains*(n: Node): bool
 proc isEqualNode*(n: Node): bool
 proc isSameNode*(n: Node): bool
 
+since (1, 3):
+  proc getRootNode*(n: Node,options: RootNodeOptions): Node
+
+  # DocumentOrShadowRoot
+  proc getSelection*(n: DocumentOrShadowRoot): Selection
+  proc elementFromPoint*(n: DocumentOrShadowRoot; x,y: float): Element
```
ringabout: Need a space between `x` and `y`.

lib/js/dom.nim
```diff
@@ -1503,6 +1523,19 @@ proc contains*(n: Node): bool
...
+  proc elementsFromPoint*(n: DocumentOrShadowRoot; x,y: float): seq[Element]
```
ringabout: Need a space between x and y too.

ringabout:
related issue:
https://github.com/nim-lang/Nim/issues/9535

Araq:
From https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/elementsFromPoint

> This is an experimental technology
> Check the Browser compatibility table carefully before using this in production.

This should go into a dom_extensions.nim module instead.

Araq:
Huh? Was that the only API that is "experimental technology"?

bung87:
that one all browsers supported, others non-standard or draft stage

lib/js/dom_extensions.nim
```diff
@@ -0,0 +1,5 @@
+import dom
+
+{.push importcpp.}
```
juancarlospaco, line 3: `{.importjs.}` ?. I think it works Ok now.

bung87: maybe , as it is a new file, maybe ok to switch to `importjs`

PR #15012, opened 2020-07-18
[Backport] Fixes callbacks being dropped on Linux/macOS/BSD.

### LINKED ISSUE #15003: AsyncSocket is failing to send data under backpressure (Mac,Linux)
`AsyncSocket.send` loses data in the presence of backpressure, i.e. when the receiver reads data more slowly than the sender sends it. At some point soon after the OS's send buffer fills up, some number of pending send calls are ignored: their data is never written to the actual socket, and their Futures are not completed.

This is quite reproducible on macOS 10.14 and 10.15; I don't have convenient access to other platforms to test on.
Occurs in Nim 1.2.2 as well as latest devel.

### Example
```nim
# AsyncSocketBug.nim
# Jens Alfke (@snej) -- 16 July 2020
# Demonstrates data loss by Nim's AsyncSocket.
# Just run it, and it will raise an assertion failure within a minute.

import asyncdispatch, asyncnet, strformat, strutils, sugar

const FrameSize = 9999   # Exact size not important, but larger sizes fail quicker

proc runServer() {.async.} =
    # Server side:
    var server = newAsyncSocket()
    server.bindAddr(Port(9001))
    server.listen()
    let client = await server.accept()
    echo "Server got client connection"
    var lastN = 0
    while true:
        let frame = await client.recv(FrameSize)
        assert frame.len == FrameSize
        let n = frame[0..<6].parseInt()
        echo "RCVD #", n, ":  ", frame[0..80], "..."
        if n != lastN + 1:
            echo &"******** ERROR: Server received #{n}, but last was #{lastN}!"
        assert n == lastN + 1
        lastN = n
        await sleepAsync 100


proc main() {.async.} =
    asyncCheck runServer()

    # Client side:
    let socket = newAsyncSocket()
    await socket.connect("localhost", Port(9001))
    echo "Client socket connected"

    var sentCount = 0
    var completedCount = 0

    while true:
        sentCount += 1
        let n = sentCount

        var message = &"{n:06} This is message #{n} of ∞. Please stay tuned for more. "
        #echo ">>> ", message
        while message.len < FrameSize:
            message = message & message
        let frame = message[0..<FrameSize]

        capture n:
            socket.send(frame).addCallback proc(f: Future[void]) =
                # Callback when the send completes:
                assert not f.failed
                echo "SENT #", n
                if n != completedCount + 1:
                    echo &"******** ERROR: Client completed #{n}, but last completed was #{completedCount}!"
                # If this assert is enabled, it will trigger earlier than the server-side assert above:
                #assert count == completedCount + 1
                completedCount = n
        await sleepAsync 1

waitFor main()
```

### Current Output
```
Client socket connected
Server got client connection
SENT #1
RCVD #1:  000001 This is message #1 of ∞. Please stay tuned for more. 000001 This is mess...
SENT #2
SENT #3
SENT #4
SENT #5
SENT #6
SENT #7
SENT #8

.....

SENT #111
SENT #2074
******** ERROR: Client completed #2074, but last completed was #111!
SENT #2075
SENT #2076

.....

SENT #6312
SENT #6313
RCVD #108:  000108 This is message #108 of ∞. Please stay tuned for more. 000108 This is me...
RCVD #109:  000109 This is message #109 of ∞. Please stay tuned for more. 000109 This is me...
RCVD #110:  000110 This is message #110 of ∞. Please stay tuned for more. 000110 This is me...
RCVD #111:  000111 This is message #111 of ∞. Please stay tuned for more. 000111 This is me...
RCVD #2074:  002074 This is message #2074 of ∞. Please stay tuned for more. 002074 This is m...
******** ERROR: Server received #2074, but last was #111!
asyncfutures.nim(431)    asyncCheckCallback
Error: unhandled exception: AsyncSocketBug.nim(25, 16) `n == lastN + 1`  [AssertionDefect]
```

Note that messages 112…2073 were dropped on the floor: neither completed nor sent over the socket.

### Expected Output
Should run forever without any assertion failures and without writing any "ERROR" lines.

### Additional Information
This completely breaks the WebSocket-based protocol implementation I'm working on, in which one side typically sends large numbers of small messages. The bug did not show up as long as I was running my tests on one machine (over the loopback interface), because the receiver process was able to keep up with the sender (about 15 megabytes/sec); but it showed up very quickly as soon as I ran any tests on two machines, where WiFi bandwidth+latency was a bottleneck.

```
$ nim -v
Compiled at 2020-07-16

```

(Was pulled and built this morning.)

dom96:
Fixes #15003.

This is a serious bug which occurs when data cannot be read/sent immediately and there are a bunch of other read/write events pending. What happens is that the new events are dropped which results in the case of the reported bug resulted in some data not being sent (!).

@narimiran if this could be backported I would be eternally grateful.

The code I used to debug this for those curious: https://gist.github.com/dom96/4a3c214ff38339ce36300549657e6e53

Tested via:

```
PATH=/mnt/c/Users/morfe/projects/nim/bin:$PATH ./koch tests r tests/async/tasynceagain.nim
```

(Verified that tests pass with change, and that they fail without the change).

### DIFF
--- a/lib/pure/asyncdispatch.nim
+++ b/lib/pure/asyncdispatch.nim
@@ -1228,14 +1228,19 @@ else:
     let newLength = max(len(curList), InitCallbackListSize)
     var newList = newSeqOfCap[Callback](newLength)
 
+    var eventsExtinguished = false
     for cb in curList:
+      if eventsExtinguished:
+        newList.add(cb)
+        continue
       if not cb(fd):
         # Callback wants to be called again.
         newList.add(cb)
         # This callback has returned with EAGAIN, so we don't need to
         # call any other callbacks as they are all waiting for the same event
         # on the same fd.
-        break
+        # We do need to ensure they are called again though.
+        eventsExtinguished = true
 
     withData(selector, fd.int, fdData) do:
       # Descriptor is still present in the queue.
--- a/tests/async/tasynceagain.nim
+++ b/tests/async/tasynceagain.nim
@@ -1,0 +1,67 @@
+discard """
+  disabled: "windows"
+  exitcode: 0
+"""
+# AsyncSocketBug.nim
+# Jens Alfke (@snej) -- 16 July 2020
+# Demonstrates data loss by Nim's AsyncSocket.
+# Just run it, and it will raise an assertion failure within a minute.
+
+import asyncdispatch, asyncnet, strformat, strutils, sugar
+
+const FrameSize = 9999   # Exact size not important, but larger sizes fail quicker
+
+proc runServer() {.async.} =
+  # Server side:
+  var server = newAsyncSocket()
+  server.bindAddr(Port(9001))
+  server.listen()
+  let client = await server.accept()
+  echo "Server got client connection"
+  var lastN = 0
+  while true:
+    let frame = await client.recv(FrameSize)
+    assert frame.len == FrameSize
+    let n = frame[0..<6].parseInt()
+    echo "RCVD #", n, ":  ", frame[0..80], "..."
+    if n != lastN + 1:
+      echo &"******** ERROR: Server received #{n}, but last was #{lastN}!"
+    assert n == lastN + 1
+    lastN = n
+    await sleepAsync 100
+
+
+proc main() {.async.} =
+  asyncCheck runServer()
+
+  # Client side:
+  let socket = newAsyncSocket(buffered = false)
+  await socket.connect("localhost", Port(9001))
+  echo "Client socket connected"
+
+  var sentCount = 0
+  var completedCount = 0
+
+  while sentCount < 2000:
+    sentCount += 1
+    let n = sentCount
+
+    var message = &"{n:06} This is message #{n} of ∞. Please stay tuned for more. "
+    #echo ">>> ", message
+    while message.len < FrameSize:
+      message = message & message
+    let frame = message[0..<FrameSize]
+
+    capture n:
+      socket.send(frame).addCallback proc(f: Future[void]) =
+        # Callback when the send completes:
+        assert not f.failed
+        echo "SENT #", n
+        if n != completedCount + 1:
+          echo &"******** ERROR: Client completed #{n}, but last completed was #{completedCount}!"
+        # If this assert is enabled, it will trigger earlier than the server-side assert above:
+        assert n == completedCount + 1
+        completedCount = n
+    await sleepAsync 1
+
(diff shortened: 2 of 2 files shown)


lib/pure/asyncdispatch.nim
```diff
@@ -1228,14 +1228,19 @@ else:
     let newLength = max(len(curList), InitCallbackListSize)
     var newList = newSeqOfCap[Callback](newLength)
 
+    var eventsExtinguished = false
     for cb in curList:
+      if eventsExtinguished:
+        newList.add(cb)
```
rayman22201, line 1234: Do you need the same `if not cb(fd)` check here?
IIUC, a previous setting of the `eventsExtinguished` flag from a possibly different cb doesn't guarantee this cb wants to be re-added to the list.

lib/pure/asyncdispatch.nim
```diff
@@ -1228,14 +1228,19 @@ else:
...
+        continue
       if not cb(fd):
         # Callback wants to be called again.
         newList.add(cb)
         # This callback has returned with EAGAIN, so we don't need to
         # call any other callbacks as they are all waiting for the same event
         # on the same fd.
-        break
```
rayman22201, line 1238: It's not sufficient to just remove the `break`? Why is the `eventsExtinguished` flag also necessary?
The `break` is what is causing the other cbs in the curList to not get run. Or are the cb's getting dropped in the next poll cycle as well?

rayman22201: oh. I think I understand. It looks like you are pushing all cbs after the triggered one onto the next poll cycle.
Why defer them instead of just continuing the loop and running them in the current poll cycle? (I hope that sentence makes sense. lol)

dom96: Read the comment:

> This callback has returned with EAGAIN, so we don't need to call any other callbacks as they are all waiting for the same event on the same fd.

lib/pure/asyncdispatch.nim
dom96: It's not that it wants to be re-added, it's that it hasn't been run. It needs to be run when the FD becomes writeable again.

lib/pure/asyncdispatch.nim
rayman22201: Ok. Cool. Makes sense. I forgot that this function is for all callbacks waiting for the same fd.

Thanks for the clarification Dom!

PR #15028, opened 2020-07-21
fix several newline problems

narimiran:


ghost:
Is there any reason checks are with `doAssert` and not `assert`?

lib/pure/asyncftpclient.nim
```diff
@@ -146,7 +146,11 @@ proc send*(ftp: AsyncFtpClient, m: string): Future[TaintedString] {.async.} =
   ## Send a message to the server, and wait for a primary reply.
   ## ``\c\L`` is added for you.
   ##
+  ## You need to make sure that the message ``m`` doesn't contain any newline
+  ## characters. Failing to do so will raise ``AssertionDefect``.
```
Varriount, line 150: In non-release mode, right? Aren't assertions turned off for release mode?

ghost: @Varriount doAssert assertions are always enabled

lib/pure/httpclient.nim
```diff
@@ -245,7 +245,8 @@ proc contentLength*(response: Response | AsyncResponse): int =
   ##
   ## A ``ValueError`` exception will be raised if the value is not an integer.
   var contentLengthHeader = response.headers.getOrDefault("Content-Length")
-  return contentLengthHeader.parseInt()
+  result = contentLengthHeader.parseInt()
```
Araq, line 248: Use ``parseSaturatedNatural`` because that cannot raise an overflow exception.

narimiran: Yes, but that changes the documented behaviour:

> A ``ValueError`` exception will be raised if the value is not an integer.

```nim
import parseutils, strutils

var a = "123.45"

var b: int
discard parseSaturatedNatural(a, b)
echo b # ==> 123

var c = parseInt(a)
echo c # ==> ValueError
```

Araq: Ok...

PR #15030, opened 2020-07-21
writing to a location counts as "side effect"; implements https://github.com/nim-lang/RFCs/issues/234

Araq:


zah:
This is a breaking change though. A lot of existing `func` usages will stop compiling.

Araq:
> This is a breaking change though. A lot of existing func usages will stop compiling.

Well it's opt-in, but yes, see also https://github.com/nim-lang/RFCs/issues/234#issuecomment-661940311

PR #15034, opened 2020-07-21
fix #15033, add T: SomeNumber to `^` in stdlib/math

### LINKED ISSUE #15033: Using `^` from stdlib/math along with converters gives a match for types that aren't SomeNumber
The `^` proc in stdlib/math doesn't restrict `T` to just floats and ints, and this gives problems when implementing `^` for custom types involving converters. In my case, it's when both operands need to be converted that I stumble upon this.

### Example
```nim
import math

type
  A = object
    val*: int
  B = object
    value*: int

converter intToA(a: int): A =
  result.val = a

converter NotAToA(a: B): A =
  result.val = a.value
  
proc `^`(a: A, b: A): A =
  result.val = a.val ^ b.val

let a = A(val: 2)
let b = B(value: 2)
echo a ^ b # one conversion, works fine 
echo a ^ 2 # one conversions, works fine
echo b ^ 2 # two conversions, does not work
```

### Current Output
```
template/generic instantiation of `^` from here
/home/hugo/.choosenim/toolchains/nim-#devel/lib/pure/math.nim(981, 18) Error: type mismatch: got <int literal(1)> but expected 'B = object'
```

### Expected Output
```
(val: 4)
(val: 4)
(val: 4)
```

### Possible Solution

The line that gives the error is [this](https://github.com/nim-lang/Nim/blob/ddcbc1a58f82b9e3931e3ef5d1a8b4337ba50ccf/lib/pure/math.nim#L985) where it sets `result = 1`. But the proc's prototype is ```proc `^`*[T](x: T, y: Natural): T``` which causes problems when T isn't int- or float-like and gives the above error message. A fix that worked for me is to narrow down what T can be:
```nim
proc `^`*[T: SomeNumber](x: T, y: Natural): T
```
This makes sure that this proc is only called when it actually works with `T`. 

### Additional Information
* I don't know if `^` is meant to work with any type except `SomeNumber` but if there is just close this.
* For the curious-minded, in my real code I'm getting this error in, I'm trying to implement a SymbolicExpression library. I have an `ExpressionType` and then I do also have a `VariableType` for symbolic variables. When doing arithmetics with variables, they get converted automatically to `ExpressionType`. The problem arises when I'm using ints as well as variables. Both need to be converted, but the compiler finds a match for it so the conversion never happens.

```
$ nim -v
```

HugoGranstrom:
* fixes #15033 
* If this would break anyone's code, I'm really interested how you use `^` with a type other then ints and floats 😉

### DIFF
--- a/lib/pure/math.nim
+++ b/lib/pure/math.nim
@@ -958,7 +958,7 @@ proc sgn*[T: SomeNumber](x: T): int {.inline.} =
 {.pop.}
 {.pop.}
 
-proc `^`*[T](x: T, y: Natural): T =
+proc `^`*[T: SomeNumber](x: T, y: Natural): T =
   ## Computes ``x`` to the power ``y``.
   ##
   ## Exponent ``y`` must be non-negative, use


Araq:
Any type compatible with the ``1`` literal (converter?) and an overloaded multiplication operator works with it but fair enough. Still something to reconsider eventually.

HugoGranstrom:
Ahh yeah, that could potentially be a way to make it work for more types. Even more converters... xD They are both a blessing and a pain

PR #15042, opened 2020-07-22
[superseded] Increase buffer size for readlink to 1024

mbargull:
If Nim is installed in a long prefix on Linux, "nim c" was not able to
read its nim.cfg since it tried to read the link target of
/proc/self/exe which may be longer than the previous 256 long buffer.

Reproducer (notice the missing `Hint: used config file ...` in the second run):
<details open><summary>

```sh
podman run --rm -t gcc sh -c '
ver=1.2.4
curl -sL "https://nim-lang.org/download/nim-${ver}-linux_x64.tar.xz" | tar -xJ
echo import os > test.nim

for len in 246 247 ; do
    prefix="/$( yes a|tr -d \\n|head "-c${len}" )"
    (cd "nim-${ver}" && ./install.sh "${prefix}")
    nim_exe="${prefix}/nim/bin/nim"
    echo "${#nim_exe} ${nim_exe}"
    PATH="${prefix}/nim/bin:${PATH}" "${nim_exe}" c test.nim
done
'
```
</summary>

```
Nim build detected
copying files...
installation successful
259 /aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa/nim/bin/nim
Hint: used config file '/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa/nim/config/nim.cfg' [Conf]
Hint: system [Processing]
Hint: widestrs [Processing]
Hint: io [Processing]
Hint: test [Processing]
Hint: os [Processing]
Hint: strutils [Processing]
Hint: parseutils [Processing]
Hint: math [Processing]
Hint: bitops [Processing]
Hint: macros [Processing]
Hint: algorithm [Processing]
Hint: unicode [Processing]
Hint: pathnorm [Processing]
Hint: osseps [Processing]
Hint: posix [Processing]
Hint: times [Processing]
Hint: options [Processing]
Hint: typetraits [Processing]
/test.nim(1, 8) Warning: imported and not used: 'os' [UnusedImport]
CC: stdlib_system.nim
CC: stdlib_times.nim
CC: stdlib_os.nim
CC: test.nim
Hint:  [Link]
Hint: 36222 LOC; 1.328 sec; 47.211MiB peakmem; Debug build; proj: /test.nim; out: /test [SuccessX]
Nim build detected
copying files...
installation successful
260 /aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa/nim/bin/nim
Hint: system [Processing]
Hint: widestrs [Processing]
Hint: io [Processing]
Hint: test [Processing]
/test.nim(1, 8) Error: cannot open file: os
```
</details>

### DIFF
--- a/lib/pure/os.nim
+++ b/lib/pure/os.nim
@@ -1213,11 +1213,11 @@ proc findExe*(exe: string, followSymlinks: bool = true;
         when not defined(windows):
           while followSymlinks: # doubles as if here
             if x.checkSymlink:
-              var r = newString(256)
-              var len = readlink(x, r, 256)
+              var r = newString(1024)
+              var len = readlink(x, r, 1024)
               if len < 0:
                 raiseOSError(osLastError(), exe)
-              if len > 256:
+              if len > 1024:
                 r = newString(len+1)
                 len = readlink(x, r, len)
               setLen(r, len)
@@ -2555,11 +2555,11 @@ proc expandSymlink*(symlinkPath: string): string {.noWeirdTarget.} =
   when defined(windows):
     result = symlinkPath
   else:
-    result = newString(256)
-    var len = readlink(symlinkPath, result, 256)
+    result = newString(1024)
+    var len = readlink(symlinkPath, result, 1024)
     if len < 0:
       raiseOSError(osLastError(), symlinkPath)
-    if len > 256:
+    if len > 1024:
       result = newString(len+1)
       len = readlink(symlinkPath, result, len)
     setLen(result, len)
@@ -2850,9 +2850,9 @@ when not weirdTarget and (defined(freebsd) or defined(dragonfly)):
 
 when not weirdTarget and (defined(linux) or defined(solaris) or defined(bsd) or defined(aix)):
   proc getApplAux(procPath: string): string =
-    result = newString(256)
-    var len = readlink(procPath, result, 256)
-    if len > 256:
+    result = newString(1024)
+    var len = readlink(procPath, result, 1024)
+    if len > 1024:
       result = newString(len+1)
       len = readlink(procPath, result, len)
     setLen(result, len)


mbargull:
I used `1024` here because it's the same size as used in https://github.com/nim-lang/Nim/pull/4937.
I would suggest to replace all those hard-coded numbers with `MAX_PATH` or a new constant or whatever seems appropriate, though. I didn't do that because I don't know what is the desired form here.

Please do feel free to edit this branch as you see fit!

ghost:
I'm not sure, but isn't the code around the lines you changed actually handles the overflow case already? Maybe it just doesn't work correctly

mbargull:
It is not correct, at least on Linux. I get:
```
$ man 3 readlink | grep -A1 RETURN
RETURN VALUE
       Upon successful completion, these functions shall return the count of bytes placed in the buffer. Otherwise, these functions shall return a value of −1, leave the buffer unchanged, and set errno to indicate the error.
```
Note `count of bytes placed in the buffer` would mean at most `256`/`1024`, depending on the input.

I have to admit, I was confused by the `if len > 256:` because it didn't make much sense to me ~to not just use `257` then~. But now I get what the intend was :).

I don't know if on other systems, e.g., \*BSDs the behavior is different, though.

mbargull:
Oh, and of course I tested this patch and it is working for me with longer paths.

mbargull:
>  not just use `257` then.

Scratch that little part of what I wrote. That doesn't make sense because the check is testing `> 256` which just won't happen.

mbargull:
If you fear it is not a good idea to use the bigger buffer from the get-go, we could also change this to
```nim
var len = readlink(..., buf, 256)
if len >= 256:  # keeping the ">" part in case non-Linux systems return bigger numbers (which IDK)
    var buf_len = max(len, BIG_NUM)
    buf = newString(len+1)
    var len = readlink(..., buf, buf_len)
```
I'm not sure when that would be necessary, though, and believe a bigger buffer would make sense, generally.

----
In any case: The code silently does the wrong thing, i.e., returns truncated paths, if the paths are larger than `256`/`1024`/`BIG_NUM`. Ideally, some kind of error handling/reporting should happen. I am no Nim developer, though and don't know other details of the code, so I cannot address that issue for you.

mbargull:
> Note `count of bytes placed in the buffer` would mean at most `256`/`1024`, depending on the input.

Confirmed the issue by compiling the following program
```c
#include <unistd.h>
#include <stdio.h>

int main(void) {
    char buf[256+1];
    ssize_t len = readlink("/proc/self/exe", buf, sizeof(buf)-1);
    if (len < 0) {
        return 1;
    }
    buf[len] = '\0';
    printf("%d\n%s\n", len, buf);
    return 0;
}
```
and placing it in a long path: This will print `256` and a truncated path.

Araq:
Thanks but isn't the real fix ``>=`` instead of ``>`` ?

mbargull:
> Thanks but isn't the real fix `>=` instead of `>` ?

No, see my comment above. On Linux `readlink` returns `256` for a path that is, e.g., `500` characters long.
Meaning the `>=` only makes sense IF (which I don't know) non-Linux systems are able to return larger numbers than the buffer size.

Araq:
Well ``>= 256`` is true for ``== 256`` too.

mbargull:
Please elaborate what you mean with "real fix `>=` instead of `>`".
From your short comments it is not clear what you are referring to.

lib/pure/os.nim
```diff
@@ -1213,11 +1213,11 @@ proc findExe*(exe: string, followSymlinks: bool = true;
...
           while followSymlinks: # doubles as if here
             if x.checkSymlink:
-              var r = newString(256)
-              var len = readlink(x, r, 256)
+              var r = newString(1024)
+              var len = readlink(x, r, 1024)
               if len < 0:
                 raiseOSError(osLastError(), exe)
-              if len > 256:
+              if len > 1024:
```
Varriount, line 1220: Shouldn't this be greater than or equal? 

From the [manual page](https://man7.org/linux/man-pages/man2/readlink.2.html)
> On success, these calls return the number of bytes placed in buf.
> (If the returned value equals bufsiz, then truncation may have occurred.)

Which means if `len == 1024` (or 256, or whatever), then you have to deal with possible truncation.

mbargull: > Which means if `len == 1024` (or 256, or whatever), then you have to deal with possible truncation.

Yes, I already stated that. Please see my other comments on the PR.
If you say `> 1024` or `>= 1024` doesn't matter much. If you write `>=` then we get a truncation to `1025` characters with this current code which is no improvement.

From one of my comments above:
> In any case: The code silently does the wrong thing, i.e., returns truncated paths, if the paths are larger than 256/1024/BIG_NUM. Ideally, some kind of error handling/reporting should happen. I am no Nim developer, though and don't know other details of the code, so I cannot address that issue for you.

mbargull: To further clarify:
The current
```nim
                r = newString(len+1)
                len = readlink(x, r, len)
```
does not make sense if there is a truncation of multiple characters.
This code should be replaced by error handling/reporting.

Varriount: Ah! I see. Thanks for your patience. It was my fault for not reading the rest of the original implementation.

Then a proper implementation would be roughly as follows, I think?
```
var buffer = newString(1024)
while true:
  var res = readlink(x, buffer, len(buffer))
  if res < 0:
    raiseOSError(osLastError()
  elif res == len(buffer):
    setLen(buffer, len(buffer) * 2)
    continue
  else:
    setLen(buffer, res)
    break
```

The loop is because it's possible (though highly unlikely) for the link to change length between calls.

mbargull:
Fixed in gh-15908.

PR #15046, opened 2020-07-23
support openssl ALPN extension

superisaac:
This pr is an implement of ALPN extension, It does't change currenty module too much.
To make alpn work the client needs to call `setAlpnProtocols()` and the server need to call `setAlpnSelectCallback()`, refer to tests/stdlib/thttpclient_ssl_alpn.nim


lib/pure/net.nim
```diff
@@ -728,6 +728,15 @@ when defineSsl:
     if SSL_set_fd(socket.sslHandle, socket.fd) != 1:
       raiseSSLError()
 
+  proc getAlpnProtocol*(socket: Socket): string =
+    ## Returns the negotiaged ALPN protocols.
+    ## If the ALPN extension is not available or no ALPN candidate
+    ## provided by clients then returns empty string.
+    if not socket.isSsl:
```
juancarlospaco: The `result` is autoinit to `""` then

```nim
if socket.isSsl: result = socket.sslHandle.getAlpnProtocol()
```

lib/pure/asyncnet.nim
```diff
@@ -772,6 +772,15 @@ when defineSsl:
     of handshakeAsServer:
       sslSetAcceptState(socket.sslHandle)
 
+  proc getAlpnProtocol*(socket: AsyncSocket): string =
+    ## Returns the negotiaged ALPN protocols.
+    ## If the ALPN extension is not available or no ALPN candidate
+    ## provided by clients then returns empty string.
+    if not socket.isSsl:
```
juancarlospaco: Same.

lib/pure/net.nim
superisaac: thank you for your comment, the fix commit is pushed as a405eec1

PR #2078, opened 2015-02-06
Add hcode.  Re-factor rawGet.  Fix infinite loop.

c-blake:
Replace state enum with a cached hash code which has the same memory overhead
and locality as the enum, but can really speed things up with non-integer-like
keys (keys for which either hash() or == take more than couple cycles, or where
the key data is "indirect" and might incur another cache miss).  To function as
both empty/filled state and a hash code cache, it only needs to be ensured that
hash codes are non-zero for any real key.  That is done at the one place in the
whole file hash() is called.  Keep convention clear via isFilled() & isEmpty().
An isDeleted state will no longer be necessary as per below excl/inf loop fix.

Since some use sites know hc and some do not, re-factor rawGet into two forms -
one with known hash code and one with an unknown HC that returns it. Both forms
still return <0 on missing, but returns the much more informative "-1 - index".
That return can be quickly inverted by -1 - result to recover the index where
insert should happen, provided no modifications are made to the table in the
meantime.  This protocol retains the prior <0 interface and also makes it easy
to avoid unnecessary duplicate search work in procs like containsOrInclImpl
(which formerly searched in the initial get and AGAIN in rawInsert).  Strip the
searching part out of rawInsert to "make it even more raw".  swap(s.data, n) a
bit earlier so rawGet and rawGetKnownHC can have similar parameter lists and
integrate well with rawInsert/code sharing between Set and OrderedSet impls.

This PR also fixes infinite looping upon too many deletes. [ The deleted state
(aka "tombstone") approach is vulnerable to the table filling up with deleted
items which forces giant scans for missing keys which could be anywhere.  In
the version prior to this PR, table wraparound wasn't even detected yielding
infinite loops. ] This PR changes excl() from marking slots as deleted to Knuth
algo 6.4R, "local/incremental moveback rehashing" - adapted from Knuth's h->h-1
to the cache-friendlier h->h+1 probe sequence and adapted from "gotos" to a new
doWhile template.  This method restores the table to a state that would have
resulted from pure inserts (in some order).  Update nextTry accordingly.  Since
linear probing can degrade a little faster, 50% rather than 66% may be a better
default growth threshold, but users should be able to adjust threshold anyway.

Old unit tests all pass.  More extensive testing in this module is probably
warranted before taking similar enhancements over to collections.tables.

PR #15047, opened 2020-07-23
deprecate tables.add

narimiran:


c-blake:
Two things.

First removing `add` does not address people like @disruptek that use & like the functionality.  Sometimes better documentation (ALL CAPS?) or [as I mentioned](https://github.com/nim-lang/Nim/issues/13499#issuecomment-662966336) advertisement via parameter names can complement such documentation.  This is especially relevant if the response to this first point is to add `MultiEverything`.

Second, you did not deprecate/remove `allValues` which should surely be paired with `add` since if there is no `add` it will make no sense to keep it.

I should say, I don't care for myself.  I just wrote my own library for this kinda stuff, and if the stdlib starts changing its post-1.0 APIs then I might just collapse sets & tables a la `critbits` with its "maybe `void`" value type to reduce the "type footprint/cognitive load" even more.

Araq:
> if the stdlib starts changing its post-1.0 APIs then I might just collapse sets & tables a la critbits with its "maybe void" value type to reduce the "type footprint/cognitive load" even more.

Well tables.add is the only controversial collections API afaict and there is no need to change anything else. However , renaming  ``add``  to ``addAnother`` or similar is also a good option.

c-blake:
I guess that "teachable moment" thing isn't working out, eh? ;-)  It's hard.  Any new prog lang is, in some sense, one ***gigantic*** teachable moment, but it's so hard to know how to "scope" that/what battles to fight. ;-)  Anyway, the only real controversy seems the failed teaching..

FWIW, I like `addAnother`.  `append` is another option.  The latter rather explicitly goes against the usual interpretation of `doc/apis.txt`.  `addDup` might be a third.  Calling it a "multimap" just in the overall/summary documentation could also help.  C++ STL names are widely disseminated..but not to everyone.

What about staying `add` but having a parameter `atFront` to whatever the new `proc` name is to emphasize the "seq nested within a collision cluster" aspect?  And maybe a paired parameter for `del`?  It makes literally zero sense to refer to front/back if there is only 1 thing.  So, the existence of those params alone could be enough to de-confuse, as well as make the API more useful.  I'm not sure if we can keep resizing efficient while preserving the order across the high index of the table.  Maybe.  That's all work involving careful thought.

hashbackup:
Isn't using a [A, seq[B]] table equivalent to an [A, B] table with duplicates, other than the API?  It seems like it is, and lets the user manage the sequence of duplicates rather than trying to do it in hash table code that in most other languages, does not allow duplicate keys.

Another consideration: when you want to add new kinds of hash tables, like Compact, there probably aren't going to be a lot of implementations that allow duplicate keys.

c-blake:
`Table[A,seq[B]]` is different from a memory management perspective.  Dunno how important it is.  People can do `Table[A,seq[B]]` right now and ignore `proc add`/`iterator allValues`.

It might reduce options a little, but `adix/olset.OLset` also supports `proc add` and `iterator allItems` using the direct-indexing/Python style ordering.  As long as good options exist, there isn't necessarily a problem.  There are always bad/worse algos.

disruptek:
I could support `append`, as it seems to capture the right meaning.  But why not just add `replace` instead?  `add` is more idiomatic as-is and no less correct for this change.

zevv:
What about a totally different idea: let `Tables` be as it is, and introduce a new `Map` type with a pretty minimal API supporting only Map-operations. `Map` gets advertized as the default container type to use for 99.9% of the use cases. It can be built very thin on top of the current Tables implementation, but also allows optimizations or a different optimized implementation in the future. 

Advantages:

- `Map` is the common name and used in a lot of other languages.  Every now and then someone stumbles into #nim if Nim has maps. "Yes, we have that, it's called tables"
- No problems anymore with `.add`, it's not available in `Map`
- `Tables` API is way too complex IMHO, I count 113 procs and 31 iterators on the type. Making `Map` would allow for a fresh start with the bare essentials.

Araq:
I don't like ``Map``, it's also used for "real" maps and whenever I program a game names like ``std::vector`` (which isn't a 2d/3d vector at all!) or ``map`` (which has nothing to do with the mini-map of my game) are annoying.

disruptek:
Naming is hard, as you found with `add`.  :+1: for a `Map`-like construct which solves all these concerns.

c-blake:
`Tab`?  `HashTab`?  Very like `HashSet`.  ;-)  `Index` is another word but people often think of only the simple `x[i]` case where `i` means index.

If there were going to be a whole new reduced thing that it should merge sets & tables capabilities the way `critbits` does, although I guess that de-reduces it. :-/  It's hard to please everyone.

I'm mostly just bike shedding here.  I'm not sure the status quo is really as bad as others seem to think.  At *some* level you must ask people to leave behind *some* preconceptions from other prog.langs/runtimes. Otherwise you aren't a different language..You're just a compiler for the other ones.  Better documentation and/or an `allowDups=true|false` parameter to init may be plenty.  I mean, even people who don't read docs will see some auto-complete or proc signature with a param name, right?  We could do an exported module `var tables.allowDupsDefault` that people could assign so they can still use `var x: Table[string,int]` for whichever. Different client code could wind up fighting wars over the value of that default, though.

Decl syntax doesn't take params outside the generic params.  So, our options are limited.  I doubt a new generic param, `Table[K, V, allowDups=false]` would be popular either, but it bears mentioning.  There might be a glitch or two needing to be ironed out, but maybe not.

Or no change at all.  I'm not sure it's current situation is worse than all these alternatives at this point.

Varriount (commented):
@narimiran: Was this talked over with anyone?

hashbackup:
Out of curiosity, since I'm not that familiar with this stuff, I wrote two test programs.  In both programs, 100K items are inserted into a hash table, and each item has 100 associated integers.  I disabled the delete loop because the existing OrderedTable would have taken forever to delete so many items.

The first test creates 100 associated values for 100K items in an ordered hash table.  (A slight bug: it actually creates 101 values for each key - oh well, bugs are everywhere!):
```
import tables
import times

const
  els = 100_000
  ndups = 100

var
  ot = initOrderedTable[int, int](els * ndups)

var t0 = cpuTime()
for i in 0 ..< els:
  ot[i] = i
  for j in 0 ..< ndups:
    ot.add i, j
echo "len = ", len(ot), ", seconds to create table: ", cpuTime() - t0

when false:
  t0 = cpuTime()
  for i in 0 ..< els:
    ot.del(i)
  echo "Seconds to delete table: ", cpuTime() - t0
  echo "len = ", len(ot)

ms:nx jim$ /usr/bin/time -l ./duptest
len = 10100000, seconds to create table: 1.350274
        1.63 real         1.45 user         0.16 sys
 537444352  maximum resident set size
```
The second test uses an ordered hash table but no duplicate keys.  It stores the 100 items of related data in a seq instead:
```
import tables
import times

const
  els = 100_000
  ndups = 100

var
  ot = initOrderedTable[int, seq[int]](els)

var t0 = cpuTime()
for i in 0 ..< els:
  var s = newSeq[int](ndups)
  for j in 0 ..< ndups:
    s.add(j)
  ot[i] = s
echo "len = ", len(ot), ", seconds to create table: ", cpuTime() - t0

when false:
  t0 = cpuTime()
  for i in 0 ..< els:
    ot.del(i)
  echo "Seconds to delete table: ", cpuTime() - t0
  echo "len = ", len(ot)

ms:nx jim$ /usr/bin/time -l ./duptestseq
len = 100000, seconds to create table: 0.170771
        0.21 real         0.14 user         0.06 sys
 215867392  maximum resident set size
```
I can't see why anyone would prefer to use dups in a hash table vs a sequence.  The sequence version runs 7.5x faster and uses half the memory of the hash table with duplicates.

hashbackup:
A more extensive test with regular Tables, including iteration and deletes:
```
import tables
import times

const
  els = 100_000
  ndups = 100

var
  ot = initTable[int, int](els * ndups)

var t0 = cpuTime()
for i in 0 ..< els:
  for j in 0 ..< ndups:
    ot.add i, j
echo "len = ", len(ot), ", seconds to create table: ", cpuTime() - t0

var sum = 0
for i in 0 ..< els:
  for j in ot.allValues(i):
    sum += j
echo "sum = ", sum, ", seconds to iterate table: ", cpuTime() - t0

t0 = cpuTime()
for i in 0 ..< els:
  for j in 0 ..< ndups:
    ot.del(i)
echo "Seconds to delete table: ", cpuTime() - t0
echo "len = ", len(ot)

ms:nx jim$ /usr/bin/time -l ./duptest
len = 10000000, seconds to create table: 1.204947
sum = 495000000, seconds to iterate table: 1.301055
Seconds to delete table: 5.813065
len = 0
        7.33 real         7.18 user         0.13 sys
 403226624  maximum resident set size
```
Same thing with sequences:
```
import tables
import times

const
  els = 100_000
  ndups = 100

var
  ot = initTable[int, seq[int]](els)

var t0 = cpuTime()
for i in 0 ..< els:
  var s = newSeq[int](ndups)
  for j in 0 ..< ndups:
    s.add(j)
  ot[i] = s
echo "len = ", len(ot), ", seconds to create table: ", cpuTime() - t0

var sum = 0
for i in 0 ..< els:
  for j in ot[i]:
    sum += j
echo "sum = ", sum, ", seconds to iterate table: ", cpuTime() - t0

t0 = cpuTime()
for i in 0 ..< els:
  ot.del(i)
echo "Seconds to delete table: ", cpuTime() - t0
echo "len = ", len(ot)

ms:nx jim$ /usr/bin/time -l ./duptestseq
len = 100000, seconds to create table: 0.169384
sum = 495000000, seconds to iterate table: 0.21271
Seconds to delete table: 0.015015
len = 0
        0.25 real         0.17 user         0.06 sys
 213774336  maximum resident set size
```
Now the test with table sequences is 29x faster than the test with table dups, and the test with dups uses 89% more memory - nearly twice.   The test with dups is 28 lines of code, the sequence test is 29.

narimiran:
> @narimiran: Was this talked over with anyone?

Yes, with @Araq himself :)

c-blake:
About those microbenchmarks - I realize you are probably just doing 100k times 100 to get times large enough to have low measurement noise or whatever, but there is a big unfairness in your tests from the particular scales you cross. The "random access" part for the table with dups needs  > L3 cache amounts of space (10M*dozen or so) while it probably fits in cache for the `seq[]` variant (100k*16 or so).  So, what your build-time tests really reflect is relative latency of DIMMs to the CPU L3 cache where, of course, the CPU cache wins.

Also, 100 is really large, more or less out of the intended range of usage.  Things get slow in the number of duplicates fast.  100 / log_2(n) in this case is > 4.  So, you've also made the average case for every element in the table much more than the expected worst case element for a table.  

>I can't see why anyone would prefer to use dups in a hash table vs a sequence.

Look, from a time performance point of view, I think this is one of those situations where benchmarks are simply less helpful than thinking about it from the type definitions, memory layout, and the algorithms.  In light of that kind of thinking, both approaches can obviously win different performance contests.  E.g., for truly giant tables, an `allValues` iteration (maybe not in a tight, prefetched loop) should have just one DIMM hit while the seq variant has two - one for the hash lookup and one for the seq data.  From a space performance, one big area padded to a po2 could be bigger (especially for big value generic types), but it is also simpler than thousands of small seqs of various lengths. That kind of fragmentation may also matter much less in a tight loop microbenchmark than a calculation doing "real" work.

So, in at least time & space dimensions one algo can be better sometimes and the other can be better other times as explained above.  I do not think the `seq` way is "always better".  I think that follows from just first principles not benchmarks, though such principles could inform better example code or sophisticated benchmarks.  Right now, people can do both/either with just one type as fits their situation/needs/expectations/hopes/dreams (which can sometimes be wrong, of course).  The idea afoot is to take one away or maybe bifurcate types.

----

This feature definitely has some "sharp edges" performance-wise. These edges are unarguably more subtle and also less well known than `seq`.  The argument made by those benchmarks is really more "this is only for 'expert' users" which may be accurate depending on 'expert'.  Most people seem to be literally unaware of its existence, never mind its appropriate usage.  Maybe the *only* users are "expert" users and that's ok?

hashbackup:
You make a good point and of course, I have to test it. :-)  Here is the same test with 10M keys and 4 values per key, first with table duplicates, then with sequences.
```
ms:nx jim$ /usr/bin/time -l ./duptestbig
len = 40000000, seconds to create table: 2.199015
sum = 60000000, seconds to iterate table: 4.192819999999999
Seconds to delete table: 3.040139
len = 0
        8.05 real         7.52 user         0.52 sys
1611202560  maximum resident set size

ms:nx jim$ /usr/bin/time -l ./duptestseqbig
len = 10000000, seconds to create table: 3.796547
sum = 60000000, seconds to iterate table: 5.451102
Seconds to delete table: 1.423804
len = 0
        7.27 real         6.74 user         0.52 sys
1356681216  maximum resident set size
```
For cases where items aren't deleted, I also ran the tests with no deletes:
```
ms:nx jim$ /usr/bin/time -l ./duptestbig
len = 40000000, seconds to create table: 2.217596
sum = 60000000, seconds to iterate table: 4.346052999999999
        5.18 real         4.64 user         0.53 sys
1611182080  maximum resident set size

ms:nx jim$ /usr/bin/time -l ./duptestseqbig
len = 10000000, seconds to create table: 3.575126
sum = 60000000, seconds to iterate table: 5.129767
        6.11 real         5.61 user         0.48 sys
1356701696  maximum resident set size
```
So you are right, the number of values does matter: for larger number of values per key, sequences are a huge win @ 29x faster for this test.  For a larger number of keys with only a few values per key, table dups are faster ... by 15%.

My personal opinion, for what it's worth, is that the problems associated with table dups, including:
* user confusion
* likely small number of users
* inconsistent operation between table types (delete all dups vs one dup for example)
* need to add more options to table operations, eg, should toTable remove or include dups
* complexity in the table code
* hard to find new table implementations that support duplicates

are not worth a 15% performance increase for a specialized, expert use case, and table duplicates should be deprecated.  For the specialized, expert user who needs the 15% performance, I'm guessing they would have no problem either copying & fine-tuning the existing table code for their use case, or using another library that specializes in multisets / multitables - perhaps adix.

c-blake:
It's not always just 15%, though..just in your one test.  It could be 2x as in my example.  Doing 4 per entry is like only 4 hash collisions for every key..some might even call that on the "high side" of what they expect.

Beats me whether the stdlib should target expert users/best in class/most flexible everything or that should be delegated to the ecosystem or what the boundaries should be.  Those are hard questions.  Support for dups is unarguably incomplete/inconsistent.  We could complete it on whatever timeline is convenient or deprecate it.  It has been around forever, though, and through 1.0.  People who seem to dislike it seem to mostly due so out of unfamiliarity/surprise.  Otherwise it doesn't really interfere with their activity. (EDIT: of course it does impact table-maintainers...but I'm not sure how much that should matter.)

dom96:
As someone that had no idea this was a thing (apparently the ``Tables`` name is supposed to have told me that duplicate keys are possible...) and implemented things *in the stdlib* as ``Table[string, seq[T]]`` all I can really say is I approve this PR.

The fact that sequences in Tables are also apparently faster puts the nail in the coffin and I agree with many of @hashbackup's points above.

disruptek:
It's not just a performance issue.  It's a semantic issue.  Sometimes, these are the semantics you want.

For example, I use a Table in a CPS dispatcher to associate semaphores with the continuations which await their signal.  The Table should be able to handle hundreds of millions of continuations.  This is a 1:N association; multiple continuations may await singular semaphores.  Of course, my insertion algorithm should not need to slow down in this case.

I think deprecating `add` is an objectively poor solution, especially when other objectively equivalent solutions exist which have none of the drawbacks.

disruptek:
Here's another proposal:

Deprecate `Table` and replace it with `Map` (or whatever you want to name it).

c-blake:
I think the plural tables refers to plural types/procs not plural keys, but maybe you're just being sarcastic.

Regarding your implementations, as I mentioned, they are ***not always faster***. Your uses of `Table[K,seq[V]]` *may well be appropriate*.  Referring to shallow performance conclusions based on just one microbenchmark concept as "The fact" even with an "apparently" qualification with nails in coffins as if it's all so "conclusive" seems simply wrong.

I already [mentioned](https://github.com/nim-lang/Nim/pull/2078#issuecomment-73744634) a probable mis-use by you of `add` in the `fsmonitor.nim` stuff waaay back when (and re-linked to it in the affiliated thread).  That possible not reading the `proc add` documentation mis-use is the *only* way this capability confusing users has much user downside, and renaming it append has been discussed here.

Otherwise keeping dups just keeps 1.0 backward compat, lives & let lives side-by-side with the regular, with no performance implication for those who don't use it, just being a performance ***option*** for those who do.  (which is ***not*** me, actually..I just defend those who use/like it such as @disruptek).

I am mostly just responding to what I feel are weak "What? Tables can do *that*??!?! How come no one else told me in my dev experience?" arguments by people.  The very same people *could* have also taken an alternate lesson: "Oh, cool!  Tables can do that, too. Ok.  Now I know!".  I feel like the latter is the response of greater humility and better personal evolution.  The "shocked *me* - remove, remove!" alternative assumes all/most/many will be just as disappointed to learn something new as "you" were (for any given value of "you" - not ad hominem at all..these "surprise"-based arguments just have this structure which I think embeds a certain superior attitude, perhaps to be discouraged).

hashbackup:
> For example, I use a Table in a CPS dispatcher to associate semaphores with the continuations which await their signal. The Table should be able to handle hundreds of millions of continuations. This is a 1:N association; multiple continuations may await singular semaphores. Of course, my insertion algorithm should not need to slow down in this case.

I'm interested since you have a real application for table dups.  From the description, the key is the semaphore and then multiple continuations are added to that one key.  And you have hundreds of millions of continuations?  For how many semaphores?  I'm asking because it lets us know on average how many values (continuations) are being stored with a single key.

One thing to keep in mind about performance of dups is that when inserting a table duplicate, the table code has to access every existing duplicate first.  Now this is usually fast because they are packed more or less together in the table, ie, they are sequential in memory.  But if you have 1000 duplicates for a key and add a new one, you still have to scan past all 1000 old ones first.

With a sequence, you append to the end without accessing any existing items in the sequence and it doesn't matter if you have 1M duplicates.

I'm also asking because you said:
> Of course, my insertion algorithm should not need to slow down in this case.

And it wouldn't for a long sequence, but definitely will for a long run of hash table dups.  The longer the run of dups, the slower your insert will be.

disruptek:
There's no concept of an average because this is a general-purpose dispatcher and I'm not aware of anyone using it.  The point I'm trying to make is that taking away my `add` necessarily changes both my code and the performance of software in the wild that may use that code.

c-blake:
For what it's worth, `proc add*[TKey, TValue](t: var TTable[TKey, TValue], key: TKey, val: TValue)` goes back to the initial August, 2008 Nim release.

juancarlospaco:
Maybe a simplified API on a separate small new module that stand on top of `tables` with more map/dictionary look and feel,
seems to work fine for `posix` / `posix_utils`, `json` / `jsonutils`, `dom` / `dom_extensions`, no breaks because is a new module.

Araq:
> goes back to the initial August, 2008 Nim release.

Yes, it's my fault. Let's deprecate it.

c-blake:
As I mentioned already right at the top here and reiterated by pairing them up, you probably also want to deprecate `iterator allValues`.

I guess I won't bother fixing `Table.==`.  You probably also want to close https://github.com/nim-lang/Nim/issues/13499 since there is no issue if duplicates cannot be `add`ed.

A path of reduced choice due to only ignorance and conformity is always a little sad, but I'll keep this feature in the https://github.com/c-blake/adix tables if anyone wants it.  Those tables have better sizing controls for "experts" anyway.  So, an assumption of higher baseline user knowledge of space-time/embedding-indirection trade-off knowledge can probably be assumed there.

hashbackup:
Also #13500

Araq:
> A path of reduced choice due to only ignorance and conformity is always a little sad, but I'll keep this feature in the https://github.com/c-blake/adix tables if anyone wants it.

Nice, thank you!

c-blake:
Also, since people seemed to think benchmarks were relevant here, [in this one](https://forum.nim-lang.org/t/5917#40869) the duplicate feature seems natural and valuable.  Of course, any one example is just that..one example.  Mostly just cross-referencing here for the curious-minded.

PR #15065, opened 2020-07-25
Fix edge-case from #15064 

### LINKED ISSUE #15064: [strscans] scanf doesn't match a single character with $+ if it's the end of the string
Self explanatory really.

### Example
```nim
import strscans

var nick, msg: string
if scanf("<abcd> a", "<$+> $+", nick, msg):
  echo nick
  echo msg
else:
  echo "failed"
```

### Current Output
```
failed
```

### Expected Output
```
abcd
o
```

```
$ nim -v
Compiled at 2020-07-23

git hash: 18a70a75638ded6ff23b2a9a57edfc7dd09bb5b1
```

ghost:
I'm not sure if this is 100% correct, but strscans tests seem to pass.
Fixes #15064

### DIFF
--- a/lib/pure/strscans.nim
+++ b/lib/pure/strscans.nim
@@ -396,7 +396,11 @@ macro scanf*(input: string; pattern: static[string]; results: varargs[typed]): b
           var resLen = genSym(nskLet, "resLen")
           conds.add newLetStmt(resLen, newCall(bindSym"parseUntil", inp,
               results[i], newLit(token), idx))
-          conds.add newCall(bindSym"!=", resLen, newLit min)
+          # Edge case for when the length of the matched token is 1
+          if token != "":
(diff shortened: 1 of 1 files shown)


ghost:
timezones failure seems unrelated - fails on devel too - https://github.com/nim-lang/Nim/runs/909341967

ghost:
Aww, seems to be incorrect:
```nim
import strscans

var nick, msg: string
if scanf("<abcd> ", "<$+> $+", nick, msg):
  echo nick
```
will work after this PR which is not correct behaviour

narimiran:
> will work after this PR which is not correct behaviour

This was incorrect even before this PR.

But I think I have a simpler/elegant solution for both problems, see my PR: https://github.com/nim-lang/Nim/pull/15223

narimiran:
#15223 was merged. Closing this.

Issue #9867, opened 2018-12-05
Long-lived application occasionally dies with SIG_PIPE when calling httpclient.request


niv:
Our serverside application, currently on 0.18, does http the occasional http request to the Google APIs. Sometimes, I think, google closes the connection on us and the application dies with SIG_PIPE:

```
...
keyprovision_google.nim(66) getAccessToken
keyprovision_google.nim(30) requestJson
keyprovision_google.nim(71) :anonymous
httpclient.nim(1196)     request
httpclient.nim(1181)     requestAux
httpclient.nim(1031)     parseResponse
httpclient.nim(883)      close
net.nim(866)             close
SIGPIPE: Pipe closed.
```

Of course, the http request() is wrapped into try/except; that does not catch the signal however.

This is running on nim 0.18 (I can't try 0.19 yet, porting it is significant effort due to libraries used), and in a docker container on Linux.

I found this thread: https://forum.nim-lang.org/t/4417, but that didn't really help. The httpclient is already configured to use a reasonable timeout (3s).

I'm fairly certain this is because Google is closing the kept-alive connection in the background; but I'm hesitant to add `Connection: Close` (or otherwise close the connection on our end) unless that's the only way to solve this.

zevv:
What is your platform? On Linux, I see that all socket sendto()s are done with the MSG_NOSIGNAL flag, instructing the kernel not to generate a SIGPIPE when the peer has closed the socket.

dom96:
huh, it's SIGPIPE'ing on a ``close``. Interesting, it appears to be OpenSSL's fault: https://github.com/nim-lang/Nim/blob/v0.18.0/lib/pure/net.nim#L866.

Anyway, please follow the advice here: https://stackoverflow.com/a/108192/492186. As a workaround you can ignore all SIGPIPEs in your program, as a proper fix I guess we need to set some OpenSSL flag or something.

niv:
Heya dom96, I tried that, like so:

```nim
from posix import signal, SIG_PIPE, SIG_IGN
signal(SIG_PIPE, SIG_IGN)
```

It seemed to have worked for a good while, but today it fell over with this (one exception per line):

```
Broken pipe Additional info: IO error has occurred in the BIO layer; will retry
not isClosed(socket) Cannot `send` on a closed socket; will retry
Broken pipe Additional info: IO error has occurred in the BIO layer; not retrying anymore
```

[...]

PR #15066, opened 2020-07-25
asyncnet, net: don't attempt SSL_shutdown if a fatal error occurred

alaviss:
Per TLS standard and SSL_shutdown(3ssl). This should prevent errors
coming from a close() after a bad event (ie. the other end of the pipe
is closed before shutdown can be negotiated).

Ref #9867

### DIFF
--- a/lib/pure/net.nim
+++ b/lib/pure/net.nim
@@ -133,6 +133,7 @@ type
       sslNoHandshake: bool # True if needs handshake.
       sslHasPeekChar: bool
       sslPeekChar: char
+      sslNoShutdown: bool # True if shutdown shouldn't be done.
     lastError: OSErrorCode ## stores the last error on this socket
     domain: Domain
     sockType: SockType
@@ -722,6 +723,7 @@ when defineSsl:
     socket.sslHandle = SSL_new(socket.sslContext.context)
     socket.sslNoHandshake = false
     socket.sslHasPeekChar = false
+    socket.sslNoShutdown = false
     if socket.sslHandle == nil:
       raiseSSLError()
 
@@ -844,6 +846,8 @@ proc socketError*(socket: Socket, err: int = -1, async = false,
         of SSL_ERROR_WANT_X509_LOOKUP:
           raiseSSLError("Function for x509 lookup has been called.")
         of SSL_ERROR_SYSCALL:
+          # SSL shutdown must not be done if a fatal error occurred.
+          socket.sslNoShutdown = true
           var errStr = "IO error has occurred "
           let sslErr = ERR_peek_last_error()
           if sslErr == 0 and err == 0:
@@ -856,6 +860,8 @@ proc socketError*(socket: Socket, err: int = -1, async = false,
           let osErr = osLastError()
           raiseOSError(osErr, errStr)
         of SSL_ERROR_SSL:
+          # SSL shutdown must not be done if a fatal error occurred.
+          socket.sslNoShutdown = true
           raiseSSLError()
         else: raiseSSLError("Unknown Error")
 
@@ -1026,7 +1032,7 @@ proc close*(socket: Socket) =
         # Don't call SSL_shutdown if the connection has not been fully
         # established, see:
         # https://github.com/openssl/openssl/issues/710#issuecomment-253897666
-        if SSL_in_init(socket.sslHandle) == 0:
+        if not socket.sslNoShutdown and SSL_in_init(socket.sslHandle) == 0:
           # As we are closing the underlying socket immediately afterwards,
           # it is valid, under the TLS standard, to perform a unidirectional
           # shutdown i.e not wait for the peers "close notify" alert with a second
@@ -1312,9 +1318,14 @@ proc recv*(socket: Socket, data: var string, size: int, timeout = -1,
   if result < 0:
     data.setLen(0)
     let lastError = getSocketError(socket)
-    if flags.isDisconnectionError(lastError): return
+    if flags.isDisconnectionError(lastError):
+      when defineSsl:
+        if socket.isSsl:
+          socket.sslNoShutdown = true
+      return
     socket.socketError(result, lastError = lastError)
-  data.setLen(result)
+  else:
+    data.setLen(result)
 
 proc recv*(socket: Socket, size: int, timeout = -1,
            flags = {SocketFlag.SafeDisconn}): string {.inline.} =
@@ -1388,7 +1399,12 @@ proc readLine*(socket: Socket, line: var TaintedString, timeout = -1,
 
   template raiseSockError() {.dirty.} =
     let lastError = getSocketError(socket)
-    if flags.isDisconnectionError(lastError): setLen(line.string, 0); return
+    if flags.isDisconnectionError(lastError):
+      setLen(line.string, 0)
+      when defineSsl:
+        if socket.isSsl:
+          socket.sslNoShutdown = true
+      return
     socket.socketError(n, lastError = lastError)
 
   var waited: Duration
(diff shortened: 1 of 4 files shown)


Araq:
builds.sr.ht job failed,  tests/stdlib/tssl.nim C

dom96:
Is there anything we can do to force OpenSSL not to use SIGPIPE?

alaviss:
> Is there anything we can do to force OpenSSL not to use SIGPIPE?

Nope (unless you're on an OS that can have SIGPIPE turned off for the socket, like macOS or FreeBSD). I have some code that can be used to catch this though, but I'll push that after I can get this test to run reliably on all OS.

euantorano:
I've been testing this locally on FreeBSD due to the test failure, and get the following:

```
(alaviss/sigpipe-ssl)> ./tests/stdlib/tssl
fish: './tests/stdlib/tssl' terminated by signal SIGSEGV (Address boundary error)
```

There is a coredump created too (`tssl.core`), which I loaded up in lldb:

```
(alaviss/sigpipe-ssl)> lldb -c tssl.core -- ./tests/stdlib/tssl
(lldb) target create "./tests/stdlib/tssl" --core "tssl.core"
Core file '/usr/home/euant/src/Nim/tssl.core' (x86_64) was loaded.
(lldb) thread backtrace all
* thread #1, name = 'tssl', stop reason = signal SIGSEGV
  * frame #0: 0x0000000000219425 tssl`GC_disable__RsY2eAkO9aEvlDhb7ZyRF8gsystem + 21
    frame #1: 0x0000000000219250 tssl`signalHandler + 16
    frame #2: 0x00000008002c93c0 libthr.so.3`___lldb_unnamed_symbol101$$libthr.so.3 + 208
    frame #3: 0x00000008002c898f libthr.so.3`___lldb_unnamed_symbol82$$libthr.so.3 + 319
    frame #4: 0x00007fffffffe193
    frame #5: 0x000000000022b34f tssl`deallocWrapper__uOonBuLtXfZwTuGZR20a6Q + 63
    frame #6: 0x00000008010983e3 libcrypto.so`___lldb_unnamed_symbol1765$$libcrypto.so + 35
    frame #7: 0x00000008002c7830 libthr.so.3`___lldb_unnamed_symbol64$$libthr.so.3 + 272
    frame #8: 0x00000008002d18e5 libthr.so.3`___lldb_unnamed_symbol185$$libthr.so.3 + 53
    frame #9: 0x00000008002d182e libthr.so.3`___lldb_unnamed_symbol183$$libthr.so.3 + 382
    frame #10: 0x00000008002d16ab libthr.so.3`pthread_exit + 11
    frame #11: 0x00000008002c373e libthr.so.3`___lldb_unnamed_symbol1$$libthr.so.3 + 334
  thread #2, name = 'tssl', stop reason = signal SIGSEGV
    frame #0: 0x00000008002d1edc libthr.so.3`___lldb_unnamed_symbol190$$libthr.so.3 + 92
    frame #1: 0x000000000023b0f0 tssl`joinThread__0HHXxflmZ5Vdaq88Fr5XxAsystem + 32
    frame #2: 0x000000000023ac90 tssl`NimMainModule + 960
    frame #3: 0x000000000023a8c9 tssl`NimMainInner + 9
    frame #4: 0x000000000023adad tssl`NimMain + 45
    frame #5: 0x000000000023ae01 tssl`main + 65
    frame #6: 0x000000000020b10f tssl`_start(ap=<unavailable>, cleanup=<unavailable>) at crt1.c:76:7
(lldb)
```

lib/pure/net.nim
```diff
@@ -1312,9 +1318,14 @@ proc recv*(socket: Socket, data: var string, size: int, timeout = -1,
   if result < 0:
     data.setLen(0)
     let lastError = getSocketError(socket)
-    if flags.isDisconnectionError(lastError): return
+    if flags.isDisconnectionError(lastError):
+      when defineSsl:
+        if socket.isSsl:
+          socket.sslNoShutdown = true
+      return
```
dom96: Would it make sense to simply put this in the `socketError` proc and avoid the duplication (+ handle other cases where an disconnection error occurs)?

dom96 (approved):
Sucks that we need this extra state, but I guess it can't be helped.

Wild idea: could we get away with simply not calling ``SSL_shutdown``? Just close the socket at the OS level instead?

alaviss:
> Wild idea: could we get away with simply not calling `SSL_shutdown`? Just close the socket at the OS level instead?

Nope, `SSL_shutdown` is a part of the TLS spec to protect against truncation attacks (by using close notification to signify session close instead of blindly disconnecting).

What will cause us trouble is that the shutdown notification can be done unidirectional, but it's recommended to be done bidirectionally, which means that if a client already signaled a shutdown, we are obligated to reply, but if they just close the socket (unidirectional shutdown) then we're in for SIGPIPE :P

lib/pure/net.nim
alaviss: We can certainly put this in `socketError`, the problem would be that we need to change the API to accept `flags`, shouldn't be too hard I suppose.

alaviss:
The test is cursed on FreeBSD... I guess I'll have to setup a VM to debug this further. This might be a FreeBSD bug.

euantorano:
I've just tested on FreeBSD again, and got the same SIGSEV. here's the backtrace:

```
> ./tests/stdlib/tssl
fish: './tests/stdlib/tssl' terminated by signal SIGSEGV (Address boundary error)

> lldb -c tssl.core -- ./tests/stdlib/tssl
(lldb) target create "./tests/stdlib/tssl" --core "tssl.core"
Core file '/usr/home/euant/src/Nim/tssl.core' (x86_64) was loaded.
(lldb) thread backtrace all
* thread #1, name = 'tssl', stop reason = signal SIGSEGV
  * frame #0: 0x000000000023411c tssl`nimZeroMem + 44
    frame #1: 0x00000000002345a9 tssl`threadVarAlloc__jhSLbTQw3j1lPALL0SGkjAsystem + 25
    frame #2: 0x0000000000234339 tssl`initThreadVarsEmulation + 9
    frame #3: 0x0000000000234310 tssl`PreMain + 32
    frame #4: 0x000000000023440d tssl`NimMain + 13
    frame #5: 0x0000000000234491 tssl`main + 65
    frame #6: 0x000000000020b10f tssl`_start(ap=<unavailable>, cleanup=<unavailable>) at crt1.c:76:7
```

This seems similar to [the issues I'm having on NetBSD](https://forum.nim-lang.org/t/6610#40964), so I've tried with TLS emulation off:

```
> bin/nim c --threads:on -d:ssl --tlsEmulation:off --gc:arc  tests/stdlib/tssl.nim
Hint: used config file '/usr/home/euant/src/Nim/config/nim.cfg' [Conf]
Hint: used config file '/usr/home/euant/src/Nim/config/config.nims' [Conf]
Hint: used config file '/usr/home/euant/src/Nim/tests/config.nims' [Conf]
............................CC: stdlib_assertions.nim
CC: stdlib_system.nim
CC: stdlib_strutils.nim
CC: stdlib_pathnorm.nim
CC: stdlib_options.nim
CC: stdlib_times.nim
CC: stdlib_os.nim
CC: stdlib_nativesockets.nim
CC: stdlib_sets.nim
CC: stdlib_monotimes.nim
CC: stdlib_dynlib.nim
CC: stdlib_openssl.nim
CC: stdlib_net.nim
CC: tssl.nim

Hint:  [Link]
Hint: 63723 lines; 0.874s; 75.844MiB peakmem; Debug build; proj: tssl; out: tssl [SuccessX]

> ./tests/stdlib/tssl
```

This completes successfully with no output.

There definitely seems to be something amiss with TLS emulation...

alaviss:
I wouldn't be too surprised, we have little to no test to stress `tlsEmulation` and most platforms out there don't use it so it has little real-world testing.

Do you think you can spin off a reproducible piece out of this test case? I'll probably disable `tlsEmulation` for this test on freebsd to make it pass, but we should track this issue so we can fix it later.

euantorano:
I'll see what I can do regarding TLS emulation, will need to work out a minimal test case for it that should work but doesn't.

Note that TLS should probably also be disabled for NetBSD - we don't have any CI for it at the moment, but I'd be very surprised if it didn't have the same behaviour.

Issue #8231, opened 2018-07-07
strutils:formatSize() is broken in JS mode


genotrance:
```nim
import strutils

echo formatSize(2462056448, '.', bpIEC, false)
```

Echos 2.293GiB for C/C++ mode but 37568YiB for JS mode. Fails test case in [strutils.nim](https://github.com/nim-lang/Nim/blob/8667899d531225e060db3effe72627b9609017ce/lib/pure/strutils.nim#L2404) but isn't getting caught since testament doesn't actually run the test, just compiles due to separate bug.

ghost:
I've investigated a bit, and it seems to be like that because bitwise operators in JS are [32-bit](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators), so `1 << 40` gives `256` and not `1099511627776`
A possible fix might be to make custom JS functions for bitwise operations so they'll work on both 32-bit and 64-bit

mratsim:
Fixed by #15070

PR #15070, opened 2020-07-25
avoid #8231, bitwise move to mul,div

bung87:
avoiding  #8231 
"bitwise operators in JS are 32-bit" as @Yardanico said
the PNode .intVal store as BigestInt which currently is int64 which I dont want touch it, instead changing bitwise move operation to mul and div.


narimiran:
Can you write a test for #8231?

compiler/jsgen.nim
```diff
@@ -602,8 +602,8 @@ proc arithAux(p: PProc, n: PNode, r: var TCompRes, op: TMagic) =
   of mMulF64: applyFormat("($1 * $2)", "($1 * $2)")
   of mDivF64: applyFormat("($1 / $2)", "($1 / $2)")
   of mShrI: applyFormat("", "")
-  of mShlI: applyFormat("($1 << $2)", "($1 << $2)")
-  of mAshrI: applyFormat("($1 >> $2)", "($1 >> $2)")
+  of mShlI: applyFormat("($1 * Math.pow(2,$2))", "($1 * Math.pow(2,$2))")
+  of mAshrI: applyFormat("Math.floor($1 / Math.pow(2,$2))", "Math.floor($1 / Math.pow(2,$2))")
```
narimiran: Should this maybe use the same function as lines 596 and 597, `Math.trunc`?

bung87: when the operands are number it always get number, so I guess no checks here .

PR #15084, opened 2020-07-26
fix #11911 overload waitFor in asyncSingleProc

### LINKED ISSUE #11911: Make usage of waitFor in {.async.} compile time error
### Summary
Make usage of waitFor in {.async.} compile time error

### Description
The docs say that waitFor shouldn't be used in async code, and Nim is a typechecked languages, so we should try to check things like that on compile time if possible 

### Alternatives
Current alternative is to expect the user to manually make sure his code doesnt use waitFor in async

### Additional Information

Probably we should just iterate through the input of the macro and check for waitFor calls, similarly to await calls: I can open a PR, but i wanted to open the issue first

bung87:


Araq:
I'm not sure this is correct, this PR has to wait a little, sorry.

PR #15085, opened 2020-07-27
Add test-cases for #12576 and #12523

### LINKED ISSUE #12523: Bug with ref type + fieldPairs + hasCustomPragma
The result of using `fieldPairs` + `hasCustomPragma` on a dereferenced ref value gives different results depending on if the ref type is declared directly or with a separate object type.

### Example
```nim
import macros

template customPragma {.pragma.}

type
  RefType = ref object
    field {.customPragma.}: int

  ObjType = object
    field {.customPragma.}: int
  RefType2 = ref ObjType

block:
  let x = RefType()
  for fieldName, fieldSym in fieldPairs(x[]):
    echo hasCustomPragma(fieldSym, customPragma)

block:
  let x = RefType2()
  for fieldName, fieldSym in fieldPairs(x[]):
    echo hasCustomPragma(fieldSym, customPragma)

```

### Current Output
```
false
true
```

### Expected Output
```
true
true
```

### Additional Information

Tested with latest devel.

### LINKED ISSUE #12576: pass varargs[seq[T]] to iterator give empty seq
When pass varargs[seq[T]] to iterator, only the first arg is valid, the following is always empty.

### Example
```nim
iterator ff(sq: varargs[seq[int]]): int =
   for x in sq:
      echo x

for x in ff(@[1,2], @[1,2,3]):
   echo x

```

### Current Output
```
@[1,2]
@[]
```

### Expected Output
```
@[1,2]
@[1,2,3]
```

```
$ nim -v
Nim Compiler Version 1.1.0 [Windows: i386]
Compiled at 2019-11-02

```

ghost:
Also changed a few comments for the issues as specified in http://nim-lang.github.io/Nim/contributing.html#writing-tests-stdlib. Maybe we should do it for the whole test suite? :) I can start with some categories

Closes #12576
Closes #12523

### DIFF
--- a/tests/pragmas/tcustom_pragma.nim
+++ b/tests/pragmas/tcustom_pragma.nim
@@ -336,7 +336,7 @@ ProcDef
   static: assert bar("x") == "x"
 
 #------------------------------------------------------
-# issue #13909
+# bug #13909
 
 template dependency*(id: string, weight = 0.0) {.pragma.}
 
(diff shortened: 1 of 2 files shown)


tests/pragmas/tcustom_pragma.nim
```diff
@@ -345,4 +345,25 @@ type
     provider*: proc(obj: string): pointer {.dependency("Data/" & obj, 16.1), noSideEffect.}
 
 proc myproc(obj: string): string {.dependency("Data/" & obj, 16.1).} =
-  result = obj
\ No newline at end of file
+  result = obj
+
+# bug 12523
+template myCustomPragma {.pragma.}
```
ghost, line 351: I changed the "customPragma" pragma name because it conflicts with the imported "custompragma" module symbol, not sure if that's a bug or not

PR #15087, opened 2020-07-27
Fix #11352 strutil.insertSep() fails on negative numbers

### LINKED ISSUE #11352: strutil.insertSep() fails on negative numbers
Function `insertSep` outputs the wrong string.

### Example

```nim
insertSep($(-100), ',')
insertSep($(-100_000), ',')
```

### Current Output
```
-,100
-,100,000
```

### Expected Output

```
-100
-100,000
```

### Possible Solution

A couple options come to mind.  Don't know which fits better with the library, but placing separators in `ints` and `floats` is something done very often.

* Put in special case code in `insertSep` to handle the case when the first character is a minus sign.
* Create an overloaded proc of `insertSep(s: int): string` that handles the special case of negative numbers

### Additional Information

```
$ nim -v
Nim Compiler Version 0.19.9 [MacOSX: amd64]
Compiled at 2019-05-21
```

bung87:
https://github.com/nim-lang/Nim/issues/11352

### DIFF
--- a/lib/pure/strutils.nim
+++ b/lib/pure/strutils.nim
@@ -2279,7 +2279,7 @@ proc insertSep*(s: string, sep = '_', digits = 3): string {.noSideEffect,
   var L = (partsLen-1) div digits + partsLen
   var j = 0
   dec(L)
-  for i in countup(0, partsLen-1):
+  for i in 0 ..< partsLen:
     if j == digits:
       result.add sep
       dec(L)
--- a/lib/pure/strutils.nim
+++ b/lib/pure/strutils.nim
@@ -2264,10 +2264,10 @@ proc insertSep*(s: string, sep = '_', digits = 3): string {.noSideEffect,
   ## if `s` contains a number.
   runnableExamples:
     doAssert insertSep("1000000") == "1_000_000"
-  var parts, prefix: string
+  var prefix: string
   let hasPrefix = isDigit(s[s.low]) == false
+  var idx:int
   if hasPrefix:
-    var idx:int
     prefix.add s[s.low]
     for i in (s.low + 1)..s.high:
       idx = i
@@ -2275,10 +2275,7 @@ proc insertSep*(s: string, sep = '_', digits = 3): string {.noSideEffect,
         prefix.add s[i]
       else:
         break
-    parts = s.substr(idx)
-  else:
-    parts = s
-  let partsLen = parts.len
+  let partsLen = s.len - idx
   var L = (partsLen-1) div digits + partsLen
   result = newString(L)
   var j = 0
@@ -2288,7 +2285,7 @@ proc insertSep*(s: string, sep = '_', digits = 3): string {.noSideEffect,
       result[L] = sep
       dec(L)
       j = 0
-    result[L] = parts[i]
+    result[L] = s[i + idx]
     inc(j)
     dec(L)
   result = prefix & result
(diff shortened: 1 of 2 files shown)


lib/pure/strutils.nim
```diff
@@ -2264,19 +2264,34 @@ proc insertSep*(s: string, sep = '_', digits = 3): string {.noSideEffect,
...
+    prefix.add s[s.low]
+    for i in (s.low + 1)..s.high:
+      idx = i
+      if not isDigit(s[i]):
+        prefix.add s[i]
+      else:
+        break
+    parts = s.substr(idx)
+  else:
+    parts.shallowCopy s
```
alaviss: Please avoid `shallowCopy`, it's semantic is confusing and ARC will feature move semantics which is superior instead.

lib/pure/strutils.nim
```diff
@@ -2264,19 +2264,34 @@ proc insertSep*(s: string, sep = '_', digits = 3): string {.noSideEffect,
...
+  if hasPrefix:
+    var idx:int
+    prefix.add s[s.low]
+    for i in (s.low + 1)..s.high:
+      idx = i
+      if not isDigit(s[i]):
+        prefix.add s[i]
+      else:
+        break
+    parts = s.substr(idx)
```
Araq: Slow.

lib/pure/strutils.nim
```diff
@@ -2264,19 +2264,34 @@ proc insertSep*(s: string, sep = '_', digits = 3): string {.noSideEffect,
...
+  else:
+    parts = s
```
Araq: Slow.

lib/pure/strutils.nim
```diff
@@ -2264,19 +2264,34 @@ proc insertSep*(s: string, sep = '_', digits = 3): string {.noSideEffect,
...
+  let partsLen = parts.len
```
Araq:
```nim

  let partsLen = s.len - idx

```

lib/pure/strutils.nim
```diff
@@ -2264,19 +2264,34 @@ proc insertSep*(s: string, sep = '_', digits = 3): string {.noSideEffect,
...
   var j = 0
   dec(L)
-  for i in countdown(len(s)-1, 0):
+  for i in countdown(partsLen-1, 0):
     if j == digits:
       result[L] = sep
       dec(L)
       j = 0
-    result[L] = s[i]
+    result[L] = parts[i]
```
Araq:
```nim

result[L] = s[i + idx]

```

lib/pure/strutils.nim
```diff
@@ -2264,19 +2264,34 @@ proc insertSep*(s: string, sep = '_', digits = 3): string {.noSideEffect,
...
     inc(j)
     dec(L)
+  result = prefix & result
```
Araq: Slow. Coding this without the copies is easy enough really.

lib/pure/strutils.nim
```diff
@@ -2264,17 +2264,27 @@ proc insertSep*(s: string, sep = '_', digits = 3): string {.noSideEffect,
...
+      if not isDigit(s[i]):
+        result.add s[i]
+      else:
+        break
+  let partsLen = s.len - idx
+  var L = (partsLen-1) div digits + partsLen
   var j = 0
   dec(L)
-  for i in countdown(len(s)-1, 0):
+  for i in countup(0, partsLen-1):
```
alaviss:
```suggestion
  for i in 0 ..< partsLen:
```

PR #15094, opened 2020-07-27
fix #15082 unittest handle check out of test scope

### LINKED ISSUE #15082: unittest prints [OK] for failed test if `check` is outside `test` block
Occasionally, I use a proc to avoid duplicated test code. I found the problem that if a `check` isn't inside a `test:` block, the assertion itself runs correctly, but a wrong status `[OK]` instead of `[FAILED]` is printed in the terminal.

This is confusing, both because of the wrong status and because it's not obvious which test the `check` output refers to. (In the example it's clear, but if the proc `sharedTestCode` is called from different tests, the output is more difficult to understand.)

### Example

File `unittest_test.nim`:

```nim
import unittest

proc sharedTestCode() =
  check 3 == 4

suite "Suite name isn't relevant":

  test "This prints [FAILED]":
    check 1 == 2

  test "This prints [OK]":
    sharedTestCode()
```

Note that `sharedTestCode` checks a false condition `3 == 4` and hence should cause a failed test.

### Current Output

```
$ nim c -r unittest_test.nim
...
[Suite] Suite name isn't relevant
    /some/path/unittest_test.nim(11, 12): Check failed: 1 == 2
  [FAILED] This prints [FAILED]
    /some/path/unittest_test.nim(5, 10): Check failed: 3 == 4
  [OK] This prints [OK]
Error: execution of an external program failed: '/some/path/unittest_test '
```

Note that unittest correctly complains about the failed check `3 == 4` but nevertheless prints `[OK]`.

### Expected Output
```
$ nim c -r unittest_test.nim
...
[Suite] Suite name isn't relevant
    /some/path/unittest_test.nim(11, 12): Check failed: 1 == 2
  [FAILED] This prints [FAILED]
    /some/path/unittest_test.nim(5, 10): Check failed: 3 == 4
  [FAILED] This prints [OK]
Error: execution of an external program failed: '/some/path/unittest_test '
```

### Possible Workarounds:

- (Try to) ignore the wrong status.
- Don't use `check` outside `test`
- Use `doAssert` in the extracted proc (although this adds a stacktrace for the assertion that may not be necessary)

It seems the output from `check` is always above the status message with the `[OK]` or `[FAILED]` tag, so it's still possible (with some effort) to see which test(s) actually failed.

### Additional Information

```
$ nim -v
Compiled at 2020-07-26

```

bung87:
fix https://github.com/nim-lang/Nim/issues/15082

dont know how to test unittest 

here is code during debuging
``` nim
import unittest

proc sharedTestCode() =
  check 3 == 4

suite "Suite name isn't relevant":

  test "This prints [FAILED]":
    check 1 == 2

  test "This prints [OK]":
    sharedTestCode()

  test "skip":
    skip()
  
  test "skip then continue":
    check 7 == 8
  
  test "see a success":
    check 9 == 9

test "single":
  check 5 == 6

```  

have to bind `contains` in devel branch, dont know why, old build nimlsp in vscode show it can find `strutils.contains` symbol.  

so we dont reset `checkpoints = @[]` in `fails` or `skip` , it's in the end of `template test`

outputs:  
```

[Suite] Suite name isn't relevant
    /Users/bung/t15082.nim(9, 12): Check failed: 1 == 2
  [FAILED] This prints [FAILED]
    /Users/bung/t15082.nim(4, 10): Check failed: 3 == 4
  [FAILED] This prints [OK]
  [SKIPPED] skip
    /Users/bung/t15082.nim(18, 12): Check failed: 7 == 8
  [FAILED] skip then continue
  [OK] see a success
/Users/bung/t15082.nim(24, 10): Check failed: 5 == 6
[FAILED] single
Error: execution of an external program failed: '/Users/bung/t15082 '
```

### DIFF
--- a/lib/pure/unittest.nim
+++ b/lib/pure/unittest.nim
@@ -499,7 +499,7 @@ template test*(name, body) {.dirty.} =
   ## .. code-block::
   ##
   ##  [OK] roses are red
-  bind shouldRun, checkpoints, formatters, ensureInitialized, testEnded, exceptionTypeName, setProgramResult
+  bind contains, shouldRun, checkpoints, formatters, ensureInitialized, testEnded, exceptionTypeName, setProgramResult
 
   ensureInitialized()
 
@@ -524,6 +524,10 @@ template test*(name, body) {.dirty.} =
       fail()
 
     finally:
+      if checkpoints.len > 0:
+        let lastCheckFailed = checkpoints[^1].contains("Check failed:")
+        if testStatusIMPL == TestStatus.OK and lastCheckFailed:
+          testStatusIMPL = TestStatus.FAILED
       if testStatusIMPL == TestStatus.FAILED:
         setProgramResult 1
       let testResult = TestResult(
@@ -578,7 +582,6 @@ template fail* =
 
   if abortOnError: quit(1)
 
-  checkpoints = @[]
 
 template skip* =
   ## Mark the test as skipped. Should be used directly
@@ -594,7 +597,6 @@ template skip* =
   bind checkpoints
 
   testStatusIMPL = TestStatus.SKIPPED
-  checkpoints = @[]
 
 macro check*(conditions: untyped): untyped =
   ## Verify if a statement or a list of statements is true.
@@ -662,7 +664,6 @@ macro check*(conditions: untyped): untyped =
 
   case checked.kind
   of nnkCallKinds:
-
     let (assigns, check, printOuts) = inspectArgs(checked)
     let lineinfo = newStrLitNode(checked.lineInfo)
     let callLit = checked.toStrLit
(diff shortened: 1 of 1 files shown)


Araq:
Strange fix, @Clyybber please investigate.

bung87:
the issue  cause by `testStatusIMPL` cant inject to undering proc , then I need remove the reset `checkpoints` it will hold duplicated messages, since checkpoints initialize is in `test` scope and reset in the bottom of `test`, so I think remove the `checkpoints = @[]` in `fails` and `skip` is fine.

lib/pure/unittest.nim
```diff
@@ -524,6 +524,10 @@ template test*(name, body) {.dirty.} =
       fail()
 
     finally:
+      if checkpoints.len > 0:
+        let lastCheckFailed = checkpoints[^1].contains("Check failed:")
```
Araq: That looks like a hack.

PR #15103, opened 2020-07-28
Count table del fixes

c-blake:
This passes `(nim r tables; nim r sharedtables)` and is probably ready for the CI tests.

PR #15104, opened 2020-07-28
Fulfill https://github.com/nim-lang/Nim/pull/14995#issuecomment-66491…

c-blake:
`(nim r tables; nim r sharedtables)` work.  Seems ready for CI testing.  (This replaces
the now defunct https://github.com/nim-lang/Nim/pull/15103)

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -140,6 +140,7 @@
 - Tables, HashSets, SharedTables and deques don't require anymore that the passed
   initial size must be a power of two - this is done internally.
   Proc `rightSize` for Tables and HashSets is deprecated, as it is not needed anymore.
+  `CountTable.inc` takes `val: int` again not `val: Positive`; I.e. it can "count down" again.
 - Removed deprecated symbols from `macros` module, deprecated as far back as `0.15`.
 
 
--- a/lib/pure/collections/sharedtables.nim
+++ b/lib/pure/collections/sharedtables.nim
@@ -138,6 +138,10 @@ proc hasKeyOrPut*[A, B](t: var SharedTable[A, B], key: A, val: B): bool =
   withLock t:
     hasKeyOrPutImpl(enlarge)
 
+template tabMakeEmpty(i) = t.data[i].hcode = 0
+template tabCellEmpty(i) = isEmpty(t.data[i].hcode)
+template tabCellHash(i)  = t.data[i].hcode
+
 proc withKey*[A, B](t: var SharedTable[A, B], key: A,
                     mapper: proc(key: A, val: var B, pairExists: var bool)) =
   ## Computes a new mapping for the ``key`` with the specified ``mapper``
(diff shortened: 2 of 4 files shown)


c-blake:
While I think it's fairly self-evident from source code and algos, some like to see benchmarks.  These are based on https://github.com/nim-lang/Nim/pull/14995#issuecomment-665051733 but reduced.
```Nim
import tables, times

template timeIt(label1, label2, value2, scale, body) =
  var t0 = epochTime()
  body
  let dt = epochTime() - t0
  echo label1, dt * 1e9 / scale, label2, value2

proc runtest(els: int) =
  var ot = initCountTable[int](els)
  timeIt("Create " & $(els div 1_000) & "Kitem table (ns/elt): ",
         " len ", ot.len, float(els)):
    for i in 0 ..< els: ot[i] = i
  timeIt("Delete99% (ns/elt): ", " len ", ot.len, float(els)):
    for i in 0 ..< els:
      if i mod 100 != 0: ot.del(i)
for els in @[ 10_000, 50_000 ]: runtest(els)
```
On an (Intel i7-6700k@4.7GHz, Linux, nim bbb250f, d:danger, etc.), I get these results.
Before (manually rounded times to 0.1 ns):
```
Create 10Kitem table (ns/elt): 21.5 len 9999
Delete99% (ns/elt): 78735.2 len 99
Create 50Kitem table (ns/elt): 26.7 len 49999
Delete99% (ns/elt): 457257.2 len 499
```
After (manually rounded times to 0.1 ns):
```
Create 10Kitem table (ns/elt): 10.7 len 9999
Delete99% (ns/elt): 28.2 len 99
Create 50Kitem table (ns/elt): 9.5 len 49999
Delete99% (ns/elt): 14.3 len 499
```
So, we see deletes get 78735/28 =~ 2800x faster and 457257/14 =~ 33,000x faster.  Obviously, since it's O(n) expected vs O(1) that could be made as big as you want.  (It's 1.7x and 1.5x less than 5000x and 10,000x since branch prediction/pre-fetching/etc.)

A careful eye might notice inserts getting faster across the PR, though no relevant code changed.  I am not sure what that is, but suspect `gcc` code auto-inlining/layout/branch prediction/instruction cache type explanations that might not even generalize to an AMD x64 CPU.  In any case, it's in the right direction, and also only O(2x).  PGO can make up to 2x deltas, anyway.  So a less half-assed study would be needed to really know.  (I mentioned it so you know I'm aware.  It might have been wiser to delete the timing for the insert phase...)

c-blake:
All tests pass.  Also, if you would like we could put a 15-20 line explanatory note in front of `delImplIdx` to perhaps prevent people from changing without understanding.  Let me know about that before you merge.

Araq:
> All tests pass. Also, if you would like we could put a 15-20 line explanatory note in front of delImplIdx to perhaps prevent people from changing without understanding. Let me know about that before you merge.

More comments are always nice but can be done in a follow-up PR. I'm merging this already because it's so beautiful. :-)

c-blake:
Ok.  Thanks.  :-)

c-blake:
https://github.com/nim-lang/Nim/pull/15108

PR #15106, opened 2020-07-29
Fix #11684 For loops over a hardcoded empty array crash the compiler

### LINKED ISSUE #11684: For loops over a hardcoded empty array crash the compiler.
A for loop over a hardcoded empty array, which is meaningless, causes the compiler to have an internal error.

### Example
```nim
for _ in []:
  discard
```

### Current Output
```
Error: internal error: getTypeDescAux(tyEmpty)
No stack traceback available
To create a stacktrace, rerun compilation with ./koch temp c <file>
```

### Expected Output
```
```

This occurs on the latest devel and stable.

bung87:
fix #11684

### DIFF
--- a/compiler/ccgexprs.nim
+++ b/compiler/ccgexprs.nim
@@ -871,6 +871,8 @@ proc genArrayElem(p: BProc, n, x, y: PNode, d: var TLoc) =
   initLocExpr(p, x, a)
   initLocExpr(p, y, b)
   var ty = skipTypes(a.t, abstractVarRange + abstractPtrs + tyUserTypeClasses)
+  if ty.kind == tyArray and ty[1].kind == tyEmpty:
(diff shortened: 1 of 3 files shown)


ghost:
What happens if you try it with the JS backend? Shouldn't that error be caught in the front-end generally?

bung87:
what I thought is it should throw during ast processing , but it not , it reach to ccgtypes.nim . it reach there should be handled in there or down stream, otherwise it shouldn't reach there.

Araq:
This needs to be fixed in the frontend.

bung87:
> This needs to be fixed in the frontend.

where's frontend related files located?

Araq:
sigmatch.nim proc ``concreteType``. It needs to say that openArray[empty] is invalid too. I think.

bung87:
give up

PR #15107, opened 2020-07-29
fix #14139

### LINKED ISSUE #14139: heapqueue pushpop() proc doesn't compile
Proc `pushpop()` in module `heapqueue` does not compile.
`swap()` proc, which is used in `pushpop()` requires inputs of type `var T`, but in `pushpop()` the `item` variable is immutable.

### Example
```nim
import heapqueue

var test_queue : HeapQueue[int]

test_queue.push(7)
test_queue.push(3)
test_queue.push(9)
let i = test_queue.pushpop(10) # This will not compile
assert i == 3
```

### Current Output
```
/Users/nproach/Documents/test.nim(8, 19) template/generic instantiation of `pushpop` from here
/usr/local/Cellar/nim/1.2.0/nim/lib/pure/collections/heapqueue.nim(153, 9) Error: type mismatch: got <int, int>
but expected one of:
proc swap[T](a, b: var T)
  required type for a: var T
  but expression 'item' is immutable, not 'var'

expression: swap(item, heap[0])
```

### Expected Output
```
```

### Possible Solution

* Unclear, adding a `var tmp = item` to the top of the `pushpop()` proc and replacing `item` with `tmp` in the rest of the proc results in a different error, which claims `expression 'heap[0]' is immutable not 'var'` , which is confusing as heap is declared in the function as `var`. Would appreciate any help on this. For now I'll just use `push()` followed by `pop()`.

```
$ nim --version
Compiled at 2020-04-12
```

ringabout:
Maybe also add `var T` version to reduce one copy?(But maybe move semantics are good enough to reduce copies?) 

```nim
proc pushpop*[T](heap: var HeapQueue[T], item: var T): T =
  ## Fast version of a push followed by a pop.
  if heap.len > 0 and heapCmp(heap.data[0], item):
    swap(item, heap.data[0])
    siftup(heap, 0)
  return item
```

### DIFF
--- a/lib/pure/collections/heapqueue.nim
+++ b/lib/pure/collections/heapqueue.nim
@@ -159,9 +159,9 @@ proc pushpop*[T](heap: var HeapQueue[T], item: T): T =
   if heap.len > 0 and heapCmp(heap.data[0], item):
     result = item
     swap(result, heap.data[0])
+    siftup(heap, 0)
   else:
     result = item
-    siftup(heap, 0)
 
 proc clear*[T](heap: var HeapQueue[T]) =
   ## Remove all elements from `heap`, making it empty.
--- a/lib/pure/collections/heapqueue.nim
+++ b/lib/pure/collections/heapqueue.nim
@@ -162,7 +162,6 @@ proc pushpop*[T](heap: var HeapQueue[T], item: T): T =
   else:
     result = item
     siftup(heap, 0)
-  return item
 
 proc clear*[T](heap: var HeapQueue[T]) =
   ## Remove all elements from `heap`, making it empty.
--- a/lib/pure/collections/heapqueue.nim
+++ b/lib/pure/collections/heapqueue.nim
@@ -156,9 +156,11 @@ proc replace*[T](heap: var HeapQueue[T], item: T): T =
 
 proc pushpop*[T](heap: var HeapQueue[T], item: T): T =
   ## Fast version of a push followed by a pop.
-  var item = item
   if heap.len > 0 and heapCmp(heap.data[0], item):
-    swap(item, heap.data[0])
+    result = item
+    swap(result, heap.data[0])
+  else:
+    result = item
     siftup(heap, 0)
   return item
 
--- a/lib/pure/collections/heapqueue.nim
+++ b/lib/pure/collections/heapqueue.nim
@@ -156,8 +156,9 @@ proc replace*[T](heap: var HeapQueue[T], item: T): T =
 
 proc pushpop*[T](heap: var HeapQueue[T], item: T): T =
   ## Fast version of a push followed by a pop.
-  if heap.len > 0 and heapCmp(heap[0], item):
-    swap(item, heap[0])
+  var item = item
+  if heap.len > 0 and heapCmp(heap.data[0], item):
+    swap(item, heap.data[0])
     siftup(heap, 0)
   return item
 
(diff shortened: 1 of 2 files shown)


Araq:
> Maybe also add var T version to reduce one copy?

Definitely not.

lib/pure/collections/heapqueue.nim
```diff
@@ -156,8 +156,9 @@ proc replace*[T](heap: var HeapQueue[T], item: T): T =
 
 proc pushpop*[T](heap: var HeapQueue[T], item: T): T =
   ## Fast version of a push followed by a pop.
-  if heap.len > 0 and heapCmp(heap[0], item):
-    swap(item, heap[0])
+  var item = item
+  if heap.len > 0 and heapCmp(heap.data[0], item):
+    swap(item, heap.data[0])
```
Araq:
```suggestion
  if heap.len > 0 and heapCmp(heap.data[0], item):
    result = item
    swap(result, heap.data[0])
  else:
    result = item
```

lib/pure/collections/heapqueue.nim
```diff
@@ -156,9 +156,11 @@ proc replace*[T](heap: var HeapQueue[T], item: T): T =
...
   ## Fast version of a push followed by a pop.
-  var item = item
   if heap.len > 0 and heapCmp(heap.data[0], item):
-    swap(item, heap.data[0])
+    result = item
+    swap(result, heap.data[0])
+  else:
+    result = item
     siftup(heap, 0)
   return item
```
ghost: maybe remove this return then? :P

lib/pure/collections/heapqueue.nim
```diff
@@ -156,10 +156,12 @@ proc replace*[T](heap: var HeapQueue[T], item: T): T =
...
 proc pushpop*[T](heap: var HeapQueue[T], item: T): T =
   ## Fast version of a push followed by a pop.
-  if heap.len > 0 and heapCmp(heap[0], item):
-    swap(item, heap[0])
+  if heap.len > 0 and heapCmp(heap.data[0], item):
+    result = item
+    swap(result, heap.data[0])
+  else:
+    result = item
     siftup(heap, 0)
```
ringabout, line 162: It seems wrong?

lib/pure/collections/heapqueue.nim
```diff
@@ -156,10 +156,12 @@ proc replace*[T](heap: var HeapQueue[T], item: T): T =
 
 proc pushpop*[T](heap: var HeapQueue[T], item: T): T =
   ## Fast version of a push followed by a pop.
-  if heap.len > 0 and heapCmp(heap[0], item):
-    swap(item, heap[0])
+  if heap.len > 0 and heapCmp(heap.data[0], item):
+    result = item
```
Clyybber, line 160:
```suggestion
  result = item
  if heap.len > 0 and heapCmp(heap.data[0], item):
```
along with the other suggestion (need to split them because of github)

lib/pure/collections/heapqueue.nim
```diff
@@ -156,10 +156,12 @@ proc replace*[T](heap: var HeapQueue[T], item: T): T =
...
+    swap(result, heap.data[0])
     siftup(heap, 0)
-  return item
+  else:
+    result = item
```
Clyybber, line 164:
```suggestion
```

PR #15123, opened 2020-07-30
Better semiStmtList parsing

Clyybber:
Related: https://forum.nim-lang.org/t/6607


compiler/parser.nim
```diff
@@ -523,17 +523,22 @@ proc parseGStrLit(p: var TParser, a: PNode): PNode =
   else:
     result = a
 
-proc complexOrSimpleStmt(p: var TParser): PNode
+proc complexOrSimpleStmt(p: var TParser, maybeExpr: static bool = false): PNode
```
Araq: Remove the ``static``.

Araq:
The PR looks fine but I need to investigate what it really implies for Nim's grammar. So this has to wait, probably for a couple of weeks, sorry!

compiler/parser.nim
```diff
@@ -2203,7 +2174,7 @@ proc complexOrSimpleStmt(p: var Parser): PNode =
   #|                     | bindStmt | mixinStmt)
   #|                     / simpleStmt
   case p.tok.tokType
-  of tkIf: result = parseIfOrWhen(p, nkIfStmt)
+  of tkIf: result = if maybeExpr: parseIfExpr(p, nkIfStmt) else: parseIfOrWhen(p, nkIfStmt)
```
Araq: This line means you implemented an "attribute grammar" but Nim's grammar does not reflect that.

Clyybber: I can move this to `semiStmtList` if you want to, since it is the only part where its actually used and `semiStmtList` seems to lack a representation in the grammar anyways.

Clyybber: Maybe I should add this:
```

  #| complexOrSimpleStmt(maybeExpr = true) = (ifExpr | ifStmt | whenStmt | whileStmt
  #|                     | tryStmt | forStmt
  #|                     | blockStmt | staticStmt | deferStmt | asmStmt
  #|                     | 'proc' routine
  #|                     | 'method' routine
  #|                     | 'func' routine
  #|                     | 'iterator' routine
  #|                     | 'macro' routine
  #|                     | 'template' routine
  #|                     | 'converter' routine
  #|                     | 'type' section(typeDef)
  #|                     | 'const' section(constant)
  #|                     | ('let' | 'var' | 'using') section(variable)
  #|                     | bindStmt | mixinStmt)
  #|                     / simpleStmt
```
to the grammar comment?

Araq: That's not how attributed grammars work... ;-)

Clyybber: Moved the condition out of `complexOrSimpleStmt` and update the grammar accordingly.

PR #15135, opened 2020-07-31
fix overloading issue with generic invocation

jcosborn:


Araq:
Have you verified that the other logic that is triggered also does `` c.inheritancePenalty += depth`` ?

jcosborn:
It seems to handle that correctly.  I've added a test.

PR #15138, opened 2020-08-01
Added the ability to initialize a deque with an openArray

archnim:
For example:
`var dq = initDeque[char](@['a', 'b', 'c'])`

### DIFF
--- a/lib/pure/collections/deques.nim
+++ b/lib/pure/collections/deques.nim
@@ -84,11 +84,11 @@ proc initDeque*[T](initialSize: int = 4): Deque[T] =
   ## The length of a newly created deque will still be 0.
   result.initImpl(initialSize)
 
-proc initDeque*[T](s: seq[T]): Deque[T] =
-  ## Create a new deque initialized with a sequence
-  result.initImpl(nextPowerOfTwo(s.len))
-  for i in 0 ..< s.len:
-    result.addLast(s[i])
+proc initDeque*[T](arr: openArray[T]): Deque[T] =
+  ## Create a new deque initialized with an open array
+  result.initImpl(nextPowerOfTwo(arr.len))
+  for i in 0 ..< arr.len:
+    result.addLast(arr[i])
 
 proc len*[T](deq: Deque[T]): int {.inline.} =
   ## Return the number of elements of `deq`.
--- a/lib/pure/collections/deques.nim
+++ b/lib/pure/collections/deques.nim
@@ -84,6 +84,12 @@ proc initDeque*[T](initialSize: int = 4): Deque[T] =
   ## The length of a newly created deque will still be 0.
   result.initImpl(initialSize)
 
+proc initDeque*[T](s: seq[T]): Deque[T] =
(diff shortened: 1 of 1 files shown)


lib/pure/collections/deques.nim
```diff
@@ -84,6 +84,12 @@ proc initDeque*[T](initialSize: int = 4): Deque[T] =
   ## The length of a newly created deque will still be 0.
   result.initImpl(initialSize)
 
+proc initDeque*[T](s: seq[T]): Deque[T] =
+  ## Create a new deque initialized with a sequence
+  result = initDeque[T]()
+  for i in 0..< s.len:
```
ghost: Nitpick: use consistent spacing around ..< - either space before/after it, or no spaces at all

lib/pure/collections/deques.nim
```diff
@@ -84,6 +84,12 @@ proc initDeque*[T](initialSize: int = 4): Deque[T] =
   ## The length of a newly created deque will still be 0.
   result.initImpl(initialSize)
 
+proc initDeque*[T](s: seq[T]): Deque[T] =
+  ## Create a new deque initialized with a sequence
+  result = initDeque[T]()
```
ghost: You should probably also use the initialSize argument for initDeque and set it to https://nim-lang.org/docs/math.html#nextPowerOfTwo,int of the length of the seq

ghost: Oh wait, you could not call the original proc at all and just use result.initImpl(nextPowerOfTwo(s.len))

archnim: Ok, i'm going to fix all that.

archnim: Please, can you explain to me the use of initImpl ? It is not commented.

ghost: @archnim just check how normal initDeque defined and use the same :)

archnim: Line 25 of lib/pure/collections/deques.nim:
`var a = initDeque[int]()`

Isn't it correct ? Sorry i'm not yet a great professional of nim.

ghost: @archnim See how initDeque is defined in https://github.com/nim-lang/Nim/blob/devel/lib/pure/collections/deques.nim#L85. You do the same but instead of using initialSize you use the next power of two of the length of the sequence so there are less allocations

archnim: Thank you. It will be done in few minutes.

alaviss:
Can't `openArray[T]` be used instead?

archnim:
> Can't `openArray[T]` be used instead?

I've just done it.

ghost:
I think this needs some documentation (and a changelog entry), but that can be done later :)

archnim:
Hello world. This was my first pull request in an open source project. Then I just want to say that I'm moved that my pr has been merged only one month after my intro to nim lang. It's a great thing for me. 😪
I'm enthusiast. If you need to translate anything from English to French, just just ask me.

PR #15140, opened 2020-08-01
export asyncdispatch handles

ringabout:
I need handles attributes to make `httpbeast` work in windows. Now It seems to be fine to use `select`(windows) with httpbeast.

```nim
  let disp = getGlobalDispatcher()

  for h in disp.handles.items:
    selector.registerHandle(nativesockets.SocketHandle(h), {Event.Read},
                            initData(Dispatcher))
```

image

image


Araq:
Document it please as an internal detail of the current implementation that might be changed.

ringabout:
I'm not sure which way is better? Export it or add a `getIoHandles` (which is specific to windows)?

Araq:
I would export it. A getter doesn't help much.

PR #15155, opened 2020-08-04
minor change in selectors

ringabout:
I don't know whether it should be changed. Maybe this is more clear.


xflywind:
What's wrong with the bsd tests?

Clyybber (approved):

PR #15158, opened 2020-08-05
Fix #5691

### LINKED ISSUE #5691: The StmtList processing of template parameters can lead to unexpected errors
Consider the following code:

```
template foo(x: stmt) =
  discard

foo:
  var a = 10

foo:
  var a = 20 # error: redefinition of `a`
```

Even though the pass-in blocks are discarded, they pollute the scope where `foo` is used and can produce redefinition errors as in the example above.

### LINKED ISSUE #15170: redefinition of 'X'; previous declaration here
This is more of a question then a bug, but I'd like to understand the underlying mechanics better: why is it not possible to rewrite a proc with a typed macro?

```nim
import macros        
                      
macro foo(n: typed) =      
  copyNimTree(n)
                       
                 
# Why is this ok
foo:       
  var a = 3
    
foo:
  let b = 3
      
# But this is not
foo:
  proc c() =
    discard 
```

This results in 

```
/tmp/t.nim(16, 1) template/generic instantiation of `foo` from here
/tmp/t.nim(17, 3) Error: redefinition of 'c'; previous declaration here: /tmp/t.nim(17, 8)
```

Clyybber:
Another option would be to open a scope, and close or merge it with the current one depending on wether this was a macro/template or a proc call. (A TExprFlag like efNoSymbols doesn't work since the stmtlist argument itself needs symbols from its own body).

@Araq Let me know which approach you prefer :)

Fixes #5691
Fixes #15170


Araq:
This is an ugly, incomplete solution. Instead ``typed`` ASTs should have a spec and side effects during overload resolution need to be undone, somehow.

Clyybber:
Yeah, I don't like this either, I should have marked it as a draft; it also has some rather fundamental issues right now. I'll try to come up with a better solution.

compiler/sigmatch.nim
```diff
@@ -2507,12 +2497,11 @@ proc matchesAux(c: PContext, n, nOrig: PNode,
...
               typeToString(n[a].typ), typeToString(formal.typ) ])
             noMatch()
-            return
         checkConstraint(n[a])
 
-    if m.state == csMatch and m.calleeSym != nil and m.calleeSym.kind in {skTemplate, skMacro}: #template or macro (and proc with static params?)
-      # Restore the original scope
-      c.currentScope[] = oldScope
+    c.closeScope
+    if m.state == csMatch and m.calleeSym != nil and m.calleeSym.kind notin {skTemplate, skMacro}:
```
disruptek, line 2503: proc please

narimiran:
@Clyybber will this also fix https://github.com/nim-lang/Nim/issues/15170?

Clyybber:
Yep :)

disruptek:
Unfortunately, this is breaking https://github.com/disruptek/criterion according to git bisect:
```
/home/runner/work/criterion/criterion/criterion/tests/test1.nim(14, 24) Error: internal error: expr: param not init input_19385020
```

```nim
import criterion
import std/sha1
import strutils

let cfg = newDefaultConfig()

benchmark cfg:
  iterator strsrc(): string =
    yield repeat('a', 20)
    yield repeat('a', 200)
    yield repeat('a', 2000)

  proc fastSHA(input: string) {.measure: strsrc.} =
    discard secureHash(input)
```

Clyybber:
@disruptek Fixed in #15252

PR #15161, opened 2020-08-05
fix sqlgetdata regression in odbc

cooldome:
Chasing regression caused by https://github.com/nim-lang/Nim/pull/14357.
Last 2 arguments of SQLGetData need have TSqlLen in its types.

Official docs:
https://docs.microsoft.com/en-us/sql/odbc/reference/syntax/sqlgetdata-function?view=sql-server-ver15


cooldome:
added more fixes. FetchScroll and ExtendedFetch use TSqlLen as well
https://docs.microsoft.com/en-us/sql/odbc/reference/syntax/sqlextendedfetch-function?view=sql-server-ver15
https://docs.microsoft.com/en-us/sql/odbc/reference/syntax/sqlfetchscroll-function?view=sql-server-ver15

PR #15166, opened 2020-08-07
deques.nim: Refactor the `toDeque` functionality

ee7:
This PR makes some small changes to the new proc introduced by #15138.

Changes:
- Rename to `toDeque` for more consistency with well-known procs like
  `toHashSet` and `toTable`.
- Rename the `openArray` parameter. The name `arr` was potentially less
  clear given that the proc can be used with a seq (or string).
- Add a `since` annotation.
- Reword the doc comment, and clarify that ordering is preserved.
- Add runnableExamples.
- Add "see also" cross linking between `initDeque` and `toDeque`.
- Remove duplicate `nextPowerOfTwo`. The `initImpl` template already
  includes it.
- Implement the proc using the `items` iterator, rather than indexing.
  This matches the implementation of `sets.toHashSet` and
  `tables.toTable`.
- Add a test within `when isMainModule`.
- Add a changelog entry.

### DIFF
--- a/lib/pure/collections/deques.nim
+++ b/lib/pure/collections/deques.nim
@@ -82,13 +82,25 @@ proc initDeque*[T](initialSize: int = 4): Deque[T] =
   ## Optionally, the initial capacity can be reserved via `initialSize`
   ## as a performance optimization.
   ## The length of a newly created deque will still be 0.
+  ##
+  ## See also:
+  ## * `toDeque proc <#toDeque,openArray[T]>`_
   result.initImpl(initialSize)
 
-proc initDeque*[T](arr: openArray[T]): Deque[T] =
-  ## Create a new deque initialized with an open array
-  result.initImpl(nextPowerOfTwo(arr.len))
-  for i in 0 ..< arr.len:
-    result.addLast(arr[i])
+proc toDeque*[T](x: openArray[T]): Deque[T] {.since: (1, 3).} =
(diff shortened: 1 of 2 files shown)


lib/pure/collections/deques.nim
```diff
@@ -82,13 +82,25 @@ proc initDeque*[T](initialSize: int = 4): Deque[T] =
...
+  ##
+  ## See also:
+  ## * `initDeque proc <#initDeque,int>`_
+  runnableExamples:
+    var a = toDeque([7, 8, 9])
+    assert len(a) == 3
+    assert a.popFirst == 7
+    assert len(a) == 2
+
+  result.initImpl(x.len)
```
ghost, line 101: Why did you remove the `nextPowerOfTwo`? `initImpl` expects the passed len to be a power of two.

ee7: > `initImpl` expects the passed len to be a power of two.

That was the case in the past, but it changed a month ago (along with similar changes in `sets` and `tables`). See [here](https://github.com/nim-lang/Nim/commit/3de5296337a0711b4d907728ac24fa07791d3224#diff-22d75f39e0a645e0c03ea53829ef7898
).

The current implementation of `initImpl` is:
https://github.com/nim-lang/Nim/blob/6f482ab9346230f8e9c24f71091bcd5ca2cb41f8/lib/pure/collections/deques.nim#L69-L72

So I think this PR is correct, unless I'm missing something.

PR #15169, opened 2020-08-08
Fix #8473

### LINKED ISSUE #5688: Compiler is confused about call convention of function with nested closure
```nimrod
import typetraits

proc foo() =
  let a = 5
  let f = (proc() = discard (proc() = echo a))
  echo name(type(f))
  echo sizeof(f)

foo()
```

returns
```
proc (){.noSideEffect, gcsafe, locks: 0.}
8
```
while in generated code, f uses closure call convention (which is correct).

### LINKED ISSUE #8473: {.nimcall.} convention won't avoid the creation of closures
Code
```nim
proc a(v: int) =
  proc b {.nimcall.} = echo v
  b()
a(5)
```
Compiles without a complaint. The opposite would add type safety against the creation of closures.
***
```nim
>nim -v
Nim Compiler Version 0.18.1 [Linux: i386]
Compiled at 2018-07-14

git hash: 9b98add6c714474e261df1a0a17819282dfd1975
```

Clyybber:
Fixes #8473
Closes #5688

### DIFF
--- a/compiler/ast.nim
+++ b/compiler/ast.nim
@@ -17,8 +17,7 @@ export int128
 
 type
   TCallingConvention* = enum
-    ccImplicit,               # proc has no explicit calling convention
-    ccNimCall,                # proc was explicitly marked as nimcall
+    ccNimCall,                # nimcall, also the default
     ccStdCall,                # procedure is stdcall
     ccCDecl,                  # cdecl
     ccSafeCall,               # safecall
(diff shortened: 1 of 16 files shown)


Araq:
<strike>Alternatively we could deprecate the explicit ``.nimcall`` annotation, right?</strike>

Araq:
Instead use a new symbol flag ``sfExplicitNimcall``.

Clyybber:
I originally used that approach, but this doesn't work with this example:
```
type A = proc() #{.nimcall.} should make this fail
let b = 1
let a = proc() = echo b
```
Currently `A` would be inferred to be a `proc() {.closure.}`, so explicitly writing `type A = proc() {.nimcall.}` should prevent that from compiling. So naturally I thought maybe it should to be a typeflag, but then making it a seperate calling convention seemed cleaner.

Araq:
I don't understand.

Clyybber:
Sorry, my example had a typo. My point was that this shouldn't compile, ~ironically enough currently it does with this PR (since it allows assigning ccImplicit to ccNimCall)~. 
~Fixing that hits a phase order problem; ccImplicit must be eliminated earlier so that its clear in sigmatch wether its a ccClosure or a ccNimCall. But then you have the issue of forward declaration...~

cooldome:
I guess @Clyybber is saying is that we have proc types without syms, yet proc types are incompatible if they have different calling conventions. Hence it can't be a sym flag.

Araq:
Ok, so please make it a typeflag then.

PR #15172, opened 2020-08-09
[FIX] strtabs interpolation in nimscript

haxscramper:
Fix `strtabs.%` interpolation in nimscript

### DIFF
--- a/tests/stdlib/tstrtabs.nims
+++ b/tests/stdlib/tstrtabs.nims
@@ -1,4 +1,5 @@
 import strtabs
 
-let t = {"name": "John", "city": "Monaco"}.newStringTable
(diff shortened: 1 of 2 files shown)


tests/stdlib/tstrtabs.nims
```diff
@@ -0,0 +1,4 @@
+import strtabs
+
+let t = {"name": "John", "city": "Monaco"}.newStringTable
+doAssert "${name} lives in ${city}" % t == "John lives in Monaco"
```
Clyybber:
```suggestion
static:
  let t = {"name": "John", "city": "Monaco"}.newStringTable
  doAssert "${name} lives in ${city}" % t == "John lives in Monaco"
```

PR #15174, opened 2020-08-11
asyncnet: don't try to close the socket again [backport]

### LINKED ISSUE #9867: Long-lived application occasionally dies with SIG_PIPE when calling httpclient.request

alaviss:
The closed flag here doesn't seem to be a a good design to me, but until
I can understand why it was needed it seems unwise to just replace it
with the invalidation of the socket.

Fixes https://github.com/nim-lang/Nim/issues/9867#issuecomment-671071188

/cc @dom96 as this `closed` flag was added by him. This flag appear to have survived the refactoring in https://github.com/nim-lang/Nim/commit/1661062ebf147e55e29a0bfb8fc62dc7959b8614, where its sister `net` module use the different approach of invalidating the `Socket` instead of a flag.

### DIFF
--- a/lib/pure/asyncnet.nim
+++ b/lib/pure/asyncnet.nim
@@ -713,8 +713,12 @@ elif defined(nimdoc):
 
 proc close*(socket: AsyncSocket) =
   ## Closes the socket.
+  if socket.closed: return
+
   defer:
     socket.fd.AsyncFD.closeSocket()
+    socket.closed = true # TODO: Add extra debugging checks for this.
+
   when defineSsl:
     if socket.isSsl:
       let res =
@@ -731,7 +735,6 @@ proc close*(socket: AsyncSocket) =
         discard
       elif res != 1:
         raiseSSLError()
-  socket.closed = true # TODO: Add extra debugging checks for this.
 
 when defineSsl:
   proc wrapSocket*(ctx: SslContext, socket: AsyncSocket) =
@@ -819,9 +822,9 @@ proc sendTo*(socket: AsyncSocket, address: string, port: Port, data: string,
   ## address or a hostname. If a hostname is specified this function will try
   ## each IP of that hostname. The returned future will complete once all data
   ## has been sent.
-  ## 
+  ##
   ## If an error occurs an OSError exception will be raised.
-  ## 
+  ##
   ## This proc is normally used with connectionless sockets (UDP sockets).
   assert(socket.protocol != IPPROTO_TCP,
          "Cannot `sendTo` on a TCP socket. Use `send` instead")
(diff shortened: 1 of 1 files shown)


lib/pure/asyncnet.nim
```diff
@@ -713,25 +713,27 @@ elif defined(nimdoc):
...
-          SSL_shutdown(socket.sslHandle)
-        else:
-          0
-      SSL_free(socket.sslHandle)
-      if res == 0:
-        discard
-      elif res != 1:
-        raiseSSLError()
-  socket.closed = true # TODO: Add extra debugging checks for this.
+  if not socket.closed:
```
dom96: Prefer: 
```
if socket.closed: return
```

Style rather than wrapping everything in more and more nested ifs.

dom96 (changes requested):
IMO an exception should be raised if you try to close an already closed socket. Maybe you can raise an exception instead of doing nothing?

alaviss:
> IMO an exception should be raised if you try to close an already closed socket. Maybe you can raise an exception instead of doing nothing.

`net` currently doesn't raise any exception so we shouldn't raise any here either.

dom96:
it always calls `close` on the fd though, maybe that will raise?

alaviss:
Nah, `nativesockets.close()` ignores all errors.

dom96:
That sounds dangerous, but alright, in any case I would still match the semantics in `net` and call `close` nevertheless.

alaviss:
Currently I'm not doing that because I'm not invalidating the socket. I can certainly do so, but IMO it's better to do that with the follow up refactoring to get rid of the `closed` flag.

lib/pure/asyncnet.nim
Araq: No, nested ifs are better.

PR #15196, opened 2020-08-17
Load libaries by absolute path on NixOS

### LINKED ISSUE #15194: Dynamic loading of libraries broken on NixOS
Nim binaries run on NixOS give an error when loading libraries that look similar to this:
`could not load: libcrypto.so(.1.1|.1.0.2|.1.0.1|.1.0.0|.0.9.9|.0.9.8|.47|.46|.45|.44|.43|.41|.39|.38|.10|)`.

This is because NixOS does not have a system-wide directory that libraries can be loaded from by name (`libcrypto.so`). Normally a binary while have a RUNPATH/rpath set in the ELF header that contains absolute paths to packages in the Nix store where the loader will look for libraries, for example: `/nix/store/…-glibc-2.31/lib:/nix/store/…-gcc-9.3.0-lib/lib`. During the Nix build process for packages these paths are added by detecting and passing the necessary `-rpath` flags to the linker.

This problem can be fixed by amending the binary RUNPATH after compilation using the `patchelf` tool, but I cannot think of any easy way to detect which libraries the binary might load at runtime, nor how to perform the patch automatically outside of building NIx packages.

I propose to adjust the Nim compiler to load libraries on NixOS by absolute path rather than simple names, and to detect the absolute at compile time. This would fix the loading problem and improve performance by avoiding the heuristic library search at runtime. This behavior could be controlled by an additional option for the Nim compiler, and the compiler take its search paths from environmental variables that are present in Nix package builds and development environments.

ghost:
If "nixbuild" is defined then choose dynamic runtime libraries by
searching $NIX_LDFLAGS at compile-time.

Fix #15194

### DIFF
--- a/lib/pure/dynlib.nim
+++ b/lib/pure/dynlib.nim
@@ -56,6 +56,9 @@
 
 import strutils
 
+when defined(nixbuild):
+  import os
+
 type
   LibHandle* = pointer ## a handle to a dynamically loaded library
 
@@ -95,6 +98,27 @@ proc libCandidates*(s: string, dest: var seq[string]) =
       libCandidates(prefix & middle & suffix, dest)
   else:
     add(dest, s)
+  when defined(nixbuild):
+    # Nix doesn't have a global library directory so
+    # load libraries using an absolute path if one
+    # can be derived from NIX_LDFLAGS.
+    #
+    # During Nix/NixOS packaging the line "define:nixbuild"
+    # should be appended to the ../../config/nim.cfg file
+    # to enable this behavior by default.
+    #
+    let nixLdflags = getEnv("NIX_LDFLAGS")
+    if nixLdflags != "":
+      var libDirs: seq[string]
+      for flag in nixLdflags.split:
+        if flag.startsWith("-L"):
+          libDirs.add(flag[2..flag.high])
+      for lib in dest:
+        for dir in libDirs:
+          let abs = dir / lib
+          if existsFile(abs):
+            dest = @[abs]
+            return
 
 proc loadLibPattern*(pattern: string, globalSymbols = false): LibHandle =
   ## loads a library with name matching `pattern`, similar to what `dynlib`


Varriount:
Is there a way to implement this that doesn't require another build flag? Does NixOS require all programs to add in special support for manual shared library loading?

ghost:
The compiler/stdlib could always check the contents of NIX_LDFLAGS, or not at all and we patch the compiler when building for Nix. I think the build flag is should be a non-issue because for Nix we can just set it permanently in nim.cfg during installation and no one needs to know it exists. Normally stuff just works on NixOS, but if something is using `dlopen` then it usually needs patched, and we like to upstream patches whenever possible.

dom96:
Why is NixOS being a special case here? Why does it need this different mechanism, it's a Linux distro after all.

ghost:
> Why is NixOS being a special case here? Why does it need this different mechanism, it's a Linux distro after all.

Because it doesn't follow FHS and each package/lib resides in it's own folder in /nix/store/

lib/pure/dynlib.nim
```diff
@@ -95,6 +98,27 @@ proc libCandidates*(s: string, dest: var seq[string]) =
...
   else:
     add(dest, s)
+  when defined(nixbuild):
+    # Nix doesn't have a global library directory so
+    # load libraries using an absolute path if one
+    # can be derived from NIX_LDFLAGS.
+    #
+    # During Nix/NixOS packaging the line "define:nixbuild"
+    # should be appended to the ../../config/nim.cfg file
+    # to enable this behavior by default.
```
alaviss, line 108: Or you can just use `--dynlibOverrideAll` then `--passL` to link as you'd like.

alaviss:
I think for Nix you can just use a custom `config.nims` that import `NIX_LDFLAGS`, then collect and convert `-L` into `-Wl,-rpath`, then add the result to `passL`. There should be no need for editing the stdlib.

Alternatively, you can use `--dynlibOverrideAll` then `--passL:"$NIX_LDFLAGS"` manually (which I presume would be more preferable).

ghost:
The problem is described in #15194.

Yes, manually passing the correct flags makes the correct thing happen, each package built from Nim sources in Nixpkgs passes special flags to fix this problem, and its the least preferable solution.

ghost:
@ehmry you can edit global nim.cfg to add these flags for the compiler and all programs (config/nim.cfg)

FRidh:
Maybe we could create a hook in Nixpkgs that would consider `NIX_LDFLAGS` and create such `nim.cfg` file. However, I am worried that (nor `config.nims`) may not be sufficient because

> Command line settings have priority over configuration file settings.

For more background, search also for `NIX_LDFLAGS` in https://nixos.org/nixpkgs/manual/#ssec-setup-hooks.

alaviss:
> However, I am worried that (nor `config.nims`) may not be sufficient because
> 
> > Command line settings have priority over configuration file settings.

Can you elaborate?

alaviss (changes requested):
I fail to see the benefit of this change over just changing all `-L` into `-Wl,-rpath`. In fact, the latter will even be more flexible as users can add into the search path with `LD_LIBRARY_PATH` as needed.

ghost:
> > However, I am worried that (nor `config.nims`) may not be sufficient because
> > > Command line settings have priority over configuration file settings.
> 
> Can you elaborate?

Can you point us to the documentation for `config.nims`.

alaviss:
> Can you point us to the documentation for `config.nims`.

Here you go: https://nim-lang.org/docs/nims.html

`config.nims` is sourced from any standard location, and for a distribution storing them in `/etc/nim/config.nims` or `$nim/config/config.nims` is a good way to ensure that they're always applied.

For packaging in a distribution I'd recommend `--dynlibOverrideAll` whenever possible, as it would allow you to reuse standard ELF tooling for dependency tracking (and also more consistent loading decisions). The auto `-L` to `-Wl,-rpath` will potentially be useful in a `nix-env` (users might not care too much about figuring all the flags for use with `--dynlibOverrideAll`), though I've not read up too much about how that environment is setup.

FRidh:
So we could modify `config/config.nims` or `config/nim.cfg` to set `-L`. Checking the default [`config/nim.cfg`](https://github.com/nim-lang/Nim/blob/devel/config/nim.cfg) I think in the future we may need to clean some things there anyway, for when using `nix-build`, e.g. use `cc %= "$CC"` instead of  the current default `cc = gcc`. However, I think also worth considering is what if you want to make a build with Nim for a non-Nix system. Not of interest to me, but worth considering.

disruptek:
It would be much better to modify a `.cfg` file since these can be interpreted without launching a NimScript interpreter with all of its coincident overhead.

FRidh:
> I think in the future we may need to clean some things there anyway, for when using nix-build, e.g. use `cc %= "$CC"` instead of the current default `cc = gcc`

`cc` is actually an object so it should probably be `cc.exe`. The annoying part is that we need to add these flags regardless of which compiler is used. Consider

> However, I think also worth considering is what if you want to make a build with Nim for a non-Nix system. Not of interest to me, but worth considering.

it may be worth creating a `cc = nix` but that seems ugly.

Edit:

There seems to be an [`envcc`](https://github.com/nim-lang/Nim/blob/49d4b50fe52f0e961ef367a44be8b1bd79994a03/compiler/extccomp.nim#L243) compiler that may be of interest as well instead.

alaviss:
> There seems to be an [`envcc`](https://github.com/nim-lang/Nim/blob/49d4b50fe52f0e961ef367a44be8b1bd79994a03/compiler/extccomp.nim#L243) compiler that may be of interest as well instead.

Note that this new `cc` will only be available for 1.4.0+ (unreleased), so you will need an another solution for older nim compilers (you can probably just backport the commit, but I haven't checked if that's possible).

Personally I would not recommend `envcc` but instead you should use `gcc` or `clang` as those may include special codegen optimized for the target compiler (currently there are none, but there might be some in the future).

For adding flags into builds you can use `passC` and `passL`, both of which are applied for all compilers and won't be overridden by the command line (they stack up instead).

FRidh:
> For adding flags into builds you can use passC and passL, both of which are applied for all compilers and won't be overridden by the command line (they stack up instead).

Good, that was what I was worried about initially!

In what format are the flags supposed to be passed in? Comma or space separated? I would expect the latter, however
```sh
$ tail -n1 config/nim.cfg
--passL="$NIX_LDFLAGS"
```
One should not use `%=` here, right?
```
Hint:  [Link]
gcc: error: /nix/store/xhdxlvh2i2pqm8vnw9f1lwlbk90hdikf-nim-1.2.6/lib64: No such file or directory
gcc: error: /nix/store/xhdxlvh2i2pqm8vnw9f1lwlbk90hdikf-nim-1.2.6/lib: No such file or directory
gcc: error: unrecognized command line option '-rpath'
gcc: error: unrecognized command line option '-rpath'
Error: execution of an external program failed: 'gcc   -o /build/nim-1.2.6/koch  /build/.cache/nim/koch_d/stdlib_assertions.nim.c.o /build/.cache/nim/koch_d/stdlib_dollars.nim.c.o /build/.cache/nim/koch_d/stdlib_io.nim.c.o /build/.cache/nim/koch_d/stdlib_system.nim.c.o /build/.cache/nim/koch_d/stdlib_parseutils.nim.c.o /build/.cache/nim/koch_d/stdlib_math.nim.c.o /build/.cache/nim/koch_d/stdlib_unicode.nim.c.o /build/.cache/nim/koch_d/stdlib_strutils.nim.c.o /build/.cache/nim/koch_d/stdlib_pathnorm.nim.c.o /build/.cache/nim/koch_d/stdlib_posix.nim.c.o /build/.cache/nim/koch_d/stdlib_times.nim.c.o /build/.cache/nim/koch_d/stdlib_os.nim.c.o /build/.cache/nim/koch_d/stdlib_parseopt.nim.c.o /build/.cache/nim/koch_d/stdlib_hashes.nim.c.o /build/.cache/nim/koch_d/stdlib_strtabs.nim.c.o /build/.cache/nim/koch_d/stdlib_streams.nim.c.o /build/.cache/nim/koch_d/stdlib_cpuinfo.nim.c.o /build/.cache/nim/koch_d/stdlib_osproc.nim.c.o /build/.cache/nim/koch_d/stdlib_sets.nim.c.o /build/.cache/nim/koch_d/@mtools@skochdocs.nim.c.o /build/.cache/nim/koch_d/stdlib_base64.nim.c.o /build/.cache/nim/koch_d/stdlib_uri.nim.c.o /build/.cache/nim/koch_d/stdlib_strformat.nim.c.o /build/.cache/nim/koch_d/@mtools@sdeps.nim.c.o /build/.cache/nim/koch_d/@mkoch.nim.c.o  -lm -lrt  $NIX_LDFLAGS  -ldl'
```
```sh
$ echo $NIX_LDFLAGS
-rpath /nix/store/hn1y9dinsrcxfbgjiymv3fhzd10k9m12-nim-1.2.6/lib64 -rpath /nix/store/hn1y9dinsrcxfbgjiymv3fhzd10k9m12-nim-1.2.6/lib -L/nix/store/a0jxdf0wc13l4270fgq9n1f42x45pq69-openssl-1.1.1g/lib -L/nix/store/5gdc4wn4nx74dx5mv2k3gqx8hh4q72mg-pcre-8.44/lib -L/nix/store/s2kyldk2s42n5z2ijjj1v5ns78n9wzr3-ncurses-6.2/lib -L/nix/store/v7b326xn4wq9n4wqjdxar3xadz73mi8f-readline-6.3p08/lib -L/nix/store/19a50vv9bv66rsw91anv8r6avnzrbwxy-boehm-gc-8.0.4/lib -L/nix/store/iwmz9cl5x7xyzdfm3phxn7y4krz27m5j-sfml-2.5.1/lib -L/nix/store/f7lws879jarjs40qznw95pwz4ngc7ib0-sqlite-3.32.3/lib -L/nix/store/a0jxdf0wc13l4270fgq9n1f42x45pq69-openssl-1.1.1g/lib -L/nix/store/5gdc4wn4nx74dx5mv2k3gqx8hh4q72mg-pcre-8.44/lib -L/nix/store/s2kyldk2s42n5z2ijjj1v5ns78n9wzr3-ncurses-6.2/lib -L/nix/store/v7b326xn4wq9n4wqjdxar3xadz73mi8f-readline-6.3p08/lib -L/nix/store/19a50vv9bv66rsw91anv8r6avnzrbwxy-boehm-gc-8.0.4/lib -L/nix/store/iwmz9cl5x7xyzdfm3phxn7y4krz27m5j-sfml-2.5.1/lib -L/nix/store/f7lws879jarjs40qznw95pwz4ngc7ib0-sqlite-3.32.3/lib
```

alaviss:
> In what format are the flags supposed to be passed in?

The same format as what the gcc driver accepts (since we pass the flags directly to gcc during link): `-Wl,<flag-here>`

ghost:
Long term we want to get rid of rpath for Nix and I don't see any reason to look up something at runtime when we can know it with certainty at compile-time. I'd rather just patch the compiler downstream than deal with this configuration stuff.

alaviss:
> Long term we want to get rid of rpath for Nix and I don't see any reason to look up something at runtime when we can know it with certainty at compile-time. 

But you are still doing lookups at runtime? All this change does is to generate library candidates for the compiler with the paths prefixed. Unless I'm reading this wrong, this means you are basically doing `dlopen()` job for searching within a set of defined path, similar to rpath, while being more inflexible.

Also note that `dynlib` module is Nim's `dlopen` and is also used by programs outside of stdlib. With this change of yours every Nim program compiled with `-d:nixbuild` will look at `NIX_LDFLAGS` whenever `loadLibPattern` is used, and that's not something you should blanketly apply. If you would like to localize this in the compiler, then do so by providing a `libCandidate` within the compiler source (it can just call `dynlib.libCandidate` then manually append the paths to the generated seq, not super efficient but not super slow either).

I'm still not convinced that this solution is better than rpath.

alaviss:
Ah, shame on me, I am reading this wrong. I didn't notice the existence check.

This does seem to be something you'd want to do in the compiler though. The `distros` module should have the necessary tooling to detect NixOS so you don't have to use a special define flag.

FedericoCeratto:
@ehmry FYI there is https://nim-lang.github.io/Nim/packaging.html - can you please keep it up to date with any relevant information? Thanks

ghost:
This is something that should be patched downstream.

PR #15201, opened 2020-08-18
Improve httpClient docs on SSL cert verification

FedericoCeratto:
Clarify that cert verification is enabled by default after CVE-2021-29495
Add example and usage instructions.

### DIFF
--- a/lib/pure/httpclient.nim
+++ b/lib/pure/httpclient.nim
@@ -115,7 +115,7 @@
 ##
 ## SSL/TLS support
 ## ===============
-## This requires the OpenSSL library, fortunately it's widely used and installed
+## This requires the OpenSSL library. Fortunately it's widely used and installed
 ## on many operating systems. httpclient will use SSL automatically if you give
 ## any of the functions a url with the `https` schema, for example:
 ## `https://github.com/`.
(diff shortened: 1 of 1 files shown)


lib/pure/httpclient.nim
```diff
@@ -129,6 +129,17 @@
 ## A set of directories and files from the `ssl_certs <ssl_certs.html>`_
 ## module are scanned to locate CA certificates.
 ##
+## .. code-block:: Nim
```
dom96, line 133: Please write a little blurb to introduce this example, otherwise it's just an example without any context.

lib/pure/httpclient.nim
```diff
@@ -129,6 +129,17 @@
...
+##    import httpclient
+##    var client = newHttpClient(sslContext=newContext(verifyMode=CVerifyPeer))
+##
+## There are three options for verify mode:
+## ``CVerifyNone``: certificates are not verified;
+## ``CVerifyPeer``: certificates are verified;
+## ``CVerifyPeerUseEnvVars``: certificates are verified and the optional
+## environment variables SSL_CERT_FILE and SSL_CERT_DIR are also used to
+## locate certificates
```
dom96:
```suggestion
##   * ``CVerifyNone``: certificates are not verified;
##   * ``CVerifyPeer``: certificates are verified;
##   * ``CVerifyPeerUseEnvVars``: certificates are verified and the optional
##     environment variables SSL_CERT_FILE and SSL_CERT_DIR are also used to
##     locate certificates
```

dom96 (changes requested):

lib/pure/httpclient.nim
FedericoCeratto: Done

PR #15206, opened 2020-08-20
Added more SSL documentation to `net` module.

JohnAD:
As promised on the Nim forum, here is a little bit more documentation to guide to the new user on how to make an non-async SSL connection via the `net` module. This includes a simple example and a notice on the version of `connect` that does not support SSL to use the other `connect` procedure.

### DIFF
--- a/lib/pure/net.nim
+++ b/lib/pure/net.nim
@@ -1954,7 +1954,8 @@ proc connect*(socket: Socket, address: string, port = Port(0),
   ## the connection to the server to be made.
   ##
   ## **Warning:** This procedure appears to be broken for SSL connections as of
-  ## Nim v1.0.2. Consider using the other `connect` procedure.
+  ## Nim v1.0.2. Consider using the other `connect` procedure. See
+  ## https://github.com/nim-lang/Nim/issues/15215 for more info.
   socket.fd.setBlocking(false)
 
   socket.connectAsync(address, port, socket.domain)
--- a/lib/pure/net.nim
+++ b/lib/pure/net.nim
@@ -38,7 +38,7 @@
 ##   var socket = newSocket()
 ##   socket.connect("google.com", Port(80))
 ##
-## For SSL, use the next example.
+## For SSL, use the following example (and make sure to compile with ``-d:ssl``):
 ##
 ## .. code-block:: Nim
 ##   var socket = newSocket()
--- a/lib/pure/net.nim
+++ b/lib/pure/net.nim
@@ -20,7 +20,8 @@
 ## ====
 ##
 ## In order to use the SSL procedures defined in this module, you will need to
-## compile your application with the ``-d:ssl`` flag.
+## compile your application with the ``-d:ssl`` flag. See the `newContext`
+## procedure for additional details.
 ##
 ## Examples
 ## ========
@@ -36,8 +37,16 @@
 ##   var socket = newSocket()
 ##   socket.connect("google.com", Port(80))
 ##
+## For SSL, use the next example.
+##
+## .. code-block:: Nim
+##   var socket = newSocket()
+##   var ctx = newContext()
+##   wrapSocket(ctx, socket)
+##   socket.connect("google.com", Port(443))
+##
 ## UDP is a connectionless protocol, so UDP sockets don't have to explicitly
-## call the ``connect`` procedure. They can simply start sending data
+## call the `connect` procedure. They can simply start sending data
 ## immediately.
 ##
 ## .. code-block:: Nim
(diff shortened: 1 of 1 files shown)


lib/pure/net.nim
```diff
@@ -36,8 +37,16 @@
...
+##
+## .. code-block:: Nim
+##   var socket = newSocket()
+##   var ctx = newContext()
+##   wrapSocket(ctx, socket)
+##   socket.connect("google.com", Port(443))
+##
 ## UDP is a connectionless protocol, so UDP sockets don't have to explicitly
-## call the ``connect`` procedure. They can simply start sending data
+## call the `connect` procedure. They can simply start sending data
```
Varriount: Doesn't this need 2 backticks? Otherwise it's just indented.

Varriount (changes requested):

lib/pure/net.nim
alaviss: Nim doc uses a RST/MD hybrid parser that can handle this, though it'd be better to link to the `proc` directly instead.

JohnAD: I used single backticks to comply with the Contributor Guidelines. As suggested I changed the `connect` and `newContext` references to internal links.

lib/pure/net.nim
```diff
@@ -36,9 +38,17 @@
...
+## .. code-block:: Nim
+##   var socket = newSocket()
+##   var ctx = newContext()
+##   wrapSocket(ctx, socket)
+##   socket.connect("google.com", Port(443))
+##
 ## UDP is a connectionless protocol, so UDP sockets don't have to explicitly
-## call the ``connect`` procedure. They can simply start sending data
-## immediately.
+## call the `connect<net.html#connect%2CSocket%2Cstring>`_ procedure. They can
```
alaviss:
```suggestion
## call the `connect <net.html#connect%2CSocket%2Cstring>`_ procedure. They can
```

Very minor styling change.

We might also want to note that on certain systems (at the very least most POSIX systems), `connect` can be used to bound UDP sockets to a peer, and which will cause all `send/recv` commands after to send/receive only to/from that peer. I've not verified whether this behavior also apply to Nim's `Socket`, so please verify before adding this into the documentation.

lib/pure/net.nim
```diff
@@ -20,7 +20,9 @@
 ## ====
 ##
 ## In order to use the SSL procedures defined in this module, you will need to
-## compile your application with the ``-d:ssl`` flag.
+## compile your application with the ``-d:ssl`` flag. See the
+## `newContext<net.html#newContext%2Cstring%2Cstring%2Cstring%2Cstring>`_
```
dom96: This doesn't lead anywhere: https://nim-lang.org/docs/net.html#newContext%2Cstring%2Cstring%2Cstring%2Cstring

lib/pure/net.nim
```diff
@@ -36,9 +38,17 @@
 ##   var socket = newSocket()
 ##   socket.connect("google.com", Port(80))
 ##
+## For SSL, use the next example.
```
dom96:
```suggestion
## For SSL, use the following example (and make sure to compile with ``-d:ssl``):
```

lib/pure/net.nim
```diff
@@ -1942,6 +1952,9 @@ proc connect*(socket: Socket, address: string, port = Port(0),
   ##
   ## The ``timeout`` parameter specifies the time in milliseconds to allow for
   ## the connection to the server to be made.
+  ##
+  ## Do not use this version of `connect` for an SSL socket. Instead use the
+  ## version with no `timeout` parameter.
```
dom96: Why?

dom96 (changes requested):

FedericoCeratto:
Related to #15201

lib/pure/net.nim
JohnAD: Not sure yet. The other `connect` is clearly marked as working with SSL and this one specifically is not. And this one creates an SSL handshake error; both in current nim and, from testing, in past versions including 1.0.2 (at least on linux.)

https://forum.nim-lang.org/t/6686#41559

lib/pure/net.nim
JohnAD: Odd. It worked on my local machine. I suspect there was a compiler change at some point for the number of parameters used in `doc` anchor generation. I will fix it manually.

lib/pure/net.nim
dom96: This one does however have code that is SSL specific. Please create an issue and change this comment to include a link to that issue. Something like "\*\*Warning:\*\* This procedure appears to be broken for SSL connections as of Nim v1.0.2. See issue https://.... for more info"

JohnAD: Done. Thanks!

PR #15211, opened 2020-08-22
fix some issues overloading with generics and inheritance

jcosborn:


compiler/sigmatch.nim
```diff
@@ -296,6 +317,9 @@ proc cmpCandidates*(a, b: TCandidate): int =
   # the other way round because of other semantics:
   result = b.inheritancePenalty - a.inheritancePenalty
   if result != 0: return
+  # check for generic subclass relation
+  result = checkGeneric(a, b)
```
Araq, line 321: Please outline why the simple counting approach that ``sumGeneric`` currently implements does not suffice for generic inheritance.

jcosborn: Currently `sumGeneric` favors `Bar[bool or int]` over either `Bar[int]` or `Bar[T]`.  That could be fixed to favor `Bar[int]`, but it couldn't really distinguish between `Bar[bool or int]` and `Bar[T]`.  Also it can easily be fooled with aliases like `type Id[T]=T` so that `Bar[Id[T]]` would win over `Bar[T]` even though they are really equivalent.  Fixing that would require more work.  In the end it seemed that just using `typeRel` would solve all of these issues and be the more robust solution.

Araq: I fear potential wasteful or infinite recursions. Can't we have some specialized code dealing with this problem?

jcosborn: `sumGeneric` isn't called from `typeRel`, so I don't think infinite recursion is possible.  We need to directly compare the two constraints, so some form of `typeRel` is needed.  I'm not sure how much could be removed and still work correctly.  If you are worried about the performance, we could try splitting out some of the most common cases from `typeRel` and make them callable directly or from within `typeRel` so that they can be reused, or maybe just make common paths in `typeRel` faster.  I'm wary of duplicating it since that just makes more code to debug later (there are several more bugs in generics that still need to be fixed).  I'd prefer to start with something that is as correct as possible before trying to optimize it too much.

Araq: Alright.

Clyybber: With this PR https://play.nim-lang.org/#ix=2vwQ hits an infinite recursion.

jcosborn: This should be fixed by #15241.

PR #15222, opened 2020-08-24
fix in doc: incomplete output

tiye:
According to https://github.com/nim-lang/Nim/commit/b40a637fac1334339c260951d0dd3fa3682de476#r41709236 .

### DIFF
--- a/lib/pure/collections/tables.nim
+++ b/lib/pure/collections/tables.nim
@@ -131,7 +131,7 @@
 ##   let myString = "abracadabra"
 ##   let letterFrequencies = toCountTable(myString)
 ##   echo letterFrequencies
-##   # 'a': 5, 'b': 2, 'c': 1, 'd': 1, 'r': 2}
+##   # output: {'a': 5, 'b': 2, 'c': 1, 'd': 1, 'r': 2}
 ##
 ## The same could have been achieved by manually iterating over a container
 ## and increasing each key's value with `inc proc


narimiran:
Other examples are the same as a pre-PR version. I don't understand why only this example should be changed.

Adding just the missing `{` should be enough.

tiye:
my original comment was to ask if we missed the `{` in the doc https://github.com/nim-lang/Nim/commit/b40a637fac1334339c260951d0dd3fa3682de476#r41709236 . I was not sure if I need to create this PR... was that by indention?...


https://nim-lang.org/docs/tables.html#basic-usage-counttable
image

ringabout:
But another `countTable` example after this used `# output: {'a': 5, 'b': 2, 'c': 1, 'd': 1, 'r': 2} `

https://github.com/nim-lang/Nim/blob/07d82b5cc4b7ba177de43a3eb46c7c645930421a/lib/pure/collections/tables.nim#L148

lib/pure/collections/tables.nim
```diff
@@ -131,7 +131,7 @@
 ##   let myString = "abracadabra"
 ##   let letterFrequencies = toCountTable(myString)
 ##   echo letterFrequencies
-##   # 'a': 5, 'b': 2, 'c': 1, 'd': 1, 'r': 2}
+##   # output: { 'a': 5, 'b': 2, 'c': 1, 'd': 1, 'r': 2}
```
ringabout: One extra space， should be `{'a': 5, 'b': 2, 'c': 1, 'd': 1, 'r': 2}`
like this: https://github.com/nim-lang/Nim/blob/07d82b5cc4b7ba177de43a3eb46c7c645930421a/lib/pure/collections/tables.nim#L148

tiye: going to force push...

(...pushed)

PR #15226, opened 2020-08-25
Add eopkg to package manager list

rocket111185:
Added 2 lines in order to avoid mistake in package manager detecting (otherwise it detects Solus package manager as "pkg" [Solaris])


Varriount:
Is this merging into the correct branch? As far as I know, devel is what PRs are normally merged into.

Araq:
It is not the correct branch indeed.

PR #15227, opened 2020-08-26
Raise KeyError if passed an invalid entry in `parsecsv.rowEntry`

dsrw:
Motivated by https://stackoverflow.com/q/63577715/14164200

`parsecsv.rowEntry` doesn't return anything if passed an invalid row header name, which leads to a SIGSEGV since (I believe) the proc is basically returning nil and strings can't be nil.

That said, should it also return `string` instead of `var string`? I can't think of any reason why the result should be mutable, and using `string` would have prevented the SIGSEGV.

Lastly (sorry!) shouldn't `var T` procs either require a result for non-nillable types , or return the default value for `T` if one is missing? Should I create an issue for this?

### DIFF
--- a/lib/pure/parsecsv.nim
+++ b/lib/pure/parsecsv.nim
@@ -322,24 +322,29 @@ proc rowEntry*(my: var CsvParser, entry: string): var string =
   ##
   ## Assumes that `readHeaderRow <#readHeaderRow,CsvParser>`_ has already been
   ## called.
+  ## 
+  ## If specified `entry` does not exist, raises KeyError.
   runnableExamples:
     import streams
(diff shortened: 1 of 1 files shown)


dom96 (commented):
Nice, but AFAIK the convention is to raise an IndexError (and only in debug mode) which I think this code will already do. Maybe this could be better documented instead?

alaviss:
> I think this code will already do

It won't, `find` returns `-1` when the needle is not found.

Araq:
Good enough for now but as you noticed there is also a compiler issue, the compiler should flag the code as invalid. (In fact, it does so but it only produces a warning, I think.)

PR #15228, opened 2020-08-26
Add converter from Option[T] to bool

ire4ever1190:
Adds a small converter for syntactic sugar in checking if `Option[T]` contains a value
```nim
var value = some("some value")
# This
if value:
    # Do work on value
# Instead of this
if value.isSome:
    # Do work on value
```

### DIFF
--- a/lib/pure/options.nim
+++ b/lib/pure/options.nim
@@ -372,6 +372,19 @@ proc unsafeGet*[T](self: Option[T]): lent T {.inline.}=
   assert self.isSome
   result = self.val
 
+converter toBool*[T](self: Option[T]): bool =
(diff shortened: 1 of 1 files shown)


bluenote10:
This looks dangerous. Doesn't it lose type safety of `Option[bool]`?

ire4ever1190:
~~it shouldn't because if a value of `Option[bool]` is `none` then it will return false since `isSome` returns false~~
never mind, I see what you are saying
yes it would mess up in that situation

alehander92:
this is a good idea if we automatically refine the type inside the branch: but I think people don't like it because it's .. a different type after all?

dom96:
big no from me on this.

PR #15233, opened 2020-08-27
Add SSL_CTX_set_session_id_context

ardek66:
Some clients might attempt to resume a session which require initialisation of the id context using `SSL_CTX_set_session_id_context` by the server. From what I noticed, openssl.nim lacks this function, resulting in possible runtime errors since the id context cannot be initialised.

https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_set_session_id_context.html

> If the session id context is not set on an SSL/TLS server and client certificates are used, stored sessions will not be reused but a fatal error will be flagged and the handshake will fail.

https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set_session_cache_mode.html

> Since a client may try to reuse a session intended for use in a different context, the session id context must be set by the server (see SSL_CTX_set_session_id_context(3)).

https://github.com/openssl/openssl/issues/10168#issuecomment-542654427

### DIFF
--- a/lib/pure/net.nim
+++ b/lib/pure/net.nim
@@ -825,6 +825,9 @@ when defineSsl:
     else:
       result = getPeerCertificates(socket.sslHandle)
 
+  proc `sessionIdContext=`*(ctx: SslContext, sidCtx: string) =
+      SSL_CTX_set_session_id_context(ctx.context, sidCtx, sidCtx.len)
+  
 proc getSocketError*(socket: Socket): OSErrorCode =
   ## Checks ``osLastError`` for a valid error. If it has been reset it uses
   ## the last error stored in the socket object.
--- a/lib/pure/net.nim
+++ b/lib/pure/net.nim
@@ -828,6 +828,9 @@ when defineSsl:
   proc `sessionIdContext=`*(ctx: SslContext, sidCtx: string) =
       ## Set the session id context, used for resuming handshakes
       ## Must be unique to application.
+      ## sidCtx must be at most 32 characters in length.
+      if sidCtx.len > 32:
+        raiseSSLError("sessionIdContext must be shorter than 32 characters")
       SSL_CTX_set_session_id_context(ctx.context, sidCtx, sidCtx.len)
   
 proc getSocketError*(socket: Socket): OSErrorCode =
--- a/lib/pure/net.nim
+++ b/lib/pure/net.nim
@@ -826,6 +826,8 @@ when defineSsl:
       result = getPeerCertificates(socket.sslHandle)
 
   proc `sessionIdContext=`*(ctx: SslContext, sidCtx: string) =
+      ## Set the session id context, used for resuming handshakes
+      ## Must be unique to application.
       SSL_CTX_set_session_id_context(ctx.context, sidCtx, sidCtx.len)
   
 proc getSocketError*(socket: Socket): OSErrorCode =
--- a/lib/pure/net.nim
+++ b/lib/pure/net.nim
@@ -826,12 +826,20 @@ when defineSsl:
       result = getPeerCertificates(socket.sslHandle)
 
   proc `sessionIdContext=`*(ctx: SslContext, sidCtx: string) =
-      ## Set the session id context, used for resuming handshakes
-      ## Must be unique to application.
-      ## sidCtx must be at most 32 characters in length.
-      if sidCtx.len > 32:
-        raiseSSLError("sessionIdContext must be shorter than 32 characters")
-      SSL_CTX_set_session_id_context(ctx.context, sidCtx, sidCtx.len)
+    ## Sets the session id context in which a session can be reused.
+    ## Used for permitting clients to reuse a session id instead of
+    ## doing a new handshake.
+    ##
+    ## TLS clients might attempt to resume a session using the session id context,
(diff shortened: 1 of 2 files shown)


ghost:
Ping @alaviss  :)

ardek66:
I made a simple example with where using the sessionId setter would be useful.
Consider we have this basic server:

```nim
import net

let ctx = newContext(certFile="mycert.pem", keyFile="mykey.pem")
var socket = newSocket()
socket.setSockOpt(OptReuseAddr, true)
socket.setSockOpt(OptReusePort, true)
socket.bindAddr(Port(1234))
socket.listen()
ctx.wrapSocket(socket)

var address = ""
while true:
  var client = newSocket(AF_INET, SOCK_STREAM, IPPROTO_TCP, true)
  socket.acceptAddr(client, address)
  let line = client.recvLine()
  if line.len > 0:
    echo address, ": ", line
    client.close()
```
We can connect to this just fine using `openssl s_client -connect localhost:1234`.
If we try to reconnect using `openssl s_client -connect localhost:1234 -reconnect`, the server will crash with the error `:SSL routines:ACCEPT_SR_CLNT_HELLO_C:session id context uninitialized`

Now, we can use the sessionIdContext setter for SslContext
```nim
import net

let ctx = newContext(certFile="mycert.pem", keyFile="mykey.pem")
ctx.sessionIdContext = "hello world"
var socket = newSocket()
socket.setSockOpt(OptReuseAddr, true)
socket.setSockOpt(OptReusePort, true)
socket.bindAddr(Port(1234))
socket.listen()
ctx.wrapSocket(socket)

var address = ""
while true:
  var client = newSocket(AF_INET, SOCK_STREAM, IPPROTO_TCP, true)
  socket.acceptAddr(client, address)
  let line = client.recvLine()
  if line.len > 0:
    echo address, ": ", line
    client.close()
```

Now, using `openssl s_client -connect localhost:1234 -reconnect` works, and the server doesn't crash.

I have personally witnessed this problem while using my nim tls server, some clients would try to resume a session, similar to the example above, but the server would crash. I made a temporary fix in the code by importing the function with `importc` pragma, but I feel like this problem might be common enough that a simple setter shoud be included.

lib/pure/net.nim
```diff
@@ -825,6 +825,14 @@ when defineSsl:
     else:
       result = getPeerCertificates(socket.sslHandle)
 
+  proc `sessionIdContext=`*(ctx: SslContext, sidCtx: string) =
+      ## Set the session id context, used for resuming handshakes
+      ## Must be unique to application.
+      ## sidCtx must be at most 32 characters in length.
```
alaviss: I think it'd be better if the docs is elaborate more about why this is needed and what is this used for.

alaviss (commented):
/cc @dom96 @FedericoCeratto since this touches `net` API.

lib/pure/net.nim
ardek66:
``` nim
      ## Sets the session context in which a session can be reused.
      ## TLS clients might attempt to resume a session using the session context, 
      ## thus it must be set if verifyMode is set to CVerifyPeer or CVerifyPeerUseEnvVars,
      ## otherwise the connection will fail and SslError will be raised if an attempt to resume is made.
      ## Should be unique per-application to prevent clients from malfunctioning.
      ## sidCtx must be at most 32 characters in length.
```
 How is this?

ardek66:
Sorry for that accidental commit, I have reverted the base to the latest related commit so I dont pollute the git history.

lib/pure/net.nim
dom96: Seems ok, but have a look at what that looks like rendered: `nim doc -d:ssl net.nim`. You need more empty lines in between, otherwise it'll be a huge paragraph and hard to read. I would also suggest using bullet points more here.

dom96: Might also be a good idea to explain what this would be used for.

dom96 (approved):
Seems like a simple enough change to not be controversial.

lib/pure/net.nim
ardek66: Ok I updated it, verified the rendered content and it looks fine to me.
```
    ## Sets the session id context in which a session can be reused.
    ## Used for permitting clients to reuse a session id instead of
    ## doing a new handshake.
    ##
    ## TLS clients might attempt to resume a session using the session id context,
    ## thus it must be set if verifyMode is set to CVerifyPeer or CVerifyPeerUseEnvVars,
    ## otherwise the connection will fail and SslError will be raised if resumption occurs.
    ##
    ## - Only useful if set server-side.
    ## - Should be unique per-application to prevent clients from malfunctioning.
    ## - sidCtx must be at most 32 characters in length.
```

dom96: It seems you haven't committed these changes?

ardek66: Yes I haven't I just wanted to check if the docs are ok before commiting :)
I have commited it now

dom96:
CI failure looks unrelated, merged.

FedericoCeratto:
`  raiseSSLError("sessionIdContext must be shorter than 32 characters")`
Is this correct? It should be "shorter or equal"

Besides, is it worth checking SSL_MAX_SSL_SESSION_ID_LENGTH instead?

PR #15234, opened 2020-08-28
Fix #15219 SQL escape in db_mysql is not enough 

### LINKED ISSUE #15219: SQL escape in db_mysql is not enough
SQL injection is possible because proc [`dbQuote`](https://github.com/nim-lang/Nim/blob/v1.2.6/lib/impure/db_mysql.nim#L117) in [`db_mysql`](https://nim-lang.org/docs/db_mysql.html) doesn't escape backslash`\`.

### Example
```nim
import db_mysql

let theDb = open("127.0.0.1", "root", "mysql", "test")

theDb.exec(sql"Drop table if exists myTestTbl")
theDb.exec(sql("create table myTestTbl (" &
    " Id    INT(11)     NOT NULL AUTO_INCREMENT PRIMARY KEY, " &
    " Name  VARCHAR(50) NOT NULL) "))

theDb.exec(sql"START TRANSACTION")
for i in 1..10:
  theDb.exec(sql"INSERT INTO myTestTbl (name) VALUES (?)",
        "Item#" & $i)
theDb.exec(sql"COMMIT")

let request = """\' or 1 = 1 -- """

for x in theDb.fastRows(sql"select * from myTestTbl where Name = ?", request):
  echo x

theDb.close()
```

### Output
```
@["1", "Item#1"]
@["2", "Item#2"]
@["3", "Item#3"]
@["4", "Item#4"]
@["5", "Item#5"]
@["6", "Item#6"]
@["7", "Item#7"]
@["8", "Item#8"]
@["9", "Item#9"]
@["10", "Item#10"]
```

### Possible Solution

- When [`NO_BACKSLASH_ESCAPES`](https://dev.mysql.com/doc/refman/5.6/en/sql-mode.html#sqlmode_no_backslash_escapes) is disable, escape backslash`\`.
- ~~Maybe [`mysql_real_escape_string()`](https://dev.mysql.com/doc/c-api/5.6/en/mysql-real-escape-string.html) is useful.~~
When NO_BACKSLASH_ESCAPES is enable, [`mysql_real_escape_string()`](https://dev.mysql.com/doc/c-api/5.7/en/mysql-real-escape-string.html) can't be used in MySQL 5.7 and after. 
 Instead, use [`mysql_real_escape_string_quote()`](https://dev.mysql.com/doc/c-api/5.7/en/mysql-real-escape-string-quote.html).

### Additional Information
```
$ nim -v
Compiled at 2020-07-29
```

bung87:
Fix #15219

### DIFF
--- a/lib/impure/db_mysql.nim
+++ b/lib/impure/db_mysql.nim
@@ -118,7 +118,7 @@ when false:
 
 proc dbQuote*(s: string): string =
   ## DB quotes the string.
-  result = newStringOfCap(s.str.len + 2)
+  result = newStringOfCap(s.len + 2)
   result.add "'"
   for c in items(s):
     # see https://cheatsheetseries.owasp.org/cheatsheets/SQL_Injection_Prevention_Cheat_Sheet.html#mysql-escaping
--- a/lib/impure/db_mysql.nim
+++ b/lib/impure/db_mysql.nim
@@ -135,7 +135,7 @@ proc dbQuote*(s: string): string =
     of '\\': result.add "\\\\"
     of '_': result.add "\\_"
     of Letters+Digits: result.add c
-    else: result.add "\\" & c
+    else: result.add "\\" & $ord(c)
   add(result, '\'')
 
 proc dbFormat(formatstr: SqlQuery, args: varargs[string]): string =
--- a/lib/impure/db_mysql.nim
+++ b/lib/impure/db_mysql.nim
@@ -118,10 +118,24 @@ when false:
 
 proc dbQuote*(s: string): string =
   ## DB quotes the string.
-  result = "'"
+  result = newStringOfCap(s.str.len + 2)
+  result.add "'"
   for c in items(s):
-    if c == '\'': add(result, "''")
-    else: add(result, c)
+    # see https://cheatsheetseries.owasp.org/cheatsheets/SQL_Injection_Prevention_Cheat_Sheet.html#mysql-escaping
+    case c:
+      of '\0': result.add "\\0"
+      of '\b': result.add "\\b"
+      of '\t': result.add "\\t"
+      of '\l': result.add "\\n"
+      of '\r': result.add "\\r"
+      of '\x1a': result.add "\\Z"
+      of '"': result.add "\\\""
+      of '%': result.add "\\%"
+      of '\'': result.add "\\'"
+      of '\\': result.add "\\\\"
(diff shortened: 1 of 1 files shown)


narimiran:
Based on the content of the link you've provided, it seems that my fix (https://github.com/nim-lang/Nim/pull/15235) is an `ANSI SQL` mode fix, while yours is `MySQL` mode fix.

And since the existing code only had a special case for `'`, I'm guessing it was trying to follow `ANSI SQL`.

bung87:
> Based on the content of the link you've provided, it seems that my fix (#15235) is an `ANSI SQL` mode fix, while yours is `MySQL` mode fix.
> 
> And since the existing code only had a special case for `'`, I'm guessing it was trying to follow `ANSI SQL`.

I see no enough reason assuming mysql server runs on `ANSI` mode, it not the default mode of mysql or mariadb. and MySQL mode also take care of `'`.

lib/impure/db_mysql.nim
```diff
@@ -118,10 +118,24 @@ when false:
...
 proc dbQuote*(s: string): string =
   ## DB quotes the string.
-  result = "'"
+  result = newStringOfCap(s.len + 2)
+  result.add "'"
   for c in items(s):
-    if c == '\'': add(result, "''")
-    else: add(result, c)
+    # see https://cheatsheetseries.owasp.org/cheatsheets/SQL_Injection_Prevention_Cheat_Sheet.html#mysql-escaping
+    case c:
```
Araq: Use 

```nim

case c
of '\0': ...
```

No colon after the "c" and less indentation.

lib/impure/db_mysql.nim
```diff
@@ -118,10 +118,24 @@ when false:
...
+    of '\l': result.add "\\n"
+    of '\r': result.add "\\r"
+    of '\x1a': result.add "\\Z"
+    of '"': result.add "\\\""
+    of '%': result.add "\\%"
+    of '\'': result.add "\\'"
+    of '\\': result.add "\\\\"
+    of '_': result.add "\\_"
+    of Letters+Digits: result.add c
+    else: result.add "\\" & c
```
Araq: This ``else`` branch must be ``result.add "\\" & $ord(c)`` or similar. Please test this better.

bung87: yeah, the original code can add Control code to string.

PR #15235, opened 2020-08-28
fix #15219, SQL escape of `'`

### LINKED ISSUE #15219: SQL escape in db_mysql is not enough

narimiran:


narimiran:
Note to reviewer: see #15234 for a more thorough solution.

Varriount:
Just a question - doesn't MySQL have built-in quoting/interpolation that we can use?

Araq:
Merged @bung87's PR instead.

PR #15236, opened 2020-08-28
fix #15035

### LINKED ISSUE #15035: [ARC] C compiler error when passing a var openArray to a sink openArray
This code results in a C compiler error. Found in https://github.com/bung87/tim_sort. I don't know if this code is supposed to work or not :) 

### Example
```nim
proc countRun(lst: sink openArray[int]): int =
  discard

proc timSort(lst: var openArray[int]) = 
  let run = countRun(lst)

var a = @[3, 5, 4, 2, 6, 1]
timSort(a)
echo a
```

### Current Output
```
/home/dian/.cache/nim/tim_sort_d/@mtim_sort.nim.c: In function ‘timSort__iUO1FYw3YnPPl6rsf8bqVg’:
/home/dian/.cache/nim/tim_sort_d/@mtim_sort.nim.c:156:54: error: ‘colontmpD_Len_0’ undeclared (first use in this function); did you mean ‘colontmpD_’?
  156 |  run = countRun__qbT9ayGxGQuz6mSoZvl6XYg(colontmpD_, colontmpD_Len_0);
      |                                                      ^~~~~~~~~~~~~~~
      |                                                      colontmpD_
```

### Expected Output
Not sure, it should either compile (if that is valid code), or result in a compiler error message (if we can't sink openArray)

```
$ nim -v
Compiled at 2020-07-21

git hash: 911e39351e34c1022b586fe9bfc3a67b2ce3bf44
```

cooldome:


Clyybber:
I think this code shouldn't be allowed. AFAICT openArray is like a view, we don't actually own the locations, so sinking it doesn't make much sense, because then we have two owners, the proc with the sink param we passed it to, and owner of the seq/array it was a view of.

PR #15240, opened 2020-08-29
template hygiene

disruptek:
Stumbled across a lack of hygiene in a template.  I think it's improved but if not, just discard this please.


Varriount:
Thanks!

PR #15241, opened 2020-08-29
fix infinite recursion in typeRel

jcosborn:


Clyybber:
@jcosborn FYI the snippet I mentionede here https://github.com/nim-lang/Nim/pull/15211#discussion_r479564544 isn't supposed to compile, its supposed to say `(13, 23) Error: expression 'prepareTileset(parsedTileset)' has no type (or is ambiguous)`

You can specify that in the test file like so:
```nim
discard """
  errormsg: "expression 'prepareTileset(parsedTileset)' has no type (or is ambiguous)"
  line: 13
"""
```

jcosborn:
It works fine for me (no error) when I run the test on its own.  I don't know why CI fails when it is included in megatest.

Clyybber:
Oh, yeah sorry. You are right it should compile just fine.

jcosborn:
`type TilesetCT[n: static[int]] = int` seems to make CI fail.  It works fine in my own tests.
Changing it to `type TilesetCT[n: static[int]] = distinct int` seems to work with CI.

compiler/sigmatch.nim
```diff
@@ -1091,7 +1091,7 @@ proc typeRel(c: var TCandidate, f, aOrig: PType,
 
   template considerPreviousT(body: untyped) =
     var prev = PType(idTableGet(c.bindings, f))
-    if prev == nil: body
+    if prev == nil or prev == f: body
```
Araq: Your PR is great! Unfortunately I don't understand it. How could it be that ``prev == f``?

jcosborn: Good point.  I had only fixed the symptom.  `TilesetCT` was being compared against itself and `trDontBind` wasn't being preserved.  This is now fixed.

PR #15245, opened 2020-08-30
add stringify for UncheckedArray[Utf16Char]

ringabout:
I need `WideCString` to do some ffis with win32 api. `WideCString` works fine as the parameter of the function. However, I have some problems with it when I use it as the element of the struct. https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-file_notify_information

I have to use `UncheckedArray[Utf16Char]`.
```nim
type
  FILE_NOTIFY_INFORMATION* = object
    NextEntryOffset*: DWORD
    Action*: DWORD
    FileNameLength*: DWORD
    FileName*: UncheckedArray[Utf16Char]
```
But there is no `$` function for `UncheckedArray[Utf16Char]`.


Araq:
1. ``UncheckedArray[Utf16Char]`` is too low level and shouldn't offer a ``$``.
2. ``UncheckedArray[Utf16Char]`` is not a valid parameter type, it should be ``ptr UncheckedArray[Utf16Char]`` (that it compiles is a compiler bug).

ringabout:
Thanks. I have a workaround with some copies.

```nim
 var tmp = newWideCString("", info.FileNameLength.int div 2)
 for idx in 0 ..< info.FileNameLength.int div 2:
    tmp[idx] = info.FileName[idx]

 let name = $tmp
```
And close now.

PR #15251, opened 2020-08-30
new-style concepts implementation, WIP

Araq:
don't merge yet. Todo:

- [x] Make new style concepts trigger implicit generics in ``proc p(c: MyConcept)``.
- [x] The Dictionary[K, V] concept must work.
- [ ] Infer ``static T`` values.


tests/concepts/tspec.nim
```diff
@@ -0,0 +1,69 @@
...
+2
+3'''
+  joinable: false
+"""
+
+import hashes
+
+type
+  Comparable = concept # no T, an atom
+    proc cmp(a, b: self): int
```
dom96: I think I suggested this already, but will repeat: I would prefer `Comparable` instead of the `self`. IMO it aids in code comprehension and removes the need for this magic `self`.

Edit: indeed, I did suggest it. https://github.com/nim-lang/RFCs/issues/168#issuecomment-530132823

dom96 (approved):
image

tests/concepts/tspec.nim
Araq: But then for non-atomic Concepts it makes things less clear, should it be

```nim

type
  Iterable[T] = concept 
    iterator items[T](x: Iterable[T]): T
```

or 

```nim

type
  Iterable[T] = concept 
    iterator items(x: Iterable): T
```

?

I really like ``self``, in fact, I think it's the key to the new concept's simplicity.

mratsim: @zah syntax is still on the table if we want a less magic self:

```Nim
type
  Comparable = concept self
    proc cmp(a, b: self): int
```

If we choose a concept with generics, it's better to have the `[T]` otherwise this would be valid
```Nim
type
  KeyValueStore[K, V] = concept
    func `[]=`[K, V](table: var KeyValueStore, key: K, value: V)
```

but AFAIK this is invalid (either all generics are specified or none):
```Nim
func `[]=`[K, V](table: var KeyValueStore, key: K, value: V) =
  discard
```

Araq: I don't want a "less magic self" though, I see no benefit in teaching "here, this trick allows you to match the concept itself that is described" vs "write 'self' to match the concept itself that is described".

dom96: > But then for non-atomic Concepts it makes things less clear, should it be

Does it even matter? Pick the top one to make things more explicit, I don't see a problem there.

c-blake: One identifier like `self` might simplify life slightly if the body of a `concept` is generic code.  `itself` might be less overloaded in the programming world at large.

Or what about the token `concept` instead of `self`?  What else could that mean in context other than the concept being defined?  It's not a nesting construct, right?  Just a thought in case no one else thunk it.  Not sure how I feel about it.

Araq: > Does it even matter? Pick the top one to make things more explicit, I don't see a problem there.

Of course it doesn't "matter", but you argue that the resulting language is "simpler" to teach and I disagree, no matter which variant I pick, the result is a bit arbitrary and hence slightly harder to teach than "self".

Araq: > Or what about the token concept instead of self? What else could that mean in context other than the concept being defined? It's not a nesting construct, right? Just a thought in case no one else thunk it. Not sure how I feel about it.

Right, it would work. (Even though ``concept`` on its own might also be interpreted as the type class of all concepts...) But I like ``self``, maybe you can convince me to pick ``itself`` instead but I don't like the other proposals so far.

c-blake: Ok.  Besides `itself` How about `Self` with a capital `S`?  Concepts are type classes.  So this would fit better, right?  I kind of like the capital-S `Self` better than `Itself` looking at it.

c-blake: Another possibility might be `Result` with a capital `'R'`.  It acts syntactically as a type, not as a value.  Lower and upper case 'r' are very visually distinct in almost any font, unlike `'s'` and `'S'`.  The stdlib seems to take the types start with capitals convention seriously.  Also, conceptually the token is the `Result` of the `concept` definition when all is said and done.  Just a thought.

dom96: > Of course it doesn't "matter", but you argue that the resulting language is "simpler" to teach and I disagree, no matter which variant I pick, the result is a bit arbitrary and hence slightly harder to teach than "self".

I disagree with you. The example you showed does not prove that my proposal is harder to teach. In fact, I would argue that the `self` introduces a special case to concepts which reduces consistency, a clear increase in learning difficulty.

Araq: >  In fact, I would argue that the self introduces a special case to concepts which reduces consistency, a clear increase in learning difficulty.

Well if you read the implementation you can see how ``self`` is a very important special case in the implementation. The implementation cannot lie, it's a special feature, it can have special syntax. And we can end the discussion here, you won't convince me.

tulayang: I also feel ``self`` looks a bit strange. I checked the specifications of rust and go, and I found that: 
  
- in rust, ``trait`` uses a ``&self`` to represent a concrete implementation 
- in go, an extra bracket ``(s MySerializer)`` to represent a concrete implementation

I suppose there are some parsing challenges for the compiler. 

```rust
trait Serializer {
  fn new() -> Self;
  fn serial(&self, data: String) -> String;
}

struct MySerializer {
  name: String
}

impl Serializer for MySerializer {
  fn serial(&self, data: String) -> String {
    format!("{}: {}", self.name, data);
  }
}
```

```go
type Serializer interface {
  serial(data string) string
}

type MySerializer struct {
  name string
}

func (self MySerializer) serial(data string) string {
  return self.name + ": " + data
}
```

```nim
type Serializer = concept
  proc serial(s: self, data: string): string

type MySerializer = object
  name: string

proc serial(self: MySerializer, data: string): string =
  self.name & ": " & data
```

BTW, from my own experience, matching a ``concept`` implicitly (similar to golang) is unfriendly to maintain large-scale code.

mratsim: > Another possibility might be `Result` with a capital `'R'`. It acts syntactically as a type, not as a value. Lower and upper case 'r' are very visually distinct in almost any font, unlike `'s'` and `'S'`. The stdlib seems to take the types start with capitals convention seriously. Also, conceptually the token is the `Result` of the `concept` definition when all is said and done. Just a thought.

I don't mind `self` or `Self`, I probably prefer the last one as it matches the convention to have types in PascalCase but `Result` is a hard no.
This would be confused with `Result` types that are widely used in Rust, Swift and OCaml and this would make our own libraries that use Result instead of Exceptions much harder to read or audit: https://github.com/status-im/nim-stew/blob/master/stew/results.nim

c-blake: Ok.  `Result` is maybe even more overloaded. :)

dom96: oh yes, can we at least get `Self` instead of `self`?

c-blake: In terms of identifier popularity estimation, I just searched my little nimbleverse archive and capitals `Self` is only used in code in 4 places  (one in Nim itself - `tests/cpp/tempty_generic_obj.nim`, then `winim` where it should probably be lower-`self` and a couple tests in `pararules` and `oop_utils`).  So, "not very popular".  Lowercase `self`, on the other hand, comes up 10s of thousands of times in like 7% of modules..So, practically, theoretically, and stylistically the lowercase version has a lot of potential for confusion.

bluenote10: I'm also in favor of `Self` because:

- If the user adheres to standard style guide, `Self` is an alias of an upper cased type.
- It makes sense to differentiate between variable name and type name, and the pattern `proc someMethod(self: Self)` is clearer that `proc someMethod(self: self)` (if that would be possible at all).
- That's the convention used in other languages as well (I'm at least aware of Rust and Swift).

> in rust, trait uses a &self to represent a concrete implementation

Note that this is actually syntactic sugar to adhere to the convention: `self` for the variable, `Self` for the type. (and the `&` isn't needed, just the passing mode)

@dom96 I think having `Self` is quite helpful, and doesn't impact learning too much. Note that with the fully explicit variant you are proposing, you'd probably have to repeat all generic parameters (possibly multiple times) like:
```nim
proc someMethod[A, B, C](self: MyConcept[A, B, C]): MyConcept[A, B, C]
proc anotherMethod[A, B, C](self: MyConcept[A, B, C]): MyConcept[A, B, C]
```
Even in clunky Rust this has been considered too much. This leaves the other alternative which magically drops the generics, which is non-standard, and thus requires learning a modified pattern. I'd say `Self` is easier to learn, because it makes the magic explicit and avoids confusion with the non-concept case.

Araq: Alright, ``Self`` with the capital S is it.

Araq: Changed it to ``Self``, closing.

Araq:
Need to reapply:

```nim

  of tyConcept:
    if kind != skParam: result = t
    else: result = nil

```

Soupertonic:
Hey, great work! What's the ETA for this to get merged? Super hyped for this!

Araq:
I hoped for somebody to take over my work as it's not scheduled.

Araq:
Will merge to keep it from bitrotting.

yrashk:
What's the plan with the new style concepts, are they still considered experimental or is there a path to maturity envisioned?

PR #15254, opened 2020-08-31
Basic VM resume support

dsrw:
Discussed briefly at https://forum.nim-lang.org/t/6724.

This stores the program counter and stack in the VM context, so the vm can be resumed after an exception. This allows a VM to be treated like a fiber, and could potentially be used for basic breakpoint support.

### DIFF
--- a/compiler/vm.nim
+++ b/compiler/vm.nim
@@ -525,6 +525,10 @@ proc rawExecute(c: PCtx, start: int, tos: PStackFrame): TFullReg =
   var savedFrame: PStackFrame
   var regs: seq[TFullReg] # alias to tos.slots for performance
   move(regs, tos.slots)
+  defer:
+    # snapshot the stack and counter for resume support
+    c.pc = pc
+    c.tos = tos
   #echo "NEW RUN ------------------------"
   while true:
     #{.computedGoto.}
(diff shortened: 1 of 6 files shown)


compiler/vm.nim
```diff
@@ -525,6 +525,10 @@ proc rawExecute(c: PCtx, start: int, tos: PStackFrame): TFullReg =
   var savedFrame: PStackFrame
   var regs: seq[TFullReg] # alias to tos.slots for performance
   move(regs, tos.slots)
+  defer:
```
dsrw, line 528: Is this the right way to do this? I could remove the `pc` and `tos` variables and use `c.pc` and `c.tos` directly everywhere instead.

compiler/vm.nim
```diff
@@ -2077,6 +2081,11 @@ proc execute(c: PCtx, start: int): PNode =
   newSeq(tos.slots, c.prc.maxSlots)
   result = rawExecute(c, start, tos).regToNode
 
+proc execFromCtx*(c: PCtx, pc = -1, tos: PStackFrame = nil): PNode =
```
dsrw, line 2084: I wanted to make this a proc called `resume` in `nimeval`, but `rawExecute` isn't exported. Should I export `rawExecute`, and get rid of this? That would also allow `Interpreter.graph` to stay private.

tests/compilerapi/tcompilerapi.nim
```diff
@@ -54,3 +56,24 @@ block issue9180:
 
   evalString("echo 10+1")
   evalString("echo 10+2")
+
+block ctxResume:
```
dsrw, line 60: Should I rename this to `issue15254`?

compiler/vm.nim
Araq: That would be worse for performance, don't use ``c.pc`` everywhere.

dsrw:
I think I want to go a different way with this. RFC at https://github.com/nim-lang/RFCs/issues/249.

Araq:
Well I liked it. :-)

PR #15263, opened 2020-09-03
[FEATURE] `strutils.dedent`

haxscramper:
Implement multiline string unindent which performs uniform whitespace prefix removal

### DIFF
--- a/lib/pure/strutils.nim
+++ b/lib/pure/strutils.nim
@@ -1566,9 +1566,9 @@ func dedent*(str: string): string =
     """.dedent()
 
     doAssert x == """
-    Hello
-      There
-    """
+  Hello
+    There
+  """
 
   let seplines = str.split("\n")
   var indent = 1000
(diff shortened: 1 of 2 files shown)


Varriount:
Why not overload unindent so that this behavior occurs when called with only a string?

haxscramper:
Overload of unindent which only takes a string already [exists](https://nim-lang.org/docs/strutils.html#unindent%2Cstring).

Varriount:
I think the old unindent proc should be deprecated, and dent be used as the replacement. The no-arguments/single argument overload behavior for unindent solves a case that can easily be performed using a single regex replacement, and I feel that the case in nearly as common as what dedent addresses.

Clyybber:
@Varriount I did that in https://github.com/nim-lang/Nim/pull/15264, but it broke niminaction tests.
Now my PR also introduces `dedent` instead of changing unindent.

haxscramper:
I don't think it is necessary to mark `unindent` as deprecated since `dedent` implements different behavior. Using regex replacement is not always possible (e.g. on nimscript target) and requiress to write `replace(re"^\s*", "")` each time you want to strip leading whitespaces instead of simply `unindent`.

PR #15264, opened 2020-09-03
Add strutils.indentation and make unindent use it

### LINKED ISSUE #161: Multi-line string literals (blocks of lines)
I would suggest -- instead of, or in addition to Python-style """ literals -- using indented block syntax for multi-line string literals. E.g.

````nim
proc foo =
  let str = ":
    Hello
      World!
  stdout.write str
````
Where str is defined equivalent to
````nim
let str = "Hello\l  World!\l"
````

This syntax avoids the indentation problem with string literals that .unindent attempts to address. Also, for clarity, all string content appears within the block, not on the opening or closing lines as is with """.

The literal terminates as soon as the block ends (i.e. a non-empty line indented less is found), avoiding the need for """ at the end. This also avoids the need to escape double quotes that belong to the string.

Whitespace at the end of any line and empty lines at the end would be omitted (and could be added via escape sequences in the rare cases where needed). Whitespace-only lines in the middle would become simply \l (no matter if there are spaces or not). This removes any ambiguity with source code formatting and makes the intention explicit.

This suggestion proposes string block to be indented by exactly two spaces (compared to the line with ": in it). Any further initial spaces would become string content.

This could still be used within parenthesis or other expression, provided that the continuation of that expression appears less indented than the string content.

Clyybber:
Closes https://github.com/nim-lang/RFCs/issues/161
Closes https://github.com/nim-lang/Nim/pull/11975

Closes #15263 , this PR is an alternative to #15263, main difference is indentation beeing a seperate proc and the loop determining the indentation of a line exiting early if the indentation is greater than or equal to the minimum indentation of the already processed lines.


dom96 (commented):
Are the compiler changes included by accident?

Clyybber:
@dom96 Yeah, more or less. This PR depends on default parameter fixes from #15270, and I accidently pushed to this branch instead of a seperate one. Should be gone now.

PR #15273, opened 2020-09-06
add getprotobyname

ringabout:
work like something in Python: https://docs.python.org/3/library/socket.html#socket.getprotobyname

Cpython implemention: https://github.com/python/cpython/blob/e822e37946f27c09953bb5733acf3b07c2db690f/Modules/socketmodule.c#L5900

From MSDN:

The protoent structure contains the name and protocol numbers that correspond to a given protocol name. Applications must never attempt to modify this structure or to **free any of its components**. Furthermore, only one copy of this structure is allocated per thread, and therefore, the application should copy any information it needs before issuing any other Windows Sockets function calls.

### DIFF
--- a/lib/pure/nativesockets.nim
+++ b/lib/pure/nativesockets.nim
@@ -186,6 +186,7 @@ proc toSockType*(protocol: Protocol): SockType =
     SOCK_RAW
 
 proc getProtoByName*(name: string): int =
+  ## Returns a protocol code from the database that matches the protocol ``name``.
   when useWinVersion:
     let protoent = winlean.getprotobyname(name.cstring)
   else:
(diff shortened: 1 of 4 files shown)


lib/pure/nativesockets.nim
```diff
@@ -185,6 +185,18 @@ proc toSockType*(protocol: Protocol): SockType =
   of IPPROTO_IP, IPPROTO_IPV6, IPPROTO_RAW, IPPROTO_ICMP, IPPROTO_ICMPV6:
     SOCK_RAW
 
+proc getProtoByName*(name: string): int =
```
Araq: Please add a ``.since`` annotation.

ringabout: done

PR #15274, opened 2020-09-06
more Protocol supports in windows

ringabout:
I have to modify the order of enum because of error: invalid order in enum.

I use `getprotobyname` to get the number of proto, it seems that only IPPROTO_RAW doesn't get supports in my win 10.

### DIFF
--- a/lib/pure/nativesockets.nim
+++ b/lib/pure/nativesockets.nim
@@ -65,14 +65,14 @@ type
     SOCK_RAW = 3,      ## raw protocols atop the network layer.
     SOCK_SEQPACKET = 5 ## reliable sequenced packet service
 
-  Protocol* = enum    ## third argument to `socket` proc
-    IPPROTO_TCP = 6,  ## Transmission control protocol.
-    IPPROTO_UDP = 17, ## User datagram protocol.
-    IPPROTO_IP,       ## Internet protocol. Unsupported on Windows.
-    IPPROTO_IPV6,     ## Internet Protocol Version 6. Unsupported on Windows.
-    IPPROTO_RAW,      ## Raw IP Packets Protocol. Unsupported on Windows.
-    IPPROTO_ICMP      ## Control message protocol. Unsupported on Windows.
-    IPPROTO_ICMPV6    ## Control message protocol for IPv6. Unsupported on Windows.
+  Protocol* = enum     ## third argument to `socket` proc
+    IPPROTO_IP = 0,    ## Internet protocol.
+    IPPROTO_ICMP = 1   ## Control message protocol.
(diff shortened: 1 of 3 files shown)


ringabout:
results with my another pr: https://github.com/nim-lang/Nim/pull/15273
```nim
import nativesockets


echo "icmpv6: ", getProtoByName("ipv6-icmp")
echo "icmp: ", getProtoByName("icmp")
echo "ipv6: ", getProtoByName("ipv6")
echo "ip: ", getProtoByName("ip")
```

```
icmpv6: 58
icmp: 1
ipv6: 41
ip: 0
```

ringabout:
Supports protocol in my win10:
```
ip         0     IP           # Internet protocol
icmp       1     ICMP         # Internet control message protocol
ggp        3     GGP          # Gateway-gateway protocol
tcp        6     TCP          # Transmission control protocol
egp        8     EGP          # Exterior gateway protocol
pup        12    PUP          # PARC universal packet protocol
udp        17    UDP          # User datagram protocol
hmp        20    HMP          # Host monitoring protocol
xns-idp    22    XNS-IDP      # Xerox NS IDP
rdp        27    RDP          # "reliable datagram" protocol
ipv6       41    IPv6         # Internet protocol IPv6
ipv6-route 43    IPv6-Route   # Routing header for IPv6
ipv6-frag  44    IPv6-Frag    # Fragment header for IPv6
esp        50    ESP          # Encapsulating security payload
ah         51    AH           # Authentication header
ipv6-icmp  58    IPv6-ICMP    # ICMP for IPv6
ipv6-nonxt 59    IPv6-NoNxt   # No next header for IPv6
ipv6-opts  60    IPv6-Opts    # Destination options for IPv6
rvd        66    RVD          # MIT remote virtual disk
```

Araq:
That's a breaking change and it can have subtle effects. We should model this is a ``distinct cint`` instead and ideally we only add code/overloads and don't touch the older stuff.

dom96:
You shouldn't need to change the values of these enums, why not just convert to whatever value they actually have? Turning this into a `distinct int` doesn't seem like the right approach.

ringabout:
So I should change `  proc toInt(p: Protocol): cint = result = cint(ord(p))`
https://github.com/nim-lang/Nim/blob/e2193b018e1551a40f9a60881b206df42ccaaac4/lib/pure/nativesockets.nim#L176
to instead of change enum values?
```nim
case p
of IPPROTO_IP:
  result = 0,     ## Internet protocol.
of IPPROTO_ICMP:
  result = 1    ## Control message protocol.
............
else:
  result = int(ord(p))
```

Araq:
> to instead of change enum values?

Yeah, that's a good solution too.

PR #15275, opened 2020-09-06
getPeerCertificates() for libressl and openssl10

ardek66:
The proc `getPeerCertificates` uses functions which are only defined in the latest OpenSSL(1.1.x). For the sake of compatibility, I made some changes that use the `-d:libressl` and `-d:openssl10` compile flags, in order to import the correct functions for these SSL libraries.
The output should, in most cases, be the same among the different versions. I have also tried to keep the code similar with the original one.

https://www.openssl.org/docs/man1.0.2/man3/SSL_get_peer_cert_chain.html
https://github.com/sergot/openssl/issues/62

### DIFF
--- a/lib/pure/net.nim
+++ b/lib/pure/net.nim
@@ -804,15 +804,33 @@ when defineSsl:
     result = newSeq[Certificate]()
     if SSL_get_verify_result(sslHandle) != X509_V_OK:
       return
-    let stack = SSL_get0_verified_chain(sslHandle)
-    if stack == nil:
-      return
-    let length = OPENSSL_sk_num(stack)
-    if length == 0:
-      return
-    for i in 0 .. length - 1:
-      let x509 = cast[PX509](OPENSSL_sk_value(stack, i))
-      result.add(i2d_X509(x509))
+    when defined(libressl) or defined(openssl10):
+      let stack = SSL_get_peer_cert_chain(sslHandle)
+      if stack == nil:
+        return
+      # SSL_get_peer_cert_chain does not return the last peer certificate
+      # if called server-side
+      let leafCert = SSL_get_peer_certificate(sslHandle)
+      if leafCert == nil:
+        return
+      if sk_value(stack, 0) != leafCert:
+        sk_insert(stack, cast[pointer](leafCert), 0)
+      let length = sk_num(stack)
+      if length == 0:
+        return
+      for i in 0 .. length - 1:
+        let x509 = cast[PX509](sk_value(stack, i))
+        result.add i2d_X509(x509)
+    else:
+      let stack = SSL_get0_verified_chain(sslHandle)
+      if stack == nil:
+        return
+      let length = OPENSSL_sk_num(stack)
+      if length == 0:
+        return
+      for i in 0 .. length - 1:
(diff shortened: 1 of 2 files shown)


ardek66:
I also believe that this function should raise a SslError if the chain is not valid. Currently it has the same output as if the chain would be empty(an empty sequence). There are cases when one would like to check if the chain is valid vs if the peer actually provided a chain, and I think that an error when the chain isn't valid would fit. I am not sure if this fits for this PR, I just wanted to bring it up. For now this commit only implements the stuff mentioned in the OP

alaviss (changes requested):
The behavior is inconsistent, on OpenSSL >= 1.1 this would return a verified chain, but on older openssl it appears that this chain is returned as is and is not verified at all.

ardek66:
I thought that
```nim
if SSL_get_verify_result(sslHandle) != X509_V_OK: return
```
would return an empty sequence if the chain wasn't succesfuly verified beforehand.
This is also mentioned in the original `getPeerCertificates` documentation.

FedericoCeratto:
As discussed, keeping compatibility with openssl 1.0 increases complexity (and therefore risk).
The 1.1 release train started in 2016 and 1.0 started in 2015. Perhaps it's not worth supporting 1.0

ardek66:
But what about LibreSSL? It is actively developed and a lot of OSes use it instead of OpenSSL(i.e OpenBSD or Void Linux), and it still uses(from what I could find) names from OpenSSL 1.0

alaviss:
> I thought that
> 
> ```nim
> if SSL_get_verify_result(sslHandle) != X509_V_OK: return
> ```
> 
> would return an empty sequence if the chain wasn't succesfuly verified beforehand.
> This is also mentioned in the original `getPeerCertificates` documentation.

Unless you can completely verify that the chain you get from `SSL_get_chain_certificate` + `SSL_get_peer_certificate` is equivalent to the chain from `SSL_get0_verified_chain` when verification success, I can't accept this implementation as OpenSSL/LibreSSL is a mess in their API and how they interact with one another.

From a quick read on OpenSSL source, the way this can be done is by creating a new `X509_STORE_CTX` with the certificate store from `SSL_CTX_get_cert_store` (which contains the loaded certificates), the peer certificate from `SSL_get_peer_certificate`, and the chain from `SSL_get_peer_cert_chain`, then verify the resulting object with `X509_verify_cert`. Then the verified chain will be presented via `X509_STORE_CTX_get1_chain`.

To be completely certain, I'd like to have some tests covering this, since this is security-sensitive code.

alaviss:
An alternative route is to hook into the verification process itself via `SSL_set_verify`'s callback and extract the verified chain there. However this requires yet-another-field in the `Socket` object for an API that might not be used frequently, so I have my concerns there as well. But this might be simpler than the route outlined earlier.

ardek66:
Closing this because of multiple reasons:

- This PR is stuck on my devel branch and I cant really work with the repo properly.
- Doesn't seem like it would get merged soon
- I don't personally need libressl support anymore, and it seems a lot of distros are giving up on LibreSSL

I am right now more interested in https://github.com/nim-lang/Nim/issues/17021, which might allow people doing low-level SSL as they wish without messing with stdlib, sorry for the time being.

PR #15278, opened 2020-09-06
don't raise index defects on malformed ast

disruptek:
Just a couple tweaks to help dust not crash so much.


Araq:
Please write it without the ``block break`` constructions.

PR #15282, opened 2020-09-07
borrow checking

Araq:
Todo: 

- [x] wiring of Nim's borrow checker to types containing a view-type.
- [x] backend extensions in order to map openArray to structs.
- [x] documentation and tests.


doc/manual_experimental.rst
```diff
@@ -1882,3 +1882,55 @@ For example:
...
+
+.. code-block:: nim
+
+  {.experimental: "views".}
+
+  proc take(a: openArray[int]) =
+    echo a.len
+
+  proc main(s: seq[int]) =
+    var x: openArray[int] = s # 'x' is a view into 's'
```
mratsim, line 1867: I think a `var x = s.view()` or `var x = s.borrow()` would be nicer but I guess we can discuss API once the low-level primitive work.

doc/manual_experimental.rst
```diff
@@ -1882,3 +1882,55 @@ For example:
...
+    # it is checked that 'x' does not outlive 's' and
+    # that 's' is not mutated.
+    for i in 0 .. high(x):
+      echo x[i]
+    take(x)
+
+    take(x.toOpenArray(0, 1)) # slicing remains possible
```
mratsim, line 1874: Similarly `[]` on sequences should now return openArray by default unless this is assigned to a returned value that escapes and maybe we need a `StringView` type?

mratsim (commented):
I can't comment on the implementation but we'll likely want to have an easy to approach API and story.

- `openArray` is long to type and very obscure apart from Oberon, Modula and Pascal devs, I would love to see it replaced either by view or span (given that slice is taken). C# and C++ also use span and call views the overarching type category that doesn't own memory just like current http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0122r1.pdf . (Rust and Swift uses Slice)
- `toOpenArray` should be using `[]` except when the compiler detects that the assignment escapes (which it now can do with borrow checking, write-tracking https://nim-lang.org/blog/2020/09/01/write-tracking.html. This has been a source of slow allocation and GC pressures in Nimbus (https://github.com/status-im/nim-beacon-chain/pull/1573) and is an easy way to get in trouble with multithreading especially OpenMP when we forget to setupForeignThreadGc
- We probably need a StringView type

I should probably write a RFC on that.

Araq:
Well I won't rename or introduce new APIs, I'm following https://github.com/nim-lang/RFCs/issues/178 

If you want new names or APIs please write an RFC and see how the community reacts. ;-)

RSDuck:
this is so cool and pretty much the only thing I felt was missing from Nim!

While this is going to be experimental at first, I think it deprecates multiple things which were more or less workarounds using pointers:
* the [byaddr pragma](https://github.com/nim-lang/Nim/pull/13508)
* *with* templates like the one for [tables](https://github.com/nim-lang/Nim/blob/devel/lib/pure/collections/tables.nim#L588)

Araq:
> While this is going to be experimental at first, I think it deprecates multiple things which were more or less workarounds using pointers

Agreed. Also the special casing for ``evalOnce`` (you cannot store an openArray in a variable) should be removable with this.

Araq:
Merging this as the dangers of merge conflicts become serious. But it's hardly usable in this state and will see follow-up PRs.

PR #15284, opened 2020-09-07
nimStackTraceOverride: enable stack traces in exceptions

### LINKED ISSUE #5: deferred stacktraces  (`getStackTrace(e: ref Exception)`) don't work (works with standard nim stacktraces)
standard nim stacktraces allow getting stacktraces stored in an Exception (eg for use in a catch statement); but libbacktrace ones don't work in this case (they only work in non-deferred mode, eg when crashing with `doAssert false`), so it's not a suitable replacement for standard stacktraces at the moment.

I had originally opened https://github.com/status-im/nim-libbacktrace/issues/4 as I assumed the code did "eager" symbolication which is not efficient but in fact I now realize it doesn't do any symbolication at all for caught exceptions; **feel free to close either this issue or https://github.com/status-im/nim-libbacktrace/issues/4 if you consider them duplicates** (although https://github.com/status-im/nim-libbacktrace/issues/4 deals with performance and this one deals with correctness).

As I mentioned in https://github.com/status-im/nim-libbacktrace/issues/4, one way to fix this would be using something like what D does (https://github.com/dlang/druntime/blob/master/src/rt/backtrace/dwarf.d) which collects backtrace (cheap) then symbolicates if needed later. However it's not yet clear to me whether libbacktrace supports deferred symbolication (see https://github.com/ianlancetaylor/libbacktrace/issues/35)

## example
```
# standard nim stacktraces: works (shows a stacktrace)
nim c -r --stacktrace:on -d:case2c -g $timn_D/tests/nim/all/t10308.nim

# libbacktrace stacktraces: doesn't work (shows empty stacktrace)
nim c -r --stacktrace:off -d:nimStackTraceOverride --import:libbacktrace -g main.nim
```

```nim
# main.nim
proc fun() =
  # doAssert false # this works, correctly showing a stacktrace
  raise newException(CatchableError, "foo")
try: fun()
except CatchableError as e:
  echo "caught:\n" & e.getStacktrace() # BUG: no stacktrace shown
echo "after"
```

this outputs:
```
caught:

after
```
instead of showing a stacktrace, like in standard nim stacktraces

## links
D20200308T163130

stefantalpalaru:
This is a two-step stack trace collection scheme, because re-raised
exceptions will collect multiple stack traces but use them rarely, when
printing info about an uncaught exception, so it makes sense to only do
the cheap stack unwinding all the time and the relatively expensive
debugging information collection on-demand.

`asyncfutures` implements its own `$` proc for printing
`seq[StackTraceEntry]`, so we have to add the debugging info there, just
like we do for the private `$` proc in `system/excpt`.

New type: `cuintptr_t` - needed because of clang++ on macOS.

Fixes https://github.com/status-im/nim-libbacktrace/issues/5

Please backport this to Nim-1.2.x .

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -148,7 +148,7 @@
 - Added `deques.toDeque`, which creates a deque from an openArray. The usage is
   similar to procs such as `sets.toHashSet` and `tables.toTable`. Previously,
   it was necessary to create an empty deque and add items manually.
-  
+
 - `std/with`, `sugar.dup` now support object field assignment expression:
   ```nim
   import std/with
@@ -168,7 +168,12 @@
   cannot be applied to every use case. The limitations and the (lack of) reliability
   of `round` are well documented.
 
+- New type: `cuintptr_t` - the same as `uintptr_t` in C.
 
+- `system/excpt`, `asyncfutures`: `-d:nimStackTraceOverride` now has exception
+  stack traces, with support for an efficient implementation that separates the
+  cheap stack unwinding step from the relatively expensive debugging info
+  collection that's done on demand.
 
 ## Language changes
 
--- a/lib/pure/asyncfutures.nim
+++ b/lib/pure/asyncfutures.nim
@@ -311,7 +311,12 @@ proc getHint(entry: StackTraceEntry): string =
     if cmpIgnoreStyle(entry.filename, "asyncmacro.nim") == 0:
       return "Resumes an async procedure"
 
-proc `$`*(entries: seq[StackTraceEntry]): string =
+proc `$`*(stackTraceEntries: seq[StackTraceEntry]): string =
+  when defined(nimStackTraceOverride):
+    let entries = addDebuggingInfo(stackTraceEntries)
+  else:
+    let entries = stackTraceEntries
+
   result = ""
   # Find longest filename & line number combo for alignment purposes.
   var longestLeft = 0
@@ -326,10 +331,10 @@ proc `$`*(entries: seq[StackTraceEntry]): string =
   # Format the entries.
   for entry in entries:
     if entry.procname.isNil:
-      if entry.line == -10:
+      if entry.line == reraisedFromBegin:
         result.add(spaces(indent) & "#[\n")
         indent.inc(2)
-      else:
+      elif entry.line == reraisedFromEnd:
         indent.dec(2)
         result.add(spaces(indent) & "]#\n")
       continue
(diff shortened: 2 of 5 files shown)


changelog.md
```diff
@@ -168,7 +168,12 @@
   cannot be applied to every use case. The limitations and the (lack of) reliability
   of `round` are well documented.
 
+- New type: `cuintptr_t` - the same as `uintptr_t` in C.
```
Araq, line 171: Well this is a (minor) feature addition, if we backport that, we don't follow semver.

Suggestion: Use the type internally but don't export it.

stefantalpalaru: > Well this is a (minor) feature addition, if we backport that, we don't follow semver.

Then I'm declaring it a bug fix for this macOS error when trying to use Nim's `uint` instead of C's `uintptr_t`:

```text
/Users/travis/build/status-im/nim-libbacktrace/libbacktrace.nim:74:31: error: no matching function for call to 'get_debugging_info_c'

        struct debugging_info* T1_ = get_debugging_info_c((&programCounters->data[((NI) 0)]), maxLength);       struct debugging_info* functionInfoPtr = T1_;

                                     ^~~~~~~~~~~~~~~~~~~~

/Users/travis/build/status-im/nim-libbacktrace/libbacktrace_wrapper.h:42:24: note: candidate function not viable: no known conversion from 'NU *' (aka 'unsigned long long *') to 'uintptr_t *' (aka 'unsigned long *') for 1st argument

struct debugging_info *get_debugging_info_c(uintptr_t *program_counters, int max_length);

                       ^

/Users/travis/build/status-im/nim-libbacktrace/libbacktrace.nim:59:6: error: cannot initialize a variable of type 'NU *' (aka 'unsigned long long *') with an rvalue of type 'uintptr_t *' (aka 'unsigned long *')

        NU* T1_ = get_program_counters_c(maxLength, ((int) 2)); NU* pcPtr = T1_;

            ^     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

2 errors generated.

Error: execution of an external compiler program 'clang++ -c  -w -I/Users/travis/build/status-im/nim-libbacktrace -I/Users/travis/build/status-im/nim-libbacktrace/install/usr/include -g -std=gnu++14 -funsigned-char  -I/Users/travis/build/status-im/nim-libbacktrace/Nim/lib -I/Users/travis/build/status-im/nim-libbacktrace/tests -o /Users/travis/.cache/nim/test2_d/@m..@slibbacktrace.nim.cpp.o /Users/travis/.cache/nim/test2_d/@m..@slibbacktrace.nim.cpp' failed with exit code: 1
```

There, problem solved ;-)

Araq:
I took over, see https://github.com/nim-lang/Nim/pull/15384/

PR #15287, opened 2020-09-08
Nil type check implementation

alehander92:


compiler/treetab.nim
```diff
@@ -10,18 +10,19 @@
 # Implements a table from trees to trees. Does structural equivalence checking.
 
 import
-  hashes, ast, astalgo, types
+  hashes, ast, astalgo, types, renderer
```
Araq: Debug leftover.

compiler/varpartitions_symbol.nim
```diff
@@ -0,0 +1,1003 @@
...
+## The algorithm is a reinvention / variation of Steensgaard's
+## algorithm.
+## The used data structure is "union find" with path compression.
+
+import ast, types, lineinfos, options, msgs, renderer
+from trees import getMagic, whichPragma
+from wordrecg import wNoSideEffect
+from isolation_check import canAlias
+<<<<<<< HEAD
+<<<<<<< HEAD
```
Araq: seriously? ;-)

compiler/nilcheck.nim
```diff
@@ -0,0 +1,928 @@
...
+
+proc checkBranch(n, ctx, map): Check =
+  result = check(n, ctx, map)
+
+
+# Faith!
+
+proc check(n: PNode, ctx: NilCheckerContext, map: NilMap): Check =
+  if map.isNil:
+    localError(ctx.config, n.info, "map is nil: something went wrong in nilcheck")
```
Araq: Instead use ``internalError``.

compiler/nilcheck.nim
```diff
@@ -0,0 +1,928 @@
...
+    quit 1
```
Araq: No ``quit`` here.

compiler/nilcheck.nim
```diff
@@ -0,0 +1,928 @@
...
+
+  map.store(context.graph(resultId), if not s.typ[0].isNil and s.typ[0].kind == tyRef: Nil else: Safe, TResult, s.ast.info)
+    
+  # echo "checking ", s.name.s, " ", filename
+
+  let res = check(body, context, map)
+  let resultGraphIndex = context.graph(resultId)
+  if res.nilability == Safe and (not res.map.history.hasKey(resultGraphIndex) or res.map.history[resultGraphIndex].len <= 1):
+    res.map.store(resultGraphIndex, Safe, TAssign, s.ast.info)
+  if not s.typ[0].isNil and s.typ[0].kind == tyRef and tfNotNil in s.typ[0].flags:
```
Araq, line 1380: What if it's not a ``tyRef`` but a tuple containing a ref?

compiler/varpartitions_symbol.nim
alehander92: ugh I forgot to make sure it's removed: I rm-ed it but probably i add-ed again

compiler/treetab.nim
alehander92: ok

compiler/nilcheck.nim
alehander92: should I just assert?

alehander92: or just wait and use NilMap not nil later

compiler/nilcheck.nim
alehander92: good point: I will think

alehander92: I am starting a commented out test for that: currently i think we should just warn for uninitialized exprs requiring non nilable and for assigning a nilable element to non-nilable (the second one seems to be sometimes happening already in the compiler)

alehander92: but i'll wait for tomorrow

compiler/nilcheck.nim
alehander92: for now i'll do so

compiler/nilcheck.nim
```diff
@@ -0,0 +1,1433 @@
+#
+#
+#           The Nim Compiler
+#        (c) Copyright 2017 Andreas Rumpf
+#
+#    See the file "copying.txt", included in this
+#    distribution, for details about the copyright.
+#
+
+import ast, renderer, intsets, tables, msgs, options, lineinfos, strformat, idents, treetab, hashes
```
alehander92, line 10: i am adding a comment so i can link

alehander92:
Hey @Araq do I need to do something right now? How to proceed, there are many possible improvements here

alehander92:
@Araq sorry, I was too lazy to write a comment here: this is ready for review since some time, but not sure what's important to fix

Araq:
Merging this now to keep it from rotting, but will create a follow-up PR.

PR #15289, opened 2020-09-09
GitHub Actions Skip CI

juancarlospaco:
- Add `[skip ci]` on the commit message to GitHub Actions.

temp

:)

### DIFF
--- a/.github/workflows/ci_docs.yml
+++ b/.github/workflows/ci_docs.yml
@@ -27,7 +27,20 @@ on:
 
 
 jobs:
+  check_skip:
+    runs-on: ubuntu-latest
+    if: |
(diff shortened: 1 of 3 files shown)


.github/workflows/ci_docs.yml
```diff
@@ -27,7 +27,20 @@ on:
 
 
 jobs:
+  check_skip:
+    runs-on: ubuntu-latest
+    if: |
+      !contains(format('{0} {1} {2}', github.event.head_commit.message, github.event.pull_request.title, github.event.pull_request.body), '[skip ci]')
```
alaviss: This condition should be inlined into `build` directly.

Also I think we should only check the commit message and PR title only. I don't think there is a good use for this feature in the PR body.

PR #15291, opened 2020-09-09
fix warnings for deprecated `low` and `high`

narimiran:


SolitudeSF:
looks like `high`/`low` on value wasn't so bad...

PR #15296, opened 2020-09-10
make 'for' a keyword operator

metagn:
Precedence is 1 like `=` and `+=`, could change if anyone has a better suggestion.

Uses are pretty obvious, it's for custom comprehension/zero overhead iterator macros. I don't think it should be defined directly as an operator macro in libraries (`comp: a for b in c` instead of directly `a for b in c`) but some people may choose to put them in their scripts or small programs as a familiar construct from other languages.

It's kind of like `=>`, except the implementation and behavior of infix `for` are pretty ambiguous compared to `=>` so defining it in sugar would be controversial. Maybe a fusion module.

### DIFF
--- a/compiler/layouter.nim
+++ b/compiler/layouter.nim
@@ -327,7 +327,8 @@ const
 
   splitters = openPars + {tkComma, tkSemiColon} # do not add 'tkColon' here!
   oprSet = {tkOpr, tkDiv, tkMod, tkShl, tkShr, tkIn, tkNotin, tkIs,
-            tkIsnot, tkNot, tkOf, tkAs, tkFrom, tkDotDot, tkAnd, tkOr, tkXor}
+            tkIsnot, tkNot, tkOf, tkAs, tkFrom, tkDotDot, tkAnd, tkOr, tkXor,
+            tkFor}
 
 template goodCol(col): bool = col >= em.maxLineLen div 2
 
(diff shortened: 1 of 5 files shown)


Araq:
But I like ``sugar.collect`` much more than Python's strange comprehensions.

disruptek:
Seems to me that `each` or `foreach` is sufficient for now and lets us worry about relaxing the spec later.

metagn:
It would actually slightly differ from Python, what would be `x+1 for y in [[1,2,3], [4,5,6], [7,8,9]] for x in y` in Python and Julia would be  `x+1 for x in y for y in [[1,2,3], [4,5,6], [7,8,9]]` under this AST. I think this way makes more sense when you try to read the code like English. Kind of like let vs where in Haskell, or Perl's postfix statements.

Speaking of Perl's postfix statements, `for` is the only keyword that can be made an operator as easy as this, if/while/when all break meaning filtering comprehensions will be hard, but you can still use those keywords for proc names via method call, so something ugly like `(x + 1).if(x % 2 == 0) for x in [1, 2, 3, 4, 5]`. I'm sure there are better alternatives.

The approach in sugar.collect is a bit off to me, normal `for` statements are never used as expressions and you have to discard any child expression. Same with `if` without `else`. I think infix `for` on top of regular statement `for`s can be added as a cosmetic AST for niche macro libraries to use and `sugar.collect` wouldn't have to adopt it and could keep its old syntax. I understand if it's a little bit of a weird change that would be hard for people to wrap their head around, but it could be clarified in the manual like "infix for is merely a cosmetic operator" etc.

Araq:
This needs an RFC and enough upvotes, I personally think it's quite terrible, sorry.

metagn:
I don't think that RFC would pass, closing

PR #15297, opened 2020-09-10
add `enumerate` macro

narimiran:


planetis-m:
it is not correct because the iteration variable must start from -1, and incremented at the start of the loop, in order to work with collect, like this: https://github.com/planetis-m/looper/blob/master/looper.nim#L112

bluenote10:
I was hoping for a solution that works not only in for statements and is fully typed to avoid any import issues -- which doesn't seem to be the case judging from the Arraymancer change.

I had the impression that an `Iterable[T]` concept isn't too far away. Wouldn't it be cleaner to wait for that?

c-blake:
Hey, @narimiran - while you are at it, why not add `toItr` (or `toIterator` or whatever people want to call it...)?  https://forum.nim-lang.org/t/6649#41330

Araq:
@narimiran we discussed to move this to a new ``std / enumerate`` module.

narimiran:
> @narimiran we discussed to move this to a new std / enumerate module.

Yes, this is just a first commit, so it can be more easily seen what has been changed since the original version.

Issue #15302, opened 2020-09-11
LwIP/FreeRTOS compile error - missing SIGPIPE and more 


elcritch:
Compiling `nim.net` since https://github.com/nim-lang/Nim/commit/c619cedd7cba60d588707fbbb2905c8ac85fd12a due to missing `SIGPIPE` and other signals. 

### Example
```sh
nim c --os:freertos --cpu:esp --gc:arc some_nim_net_example.nim 
```

### Current Output
Build error. 

> /home/elcritch/.asdf/installs/nim/devel/compiler/nimblecmd.nim(117, 12) compiler msg initiated here [MsgOrigin]
> ..................................
> /home/elcritch/.asdf/installs/nim/devel/lib/pure/net.nim(1066, 35) Error: undeclared identifier: 'SIGPIPE'
> /home/elcritch/.asdf/installs/nim/devel/compiler/lookups.nim(294, 13) compiler msg initiated here [MsgOrigin]
> make: *** [Makefile:16: nim] Error 1
> 

### Expected Output
Compiles. 

### Possible Solution

I've already fixed the issue. I'm filing this to tag in a PR. 

That'll solve the current issue. I'd like to figure out how/where to add a test for builds. There's a number of places where LwIP differs from SysV POSIX. The general pattern is to test for `when defined(posix) and not defined(lwip)` in cases where `lwip` doesn't support the posix features. Generally those are just signals. 

In order to reduce maintenance there should be some tests for compiling basic FreeRTOS networking examples.

elcritch:
See https://github.com/elcritch/esp32_nim_net_example/issues/1 for more information.

elcritch:
Closing this pending https://github.com/nim-lang/Nim/pull/15303.

PR #15303, opened 2020-09-11
Fixing issue #15302 -- lwip doesn't support signals

elcritch:
Fixing issue #15302 by excluding lwip from posix signal support.

### DIFF
--- a/compiler/platform.nim
+++ b/compiler/platform.nim
@@ -179,7 +179,7 @@ const
       props: {ospNeedsPIC, ospPosix}),
      (name: "FreeRTOS", parDir: "..", dllFrmt: "lib$1.so", altDirSep: "/",
       objExt: ".o", newLine: "\x0A", pathSep: ":", dirSep: "/",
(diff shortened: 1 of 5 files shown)


elcritch:
Added test to compile a FreeRTOS/LwIP. Tested the new test `stdlib/tlwip.nim` with and without the fix to `net.nim`. Without the fix it fails while with it passes. 

It'll compile and run on POSIX platforms but not Windows. Though it should be possible to compile Nim for FreeRTOS on Windows, but it'll need to be built within `esp-idf`.

elcritch:
Testing the `tlwip.nim` test with compile only on Windows and Linux.

elcritch:
Trying to compile on Windows doesn't appear to work but it's not clear why from the logs. I modified the test to only run on Linux for now. It'd be nice to run on Windows, and in theory it should work like any other cross compiling. I don't have Windows. I'll include that in a PR for the documentation when I get a chance.

elcritch:
@Araq -- this is ready for merging. I've squashed the commits down a bit. Only Linux / MacOS is currently being tested but they pass!

PR #15311, opened 2020-09-12
Fix #15305

### LINKED ISSUE #15305: [Regression] Parser regression
Use to compile previosly:
### Example
```nim
var a, b: int
let x = if a > b: 
    0
    else: 1
```
Now fails with invalid indentation

Clyybber:
Fixes #15305


Araq:
Please add a test case that ``if`` as a statement requires precise indentation for ``elif`` and ``else``.

PR #15318, opened 2020-09-13
Fix Nimgrep bug

juancarlospaco:
- Fix Nimgrep invalid input bug.
- Nimgrep parameter takes negative values that make no sense and produce buggy output in loop.
- 3 line diff.

```console
$ nimgrep --afterContext:-9 --beforeContext:-9 --context:-9 linesBefore nimgrep.nim
```

Prints the whole file several times in loop.

:)


Araq:
Input validation shouldn't misuse language features.

PR #15319, opened 2020-09-13
fix cookie with comma

ringabout:
issues:
https://github.com/planety/prologue/issues/38
https://github.com/dom96/jester/issues/254

Don't process the field of the HTTP header when the name is `cookie`.

Or we can document that `asynchttpserver` doesn't support cookie with comma.

I don't know how to test `asynchttpserver` with `cookie`.


ringabout:
Oh i forget that I can write some tests for `httpcore`. Will do.

PR #15327, opened 2020-09-15
dom.Window add missing methods

juancarlospaco:
- Add missing attributes and methods to `dom.Window`. Tiny diff.

:)


dom96 (changes requested):
You should define `Crypto` and define it as a field on `window` instead of this shortcut.

dom96:
Why the close?

PR #15330, opened 2020-09-15
Introduce explicit copy

cooldome:
The way to manage the performance warning. Also makes performance predictable.


juancarlospaco:
Should it be `{.inline.}` ?.

cooldome:
added inline

Clyybber:
Maybe this should warn when we do an eplicit but unneccessary copy?

I think a better idea is to not emit the warnings unless requested, as this solution could lead to library authors not using sink so that the API is nicer.

Issue #15324, opened 2020-09-14
Symbols lookup regression


cooldome:
Used to compile before:
### Example
```nim
import macros

var interesting_funcs {.compileTime.}: seq[NimNode] = @[]

macro interest(fns: varargs[typed]): typed =
  for fn in fns:
    interesting_funcs.add fn[0]

proc myproc(x: float): float {.interest.} =
  2*x

myproc(3.0)
```

Now fails on the last line with error: `undeclared identifier 'myproc'`

cooldome:
Regression is caused by this change: https://github.com/nim-lang/Nim/commit/fb58066b61b14f4a1d6cdb0f4a8f0a9ea4174d82

@Clyybber please have a look

Clyybber:
This is intended. Previously calling it would have just created cgen errors (at least with `fns: typed`, not sure if varargs elided the bug).

Clyybber:
Adapted code would be:
```nim
import macros

var interesting_funcs {.compileTime.}: seq[NimNode] = @[]

macro interest(fns: varargs[typed]): typed =
  result = newNimNode(nnkStmtList)
  for fn in fns:
    interesting_funcs.add fn[0]
    result.add fn

proc myproc(x: float): float {.interest.} =
  2*x

echo myproc(3.0)
```
but this fails currently, due to a bug thats fixed with #15252

cooldome:
Hi @Clyybber, 
cgen is not related at all.

The problem is that `myproc` now disappears from symbol list if you apply typed macro to it.
`interest` here plays a role of pragma that should not affect the visibility of the `myproc`.

This is simplified example that fails with the same error:
```nim
import macros

macro interest(fn: typed) =
  discard

proc myproc(x: float): float {.interest.} =
  2*x

myproc(3.0)
```

Clyybber:
Yeah, this is intended. cgen is related because before my PR this would just call a proc that codegen would never encounter, thus generate invalid C code.

In this case the proc is generated anyways on older Nim (not sure why, I haven't investigated), but other cases like
```nim
import macros

macro interest(fn: typed) =
  discard

interest:
  var x = 1

echo x
```
show why the previous behaviour was broken.

cooldome:
Agree you change is useful and I am not suggesting reverting it, but it has unintended consequences that we should fix. 

Possibly the issue is in the way sempass is handling pragmas:
```nim
proc p {.mymacro.}
```
internally is rewritten into:
```nim
mymacro:
[...]

PR #15332, opened 2020-09-15
proc params as syms

cooldome:
attempt to fix test failures for https://github.com/nim-lang/Nim/pull/15252


Clyybber (approved):
Thanks for fixing the tests/packages!

cooldome:
@Clyybber  happy to help.

Since you are here. Could you please review https://github.com/nim-lang/Nim/pull/15341. 
It is also required for the workaround https://github.com/nim-lang/Nim/issues/15324

PR #15336, opened 2020-09-16
fix #15333

### LINKED ISSUE #15333: uri.nim url with literal ipv6 address is printed wrong, and cannot parsed again
When url with ipv6 address is converted to string, the hostname enclosing
"[", and "]" are missing.
This makes the url unparseable.

### Example
```nim
import uri

let org = "udp://[2001:0db8:85a3:0000:0000:8a2e:0370:7334]:8080"
let url = parseUri(org)
echo $url # udp://2001:0db8:85a3:0000:0000:8a2e:0370:7334:8080
echo $url.hostname == "2001:0db8:85a3:0000:0000:8a2e:0370:7334" # true
echo parseUri($url) # udp://2001:0db885a3000000008a2e037073348080
```

```
Compiled at 2020-09-13

```

ringabout:


lib/pure/uri.nim
```diff
@@ -52,10 +52,14 @@ include includes/decode_helpers
 type
   Url* = distinct string
 
+  IpKind* = enum
```
Araq: You unconditionally export this symbol.

ringabout: I will remove `IpKind` and use `initUri(useIpv6: bool)` for compatibility.

PR #15338, opened 2020-09-16
Methods docs improvement

ringabout:
The title in manual is wrong. Because the example is single dispatch.


juancarlospaco:
I think the `{.inline.}` should be removed to make the example as minimal as possible, is kinda off-topic also.
:)

ringabout:
I just steal the examples from Nim Tutorial (Part II) :)

PR #15341, opened 2020-09-16
fix #15326

### LINKED ISSUE #15326: typed macros break generic proc definitions
### Example
```nim
import macros
macro m(n:typed):auto =
  result = n
proc f[T](x:T):T {.m.}= x
echo f 3
```

### Current Output
```
/Users/jin/Work/npg/mtypedgen.nim(4, 7) Error: illformed AST: [T]
```

### Additional Information

```
$ nim -v
Compiled at 2020-09-14

git hash: c38487aa225674f67659c28ce171fd93e2231ad1
```

cooldome:
fix #15326

Allow for sem checking already typed procs.

Changes are a lot smaller than `git diff` shows.

### DIFF
--- a/compiler/semtypes.nim
+++ b/compiler/semtypes.nim
@@ -2054,72 +2054,80 @@ proc semGenericConstraints(c: PContext, x: PType): PType =
   result = newTypeWithSons(c, tyGenericParam, @[x])
 
 proc semGenericParamList(c: PContext, n: PNode, father: PType = nil): PNode =
+
+  template addSym(result: PNode, s: PSym): untyped = 
+    if father != nil: addSonSkipIntLit(father, s.typ)
+    if sfGenSym notin s.flags: addDecl(c, s)
+    result.add newSymNode(s)
+
   result = copyNode(n)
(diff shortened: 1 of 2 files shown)


cooldome:
Please review.  required for https://github.com/nim-lang/Nim/issues/15324

compiler/semtypes.nim
```diff
@@ -2054,72 +2054,80 @@ proc semGenericConstraints(c: PContext, x: PType): PType =
...
     return
   for i in 0..<n.len:
-    var a = n[i]
-    if a.kind != nkIdentDefs: illFormedAst(n, c.config)
-    var def = a[^1]
-    let constraint = a[^2]
-    var typ: PType
+    var a = n[i]    
+    case a.kind    
+    of nkSym: result.addSym(a.sym)
```
Clyybber, line 2070: Won't this duplicate the generic params, if we previously semmed this node?

cooldome: No, result is empty and we recreate it here reusing symbols

PR #15346, opened 2020-09-16
Revert "Introduce explicit copy"

Araq:
Reverts nim-lang/Nim#15330


Araq:
Reason: 1.4 is just around the corner and this didn't go through the RFC process.

PR #15347, opened 2020-09-17
Documentation strformat

juancarlospaco:
- Documentation only.
- Document 2 features/differences details of `strformat`.

:)

### DIFF
--- a/lib/pure/strformat.nim
+++ b/lib/pure/strformat.nim
@@ -305,6 +305,9 @@ Nim ``strformat`` supports Backslash ``\``:
 
   echo fmt"""{ "yep\nope" }"""  ## Works Ok.
 
+This is a feature of the triple string literals, not of ``strformat`` itself,
+but they can be used togheter to support the Backslash ``\`` on ``strformat``.
+
 
 Backtick and whitespace support
 ===============================
--- a/lib/pure/strformat.nim
+++ b/lib/pure/strformat.nim
@@ -299,24 +299,24 @@ everything is wrapped in a ``block``.
 Backslash support
 =================
 
-Different from Python F-string, Nim ``strformat`` supports Backslash ``\``:
+Nim ``strformat`` supports Backslash ``\``:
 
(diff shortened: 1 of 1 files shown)


lib/pure/strformat.nim
```diff
@@ -296,6 +296,29 @@ to be injected into the context where ``myTemplate`` is expanded,
 everything is wrapped in a ``block``.
 
 
+Backslash support
+=================
+
+Nim ``strformat`` supports Backslash ``\``:
```
Araq: No, it doesn't. That's a feature of the "triple string literal", not of ``strformat``.

juancarlospaco: I know, the point is that it can be used, python one does not support that, it exits with error:

```python
>>> print( f"""{ "yep\nope" }""" )
Error: f-string expression part cannot include a backslash.
```

:)

lib/pure/strformat.nim
```diff
@@ -296,6 +296,32 @@ to be injected into the context where ``myTemplate`` is expanded,
...
+=================
+
+Nim ``strformat`` supports Backslash ``\``, thanks to triple string literals:
+
+.. code-block:: nim
+
+  echo fmt"""{ "yep\nope" }"""  ## Works Ok.
+
+This is a feature of the triple string literals, not of ``strformat`` itself,
+but they can be used togheter to support the Backslash ``\`` on ``strformat``.
```
Araq, line 309: typo: "together"

Araq: "the Backslash"? how about "the backslash"

Clyybber: I think that section should be removed entirely, it belongs to https://github.com/nim-lang/Nim/wiki/Nim-for-Python-Programmers where it already is :)

Clyybber: @juancarlospaco Sorry, I didn't mean the whole PR, just this `Backslash support` section

PR #15352, opened 2020-09-18
string is not nil anymore

ringabout:


lib/pure/strtabs.nim
```diff
@@ -141,10 +141,8 @@ template get(t: StringTableRef, key: string) =
   var index = rawGet(t, key)
   if index >= 0: result = t.data[index].val
   else:
-    when compiles($key):
-      raise newException(KeyError, "key not found: " & $key)
-    else:
-      raise newException(KeyError, "key not found")
+    raise newException(KeyError, "key not found: " & key)
```
Varriount, line 144: Won't this fail to compile if `key` doesn't have a `$` method?

Varriount (changes requested):

lib/pure/strtabs.nim
ringabout: But `key` is always `string`, it doesn't need `$`?

narimiran: I think @Varriount meant to write this comment under the other change - in `critbits.nim`, where `key` might not be a string?

PR #15356, opened 2020-09-18
Test original snippet from #15241

Clyybber:


Clyybber:
@jcosborn I can reproduce the failure locally by running `./koch temp` followed by `./testament/testament --nim:nim_temp c megatest`, does that work for you?

Clyybber:
ping @jcosborn

Clyybber:
ping @jcosborn

jcosborn:
Ok, I got it to work now.  It seems that aliasing `TilesetCT` to `int` causes it to be substituted for `int` later even if that definition is in a different block scope.  The error comes because a `TilesetCT` object gets used in the vm without being properly instantiated, so the size isn't available when constructing the vm code.  Here's a simple reproducer (tested on 1.4.0):
```Nim
block:
  type
    Foo[n: static int] = int

block:
  static:
    let x = int 1
    echo x.type   # Foo
    let y = x + 1
    # Error: unhandled exception: value out of range: -8 notin 0 .. 65535 [RangeDefect]
    echo y
```

PR #15374, opened 2020-09-20
async: removed/disabled some code

Araq:


dom96:
>  removed/disabled some code

That makes it sound like you're disabling a major part of async lol. Please be more specific: removed `unown` in async.

PR #15377, opened 2020-09-21
Documentation only

juancarlospaco:
- Documentation only.
- This is the second part of https://github.com/nim-lang/Nim/pull/15299#issue-484265994

### DIFF
--- a/doc/prelude.rst
+++ b/doc/prelude.rst
@@ -1,0 +1,29 @@
+Prelude
+=======
+
+This is an include file that simply imports common modules for your convenience:
+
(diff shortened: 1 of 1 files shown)


Araq:
Great but how do we link to it? Update tutorial 1 or similar?

juancarlospaco:
I do not understand the question, if Merged it will appear automatically on [TheIndex](https://nim-lang.github.io/Nim/theindex.html), the [Testament doc](https://nim-lang.github.io/Nim/testament.html) is already there as example.
:)

PR #15381, opened 2020-09-21
os: make getApplFreebsd available for NetBSD

alaviss:
NetBSD was switched to getApplFreebsd() in #15359, however the proc was
not made available for the said target.

This commit exposes `getApplFreebsd()` to NetBSD.


narimiran:
@Araq this (hopefully) fixes the nightlies (they started to fail 3 days ago when #15359 was merged) and it is necessary to merge this before we can proceed with any release-related stuff.

PR #15383, opened 2020-09-22
deinitLock

ringabout:
I'm not sure is it necessary.

In windows

Deleting a critical section object releases all system resources used by the object. The caller is responsible for ensuring that the critical section object is unowned and the specified CRITICAL_SECTION structure is not being accessed by any critical section functions called by other threads in the process.

https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-deletecriticalsection

In linux, So the simple answer is yes; you must call this API - and here's the thing - even if the API does nothing at the moment, because although the API itself is fixed forever into the future, the implementation behind the API is not.

https://stackoverflow.com/questions/14721229/is-it-necessary-to-call-pthread-mutex-destroy-on-a-mutex#:~:text=From%20IEEE%20documentation%20which%20is,becomes%2C%20in%20effect%2C%20uninitialized.&text=The%20documentation%20does%20not%20say,good%20practice%20to%20do%20so.

### DIFF
--- a/lib/pure/ioselects/ioselectors_select.nim
+++ b/lib/pure/ioselects/ioselectors_select.nim
@@ -110,6 +110,7 @@ proc close*[T](s: Selector[T]) =
   when hasThreadSupport:
     deallocSharedArray(s.fds)
     deallocShared(cast[pointer](s))
+    deinitLock(s.lock)
 
 when defined(windows):
   proc newSelectEvent*(): SelectEvent =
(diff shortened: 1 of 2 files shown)


ringabout:
A lot of global locks variables in `threadpool.nim`, what should I do?

Araq:
> A lot of global locks variables, what should I do?

Usually they are fine since they are alive for as long as the program runs. The OS cleans up everything after the process stopped. Yes, I know there are systems where this doesn't happen but these systems run so low on resources that in practice nothing needs to be done either.

ringabout:
Thanks, sounds reasonable.

PR #15384, opened 2020-09-22
better nativestacktrace support; refs #15284; backport [1.2]

Araq:


stefantalpalaru:
Nice reorganisation, but you did not test it with https://github.com/status-im/nim-libbacktrace so you did not notice that removing "noinline" from the exported proc types broke it.

A more subtle breakage is not exporting "stacktraces" from "system/excpt.nim" which breaks backwards compatibility in nim-libbacktrace which can't just stick an "import ..." in a "when compiles(...)".

This fixes it:

```diff
diff --git a/lib/system/excpt.nim b/lib/system/excpt.nim
index 1a7473a76..feea48c0c 100644
--- a/lib/system/excpt.nim
+++ b/lib/system/excpt.nim
@@ -12,6 +12,7 @@
 
 import std/private/miscdollars
 import stacktraces
+export stacktraces
 
 var
   errorMessageWriter*: (proc(msg: string) {.tags: [WriteIOEffect], benign,
diff --git a/lib/system/stacktraces.nim b/lib/system/stacktraces.nim
index dce25fdc0..e5998184b 100644
--- a/lib/system/stacktraces.nim
+++ b/lib/system/stacktraces.nim
@@ -22,26 +22,26 @@ when defined(nimStackTraceOverride):
       ## This is the same as the type ``uintptr_t`` in C.
 
     StackTraceOverrideGetTracebackProc* = proc (): string {.
-      nimcall, gcsafe, locks: 0, raises: [], tags: [].}
+      nimcall, noinline, gcsafe, locks: 0, raises: [], tags: [].}
     StackTraceOverrideGetProgramCountersProc* = proc (maxLength: cint): seq[cuintptr_t] {.
-      nimcall, gcsafe, locks: 0, raises: [], tags: [].}
+      nimcall, noinline, gcsafe, locks: 0, raises: [], tags: [].}
     StackTraceOverrideGetDebuggingInfoProc* =
       proc (programCounters: seq[cuintptr_t], maxLength: cint): seq[StackTraceEntry] {.
-        nimcall, gcsafe, locks: 0, raises: [], tags: [].}
+        nimcall, noinline, gcsafe, locks: 0, raises: [], tags: [].}
 
   # Default procedures (not normally used, because people opting in on this
   # override are supposed to register their own versions).
   var
     stackTraceOverrideGetTraceback: StackTraceOverrideGetTracebackProc =
-      proc (): string {.nimcall, gcsafe, locks: 0, raises: [], tags: [].} =
+      proc (): string {.nimcall, noinline, gcsafe, locks: 0, raises: [], tags: [].} =
         discard
         #result = "Stack trace override procedure not registered.\n"
     stackTraceOverrideGetProgramCounters: StackTraceOverrideGetProgramCountersProc =
-      proc (maxLength: cint): seq[cuintptr_t] {.nimcall, gcsafe, locks: 0, raises: [], tags: [].} =
+      proc (maxLength: cint): seq[cuintptr_t] {.nimcall, noinline, gcsafe, locks: 0, raises: [], tags: [].} =
         discard
     stackTraceOverrideGetDebuggingInfo: StackTraceOverrideGetDebuggingInfoProc =
       proc (programCounters: seq[cuintptr_t], maxLength: cint): seq[StackTraceEntry] {.
-        nimcall, gcsafe, locks: 0, raises: [], tags: [].} =
+        nimcall, noinline, gcsafe, locks: 0, raises: [], tags: [].} =
           discard
 
   # Custom procedure registration.
```

mratsim:
I'm surprised we can even inline callbacks. What does the C compiler even produce in the current PR?

stefantalpalaru:
> I'm surprised we can even inline callbacks.

I don't think we can. `getBacktrace()` from nim-libbacktrace can also be used directly and, in that case, I need it to never be inlined so I know exactly how many functions I need to skip from the backtrace: https://github.com/status-im/nim-libbacktrace/blob/c21107e34c9f5d2c9fd5d313a6a5b554ddafa141/libbacktrace.nim#L44

> What does the C compiler even produce in the current PR?

We don't get to the C compiler, because the Nim compiler complains about a type mismatch when trying to register nim-libbacktrace procs.

Araq:
Well avoiding ``export stacktraces`` was the purpose of the reorganization. Any other ideas?

stefantalpalaru:
> Well avoiding export stacktraces was the purpose of the reorganization. Any other ideas?

This seems to do the trick:

```diff
diff --git a/libbacktrace.nim b/libbacktrace.nim
index 437a243..5d9d989 100644
--- a/libbacktrace.nim
+++ b/libbacktrace.nim
@@ -6,6 +6,9 @@
 # This file may not be copied, modified, or distributed except according to
 # those terms.
 
+when (NimMajor, NimMinor, NimPatch) >= (1, 2, 8):
+  import system/stacktraces
+
 # Don't warn that this module is unused (e.g.: when the Nim compiler supports it
 # and users need to import it, even if they don't call getBacktrace() manually).
 {.used.}
```

Araq:
You can now also use ``when defined(nimHasStacktracesModule)``

stefantalpalaru:
Thanks! I just did that in nim-libbacktrace-0.0.4 and everything works as expected.

PR #15386, opened 2020-09-22
Fix forward declarations in shadow scope contexts

### LINKED ISSUE #15385: Regression. Closure variable lifting is affected by recent changes in scope handling
Test case
### Example
```nim

import macros

var captured_funcs {.compileTime.}: seq[NimNode] = @[]

macro aad*(fns: varargs[typed]): typed =
  result = newStmtList()
  for fn in fns:
    captured_funcs.add fn[0]
    result.add fn

func exp*(x: float): float ## get different error if you remove forward declaration

func exp*(x: float): float {.aad.} =
  var x1 = min(max(x, -708.4), 709.8)
  var result: float   ## looks weird because it is taken from template expansion
  result = x1 + 1.0
  result

 
template check_accuracy(f: untyped, rng: Slice[float], n: int, verbose = false): auto =

  proc check_accuracy: tuple[avg_ulp: float, max_ulp: int] {.gensym.} =
    let k = (rng.b - rng.a) / (float) n
    var
      res, x: float
      i, max_ulp = 0
      avg_ulp = 0.0

    x = rng.a
    while (i < n):
      res = f(x)
      i.inc
      x = x + 0.001
    (avg_ulp, max_ulp)
  check_accuracy()

discard check_accuracy(exp, -730.0..709.4, 40000)
```
Compilation fails with error:
```
Error: 'result' is of type <float> which cannot be captured as it would violate memory safety, declared here: C:\Nim\t7.nim(14, 1); 
```

Without forward declaration the error is:
```
Error: internal error: expr: param not init x_4871080
```

Clyybber:
Fixes #15385


compiler/lookups.nim
```diff
@@ -91,13 +91,15 @@ proc skipAlias*(s: PSym; n: PNode; conf: ConfigRef): PSym =
       message(conf, n.info, warnDeprecated, "use " & result.name.s & " instead; " &
               s.name.s & " is deprecated")
 
+proc isShadowScope*(s: PScope): bool = s.parent != nil and s.parent.depthLevel == s.depthLevel
```
Araq: Add an ``.inline`` here to please more primitive compilers.

PR #15390, opened 2020-09-22
Fix #15389

### LINKED ISSUE #15389: Regression. Double sem check for procs.
Test case
```nim
import macros
var captured_funcs {.compileTime.}: seq[NimNode] = @[]

macro aad*(fns: varargs[typed]): typed =
  result = newStmtList()
  for fn in fns:
    captured_funcs.add fn[0]
    result.add fn

func exp*(x: float): float {.aad.} =
  var x1 = min(max(x, -708.4), 709.8)
  if x1 > 0.0:
    return x1 + 1.0
  result = 10.0

discard exp(5.0)
```

Compilation fails with error:
```
 expression 'result = x1 + 1.0' has no type (or is ambiguous)
```

cooldome:
allow reuse of result symbols in double sem passes.
`addResult` and `addResultNode` procs replaced with one function `addResult`.

### DIFF
--- a/compiler/semexprs.nim
+++ b/compiler/semexprs.nim
@@ -1754,19 +1754,21 @@ proc semReturn(c: PContext, n: PNode): PNode =
   if c.p.owner.kind in {skConverter, skMethod, skProc, skFunc, skMacro} or
       (not c.p.owner.typ.isNil and isClosureIterator(c.p.owner.typ)):
     if n[0].kind != nkEmpty:
-      if c.p.resultSym != nil:
+      if n[0].kind == nkAsgn and n[0][0].kind == nkSym and c.p.resultSym == n[0][0].sym:
+        discard "return is already transformed"
+      elif c.p.resultSym != nil:
         # transform ``return expr`` to ``result = expr; return``
-        var asgn = newNodeI(nkAsgn, n[0].info)
-        asgn.add newSymNode(c.p.resultSym)
-        asgn.add n[0]        
-        asgn = semAsgn(c, asgn)
-        if asgn[1].kind == nkSym and asgn[1].sym == c.p.resultSym:
-          # optimize away ``result = result``:
-          result = newTreeI(nkReturnStmt, n.info, c.graph.emptyNode)
-        else:
-          result = newTreeI(nkStmtList, n.info, asgn, newTreeI(nkReturnStmt, n.info, c.graph.emptyNode))        
+        var a = newNodeI(nkAsgn, n[0].info)
+        a.add newSymNode(c.p.resultSym)
+        a.add n[0]
+        n[0] = a
       else:        
         localError(c.config, n.info, errNoReturnTypeDeclared)
+        return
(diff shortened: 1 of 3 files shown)


compiler/semexprs.nim
```diff
@@ -1754,17 +1754,21 @@ proc semReturn(c: PContext, n: PNode): PNode =
   if c.p.owner.kind in {skConverter, skMethod, skProc, skFunc, skMacro} or
       (not c.p.owner.typ.isNil and isClosureIterator(c.p.owner.typ)):
     if n[0].kind != nkEmpty:
-      # transform ``return expr`` to ``result = expr; return``
-      if c.p.resultSym != nil:
+      if n[0].kind == nkAsgn and n[0][0].kind == nkSym and c.p.resultSym == n[0][0].sym:
+        discard "return is already transformed"
+      elif c.p.resultSym != nil:
+        # transform ``return expr`` to ``result = expr; return``
```
Clyybber, line 1760: Maybe the simpler fix is to transform this according to the comment, so `result = expr; return` instead of the current `return result = expr`

cooldome: then how you can avoid double transformation in second sempass?

Clyybber: second sempass won't touch the `return` because `n[0].kind == nkEmpty`.

compiler/semexprs.nim
```diff
@@ -1754,16 +1754,18 @@ proc semReturn(c: PContext, n: PNode): PNode =
...
-        n[0] = semAsgn(c, a)
-        # optimize away ``result = result``:
-        if n[0][1].kind == nkSym and n[0][1].sym == c.p.resultSym:
-          n[0] = c.graph.emptyNode
-      else:
+        # transform ``return expr`` to ``result = expr; return``
+        var asgn = newNodeI(nkAsgn, n[0].info)
+        asgn.add newSymNode(c.p.resultSym)
+        asgn.add n[0]        
+        asgn = semAsgn(c, asgn)
```
Clyybber, line 1762:
```suggestion
        let asgn = semAsgn(c, newNodeI(nkAsgn, n[0].info, newSymNode(c.p.resultSym), n[0]))
```
Btw, we now have `newTreeI`/`newTreeIT` similar to `newNodeI` and `newNodeIT` for convinience: https://github.com/nim-lang/Nim/blob/devel/compiler/ast.nim#L1151

compiler/semexprs.nim
Clyybber: I think the tests need to be adapted (macros/tmacrostmt.nim)

cooldome:
@Clyybber, tried your approach. It introduces many breaking changes.

cooldome:
I can confirm that with this fix the tail of regressions caused by double sempass for procs is now over.
I can now compile my project.

Please review the fix

cooldome:
fixes #15389 for github link

Clyybber (approved):

PR #15399, opened 2020-09-24
fix #11537, correctly parse inline code without surrounding spaces

### LINKED ISSUE #11537: DocGen HTML output appears improperly when encountering text immediately after/before inline monospace; in some cases won't compile
When the documentation generator encounters a non-whitespace character (other than `\`) immediately after properly closed inline monospace styling, the HTML either:
- fails to compile (if \`\` is used, or if there is an empty line after the offending text)
- doesn't actually close the styling until multiple \` characters later

If a non-whitespace character occurs immediately *before* inline monospace styling, the styling in question is not applied at all.

[Pandoc](https://pandoc.org)'s reStructuredText implementation doesn't appear to have the same problem.

#### Single backtick, no empty line, one character after
```nim
type SampleType* = int
  ## you should use hundreds of these `SampleType`s shouldn't you
  ## think `about` it
```
#### Same as above, but with double backticks on first line
```nim
type SampleType* = int
  ## you should use hundreds of these ``SampleType``s shouldn't you
  ## think `about` it
```
#### Double backticks, no empty line, characters before
```nim
proc sampleProc*(a, b: int): int =
  ## This procedure adds two``integer`` parameters together and returns only if
  ## you truly want it to.
  result = a + b
```
#### Single backticks, empty line, one character after
```nim
proc sampleProc*(a, b: int): int =
  ## This procedure adds two `integer`s together and never returns, even if you
  ##
  ## really `want that` to `happen`.
  discard
```

##### Compilation
Successful.
##### Result
docissue1

##### Compilation
Unsuccessful.
```
docissue.nim(1, 51) Error: '``' expected
C:\Users\travis\build\nim-lang\nightlies\nim\e7471cebae2a404f3e4239f199f5a0c422484aac\compiler\lineinfos.nim(229) raiseRecoverableError
Error: unhandled exception: '``' expected [ERecoverableError]
```
##### Compilation
Successful.
##### Result
docissue1

##### Compilation
Unsuccessful.
```
docissue.nim(8, 27) Error: '`' expected
C:\Users\travis\build\nim-lang\nightlies\nim\e7471cebae2a404f3e4239f199f5a0c422484aac\compiler\lineinfos.nim(229) raiseRecoverableError
Error: unhandled exception: '`' expected [ERecoverableError]
```

### Expected Output
All examples should successfully compile and display as the samples under the Possible Solution header do.

### Possible Solution
Inserting a backslash (`\`) character before the character that would immediately follow the monospace results in normal looking output. For instance...

```nim
## you should use hundreds of these `SampleType`\s shouldn't you
## think `about` it
```
results in:
docissue1

And the following, which would fail to compile without the backslash...
```nim
## This procedure adds two `integer`\s together and never returns, even if you
##
## really `want that` to `happen`.
```
results in:
docissue1

### Additional Information
```
$ nim -v
Nim Compiler Version 0.20.0 [Windows: amd64]
```
I am unaware if previous versions have this same issue.

narimiran:


doc/advopt.txt
```diff
@@ -13,7 +13,7 @@ Advanced commands:
   //genDepend               generate a DOT file containing the
                             module dependency graph
   //dump                    dump all defined conditionals and search paths
-                            see also: --dump.format:json (useful with: ` | jq`)
+                            see also: --dump.format:json (useful with: `| jq`)
```
Araq, line 16: So it's a breaking change?

narimiran: This wasn't rendered properly before: it wasn't an inline code but a normal text containing backticks: ``(useful with: ` | jq`)``. (See the existing documentation: https://nim-lang.github.io/Nim/nimc.html#compiler-usage-command-line-switches)

Per rule 2 (line 531, https://github.com/nim-lang/Nim/blob/b552cf722ccdf7bf567aea69f3286dd53066f3b1/lib/packages/docutils/rst.nim#L531), we don't allow whitespace after the starting `` ` `` to be considered as a valid inline code block.
(And yes, keeping it as it was before would now cause a failure)

PR #15410, opened 2020-09-24
fix #15405. deepcopy arc

### LINKED ISSUE #15405: Deepcopy in arc crashes
Deepcopy recently appeared  for arc runtime, first test results.
Test case:
```nim
import lexbase, streams
const test = "<A><B>value</B></A>"
var stream = newStringStream(test)
var lex: BaseLexer
open(lex, stream)
var lex2 = deepCopy(lex)
```
Crashes on the nil pointer dereference with stacktrace:
```nim
C:\Nim\t8.nim(24)        t8
C:\Nim\lib\system.nim(2881) deepCopy
C:\Nim\lib\system\deepcopy.nim(192) genericDeepCopy
C:\Nim\lib\system\deepcopy.nim(130) genericDeepCopyAux
C:\Nim\lib\system\deepcopy.nim(72) genericDeepCopyAux
C:\Nim\lib\system\deepcopy.nim(68) genericDeepCopyAux
C:\Nim\lib\system\deepcopy.nim(168) genericDeepCopyAux
```

I have done a bit of investigation.: `.typeInfoV1` member is nil for the `NTIv2` struct of the `StringStreamObj` object.

`typeInfoV1` is not nil for `StreamObj` which is inheritance base for `StringStreamObj`. AFAIK,  `genTypeInfoV1` codegen is done only for type `BaseLexer` that includes reference to  `StreamObj`. Because `StringStreamObj` is not directly referenced, `genTypeInfoV1` is not executed for it.

cooldome:
Fix makes TypeInfoV1 emitted for all inherited objects.
Not the most efficient approach, but it is ok until V1 type info is in use.


Araq:
That defeats the purpose of the new-styled RTTI quite a bit. The new RTTI was designed for code sizes working on micro-controllers. More philosophically speaking, preparing every type for ``deepCopy`` which is almost impossible to strip out of the binary is against Nim's design. I vote for an explicit ``enableDeepCopy`` solution.

PR #15412, opened 2020-09-26
Fix proc generic params ident defs, missing empty

metagn:


disruptek:
Can we also apply this to procs (and, really, assert that it applies everywhere else)?  I never bothered to figure out why the minor form exists, but since you brought it up... 😉

metagn:
Are you asking to write a test or to make the same change to another wrong example? Because I couldn't find another wrong example in the doc, and writing a test seems like too much for such a small thing. If you're asking for general proof whether this is correct, [here is a playground snippet](https://play.nim-lang.org/#ix=2yQ9)

disruptek:
It’s sometimes the case that IdentDefs exist with only two nodes; I’m just wondering if we cannot remove this form that omits a final Empty node.  Is there a use-case for the short form or can we make the AST more consistent here?

Because it’s something I have to code around frequently and I’d rather the compiler did it for me...

PR #15417, opened 2020-09-28
odbc regression from #14357

shirleyquirk:
[forum post that got me here](https://forum.nim-lang.org/t/6860)
[commit responsible for this regression](https://github.com/nim-lang/Nim/pull/14357/commits/fe51c23bd22cf8531d849d00308395eb272a8e03)

third argument to SqlSetEnvAttr, ValuePtr described in Microsoft docs:
>ValuePtr
>[Input] Pointer to the value to be associated with Attribute. Depending on the value of Attribute, ValuePtr will be a 32-bit integer value or point to a null-terminated character string.

before the offending commit, 'val', a TSqlInteger, was being passed in, which is what SqlSetEnvAttr expects.  I get why the committer couldn't bear to cast a literal 3 to a pointer, but, hey, that's how she rolls.


shirleyquirk:
sanity check: search results through github for usage:
https://github.com/search?q=SqlSetEnvAttr+SQL_OV_ODBC3&type=Code
it's passed in straight ``SQL_OV_ODBC3`` or cast to ``(void *)SQL_OV_ODBC3`` or ``(SQLPOINTER)SQL_OV_ODBC3``

Araq:
@cooldome I'm merging this for now. Feel free to review and complain later.

PR #15421, opened 2020-09-28
Add parseFloatThousandSep

juancarlospaco:
- Convenience func for `parseFloat` designed to parse floats with thousand separators as found in the wild formatted for humans.
- Documentation, `runnableExamples` with `doAssert`, `since`, changelog, etc.

```nim
doAssert parseFloatThousandSep("1,000,000.000") == 1000000.0
doAssert parseFloatThousandSep("1'000'000.000", thousandSep = '\'') == 1000000.0

echo parseFloat("1,000,000.000")
  Error: unhandled exception: invalid float: 1,000,000.000 [ValueError]
echo parseFloat("1'000'000.000")
  Error: unhandled exception: invalid float: 1'000'000.000 [ValueError]
```

:)

## links
* [Formatting a float to currency - Nim forum](https://forum.nim-lang.org/t/8162)


lib/pure/strmisc.nim
```diff
@@ -85,6 +86,41 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+    doAssert parseFloatThousandSep("-1.0") == -1.0
+    doAssert parseFloatThousandSep("1.000") == 1.0
+    doAssert parseFloatThousandSep("1,000") == 1.0
+    doAssert parseFloatThousandSep("10,000.000") == 10000.0
+    doAssert parseFloatThousandSep("10.000,000") == 10.0
+    doAssert parseFloatThousandSep("1,000,000.000") == 1000000.0
+    doAssert parseFloatThousandSep("1.000.000,000") == 1000000.0
+    doAssert parseFloatThousandSep("0000000000000001.00,0,000,00000,,00") == 1.0
+    doAssert parseFloatThousandSep("0000000000000001,00.0.000.00000..00") == 1.0
+    doAssert parseFloatThousandSep("000,1.000,,,,,,,,,,,,,,000,,,,,0000") == 1.0
```
Araq: er... what? Since it performs some guesswork already, it should be stricter and don't allow ",,," at all.

Vindaar:
I think this might be a good addition.

However, if we are to have something like this I would trade correctness for speed. Perform a first pass over the string and make explicit checks, namely:
- no separator before a digit
- first separator can be anywhere after first digit
- there *has* to be 3 digits between successive separators and between the last separator and the decimal dot
- no separator after decimal dot
- ...? (maybe I forgot something else)

Have that pass over the string remove the separators and hand the result to normal `parseFloat`.

juancarlospaco:
@Vindaar 
Have any pseudocode, code or github suggestion on how you imagine it?, your comment makes sense.

juancarlospaco:
This is for strings as formatted for humans with quirky punctuation, 
sometimes used for money, coords, big floats on sciences, etc etc.

Thats why I named differently, because does different things than `parseFloat`.
Can be made stricter and have more options yeah...

lib/pure/strmisc.nim
juancarlospaco: Maybe I exaggerated the example too much to try to break it lol :laughing:

Vindaar:
@juancarlospaco I'll write a couple of lines later if I don't forget (ping me otherwise). Not a parsing expert though. :P

Vindaar:
Probably neither the smartest, nor most efficient way, but this is what I came up with:

edit: haha, I just noticed I forgot about both minus and exp notation. The latter isn't important (who in their right mind would combine the two, but the former kinda is). If you think this is useful, I can add it. But won't do it now. Doesn't change the idea / approach.

```nim
import parseutils, strutils

proc parseFloatThousandSep(s: string,
                           sep: static char = ',',
                           decimalDot: static char = '.'): float = # maybe should not have a default sep?
  ## version of `parseFloat` which allows for thousand separators.
  ## The following assumptions / requirements must be met by the string
  ## - no separator before a digit
  ## - first separator can be anywhere after first digit, but no more than 3 characters
  ## - there ``has`` to be 3 digits between successive separators
  ##   - and between the last separator and the decimal dot
  ## - no separator after decimal dot
  ## - no duplicate separators
  ## - floats without separator allowed
  var
    buf = s
    idx = 0
    successive = 0
    afterDot = false
    lastWasDot = false
    lastWasSep = false
    hasAnySep = false
  template bail(msg: untyped): untyped =
    raise newException(ValueError, "Invalid float containing thousand separator." &
      " Reason: " & $msg)

  while idx < buf.len:
    case buf[idx]
    of sep:
      if idx == 0:
        bail("String starts with thousand separator.")
      elif lastWasSep:
        bail("Two separators in a row.")
      elif afterDot:
        bail("Separator found after decimal dot.")
      buf.delete(idx, idx)
      lastWasSep = true
      hasAnySep = true
      successive = 0
    of '0' .. '9':
      if hasAnySep and successive > 2:
        bail("More than 3 digits between thousand separators.")
      lastWasSep = false
      lastWasDot = false
      inc successive
      inc idx
    of decimalDot:
      if idx == 0:
        bail("String starts with decimal dot.")
      elif hasAnySep and successive != 3:
        bail("Not 3 successive digits before decimal point, despite larger 1000!")
      successive = 0
      lastWasDot = true
      afterDot = true
      inc idx
    else:
      # NOTE: could also move separator logic here if we wanted runtime separator
      # selection. Case needs CT info
      bail("Invalid character in float: " & $buf[idx])
  result = buf.parseFloat


doAssert parseFloatThousandSep("1.0") == 1.0
doAssert parseFloatThousandSep("1.000") == 1.0
doAssert parseFloatThousandSep("1,000") == 1000.0
doAssertRaises(ValueError):
  # invalid because , not the sep
  discard parseFloatThousandSep("1,000", sep = '\'')
# compile time error due to duplicate case label
# parseFloatThousandSep("1,000", sep = '.')
doAssert parseFloatThousandSep("10,000.000") == 10000.0
doAssertRaises(ValueError):
  # thousand sep after decimal dot
  discard parseFloatThousandSep("10.000,000")
doAssert parseFloatThousandSep("1,000,000.000") == 1000000.0
doAssert parseFloatThousandSep("10,000,000.000") == 10000000.0
doAssertRaises(ValueError):
  # starts with sep
  discard parseFloatThousandSep(",123.000")
doAssertRaises(ValueError):
  # starts with decimal dot
  discard parseFloatThousandSep(".000")
doAssertRaises(ValueError):
  # duplicate thousand sep
  discard parseFloatThousandSep("123,,100.0")
doAssertRaises(ValueError):
  # sep before dot
  discard parseFloatThousandSep("123,.0")
```

Up to someone else to decide if the separator should be compile time, but this made the code a lot nicer imo. Feel free to take it or leave it. :)

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,137 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+      of '0' .. '9':  # Digits
+        if hasAnySep and successive > 2:
+          bail("more than 3 digits between thousand separators.")
+        else:
+          lastWasSep = false
+          lastWasDot = false
+          inc successive
+          inc idx
+      of sep:  # Thousands separator
+        if idx == 0:
```
Vindaar: Since you now added the handling of `-`, the `idx == 0` check isn't enough anymore. In case of `isNegative` it's gotta be idx 1.

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,137 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+          bail("two separators in a row.")
+        elif afterDot:
+          bail("separator found after decimal dot.")
+        else:
+          s.delete(idx, idx)
+          lastWasSep = true
+          hasAnySep = true
+          successive = 0
+      of decimalDot:
+        if idx == 0:  # Wont allow .1
```
Vindaar: Same as above.

dom96:
Cool, can we also get the opposite of this? :)

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,139 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+    doAssert parseFloatThousandSep("1,000,000.000") == 1000000.0
+    doAssert parseFloatThousandSep("10,000,000.000") == 10000000.0
+    doAssert parseFloatThousandSep("10.000,0", '.', ',') == 10000.0
+    doAssert parseFloatThousandSep("1'000'000,000", '\'', ',') == 1000000.0
+  var copiedString = s
+  parseFloatThousandSepImpl(copiedString, sep, decimalDot)
+
+
+func parseFloatThousandSep*(s: var string; sep: static char = ','; decimalDot: static char = '.'): float {.since: (1, 3).} =
+  ## In-place version of `parseFloatThousandSep`, does not copy the string internally.
```
Araq: Why does parsing require an internal copy? Remove this overload.

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,139 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+          bail("the '-' character can only be at the start of the string.")
+        else:
+          isNegative = true
+          inc idx
+      else:
+        bail("invalid character in float: " & $s[idx])
+  parseFloat(s)
+
+
+func parseFloatThousandSep*(s: string; sep: static char = ','; decimalDot: static char = '.'): float {.since: (1, 3).} =
```
Araq: Remove the ``static`` keywords, it's bad API design.

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,139 @@ proc rpartition*(s: string, sep: string): (string, string, string)
 
   return partition(s, sep, right = true)
 
+
+template parseFloatThousandSepImpl(s: var string; sep: static char; decimalDot: static char): float =
+  assert sep notin {'-', ' '} and decimalDot notin {'-', ' '} and sep != decimalDot
+
+  template bail(m: string) =
```
Araq: ``template bail`` should be ``proc bail``.

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,139 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+    raise newException(ValueError, "Invalid float containing thousand separators, " & m)
+
+  if likely(s.len > 1): # Allow "0" thats valid, is 0.0
```
Araq: How do you know what is "likely"? Remove these silly likely annotations.

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,139 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+          inc idx
+      of sep:  # Thousands separator
+        if unlikely(isNegative and idx == 1 or idx == 0):
+          bail("string starts with thousand separator.")
+        elif lastWasSep:
+          bail("two separators in a row.")
+        elif afterDot:
+          bail("separator found after decimal dot.")
+        else:
+          s.delete(idx, idx)
```
Araq: Instead of ``s.delete`` copy ``s`` into a sanitized buffer like so:

```nim

  var buf = newStringOfCap(s.len)
  var i = 0
  while i < s.len:
     ...
     buf.add s[i]
  ...
  parseFloat(buf)
```

juancarlospaco:
@Araq This is ready for Merge. :slightly_smiling_face:

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,102 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
 
+
+func parseFloatThousandSep*(str: string; sep = ','; decimalDot = '.'): float {.since: (1, 3).} =
+  ## Convenience func for `parseFloat` which allows for thousand separators,
+  ## this is designed to parse floats as found in the wild formatted for humans.
+  ##
+  ## The following assumptions and requirements must be met:
+  ## - String must not be empty.
+  ## - String must be stripped of trailing and leading whitespaces.
+  ## - `sep` must not be `'-'` nor `' '`.
```
Clyybber: space as seperator should also be supported for ISO 31-0.

changelog.md
```diff
@@ -31,6 +31,152 @@
 - Removed deprecated `iup` module from stdlib, it has already moved to
   [nimble](https://github.com/nim-lang/iup).
 
+  The following modules now compile on both JS and NimScript: `parsecsv`,
```
timotheecour: changelog needs fixing

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,102 @@ proc rpartition*(s: string, sep: string): (string, string, string)
 
   return partition(s, sep, right = true)
 
+
+func parseFloatThousandSep*(str: string; sep = ','; decimalDot = '.'): float {.since: (1, 3).} =
+  ## Convenience func for `parseFloat` which allows for thousand separators,
```
timotheecour: must be cross referenced from `parseFloat`

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,102 @@ proc rpartition*(s: string, sep: string): (string, string, string)
 
   return partition(s, sep, right = true)
 
+
+func parseFloatThousandSep*(str: string; sep = ','; decimalDot = '.'): float {.since: (1, 3).} =
```
timotheecour: fix since

changelog.md
juancarlospaco: Yeah I noticed it still conflicting, even if it suppose not to :(

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,102 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+            "Invalid float containing thousand separators, more than 3 digits between thousand separators.")
+        else:
+          s.add c
+          lastWasSep = false
+          lastWasDot = false
+          inc successive
+          inc idx
+      if c == sep:  # Thousands separator, this is NOT the dot
+        if unlikely(isNegative and idx == 1 or idx == 0):
+          raise newException(ValueError,
```
timotheecour: need to report relevant runtime values on each error: str + index

changelog.md
timotheecour: then vote for re-opening https://github.com/nim-lang/Nim/pull/14025

juancarlospaco: I like the idea as starter for a change, but ultimately a script should do it automatically like my example, 
1 tiny script that does 1 thing, I mean [theres even less useful dead stuff in the repo anyways](https://github.com/nim-lang/Nim/blob/devel/tools/nimrepl.nim).

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,110 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+  ## - There has to be 3 digits between the last separator and the decimal dot.
+  ## - No separator after decimal dot.
+  ## - No duplicate separators.
+  ## - Floats without separator allowed.
+  ##
+  ## See also:
+  ## * `strutils <strutils.html>`_
+  runnableExamples:
+    doAssert parseFloatThousandSep("0") == 0.0
+    doAssert parseFloatThousandSep("-0") == -0.0
```
timotheecour: put everything that's obvious (for humans) but that still needs testing into tests/stdlib/tstrmisc.nim

this includes things like `doAssert parseFloatThousandSep("-0.0") == -0.0` (note that -0.0 == 0.0)

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,110 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+          raise newException(ValueError,
+            "Invalid float containing thousand separators, string starts with thousand separator '" &
+            c & "' at index " & $idx)
+        elif lastWasSep:
+          raise newException(ValueError,
+            "Invalid float containing thousand separators, two separators in a row '" &
+            c & "' at index " & $idx)
+        elif afterDot:
+          raise newException(ValueError,
+            "Invalid float containing thousand separators, separator found after decimal dot '" &
```
timotheecour: * as @araq would say this causes code bloat; (as a consequence, this can lead to worse performance because of instruction cache)

this can be seen by comparing codegen for:
```nim
when defined case1:
  proc main(a: int)=
    if a == 2: raise newException(ValueError, "ok2")
    elif a == 3: raise newException(ValueError, "ok3")
    elif a == 1: raise newException(ValueError, "ok1")
  main(1)

when defined case2:
  proc raise2(msg: string) =
    raise newException(ValueError, msg)
  proc main(a: int)=
    if a == 2: raise2("ok2")
    elif a == 3: raise2("ok2")
    elif a == 1: raise2("ok1")
  main(1)
```

* you can fix via an auxiliary proc:

```nim
proc raiseError(index: int, input: string) {.noinline, noreturn.} =
  raise newException(ValueError, "Invalid char at position: " & $index & " for input: " & input)
```

note, I'm not sure you really need to specify more detailed error messages (ie 1 generic one like above should suffice), but if you do, then just pass an additional `msg = ""` for more info for cases that need it)

lib/pure/strutils.nim
```diff
@@ -1150,6 +1150,9 @@ proc parseFloat*(s: string): float {.noSideEffect,
   ##
   ## If `s` is not a valid floating point number, `ValueError` is raised.
   ##``NAN``, ``INF``, ``-INF`` are also supported (case insensitive comparison).
+  ##
+  ## See also:
+  ## * `strmisc <strmisc.html>`_
```
timotheecour: `parseFloatThousandSep` should appear in the link; if annoying to type (which I understand, see https://github.com/nim-lang/RFCs/issues/125 for better doc links), just use:
```See also: `strmisc.parseFloatThousandSep` ```

File: lib/pure/strmisc.nim [hunk 17]
```diff
@@ -85,6 +86,102 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+  ## - There has to be 3 digits between successive separators.
+  ## - There has to be 3 digits between the last separator and the decimal dot.
+  ## - No separator after decimal dot.
+  ## - No duplicate separators.
+  ## - Floats without separator allowed.
+  ##
+  ## See also:
+  ## * `strutils <strutils.html>`_
+  runnableExamples:
+    doAssert parseFloatThousandSep("1,000") == 1000.0
```
timotheecour: add tests/stdlib/tstrmisc.nim and add at least a few error cases eg:
```nim
doAssertRaises(ValueError): parseFloatThousandSep("1,0000.000")
```
(+ whatever non-error cases you judge as important to test)

tests/stdlib/tstrmisc.nim
```diff
@@ -0,0 +1,20 @@
+discard """
+  action: "run"
```
timotheecour: remove this discard section, it's superfluous in this case

tests/stdlib/tstrmisc.nim
```diff
@@ -10,6 +10,12 @@ func main() =
   doAssert parseFloatThousandSep("-1.0") == -1.0
   doAssert parseFloatThousandSep("1.000") == 1.0
   doAssert parseFloatThousandSep("-1.000") == -1.0
+  doAssertRaises(ValueError):
```
timotheecour: weak test; will succeed even if only 1 stmt fails (instead, ensure all stmts fail)
```nim
for s in ["bad1", "bad2", ... ]:
  doAssertRaises(ValueError): discard parseFloatThousandSep(s)
```

lib/pure/strmisc.nim
```diff
@@ -116,63 +116,44 @@ func parseFloatThousandSep*(str: string; sep = ','; decimalDot = '.'): float {.s
     doAssert parseFloatThousandSep("10.000,0", '.', ',') == 10000.0
     doAssert parseFloatThousandSep("1'000'000,000", '\'', ',') == 1000000.0
   assert sep != '-' and decimalDot notin {'-', ' '} and sep != decimalDot
+
+  proc raiseError(i: int, c: char) {.noinline, noreturn.} =
```
timotheecour: `str` must be reported (c can be derived from i + str, if you want to also show `c`), otherwise  user won't have enough context to understand what triggered the exception in the general case where str isn't some litteral but is some arbitrary value

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,84 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+  ## - String must be stripped of trailing and leading whitespaces.
+  ## - `sep` must not be `'-'`.
+  ## - `decimalDot` must not be `'-'` nor `' '`.
+  ## - `sep` and `decimalDot` must be different.
+  ## - No separator before a digit.
+  ## - First separator can be anywhere after first digit, but no more than 3 characters.
+  ## - There has to be 3 digits between successive separators.
+  ## - There has to be 3 digits between the last separator and the decimal dot.
+  ## - No separator after decimal dot.
+  ## - No duplicate separators.
```
timotheecour: can all these be replaced by the following rule, which seems clearer?
`exactly 1 separator must appear after each 3 consecutive inner digits to the left of the dot, and nowhere else`
(unless I'm missing something)

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,84 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+  ## - Floats without separator allowed.
```
timotheecour: not reflected in your examples; did you mean s/separator/decimalDot ?

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,84 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+    doAssert parseFloatThousandSep("10,000.000") == 10000.0
+    doAssert parseFloatThousandSep("1,000,000.000") == 1000000.0
+    doAssert parseFloatThousandSep("10,000,000.000") == 10000000.0
+    doAssert parseFloatThousandSep("10.000,0", '.', ',') == 10000.0
+    doAssert parseFloatThousandSep("1'000'000,000", '\'', ',') == 1000000.0
+  assert sep != '-' and decimalDot notin {'-', ' '} and sep != decimalDot
+
+  proc raiseError(i: int; c: char) {.noinline, noreturn.} =
+    raise newException(ValueError,
+      "Invalid float containing thousand separators, invalid char '" &
```
timotheecour: `"xxx" % [c, $i, str]`

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,84 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+      c & "' at index " & $i & " for input " & str)
+
+  if str.len > 1: # Allow "0" thats valid, is 0.0
```
timotheecour: grammar+clarity
Allow "0" which is valid, equal to 0.0

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,84 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+          inc idx
+      if c == sep:  # Thousands separator, this is NOT the dot
+        if lastWasSep or afterDot or (isNegative and idx == 1 or idx == 0):
+          raiseError(idx, c)
+        else:
+          lastWasSep = true # Do NOT add the Thousands separator here.
+          hasAnySep = true
+          successive = 0
+      if c == decimalDot:  # This is the dot
+        if (isNegative and idx == 1 or idx == 0) or (hasAnySep and successive != 3):  # Wont allow .1
```
timotheecour: Won't

tests/stdlib/tstrmisc.nim
```diff
@@ -0,0 +1,21 @@
...
+  doAssert parseFloatThousandSep("0.0") == 0.0
+  doAssert parseFloatThousandSep("1.0") == 1.0
+  doAssert parseFloatThousandSep("-0.0") == -0.0
+  doAssert parseFloatThousandSep("-1.0") == -1.0
+  doAssert parseFloatThousandSep("1.000") == 1.0
+  doAssert parseFloatThousandSep("-1.000") == -1.0
+  doAssertRaises(ValueError): discard parseFloatThousandSep("1,0000.000")
+  doAssertRaises(ValueError): discard parseFloatThousandSep("--")
+  doAssertRaises(ValueError): discard parseFloatThousandSep("..")
+  doAssertRaises(ValueError): discard parseFloatThousandSep("000,1.000,,,,,,,,,,,,,,000,,,,,0000")
```
timotheecour: always show the "least extreme" failure modes (ie at the boundary) otherwise this weakens the test (and also confuses the reader);
this is not a great example because it has multiple errors; instead maybe:
`parseFloatThousandSep("1,,000")`

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,78 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+    doAssert parseFloatThousandSep("1,000,000.000") == 1000000.0
+    doAssert parseFloatThousandSep("10,000,000.000") == 10000000.0
+    doAssert parseFloatThousandSep("10.000,0", '.', ',') == 10000.0
+    doAssert parseFloatThousandSep("1'000'000,000", '\'', ',') == 1000000.0
+  assert sep != '-' and decimalDot notin {'-', ' '} and sep != decimalDot
+
+  proc raiseError(i: int; c: char; s: string) {.noinline, noreturn.} =
+    raise newException(ValueError,
+      "Invalid float containing thousand separators, invalid char '$1' at index $2 for input $3" %
+      [$c, $i, s])
```
timotheecour: optional: how about `c.repr` ? (eg in cases where c is special (control char, `\0` or other), will be easier to debug via repr)

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,78 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+          successive = 0
+      if c == decimalDot:  # This is the dot
+        if (isNegative and idx == 1 or idx == 0) or (hasAnySep and successive != 3):  # Disallow .1
+          raiseError(idx, c, str)
+        else:
+          s.add '.' # Replace decimalDot to '.' so parseFloat can take it.
+          successive = 0
+          lastWasDot = true
+          afterDot = true
+          inc idx
```
timotheecour: I don't follow the logic, `idx` can get updated a variable number of times in the loop; this will give incorrect error msgs in `raiseError(idx, c, str)` in some cases; instead why not:

`for idx, c in str: ...`

I don't know if you need idx for other purposes that would be different form that `idx`, but if so use another variable; the error msg should refer to the input (str) + position wrt input (idx) so user can easily understand errors

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,78 @@ proc rpartition*(s: string, sep: string): (string, string, string)
 
   return partition(s, sep, right = true)
 
+
+func parseFloatThousandSep*(str: string; sep = ','; decimalDot = '.'): float {.since: (1, 5).} =
```
timotheecour: can this take `openArray[char]` ?

juancarlospaco: But why?, `parseFloat` takes `string`. :thinking: Not sure I understand.

timotheecour: it avoids un-necessary copies in some cases (note that parseFloat should also take openArray, but even if it didn't, you'd still avoid copies in some cases by taking openArray[char])

juancarlospaco: You mean 2 overloads?.

timotheecour:
> Cool, can we also get the opposite of this? :)

+1 but in a different PR; and that one should also be cross-referenced with `formatEng`

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,85 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+    doAssert parseFloatThousandSep("-1,000") == -1000.0
+    doAssert parseFloatThousandSep("10,000.000") == 10000.0
+    doAssert parseFloatThousandSep("1,000,000.000") == 1000000.0
+    doAssert parseFloatThousandSep("10,000,000.000") == 10000000.0
+    doAssert parseFloatThousandSep("10.000,0", '.', ',') == 10000.0
+    doAssert parseFloatThousandSep("1'000'000,000", '\'', ',') == 1000000.0
+  if s.len > 1: parseFloatThousandSepImpl(s, sep, decimalDot) else: parseFloat(s)
+
+
+func parseFloatThousandSep*(s: openArray[char]; sep = ','; decimalDot = '.'): float {.since: (1, 5).} =
```
timotheecour: why a separate overload? `openArray[char]` accepts string as well; the point of using `openArray[char]` instead of string is it allows client code to use slices eg:

`parseFloatThousandSep(input.toOpenArray(1, 4))`

you probably don't even need to test things like
```nim
parseFloatThousandSep(['1', ',', '0', '0', '0']) 
```
where you pass an array of char; that's just testing the semantics of openArray[char], which is already being tested elsewhere

juancarlospaco: Oh, you mean to remove the `string` one?.

timotheecour: yes

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,74 @@ proc rpartition*(s: string, sep: string): (string, string, string)
 
   return partition(s, sep, right = true)
 
+
+func parseFloatThousandSep*(str: openArray[char]; sep = ','; decimalDot = '.'): float {.since: (1, 5).} =
+  ## Convenience func for `parseFloat` which allows for thousand separators,
+  ## this is designed to parse floats as found in the wild formatted for humans.
+  ##
+  ## The following assumptions and requirements must be met:
+  ## - String must not be empty.
```
timotheecour: nit: `s/String/str/` (ditto right below)

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,74 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+  ## - String must be stripped of trailing and leading whitespaces.
+  ## - `sep` must not be `'-'`.
+  ## - `decimalDot` must not be `'-'` nor `' '`.
+  ## - `sep` and `decimalDot` must be different.
+  ## - Exactly 1 separator must appear after each 3 consecutive inner digits to the left of the dot and nowhere else.
+  ##
+  ## See also:
+  ## * `parseFloat <strutils.html#parseFloat,string>`_
+  runnableExamples:
+    doAssert parseFloatThousandSep("-1,000") == -1000.0
```
timotheecour: * add negative test to tstrmisc:
parseFloatThousandSep("1,000000")

* add positive test to examples:
parseFloatThousandSep("1000,000")
parseFloatThousandSep("1000000")

are these supposed to work? from current spec they shouldn't (but maybe my `Exactly 1 separator must appear after each 3 consecutive inner digits to the left of the dot and nowhere else.` wording from a previous comment didn't consider this was valid)

I'm not super sure whether these should work or not, is there some external reference/ how are other standard libraries handling those cases?

juancarlospaco: Yes, because is the thousand separator not the dot.

```nim
doAssert parseFloat("1000000") == 1000000.0
```

timotheecour: * what i mean is that `1000,000` (or `1000000,000` ) violates this:

`Exactly 1 separator must appear after each 3 consecutive inner digits to the left of the dot and nowhere else.`

but are accepted by `parseFloatThousandSep`, so that wording needs to be updated.

tests/stdlib/tstrmisc.nim
```diff
@@ -0,0 +1,22 @@
+import strmisc
+
+
+func main() =
+  doAssert parseFloatThousandSep("0") == 0.0
```
timotheecour: are leading `0` supposed to work? eg `parseFloatThousandSep("001")` currently works

juancarlospaco: Yes, feel free to send PR to `parseFloat` if you think otherwise.

```nim
doAssert parseFloat("0") == 0.0
doAssert parseFloat("001") == 1.0
```

juancarlospaco: https://tio.run/##K6gsycjPM/7/v6AoM69EIy0nP7FEQ8lASVOTC1XEwBBDzNAADIDi//8DAA

lib/pure/strmisc.nim
juancarlospaco: I can revert to the old wording, more verbose but more direct explicit. :thinking:

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,76 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+    doAssert parseFloatThousandSep("10,000.000") == 10000.0
+    doAssert parseFloatThousandSep("1,000,000.000") == 1000000.0
+    doAssert parseFloatThousandSep("10,000,000.000") == 10000000.0
+    doAssert parseFloatThousandSep("10.000,0", '.', ',') == 10000.0
+    doAssert parseFloatThousandSep("1'000'000,000", '\'', ',') == 1000000.0
+    doAssert parseFloatThousandSep("1000,000") == 1000000.0
```
timotheecour: add a comment here, it'll avoid a lengthy/hard to read description at the top:
```nim
you can omit `sep`, but then all subsequent `sep` to the left must also be omitted.
```

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,76 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+    doAssert parseFloatThousandSep("1000000") == 1000000.0
+  assert sep != '-' and decimalDot notin {'-', ' '} and sep != decimalDot
+
+  proc raiseError(i: int; c: char; s: openArray[char]) {.noinline, noreturn.} =
+    raise newException(ValueError,
+      "Invalid float containing thousand separators, invalid char $1 at index $2 for input $3" %
+      [c.repr, $i, s.repr])
+
+  if str.len > 1:
```
timotheecour: why distinguish `if str.len > 1:` ? seems to work without this special case (so you can also remove `result = parseFloat($str[0])` below)

juancarlospaco: To skip all the logic if its just `"0"` ?.

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,84 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+  ## this is designed to parse floats as found in the wild formatted for humans.
+  ##
+  ## The following assumptions and requirements must be met:
+  ## - `str` must not be empty.
+  ## - `str` must be stripped of trailing and leading whitespaces.
+  ## - `sep` must not be `'-'`.
+  ## - `decimalDot` must not be `'-'` nor `' '`.
+  ## - `sep` and `decimalDot` must be different.
+  ## - No separator before a digit.
+  ## - First separator can be anywhere after first digit, but no more than 3 characters.
```
timotheecour: i was confused before, and I'm still confused now that you put this line back; what does it mean?
first is vague (leftmost or rightmost?)
that would kind of imply one of those 2 is valid (depending on leftmost or rightmost) but that's not the case:

parseFloatThousandSep("10000,00.0")
parseFloatThousandSep("10,00000.0")

juancarlospaco: Ultimately user can just look at the code anyways... :shrug:

lib/pure/strmisc.nim
timotheecour: ok fine; at least add a code comment "for performance" (otherwise seems like it's for semantics)

timotheecour: EDIT: meh, I'm not sure singling this case out is really needed, but either way this leads to a bug in your code that needs fixing:
nim r --lib:lib --eval:'import strmisc; echo parseFloatThousandSep("")'

gives IndexDefect instead of ValueError
(because of $str[0])

tests/stdlib/tstrmisc.nim
```diff
@@ -0,0 +1,22 @@
...
+  doAssert parseFloatThousandSep("-0") == -0.0
+  doAssert parseFloatThousandSep("0.0") == 0.0
```
timotheecour, line 5: are these supposed to be valid? currently they are:
```
nim> echo parseFloatThousandSep("1,")
1.0
nim> echo parseFloatThousandSep("1,1")
11.0
```

juancarlospaco: I think so. :slightly_smiling_face:

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,80 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+    doAssert parseFloatThousandSep("1000000") == 1000000.0
+    ## You can omit `sep`, but then all subsequent `sep` to the left must also be omitted:
+    doAssert parseFloatThousandSep("1000,000") == 1000000.0
+
+  assert sep != '-' and decimalDot notin {'-', ' '} and sep != decimalDot
+
+  proc raiseError(i: int; c: char; s: openArray[char]) {.noinline, noreturn.} =
+    raise newException(ValueError,
+      "Invalid float containing thousand separators, invalid char $1 at index $2 for input $3" %
+      [c.repr, $i, s.repr])
```
timotheecour: TODO for future PR:
this will show weird looking errors:
`input ['1', '2', ',', '1', '.'] [ValueError]`
instead of:
`input "12,1." [ValueError]`

I sent out https://github.com/nim-lang/Nim/pull/15951 ; when this is merged we can revisit this using `add`

tests/stdlib/tstrmisc.nim
timotheecour: why though? seems too lax, violating the "thousands" aspect (ie, won't catch bugs).

* note that `locale.atof` accepts it but also accepts arbitrary gargage like `locale.atof('12,3,,4')` so isn't a comparison point
* you previously compared against python3 `float()` but also not a comparison point since `float('1,000')` isn't accepted
* is there another reference to justify these ?

note: marked as unresolved; assuming you marked as resolved by accident? this still seems like a weird behavior

lib/pure/strmisc.nim
Vindaar: @timotheecour It means that a *thousand* separator can appear anywhere after the first digit (referring to one, ten, one hundred thousands) but of course no more than 3 characters, because at that point it wouldn't be a thousand separator anymore. 
Of course once the first thousand separator has appeared there *has* to be 3 characters between each thousand separator and between the last and a possible decimal dot.

tests/stdlib/tstrmisc.nim
Vindaar: I agree that `1,` or `1,1` by default should not be accepted (both can only work if the decimal "point" is supposed to be `,` and the first even only in a more relaxed way than I'm personally happy with).

lib/pure/strmisc.nim
Vindaar: `parseFloatThousandSep("1000,000")` should certainly not work. Since by default `,` is the thousand separator having it after 4 digits doesn't make sense.

lib/pure/strmisc.nim
Vindaar: As said above, this example shouldn't work.

timotheecour:
ping @juancarlospaco (at your leisure); it's getting close to LGTM but the big blocker is https://github.com/nim-lang/Nim/pull/15421#discussion_r522699845; thanks for your patience!

tests/stdlib/tstrmisc.nim
```diff
@@ -0,0 +1,32 @@
...
+  doAssertRaises(ValueError): discard parseFloatThousandSep("--")
+  doAssertRaises(ValueError): discard parseFloatThousandSep("..")
+  doAssertRaises(ValueError): discard parseFloatThousandSep("1,,000")
+  doAssertRaises(ValueError): discard parseFloatThousandSep("1..000")
+  doAssertRaises(ValueError): discard parseFloatThousandSep("1,000000")
+
+  doAssertRaises(ValueError): discard parseFloatThousandSep(",1")
+  doAssertRaises(ValueError): discard parseFloatThousandSep("1,")
+  doAssertRaises(ValueError): discard parseFloatThousandSep("1.")
+  doAssertRaises(ValueError): discard parseFloatThousandSep(".1")
```
timotheecour: false negative:
echo parseFloatThousandSep("1,222.0001")
Error: unhandled exception: Invalid float containing thousand separators

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,130 @@ proc rpartition*(s: string, sep: string): (string, string, string)
 
   return partition(s, sep, right = true)
 
+
+since (1, 5):
+  type ParseFloatOptions* = enum  ## Options for `parseFloatThousandSep`.
+    pfLeadingSep,    ## Allow leading separator, like ",9" and similar.
```
timotheecour: * when is this ever useful?
I've never seen `,9` in the wild, in the rare case someone needs it, he can use a custom pre-processing?

unlike `.9` which does happen in the wild

* ditto, `pfTrailingSep` doesn't seem ever useful in practice
* ditto pfMultipleMinus

There's a difference between "1 size fits all" (bad) and supporting exotic things that never occur in practice

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,124 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+since (1, 5):
+  type ParseFloatOptions* = enum  ## Options for `parseFloatThousandSep`.
+    pfLeadingDot,    ## Allow leading dot, like ".9" and similar.
+    pfTrailingDot,   ## Allow trailing dot, like "9." and similar.
+    pfSepAnywhere,   ## Allow separator anywhere in between, like "9,9", "9,99".
+    pfDotOptional,   ## Allow "9", "-0", integers literals, etc.
+    pfEmptyString    ## Allow "" to return 0.0, so you do not need to do
+                     ## try: parseFloatThousandSep(str) except: 0.0
+
+  func parseFloatThousandSep*(str: openArray[char]; options: set[ParseFloatOptions];
```
timotheecour: `options: set[ParseFloatOptions] = {}` since you've (wisely) used sane defaults

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,124 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+    ## - `sep` must not be `'-'`.
+    ## - `decimalDot` must not be `'-'` nor `' '`.
+    ## - `sep` and `decimalDot` must be different.
+    ## - `str` must be stripped of trailing and leading whitespaces.
+    ## - `pfTrailingDot` and `pfLeadingDot` must not be used together.
+    ##
+    ## See also:
+    ## * `parseFloat <strutils.html#parseFloat,string>`_
+    runnableExamples:
+      doAssert parseFloatThousandSep("10,000.000", {}) == 10000.0
```
timotheecour: * 1st 2 lines redundant with 3rd line from a user perspective (but fine to include in ttstrmisc)
* remove `, {}` assuming you've followed https://github.com/nim-lang/Nim/pull/15421/files#r526538750

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,124 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+    ##
+    ## `parseFloatThousandSep` "prepares" `str` and then calls `parseFloat`,
+    ## consequently `parseFloatThousandSep` by design is slower than `parseFloat`.
+    ##
+    ## The following assumptions and requirements must be met:
+    ## - `sep` must not be `'-'`.
+    ## - `decimalDot` must not be `'-'` nor `' '`.
+    ## - `sep` and `decimalDot` must be different.
+    ## - `str` must be stripped of trailing and leading whitespaces.
+    ## - `pfTrailingDot` and `pfLeadingDot` must not be used together.
```
timotheecour: remove this (and the check); 
this should be legal:
```nim
echo parseFloatThousandSep(x,{pfLeadingDot,pfTrailingDot})
```
with semantics of accepting both `1.0`, `1.`, `.1`, but obviously not `.` (and `.` should be illegal and checked for)

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,124 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+      doAssert parseFloatThousandSep("10.000,0", {}, '.', ',') == 10000.0
+      doAssert parseFloatThousandSep("1'000'000,000", {}, '\'', ',') == 1000000.0
+      doAssert parseFloatThousandSep("1000000", {pfDotOptional}) == 1000000.0
+      doAssert parseFloatThousandSep("-1,000", {pfDotOptional}) == -1000.0
+      ## You can omit `sep`, but then all subsequent `sep` to the left must also be omitted:
+      doAssert parseFloatThousandSep("1000,000", {pfDotOptional}) == 1000000.0
+      ## Examples using different ParseFloatOptions:
+      doAssert parseFloatThousandSep(".1", {pfLeadingDot}) == 0.1
+      doAssert parseFloatThousandSep("1", {pfDotOptional}) == 1.0
+      doAssert parseFloatThousandSep("1.", {pfTrailingDot}) == 1.0
```
timotheecour, line 130: on the fence about making `01.00` accepted by default, but both parseFloat and python accept it so maybe it's fine to not have some yet another `pfLeading0`/`pfTrailing0` flag (that is easy to check in user code, unlike thousand sep);
however, there needs to be an example eg:
`doAssert parseFloatThousandSep("01.00",{}) == 1.0`

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,124 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+      doAssert parseFloatThousandSep("1.0,0,0", {pfSepAnywhere}) == 1.0
```
timotheecour: `doAssert parseFloatThousandSep("10,0.0,0,0", {pfSepAnywhere}) == 100.0` to make semantics crystal clear

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,124 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+        "Invalid float containing thousand separators, invalid char $1 at index $2 for input $3" %
+        [c.repr, $i, s.repr])
+
+    # Fail fast, before looping.
+    let strLen = str.len
+    if strLen == 0: # Empty string.
+      if pfEmptyString notin options:
+        parseFloatThousandSepRaise(0, ' ', "empty string")
+      else:
+        return 0.0 # So user dont need to do `(try: parseFloat(str) except: 0.0)` etc
```
timotheecour: sounds too niche and trivial to work without this flag in the rare case where needed, eg:
`let s = if str.len == 0: 0.0 else: parseFloatThousandSep(str)`

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,124 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+    if str[0] == sep:          # ",1"
+      parseFloatThousandSepRaise(0, sep, str)
+    if pfLeadingDot notin options and str[0] == decimalDot:   # ".1"
+      parseFloatThousandSepRaise(0, decimalDot, str)
+    if str[^1] == sep:        # "1,"
+      parseFloatThousandSepRaise(strLen, sep, str)
+    if pfTrailingDot notin options and str[^1] == decimalDot: # "1."
+      parseFloatThousandSepRaise(strLen, decimalDot, str)
+    if pfSepAnywhere notin options and (not(str.len > 4) and sep in str):
```
timotheecour: not(str.len > 4)
=>
str.len <= 4

tests/stdlib/tstrmisc.nim
```diff
@@ -0,0 +1,47 @@
...
+  doAssert parseFloatThousandSep("1", {pfDotOptional}) == 1.0
+  doAssert parseFloatThousandSep("1.", {pfTrailingDot}) == 1.0
+  doAssert parseFloatThousandSep("1.0,0,0", {pfSepAnywhere}) == 1.0
+  doAssert parseFloatThousandSep("", {pfEmptyString}) == 0.0
+  doAssert parseFloatThousandSep(".10", {pfLeadingDot}) == 0.1
+  doAssert parseFloatThousandSep("10.", {pfTrailingDot}) == 10.0
+  doAssert parseFloatThousandSep("10", {pfEmptyString, pfDotOptional, pfSepAnywhere}) == 10.0
+  doAssert parseFloatThousandSep("1.0,0,0,0,0,0,0,0", {pfSepAnywhere}) == 1.0
+  doAssert parseFloatThousandSep("0,0,0,0,0,0,0,0.1", {pfSepAnywhere}) == 0.1
+
```
timotheecour, line 56: this code is still buggy.

`parseFloatThousandSep("10,00.0",{})` is accepted but shouldn't

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,115 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+      doAssert parseFloatThousandSep("1.", {pfTrailingDot}) == 1.0
+      doAssert parseFloatThousandSep("10,0.0,0,0", {pfSepAnywhere}) == 100.0
+      doAssert parseFloatThousandSep("01.00") == 1.0
+
+    assert sep != '-' and decimalDot notin {'-', ' '} and sep != decimalDot
+
+    proc parseFloatThousandSepRaise(i: int; c: char; s: openArray[char]) {.noinline, noreturn.} =
+      raise newException(ValueError,
+        "Invalid float containing thousand separators, invalid char $1 at index $2 for input $3" %
+        [c.repr, $i, s.repr])
```
Araq: ``repr`` is for debugging, not for stdlib code.

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,116 @@ proc rpartition*(s: string, sep: string): (string, string, string)
 
   return partition(s, sep, right = true)
 
+
+since (1, 5):
+  type ParseFloatOptions* = enum  ## Options for `parseFloatThousandSep`.
+    pfLeadingDot,    ## Allow leading dot, like ".9" and similar.
+    pfTrailingDot,   ## Allow trailing dot, like "9." and similar.
+    pfSepAnywhere,   ## Allow separator anywhere in between, like "9,9", "9,99".
+    pfDotOptional    ## Allow "9", "-0", integers literals, etc.
```
timotheecour: integer literals

lib/pure/strmisc.nim
```diff
@@ -85,6 +86,116 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+    for idx, c in str:
+      if c in '0' .. '9':  # Digits
+        if hasAnySep and not afterDot and successive > 2:
+          parseFloatThousandSepRaise(idx, c, str)
+        else:
+          s.add c
+          lastWasSep = false
+          lastWasDot = false
+          inc successive
+      if c == sep:  # Thousands separator, this is NOT the dot
```
timotheecour: elif? (ditto elsewhere)

tests/stdlib/tstrmisc.nim
```diff
@@ -0,0 +1,51 @@
...
+  doAssert parseFloatThousandSep("1.", {pfLeadingDot,pfTrailingDot}) == 1.0
+  doAssert parseFloatThousandSep("1", {pfDotOptional}) == 1.0
+  doAssert parseFloatThousandSep("1.0,0,0", {pfSepAnywhere}) == 1.0
+  doAssert parseFloatThousandSep(".10", {pfLeadingDot}) == 0.1
+  doAssert parseFloatThousandSep("10.", {pfTrailingDot}) == 10.0
+  doAssert parseFloatThousandSep("10", {pfDotOptional, pfSepAnywhere}) == 10.0
+  doAssert parseFloatThousandSep("1.0,0,0,0,0,0,0,0", {pfSepAnywhere}) == 1.0
+  doAssert parseFloatThousandSep("0,0,0,0,0,0,0,0.1", {pfSepAnywhere}) == 0.1
+
+
```
timotheecour, line 57: `parseFloatThousandSep("1e02.2")` gives 102.2
incorrect, should give ValueError

Araq:
Why does this return a ``float``? Seems useful for money too so it should return a ``Rational[T]``. And maybe move the proc to ``rationals.nim``.

timotheecour:
> Seems useful for money too so it should return a Rational[T]. And maybe move the proc to rationals.nim.

I don't understand this, having `sep` is orthogonal to whether user would want a rational. And rationals would overfit for large mantissa.

> Why does this return a float? 

that I agree, but separate overloads make more sense (eg because `e12` mantissa, nan, inf don't make sense for types other than float, and because some options (eg `dotOptional`) don't make sense for parsing an int64, eg `1,000,000`

lib/pure/strmisc.nim
```diff
@@ -84,3 +85,154 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
   return partition(s, sep, right = true)
+
+
+since (1, 5):
+  type ParseFloatOptions* = enum  ## Options for `parseFloatThousandSep`.
+    pfLeadingDot,    ## Allow leading dot, like ".9" and similar.
+    pfTrailingDot,   ## Allow trailing dot, like "9." and similar.
+    pfSepAnywhere,   ## Allow separator anywhere in between, like "9,9", "9,99".
+    pfDotOptional    ## Allow "9", "-0", integer literals, etc.
+    pfScientific     ## Allow Scientific Notation "1.0e9", "1.0e-9", etc.
```
timotheecour: shouldn't pfScientific always be implicitly allowed?

tests/stdlib/tstrmisc.nim
```diff
@@ -0,0 +1,73 @@
...
+  doAssert parseFloatThousandSep("0,0,0,0,0,0,0,0.1", {pfSepAnywhere}) == 0.1
+  doAssert parseFloatThousandSep("1.0e9", {pfScientific}) == 1000000000.0
+  doAssert parseFloatThousandSep("1.0e-9", {pfScientific}) == 1e-09
+  doAssert parseFloatThousandSep("1,000.000e9", {pfScientific}) == 1000000000000.0
+  doAssert parseFloatThousandSep("1e9", {pfScientific, pfDotOptional}) == 1000000000.0
+  doAssert parseFloatThousandSep("1.0E9", {pfScientific}) == 1000000000.0
+  doAssert parseFloatThousandSep("1.0E-9", {pfScientific}) == 1e-09
+  doAssert parseFloatThousandSep("Inf", {pfNanInf}) == Inf
+  doAssert parseFloatThousandSep("-Inf", {pfNanInf}) == -Inf
+  doAssert parseFloatThousandSep("+Inf", {pfNanInf}) == +Inf
```
timotheecour, line 51:
```nim
nim> echo parseFloatThousandSep("1,000.000,000,E,+,9,0", {pfScientific,pfSepAnywhere})
1e+93
```

the last 4 `,` should be disallowed (including the 1st of those 4, `0,E`)

tests/stdlib/tstrmisc.nim
```diff
@@ -0,0 +1,75 @@
...
+  doAssert parseFloatThousandSep("-1.0") == -1.0
+  doAssert parseFloatThousandSep("1.000") == 1.0
+  doAssert parseFloatThousandSep("1.000") == 1.0
+  doAssert parseFloatThousandSep("-1.000") == -1.0
+  doAssert parseFloatThousandSep("-1,222.0001") == -1222.0001
+  doAssert parseFloatThousandSep("3.141592653589793") == 3.141592653589793
+  doAssert parseFloatThousandSep("6.283185307179586") == 6.283185307179586
+  doAssert parseFloatThousandSep("2.718281828459045") == 2.718281828459045
+
+  doAssertRaises(ValueError): discard parseFloatThousandSep("1,11")
```
timotheecour: can be done in followup PR:
```nim
template reject(a) =
  doAssertRaises(ValueError): discard parseFloatThousandSep(a)
reject "1,1"
...
```

tests/stdlib/tstrmisc.nim
```diff
@@ -0,0 +1,75 @@
...
+  doAssert parseFloatThousandSep("1e9", {pfDotOptional}) == 1000000000.0
+  doAssert parseFloatThousandSep("1.0E9") == 1000000000.0
+  doAssert parseFloatThousandSep("1.0E-9") == 1e-09
+  doAssert parseFloatThousandSep("Inf", {pfNanInf}) == Inf
+  doAssert parseFloatThousandSep("-Inf", {pfNanInf}) == -Inf
+  doAssert parseFloatThousandSep("+Inf", {pfNanInf}) == +Inf
+  doAssert parseFloatThousandSep("1000.000000E+90") == 1e93
+  doAssert parseFloatThousandSep("-10 000 000 000.0001", sep=' ') == -10000000000.0001
+  doAssert parseFloatThousandSep("-10 000 000 000,0001", sep=' ', decimalDot = ',') == -10000000000.0001
+  discard parseFloatThousandSep("NaN", {pfNanInf})
```
timotheecour: doAssert `classify(...) == fcNan`

tests/stdlib/tstrmisc.nim
```diff
@@ -0,0 +1,59 @@
...
+  doAssert parseFloatThousandSep("1.0e-9") == 1e-09
+  doAssert parseFloatThousandSep("1,000.000e9") == 1000000000000.0
+  doAssert parseFloatThousandSep("1e9", {pfDotOptional}) == 1000000000.0
+  doAssert parseFloatThousandSep("1.0E9") == 1000000000.0
+  doAssert parseFloatThousandSep("1.0E-9") == 1e-09
+  doAssert parseFloatThousandSep("Inf", {pfNanInf}) == Inf
+  doAssert parseFloatThousandSep("-Inf", {pfNanInf}) == -Inf
+  doAssert parseFloatThousandSep("+Inf", {pfNanInf}) == +Inf
+  doAssert parseFloatThousandSep("1000.000000E+90") == 1e93
+  doAssert parseFloatThousandSep("-10 000 000 000.0001", sep=' ') == -10000000000.0001
```
timotheecour, line 53: `parseFloatThousandSep("1e1")` raises but should be accepted even without  `{pfDotOptional}` because:
* a user may not want to set `{pfDotOptional}` (too loose, eg would allow integers to be parsed as float)
* nim: strutils.parseFloat accepts it
* D: accepts it too (`rdmd --eval 'writeln("1e1".to!double);'`)
* python3: accepts it (`float("1e1")`)

timotheecour:
@juancarlospaco I think https://github.com/nim-lang/Nim/pull/15421/files#r534562246 is my last comment, after that LGTM finally... unless i've missed some other case

EDIT: just found another bug: https://github.com/nim-lang/Nim/pull/15421/files#r534567618

please run sanity checks (or rather increase test coverage) before the next PTAL

tests/stdlib/tstrmisc.nim
```diff
@@ -0,0 +1,59 @@
...
+  doAssert parseFloatThousandSep("-10 000 000 000,0001", sep=' ', decimalDot = ',') == -10000000000.0001
+  doAssert classify(parseFloatThousandSep("NaN", {pfNanInf})) == fcNan
```
timotheecour, line 55: bug, this shouldn't be accepted:
```nim
nim> echo parseFloatThousandSep("inf.0", {pfNanInf})
0.0
```

tests/stdlib/tstrmisc.nim
```diff
@@ -0,0 +1,59 @@
+import strmisc, math
```
timotheecour, line 1: please move the content of `tests/stdlib/tstrmiscs.nim` into here (tstrmiscs.nim was badly named)

lib/pure/strmisc.nim
```diff
@@ -84,3 +85,134 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+      doAssert parseFloatThousandSep("01.00") == 1.0
+      doAssert parseFloatThousandSep("1,000.000e-9") == 1e-06
+
+    assert decimalDot notin {'-', '+', 'e', 'i', 'n', 'f', 'a', ' ', '\t', '\v', '\c', '\n', '\f'}
+    assert sep notin {'-', '+', 'e', 'i', 'n', 'f', 'a', '\n'}
+    assert sep != decimalDot
+
+    proc parseFloatThousandSepRaise(i: int; c: char; s: openArray[char]) {.noinline, noreturn.} =
+      raise newException(ValueError,
+        "Invalid float containing thousand separators, invalid char $1 at index $2 for input $3" %
```
timotheecour, line 141:
```nim
      raise newException(ValueError,
        "Invalid float containing thousand separators, invalid char $1 at index $2 for input '$3'" % [$c, $i, s.join])
```

so that it shows as 
```
'1,0000'
```
instead of
```
['1', ',', '0', '0', '0', '0'] 
```

lib/pure/strmisc.nim
```diff
@@ -84,3 +85,134 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+      parseFloatThousandSepRaise(0, decimalDot, str)
+    if str[^1] == sep:                                        # "1,"
+      parseFloatThousandSepRaise(strLen, sep, str)
+    if pfTrailingDot notin options and str[^1] == decimalDot: # "1."
+      parseFloatThousandSepRaise(strLen, decimalDot, str)
+    if pfSepAnywhere notin options and (str.len <= 4 and sep in str):
+      parseFloatThousandSepRaise(0, sep, str)                 # "1,1"
+
+    if (strLen == 3 or strLen == 4) and (
+      (str[0] in {'i', 'I'} and str[1] in {'n', 'N'} and str[2] in {'f', 'F'}) or
```
timotheecour, line 160: cna you simplify this logic? it's buggy for 2 reasons:
* https://github.com/nim-lang/Nim/pull/15421/files#r534567618 wrongly accepted
* inconsistently defers error handling to parseFloat:
```
nim> echo parseFloatThousandSep("infx", {pfNanInf})
Error: unhandled exception: invalid float: infx [ValueError]
nim> echo parseFloatThousandSep("infxx", {pfNanInf})
Error: unhandled exception: Invalid float containing thousand separators, invalid char , at index 0 for input 'infxx' [ValueError]
```

lib/pure/strmisc.nim
```diff
@@ -84,3 +85,134 @@ proc rpartition*(s: string, sep: string): (string, string, string)
...
+since (1, 5):
+  type ParseFloatOptions* = enum  ## Options for `parseFloatThousandSep`.
+    pfLeadingDot,    ## Allow leading dot, like ".9" and similar.
+    pfTrailingDot,   ## Allow trailing dot, like "9." and similar.
+    pfSepAnywhere,   ## Allow separator anywhere in between, like "9,9", "9,99".
+    pfDotOptional    ## Allow "9", "-0", integer literals, etc.
+    pfNanInf         ## Allow "NaN", "Inf", "-Inf", etc.
+
+  func parseFloatThousandSep*(str: openArray[char]; options: set[ParseFloatOptions] = {};
+      sep = ','; decimalDot = '.'): float =
```
timotheecour, line 99: can be discussed in future work, but just curious if there's a reverse proc for this?
I just found about `insertSep`, maybe they should cross-reference each other?
I'm not sure how robust `insertSep` is though; there's also `formatFloat` and `formatEng`, maybe they should be extended to support formatting with thousand separators

Araq:
Sorry, rejected. For multiple reasons: 

- By design it cannot handle money which shouldn't be stored as a "float".
- The ``openArray[char]`` interface is alien and not good enough for general parsing. (Hint: It doesn't return how many characters it returned so you're still left with a preprocessing step. Even if the string comes from an input field, you generally want to accept leading or trailing whitespace. So that would be yet another option ``pfAllowWhitespace``)
- The code is harder to use than the more hacky ``s.multiReplace({".": "", ",": ".").parseFloat`` (which would be quite appropriate for German btw).
- The amount of options you can pass to the proc indicates that nobody really knows the use cases or that "every" use case should be covered. And yet, "money" is not among these.

timotheecour:
/cc @Araq 
> For multiple reasons:

these are good arguments; how about the following instead:
```nim
proc parseIntCustom(a: string, start: int, T: typedesc[SomeInt], options: ParseOpt = {}): tuple[val: T, num: int] =
  ## val: parsed value; num: number of characters parsed on success or 0 on error
  runnableExamples:
    assert parseIntCustom("score: -1,234,567 name: foo", start = 5, int) == (-1,234,567, 10)
    assert parseIntCustom("--1", start = 0, int) == (0, 0) # because --
    assert parseIntCustom("-1", start = 0, uint) == (0, 0) # because uint
```

> By design it cannot handle money which shouldn't be stored as a "float".

=> an API for that can be built on top of `parseIntCustom` to parse the integral part of the amount

> The openArray[char] interface is alien and not good enough for general parsing

proposed API fixes that

> The code is harder to use than the more hacky

but the more hacky way doesn't help with input validation

> The amount of options you can pass to the proc indicates

I think `parseIntCustom` hits a sweet spot between complexity and usefulness; by restricting it to just integers (signed/unsigned), you avoid complexity of handling FP numbers; note that parsing FP numbers with thousand sep can use `parseIntCustom` as a building block, so can parsing of money.

Issue #15243, opened 2020-08-29
defer doesnt work with block, break and await


gabbhack:
### Example
```nim
import asyncdispatch

proc foo() {.async.} =
  echo "foo"

proc main() {.async.} =
  block:
    defer: await foo()
    break

waitFor main()
```

### Current Output
```
```

### Expected Output
```
foo
```

```
$ nim -v
Nim Compiler Version 1.2.6 [Windows: amd64]
Compiled at 2020-07-29
Copyright (c) 2006-2020 by Andreas Rumpf

git hash: bf320ed172f74f60fd274338e82bdc9ce3520dd9
active boot switches: -d:release
```

Araq:
Other example program:

```nim
import asyncdispatch

proc f() {.async.} =
  while true:
    try:
      await sleepAsync(400)
      break
    finally:
      echo "finally handler"
      inc j

when isMainModule:
  waitFor f()
```

Clyybber:
The best way to fix this IMO is to bind unnamed breaks in sem1 and make the async macro take a typed param.

Issue #15242, opened 2020-08-29
template with "return" cause a type mismatch error in async proc


gabbhack:
### Example
```nim
import asyncdispatch

template test: untyped =
  return true

proc main(): Future[bool] {.async.} =
  test

echo waitFor main()
```

### Current Output
```
Error: type mismatch: got <bool> but expected 'FutureBase = ref FutureBase:ObjectType'
```

### Expected Output
```
true
```

```
$ nim -v
Nim Compiler Version 1.2.6 [Windows: amd64]
Compiled at 2020-07-29
Copyright (c) 2006-2020 by Andreas Rumpf

git hash: bf320ed172f74f60fd274338e82bdc9ce3520dd9
active boot switches: -d:release
```

Araq:
The ``async`` macro cannot rewrite code resulting from template expansions. That's a Nim language restriction that we're solving by "better typed ASTs". But it's unclear if and when we'll update our ``async`` macro so that it is based on ``typed`` ASTs.

PR #15425, opened 2020-09-28
WIP Fixing #15243

yglukhov:
First part of fixing state split transformation.
- The algorithm is rethought from ground up
- The algorithm now has a small prepass to mark the nodes to split. This is mostly required to differentiate nkTryes between those we want to split and those we don't. E.g. nkTry normally doesn't need to be split if it contains no yields, but it must be split if it contains a break of a block that needs a split.
- I thought it would be nice to not support nkContinues and nkBreaks to whiles, and only support properly labeled breaks, but apparently unlabeled breaks are not always lowered before the transformation. There's a commented-out test in the middle of `tyieldintry.nim` that demonstrates just that. @Araq need your input here.

Still TODO:
- Always lower breaks/continues before my state split tranformation.
- Emit chains of duplicated finallies in place of breaks.

### DIFF
--- a/compiler/ast.nim
+++ b/compiler/ast.nim
@@ -487,6 +487,7 @@ type
     nfDefaultRefsParam # a default param value references another parameter
                        # the flag is applied to proc default values and to calls
     nfExecuteOnReload  # A top-level statement that will be executed during reloads
+    nfStateSplit # whether a statement needs to be split into states (used in closure iterator transformation)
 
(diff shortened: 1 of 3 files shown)


Clyybber:
I think this intends to fix #15243 not #15242, right?

yglukhov:
Indeed, corrected commit message

Araq:
Please get back to this, we need your better solution.

PR #15429, opened 2020-09-29
Add first draft of new osproc.readLines

c-blake:
The example in the doc comment also serves as motivation (and could be converted into a test with just an `import osproc` and a `discard` of output up top).  { My actual use-case is different, but less self-contained.  With so many idle cores these days there are many circumstances where launching programs in parallel makes sense. }

The `seq[string]` return makes this new `readLines(Process)` call side-step CRLF vs LF conversion questions since `lines`-like calls in Nim all remove the ending.  So there should be no caller expectation of non-conversion (unlike immediately subsequent `execProcess`).  An `echo` loop can put back the OS-local line-endings as desired.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -199,6 +199,8 @@
 
 - Add `initUri(isIpv6: bool)` to `uri` module, now `uri` supports parsing ipv6 hostname.
 
+- Add `readLines(p: Process)` to `osproc` module for `startProcess` convenience.
+
 ## Language changes
 
 - The `=destroy` hook no longer has to reset its target, as the compiler now automatically inserts
--- a/lib/pure/osproc.nim
+++ b/lib/pure/osproc.nim
@@ -444,6 +444,31 @@ proc execProcesses*(cmds: openArray[string],
       if afterRunEvent != nil: afterRunEvent(i, p)
       close(p)
 
+proc readLines*(p: Process): (seq[string], int) =
+  ## Convenience function for working with `startProcess` to read data from a
+  ## background process.
+  ##
+  ## Example:
+  ##
+  ## .. code-block:: Nim
+  ##  const opts = {poUsePath, poDaemon, poStdErrToStdOut}
+  ##  var ps: seq[Process]
+  ##  for prog in ["a", "b"]: # run 2 progs in parallel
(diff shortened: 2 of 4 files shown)


c-blake:
This seems good to go from my end.  Another application of this is with `execProcesses`, as in (fill in YADDA & BLAH as desired...):
```Nim
import osproc
const opts = {poUsePath, poStdErrToStdOut}
var nms, cmds: seq[string]
for e in YADDA:
  nms.add e
  cmds.add "BLAH " & e
let x = execProcesses(cmds, opts, n=24, afterRunEvent =
  proc(i: int, p: Process) =
    let (lines, exCode) = p.readLines ## <-- ANOTHER E.G. ##
    if lines.len > 0 or exCode != 0:
      echo nms[i], ":(exCode=", exCode, ")"
      for ln in lines: echo "  ", ln
    p.close)
quit(x)
```
While this can be implemented outside `osproc`, it applies to any situation where you want to read background process output all at once which is a very common case.  It makes sense to me for `osproc` to just provide it.

Doing it like this is also an easy way to keep output from multiple subprograms organized (rather than being all mixed together through a shared file handle).  As in this example, you can even indent the output relative to a label and conditionally output nothing (not even a "label:").

Varriount:
While this works for the case of a process that has a limited amount of output, it will fail for any process that runs forever. I believe it would be better if this were an iterator.

c-blake:
While I agree it would be bad on unbounded output, the same could be said of both `execProcess` & `execCmdEx` (as well as every backtick or `$()` command substitution ever done on Unix).  Many programs produce a deterministically small amount of output (such as one error message or not, etc.).  So, we could *also* provide a `lines` iterator, but both kinds of control flow make sense.  Do you think the iterator should return the exit code via a passed `var int` or should callers have to invoke `peekExitCode` or `waitForExit` themselves?

c-blake:
Well, I've addressed @Varriount's critique and all tests still pass.  Let me know what else may matter.

Araq:
Needs a ``.since`` annotation.

c-blake:
Ok.  I also added a `ReadIOEffect` tag on the iterator.

PR #15434, opened 2020-09-30
[docs minor] unify generates and Generates

ringabout:


lib/pure/htmlgen.nim
```diff
@@ -17,7 +17,7 @@
 ##
 ## This module implements a simple `XML`:idx: and `HTML`:idx: code
 ## generator. Each commonly used HTML tag has a corresponding macro
-## that generates a string with its HTML representation.
+## that Generates a string with its HTML representation.
```
Araq: Typo here.

PR #15439, opened 2020-09-30
Add 1 overload to apply

juancarlospaco:
- Add 1 overload to `sequtils.apply`
- https://play.nim-lang.org/#ix=2zgP :arrow_right:  https://play.nim-lang.org/#ix=2zgS
- Sometimes you just want to do stuff and echo it or write it to file, etc.
- Tiny diff.


alaviss:
I don't see how this is any better than just writing a for loop... they are not even that long.

Araq:
Acceptable because it's consistent with the other (terrible) apply procs and fills a hole.

PR #15409, opened 2020-09-24
koch, compiler: bundle fusion as part of the source archive [backport:1.4]

alaviss:
This allows distributions to build Nim from the downloaded source
archive without an Internet connection.

PR #15443, opened 2020-09-30
koch: unify nimble building scripts [backport:1.4]

alaviss:
This commit gets rid of buildNimble in favor of the simpler
bundleNimbleExe, which does pretty much the same thing.

This fixes nightlies.

### DIFF
--- a/koch.nim
+++ b/koch.nim
@@ -145,35 +145,11 @@ proc bundleC2nim(args: string) =
 
 proc bundleNimbleExe(latest: bool, args: string) =
   let commit = if latest: "HEAD" else: NimbleStableCommit
-  cloneDependency(distDir, "https://github.com/nim-lang/nimble.git", commit = commit)
+  cloneDependency(distDir, "https://github.com/nim-lang/nimble.git",
+                  commit = commit, allowBundled = true)
   # installer.ini expects it under $nim/bin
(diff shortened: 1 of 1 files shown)


koch.nim
```diff
@@ -145,35 +145,11 @@ proc bundleC2nim(args: string) =
...
-  cloneDependency(distDir, "https://github.com/nim-lang/nimble.git", commit = commit)
+  cloneDependency(distDir, "https://github.com/nim-lang/nimble.git",
+                  commit = commit, allowBundled = true)
   # installer.ini expects it under $nim/bin
   nimCompile("dist/nimble/src/nimble.nim",
-             options = "-d:release --nilseqs:on " & args)
-
-proc buildNimble(latest: bool, args: string) =
-  # if koch is used for a tar.xz, build the dist/nimble we shipped
-  # with the tarball:
```
Araq, line 155: Read this please:

> if koch is used for a tar.xz, build the dist/nimble we shipped
> with the tarball

With your patch applied, this is not done anymore!

alaviss: It's still being done? `bundleNimbleExe` is functionally equal to `buildNimble`, with less lines. Note that since #15409, `cloneDependency` can respect a bundled directory.

PR #15444, opened 2020-10-01
Add critbits.toCritBitTree

juancarlospaco:
- Add `critbits.toCritBitTree`, similar to `tables.toTable`.
- runnableExamples with doAssert, since, changelog, etc.

```nim
doAssert {"a": "0", "b": "1", "c": "2"}.toCritBitTree is CritBitTree[string]
```

### DIFF
--- a/lib/pure/collections/critbits.nim
+++ b/lib/pure/collections/critbits.nim
@@ -33,7 +33,7 @@ type
     root: Node[T]
     count: int
 
-proc len*[T](c: CritBitTree[T]): int =
+func len*[T](c: CritBitTree[T]): int {.inline.} =
   ## Returns the number of elements in `c` in O(1).
   runnableExamples:
     var c: CritBitTree[void]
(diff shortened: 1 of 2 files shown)


lib/pure/collections/critbits.nim
```diff
@@ -531,6 +531,18 @@ proc commonPrefixLen*[T](c: CritBitTree[T]): int {.inline, since((1, 3)).} =
     else: c.root.byte
   else: 0
 
+proc toCritBitTree*[A, B](pairs: openArray[(A, B)]): CritBitTree[A] {.inline, since: (1, 3).} =
```
Araq: I think this can be a ``func`` too. Also, no ``.inline`` for this.

lib/pure/collections/critbits.nim
```diff
@@ -531,6 +531,18 @@ proc commonPrefixLen*[T](c: CritBitTree[T]): int {.inline, since((1, 3)).} =
...
+  ## Creates a new `CritBitTree` that contains the given `pairs`.
+  runnableExamples:
+    doAssert {"a": "0", "b": "1", "c": "2"}.toCritBitTree is CritBitTree[string]
+  for item in pairs: result.incl item[0], item[1]
+
+proc toCritBitTree*[T](items: openArray[T]): CritBitTree[void] {.inline, since: (1, 3).} =
```
Araq: I think this can be a ``func`` too. Also, no ``.inline`` for this.

Issue #14405, opened 2020-05-20
`dup` documentation in release note is better than in sugar


mratsim:
The `dup` documentation in sugar is not detailed enough:

![image](https://user-images.githubusercontent.com/22738317/82477911-682ee400-9ad0-11ea-923a-8f01419865e4.png)

Compared to the release notes:

![image](https://user-images.githubusercontent.com/22738317/82477955-767d0000-9ad0-11ea-9882-92fea8c8c9eb.png)

it's missing the part of the underscore placeholder

Furthermore, the block syntax with one transformation per line is quite appealing visually and missing from the documentation as well.

luismanfroni:
Doing this task, could i be assigned to it?

luismanfroni:
Was merged in PR: https://github.com/nim-lang/Nim/pull/15455
Review if can be closed

PR #15455, opened 2020-10-02
doc(sugar): added description and examples to dup

luismanfroni:
## Solving issue #14405

Added more description and examples to sugar/dup documentation.

#### Before:
image

#### After:
image

#### Before:
image

#### After
image

### DIFF
--- a/lib/pure/sugar.nim
+++ b/lib/pure/sugar.nim
@@ -217,7 +217,10 @@ since (1, 1):
 
   macro dup*[T](arg: T, calls: varargs[untyped]): T =
     ## Turns an `in-place`:idx: algorithm into one that works on
-    ## a copy and returns this copy.
+    ## a copy and returns this copy, without modifying its input.
+    ##
+    ## This macro also allows for (otherwise in-place) function chaining.
+    ##
     ## **Since**: Version 1.2.
     runnableExamples:
       import algorithm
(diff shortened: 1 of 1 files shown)


ringabout:
Hi @luismanfroni
Could you close and reopen this pr to rerun CI?

Araq:
Expanding the explanation for "dup" is nice but I think even more examples overdo it.

lib/pure/sugar.nim
```diff
@@ -234,6 +242,21 @@ since (1, 1):
...
+      var c = "xyz"
+
+      # An underscore (_) can be used to denote the place of the argument you're passing:
+      # b = "xyz"
+      var d = dup c:
+        makePalindrome # xyzyx
+        sort(_, SortOrder.Descending) # zyyxx
+        makePalindrome # zyyxxxyyz
+
+      doAssert e == "zyyxxxyyz"
```
Varriount: This appears to be using an undeclared variable

lib/pure/sugar.nim
```diff
@@ -217,7 +217,15 @@ since (1, 1):
 
   macro dup*[T](arg: T, calls: varargs[untyped]): T =
     ## Turns an `in-place`:idx: algorithm into one that works on
-    ## a copy and returns this copy.
+    ## a copy and returns this copy, without modifying its input.
+    ##
+    ## The idea is that future versions of Nim will offer only
```
Varriount: I'm not sure how appropriate it is to mention standard library motives in function documentation, especially when that motive hasn't been codified anywhere dedicated and explicit.

Varriount (changes requested):

lib/pure/sugar.nim
luismanfroni: My bad, did push without testing locally.

lib/pure/sugar.nim
luismanfroni: Ok, removed the second part from the description. Thanks for the feedback!

luismanfroni:
> Expanding the explanation for "dup" is nice but I think even more examples overdo it.

Changed doc to only have the underscore example.

PR #15456, opened 2020-10-02
Add compile-time unrolled for loops

juancarlospaco:
- Add compile-time unrolled for loops.
- Documentation, examples, links, runnableExamples, changelog, etc.
- Tiny diff.

### DIFF
--- a/lib/system/iterators_1.nim
+++ b/lib/system/iterators_1.nim
@@ -1,3 +1,5 @@
+import std/private/since
+
 when sizeof(int) <= 2:
   type IntLikeForCount = int|int8|int16|char|bool|uint8|enum
 else:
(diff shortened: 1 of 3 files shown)


lib/system/iterators_1.nim
```diff
@@ -178,7 +180,7 @@ else: # not defined(nimNewRoof)
       inc i
 
 
-iterator `||`*[S, T](a: S, b: T, annotation: static string = "parallel for"): T {.
+iterator `||`*[S, T](a: S, b: T, annotation: static string = "omp parallel for"): T {.
```
alaviss, line 181: If you need a new iterator type, make one, don't abuse the parallel for iterator.

Removing the `omp` as a prefix of the iterator annotation is a breaking change for anyone using this iterator with a custom annotation.

alaviss (changes requested):

lib/system/iterators_1.nim
juancarlospaco: I dont understand what you mean, the `||` keeps working.

alaviss: Now anyone that use `||` with a custom annotation must prefix their annotation with `omp`, so this is a breaking change.

@mratsim's laser is [a big user of this iterator](https://github.com/numforge/laser/blob/master/laser/openmp.nim), and this change will cause it not to compile.

alaviss: My advise would be to create a brand new loop type. And note that `#pragma unroll` can unroll any loop, so a solution that allows tagging a generated loop would be much more efficient. Basically implementing the `{.unroll.}` pragma, if you may.

PR #15457, opened 2020-10-02
use func in httpcore

ringabout:


ghost:
can you please also verify that it works with {.experimental: "strictFuncs".} ?

ringabout:
How can I verify that? I use Nim devel@9.26
```nim
func add*(headers: HttpHeaders, key, value: string) =
  ## Adds the specified value to the specified key. Appends to any existing
  ## values associated with the key.
  if not headers.table.hasKey(headers.toCaseInsensitive(key)):
    headers.table[headers.toCaseInsensitive(key)] = @[value]
  else:
    headers.table[headers.toCaseInsensitive(key)].add(value)
```
But when I change this to `func`, `{.experimental: "strictFuncs".}` can't catch this?

ringabout:
I only use `testament r --nim:bin/nim_temp thttpcore.nim`  with `{.experimental: "strictFuncs".}`

PR #15460, opened 2020-10-02
intsets.nim: Add `toIntSet` proc

ee7:
Similar to:
- `critbits.toCritBitTree`
- `deques.toDeque`
- `sets.toHashSet`
- `tables.toTable`

### DIFF
--- a/lib/pure/collections/intsets.nim
+++ b/lib/pure/collections/intsets.nim
@@ -18,9 +18,8 @@
 ## **See also:**
 ## * `sets module <sets.html>`_ for more general hash sets
 
-
-import
-  hashes
+import std/private/since
+import hashes
 
 type
   BitScalar = uint
@@ -161,6 +160,9 @@ iterator items*(s: IntSet): int {.inline.} =
 
 proc initIntSet*: IntSet =
   ## Returns an empty IntSet.
+  ##
+  ## See also:
+  ## * `toIntSet proc <#toIntSet,openArray[int]>`_
   runnableExamples:
     var a = initIntSet()
     assert len(a) == 0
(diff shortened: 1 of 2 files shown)


lib/pure/collections/intsets.nim
```diff
@@ -251,6 +253,24 @@ proc incl*(s: var IntSet, other: IntSet) =
 
   for item in other: incl(s, item)
 
+proc toIntSet*(x: openArray[int]): IntSet {.since: (1, 3).} =
```
ee7, line 256:
```suggestion
proc toIntSet*(x: openArray[SomeSignedInt]): IntSet {.since: (1, 3).} =
```

Is `SomeSignedInt` preferred here?

Then users on 64-bit systems could write e.g. `[7'i32, 8, 9].toIntSet`, which might be reasonable to support given that they can currently write e.g. `myintset.incl(7'i32)` (and similar for `contains`, `containsOrIncl`, `excl` and `missingOrExcl`).

For such users, note that e.g. `myintset.incl(7'i64)` doesn't compile, so `[7'i64, 8, 9].toIntSet` wouldn't work either way.

lib/pure/collections/intsets.nim
```diff
@@ -161,6 +160,9 @@ iterator items*(s: IntSet): int {.inline.} =
 
 proc initIntSet*: IntSet =
   ## Returns an empty IntSet.
+  ##
+  ## See also:
+  ## * `toIntSet proc <#toIntSet,openArray[int]>`_
```
ee7, line 165:
```suggestion
  ## * `toIntSet proc <#toIntSet,openArray[SomeSignedInt]>`_
```

Araq:
I like the ``int`` version better.

Issue #7789, opened 2018-05-06
Ability to check for effects at compile-time


dom96:
```nim
proc foo() {.tags: [MyTag].} = ...

proc bar() = foo()

when MyTag in bar.tags:
  echo("Bar has MyTag!")
```

same for ``raises``. That would make the effect system much more useful.

twetzel59:
Is ``tags`` a magical field?

arkanoid87:
this should be closed as effecttraits seems to cover the required checks

https://play.nim-lang.org/#ix=3ud2

ringabout:
See also https://github.com/nim-lang/Nim/pull/15503

PR #15462, opened 2020-10-02
Added std/effecttraits.nim

Araq:
Note: This has had no RFC process but I hope it's tiny enough to be harmless.


metagn:
There is an RFC that would be closed after this: https://github.com/nim-lang/RFCs/issues/91 but only if a `tags` list is added too, same for #7789

juancarlospaco:
Needs `.since`.

Araq:
A completely new module doesn't need ``.since``.

PR #15467, opened 2020-10-02
fix adding empty sequence to HTTP headers causes some problems

ringabout:
Allowing adding empty sequence causes some problems when visiting it. 
```nim
import httpcore

var headers = newHttpHeaders()
headers["empty"] = @[]


if headers.hasKey("empty"):
  # Error: unhandled exception: index out of bounds, the container is empty [IndexError]
  echo headers["empty"]
```

### DIFF
--- a/compiler/astalgo.nim
+++ b/compiler/astalgo.nim
@@ -591,9 +591,16 @@ proc value(this: var DebugPrinter; value: PNode) =
   if this.conf != nil:
     this.key "info"
     this.value $lineInfoToStr(this.conf, value.info)
-  if card(value.flags) > 0:
+  if value.flags != {}:
     this.key "flags"
     this.value value.flags
+
+  if value.typ != nil:
+    this.key "typ"
+    this.value value.typ.kind
+  else:
+    this.key "typ"
+    this.value "nil"
 
   case value.kind
   of nkCharLit..nkUInt64Lit:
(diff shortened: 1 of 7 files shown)


lib/pure/httpcore.nim
```diff
@@ -166,9 +166,10 @@ proc `[]=`*(headers: HttpHeaders, key, value: string) =
 
 proc `[]=`*(headers: HttpHeaders, key: string, value: seq[string]) =
   ## Sets the header entries associated with ``key`` to the specified list of
-  ## values.
-  ## Replaces any existing values.
-  headers.table[headers.toCaseInsensitive(key)] = value
+  ## values. Replaces any existing values.
+  ## If ``value`` is an empty sequence, the key will be ignored.
+  if value.len > 0:
+    headers.table[headers.toCaseInsensitive(key)] = value
```
dom96: IMO you should remove the key instead.

dom96 (changes requested):

lib/pure/httpcore.nim
ringabout: yea, I agree.

juancarlospaco: @xflywind 
Check that some fields can be empty according to the standard, content length can be empty I think, is optional, I dont remember if theres others too.

ringabout: yea, it can be empty, but should use `headers.add(@[""])` instead.

juancarlospaco: :+1:

PR #15468, opened 2020-10-02
parser hotfix: don't run into endless loops; regression

Araq:


Araq:
@Clyybber See, this is why we don't touch the parser... ;-)

PR #15477, opened 2020-10-03
Provide a configurable progressInterval for asyncftpclient.

supakeen:
Make the progressInterval used to fire callbacks in the asyncftpclient configurable.

PR #15483, opened 2020-10-03
asyncftpclient onProgressHandler always start/end.

supakeen:
Currently the ``onProgressHandler`` in asyncftpclient is only called at a timed interval, this PR always calls onProgressHandler at the start and end of a transfer while other calls are the same as before. This allows for easier progress tracking of transfers (say if using this to render a progress bar).

Note that this PR might need to be rebased on https://github.com/nim-lang/Nim/pull/15477 if merged.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,215 @@
 
 ## Standard library additions and changes
 
+- Added some enhancements to `std/jsonutils` module.
+  * Added a possibility to deserialize JSON arrays directly to `HashSet` and
+    `OrderedSet` types and respectively to serialize those types to JSON arrays
+    via `jsonutils.fromJson` and `jsonutils.toJson` procedures.
+  * Added a possibility to deserialize JSON `null` objects to Nim option objects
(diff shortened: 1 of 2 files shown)


dom96 (commented):
FWIW you know when the progress starts and ends (when you call getFile and when getFile completes) so is this really worth the extra complexity?

supakeen:
While it is certainly possible to know when the progress starts and ends I feel it is nice to have progress events always, not just when a configurable timeout passes (it might not pass during the transfer window at all).

I'm pretty neutral about this but during some work on the other PRs this was nice to whip up both a quick test script and it seems to be it might be useful for others who want to drive something (progress bar, etc) fully off events.

juancarlospaco:
Have you tried any benchmark on it?, maybe it was for performance reasons, 
on fast connections reporting the status too much can cause slowdowns.

supakeen:
I could do a benchmark on it, something with a lot of small files would hit this path the most, I'll do that and get back with results. I've also rebased this branch since the ``progressInterval`` branch was merged!

PR #15486, opened 2020-10-04
use func in uri module

ringabout:


dom96 (commented):
Hmm, do these need to be called for the Nim compiler to test whether they have side effects? If so you will need to test each of them by calling them.

ringabout:
Yea, there are `when isMainModule` tests in this module, I remove all the export symbols and only one function `initUri` is not covered. I call it by hand.

Run the test, we can get the successful msgs: "All good!"

Araq:
Whenever we switch from ``proc`` to ``func`` we should also test with ``--experimental:strictFuncs``

ringabout:
Is `{.experimental:strictFuncs.}` at the top of `uri` for testing module enough?
I will add test option in testsament.

Araq:
Instead patch ``tests/effects/tstrict_funcs``, add ``uri`` to the ``import`` statement.

PR #15497, opened 2020-10-05
Iterate over smaller set when computing intersection

### LINKED ISSUE #15496: Argument order affects set intersection performance
I'm working on some scientific that involves performing a large number of set intersections between very large (~500,000 elements) and very small (~10 elements) sets, though the sizes are variable within an order of magnitude or so. 

From looking at the source code and from testing, which set is s1 and which is s2 can have a dramatic impact on performance:

https://github.com/nim-lang/Nim/blob/4e438f9096fc413824bdbe8377a4ca376d63be35/lib/pure/collections/sets.nim#L425-L426

After experimenting with my code in which the only change is the argument order, here is the runtime performance:

`bigSet.intersection(littleSet)` => 11,400 seconds
`littleSet.intersection(bigSet)` => 127 seconds

That's a **90x** performance difference. I propose fixing this with a check to ensure that the set being iterated over is the smaller of the two.

Benjamin-Lee:
Closes #15496

### DIFF
--- a/lib/pure/collections/sets.nim
+++ b/lib/pure/collections/sets.nim
@@ -422,8 +422,15 @@ proc intersection*[A](s1, s2: HashSet[A]): HashSet[A] =
     assert c == toHashSet(["b"])
 
   result = initHashSet[A](max(min(s1.data.len, s2.data.len), 2))
-  for item in s1:
-    if item in s2: incl(result, item)
+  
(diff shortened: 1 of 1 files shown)


Benjamin-Lee:
Would some tests for set intersections be a worthwhile addition?

juancarlospaco:
Tests and Documentation are always welcome.

PR #15498, opened 2020-10-05
inline tiny proc

juancarlospaco:
- `{.inline.}` tiny `proc` on `httpclient`.
- Continuation of https://github.com/nim-lang/Nim/pull/15480#issue-497289333

### DIFF
--- a/lib/pure/httpclient.nim
+++ b/lib/pure/httpclient.nim
@@ -232,7 +232,7 @@ proc code*(response: Response | AsyncResponse): HttpCode
   ## corresponding ``HttpCode``.
   return response.status[0 .. 2].parseInt.HttpCode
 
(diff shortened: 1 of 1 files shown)


dom96 (commented):
Is doing this really worthwhile? The C compiler will likely auto-inline this for us so this just seems like a bunch of noise.

juancarlospaco:
I am reading the diff of the generated C and I can confirm `{.inline.}` works Ok. :shrug:

ghost:
> I am reading the diff of the generated C and I can confirm `{.inline.}` works Ok. :shrug: 
> 

C code is not what you should read, you should read assembly of the binary :)

Araq:
Accepting for now but eventually we should enable LTO with -d:release and call it a day, time is money.

PR #15513, opened 2020-10-07
Update parsecfg.nim

lihf8515:
Returns the specified default value if the specified key value does not exist.

### DIFF
--- a/lib/pure/parsecfg.nim
+++ b/lib/pure/parsecfg.nim
@@ -562,7 +562,7 @@ proc writeConfig*(dict: Config, filename: string) =
 
 proc getSectionValue*(dict: Config, section, key: string, defaultVal = ""): string =
   ## Gets the Key value of the specified Section.
-  ## Returns the specified default value if the specified key value does not exist.
+  ## Returns the specified default value if the specified key does not exist.
   if dict.hasKey(section):
     if dict[section].hasKey(key):
       result = dict[section][key]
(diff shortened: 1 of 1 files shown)


lib/pure/parsecfg.nim
```diff
@@ -560,15 +560,16 @@ proc writeConfig*(dict: Config, filename: string) =
   let fileStream = newFileStream(file)
   dict.writeConfig(fileStream)
 
-proc getSectionValue*(dict: Config, section, key: string): string =
-  ## Gets the Key value of the specified Section, returns an empty string if the key does not exist.
+proc getSectionValue*(dict: Config, section, key: string, defaultVal: string = ""): string =
```
ringabout:
```suggestion
proc getSectionValue*(dict: Config, section, key: string, defaultVal = ""): string =
```

lib/pure/parsecfg.nim
```diff
@@ -560,15 +560,16 @@ proc writeConfig*(dict: Config, filename: string) =
...
+  ## Gets the Key value of the specified Section.
+  ## Returns the specified default value if the specified key value does not exist.
```
ringabout:
```suggestion
  ## Returns the specified default value if the specified key does not exist.
```

lib/pure/parsecfg.nim
```diff
@@ -560,15 +560,16 @@ proc writeConfig*(dict: Config, filename: string) =
   let fileStream = newFileStream(file)
   dict.writeConfig(fileStream)
 
-proc getSectionValue*(dict: Config, section, key: string): string =
-  ## Gets the Key value of the specified Section, returns an empty string if the key does not exist.
+proc getSectionValue*(dict: Config, section, key: string, defaultVal: string = ""): string =
+  ## Gets the Key value of the specified Section.
```
ringabout:
```suggestion
  ## Gets the key value of the specified Section.
```

PR #15518, opened 2020-10-07
fix #15516

### LINKED ISSUE #15516: Sizeof of case object is incorrect. Showstopper
Test case for 64bit arch:
```nim
type
  Kind = enum
    K1, K2
  
  AnotherEnum = enum
    X1, X2, X3

  MyObject = object
    s: string
    case k: Kind
    of K1: nil
    of K2:
        x: float
        y: int32
    z: AnotherEnum

import macros

macro c_sizeof(a: typed): int32 =
  result = quote do:
    var res: int32
    {.emit: [res, " = sizeof(", `a`, ");"] .}
    res

proc test =
  echo sizeof(MyObject)
  echo c_sizeof(MyObject)

test()
```
Nim size is 32 bytes, while C compiler size is 40 bytes.

It took me a week to trace it down. Somehow crashes were happening only in `--gc:arc` mode where Nim size is 40 bytes while C size is 48 bytes.

I believe nim doesn't take into account that elements `x` and `y` are wrapped into struct that needs its own 4 byte padding at its end.

cooldome:
fix sizeof bug #15516

### DIFF
--- a/compiler/sizealignoffsetimpl.nim
+++ b/compiler/sizealignoffsetimpl.nim
@@ -138,12 +138,13 @@ proc computeObjectOffsetsFoldFunction(conf: ConfigRef; n: PNode, packed: bool, a
       accum.offset  = szUnknownSize
       accum.maxAlign = szUnknownSize
     else:
-      # the union neds to be aligned first, before the offsets can be assigned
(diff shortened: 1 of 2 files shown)


cooldome:
test failures are unrelated, one the packages has changed its tests structure

PR #15519, opened 2020-10-08
Fix #12027

### LINKED ISSUE #12027: Csources: huge size increase (x2.3) in 0.20
Here are the sizes of the csources .zip files from the past 2 years

| Version | Commit  | Date       | Size      |
|---------|---------|------------|-----------|
| 0.17.0  | 9f7c507 | 2017-05-18 | 51.8 MiB  |
| 0.17.2  | 08c310b | 2017-09-07 | 64.8 MiB  |
| 0.18.0  | da97f85 | 2018-03-01 | 69.9 MiB  |
| 0.19.0  | b56e49b | 2018-09-26 | 80.8 MiB  |
| 0.20.0  | f72f471 | 2019-06-05 | **186.4 MiB** |

0.20.0 increased **compressed** csources size by 2.3x compared to before.

This is compounded when cloning the git repository (which might require some git garbage collection).

#### Potential rootcause

This might be related to the split of `-d:release` into `-d:release` and `-d:danger`. That would introduce lots of extra checks in the generated code if the sources are only built with `-d:release`. See also https://github.com/nim-lang/Nim/issues/11860.

genotrance:
This is not really required since the nightlies [build](https://github.com/nim-lang/nightlies/blob/master/.github/workflows/nightlies.yml#L210) csources with -d:danger so this is just for completeness.

Note that nightlies no longer bundle the csources directory in the tarball - not sure if that is intentional.

cc @alaviss

### DIFF
--- a/koch.nim
+++ b/koch.nim
@@ -78,7 +78,7 @@ Boot options:
 Commands for core developers:
   runCI                    runs continuous integration (CI), eg from travis
   docs [options]           generates the full documentation
-  csource -d:release       builds the C sources for installation
+  csource -d:danger        builds the C sources for installation
   pdf                      builds the PDF documentation
   zip                      builds the installation zip package
   xz                       builds the installation tar.xz package
@@ -416,7 +416,7 @@ proc winRelease*() =
                   "web/upload/download/docs-$1.zip" % VersionAsString
   when true:
     inFold "winrelease csource":
-      csource("-d:release")
+      csource("-d:danger")
   when sizeof(pointer) == 4:
     winReleaseArch "32"
   when sizeof(pointer) == 8:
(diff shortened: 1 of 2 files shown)


alaviss:
csources is still bundled, but I don't know about this `csources` directory. If you mean that binaries don't have csources bundled, then this is intentional as they're prebuilts, it makes little sense to bundle 1GB worth of C code that you're not going to use. 

Nightlies also produces source-only tarballs, which has csources bundled inside.

genotrance:
Sounds fair - as long as it is intentional.

genotrance:
Failures look unrelated.

Araq:
I think it's more important to ensure people have a ``release`` compiler and not a ``danger``'ous compiler.

genotrance:
This is just for generating csources - not the bootstrap.

PR #15525, opened 2020-10-08
Fix --gc:arc in nimscript

### LINKED ISSUE #14813: Error compiling with gc:orc and a config.nims file: "system module needs: nimDestroyAndDispose"
The compiler immediately fails when compiling with `--gc:orc` if there is a `config.nims` file present (even if that file is empty!) It outputs:
```
...nim-#devel/lib/system/assertions.nim(23, 11) template/generic instantiation of `sysFatal` from here
...nim-#devel/lib/system/fatal.nim(49, 5) Error: system module needs: nimDestroyAndDispose
```
This happens in 1.2.2 and in the current devel build.

### Example

(Using dev channel, checked out & built this morning with `choosenim devel`, on macOS 10.15.5)

```bash
$  tree
.
├── config.nims
├── foo.nim
└── nim.cfg

0 directories, 3 files
$  cat config.nims
$  cat foo.nim
echo "foo!"
$  cat nim.cfg
--gc:orc
$  nim c foo.nim
Hint: used config file '/Users/snej/.choosenim/toolchains/nim-#devel/config/nim.cfg' [Conf]
Hint: used config file '/Users/snej/.choosenim/toolchains/nim-#devel/config/config.nims' [Conf]
Hint: used config file '/private/tmp/oops/nim.cfg' [Conf]
Hint: used config file '/private/tmp/oops/config.nims' [Conf]
.
/Users/snej/.choosenim/toolchains/nim-#devel/lib/system/assertions.nim(23, 11) template/generic instantiation of `sysFatal` from here
/Users/snej/.choosenim/toolchains/nim-#devel/lib/system/fatal.nim(49, 5) Error: system module needs: nimDestroyAndDispose
```

### Current Output

See error at end of above transcript.

### Expected Output

Should compile without errors.

### Additional Information

* Removing the [empty] config.nims file makes the problem go away.
* Removing the line `--gc:orc` from nim.cfg also makes the problem go away.

```
$ nim -v
Compiled at 2020-06-25

```

cooldome:
I have a weird issue that adding `switch("gc", "arc")` to nim script doesn't work, in command line it does work.
However I can't replicate this problem in the small test case for unknown reason.

This PR fixes the issue. Nimscipt undefs the "nimV2" and that causes the problem.

P.S.
I am now moved to `--gc:arc` with `--sinkInference:off --cursorInference:off --deepcopy:on` but it works.

### DIFF
--- a/compiler/scriptconfig.nim
+++ b/compiler/scriptconfig.nim
@@ -228,11 +228,14 @@ proc runNimScript*(cache: IdentCache; scriptName: AbsoluteFile;
 
   # watch out, "newruntime" can be set within NimScript itself and then we need
   # to remember this:
+  if conf.selectedGC == gcUnselected:
+    conf.selectedGC = oldSelectedGC
   if optOwnedRefs in oldGlobalOptions:
     conf.globalOptions.incl {optTinyRtti, optOwnedRefs, optSeqDestructors}
     defineSymbol(conf.symbols, "nimv2")
-  if conf.selectedGC == gcUnselected:
-    conf.selectedGC = oldSelectedGC
+  if conf.selectedGC in {gcArc, gcOrc}:
+    conf.globalOptions.incl {optTinyRtti, optSeqDestructors}
+    defineSymbol(conf.symbols, "nimv2")
 
   # ensure we load 'system.nim' again for the real non-config stuff!
   resetSystemArtifacts(graph)


ghost:
Thanks :) This PR closes https://github.com/nim-lang/Nim/issues/14813 and now I can put "gc = orc" in my config globally for easier testing of packages with ORC :)

cooldome:
@Yardanico 
 nice


fixes #14813 for github

ghost:
@cooldome there's one more problem left (don't know if it's a bug):
If you have "gc = orc" or "gc = arc" in your global config but then use "nim c --gc:refc file.nim" it breaks with
```
Error: system module needs: nimGCvisit
```

PR #15531, opened 2020-10-09
update parsecfg

lihf8515:
A simplified version of the name for the `getSectionValue` , `setSectionValue` and `delSection` function.


Araq:
Sorry but you can always introduce shortcuts in your own code, but having them in the stdlib means everybody needs to learn about them, a bad trade-off.

PR #15555, opened 2020-10-12
suggest: try to find the implementation of a symbol when def is used

### LINKED ISSUE #79: `def` should go to definition, not forward declaration
For scenarios when a signature of a proc is defined without implementation and additionally the implemented proc is defined too, and the order of the code is like the following
```
proc test(i: var int)

proc test2(i: var int) =
  i.test() # <-- click here on goto goes to line 1; more interesting would be line 6

proc test(i: var int) =
  i.inc(i)
```

`nimsuggest` will go to the signature definiton of the unimplemented proc although the implementation is defined below.
Especially for a bigger source code it's a bit annoying to manually search for the place where it's implemented.

Araq mentioned this can be changed in `semProcAux`.
(https://github.com/nim-lang/Nim/blob/1d1ba4481fb72350f0ee9ea17fa23a04c425629a/compiler/semstmts.nim#L1437)

See also:
- https://github.com/nim-lang/Nim/issues/6867
- https://github.com/nim-lang/Nim/issues/7237

alaviss:
Fixes nim-lang/nimsuggest#79.

Known issues:
~~- The forward declaration of the symbol is now inaccessible.~~

### DIFF
--- a/compiler/suggest.nim
+++ b/compiler/suggest.nim
@@ -119,7 +119,8 @@ proc getTokenLenFromSource(conf: ConfigRef; ident: string; info: TLineInfo): int
 
 proc symToSuggest(conf: ConfigRef; s: PSym, isLocal: bool, section: IdeCmd, info: TLineInfo;
                   quality: range[0..100]; prefix: PrefixMatch;
-                  inTypeContext: bool; scope: int): Suggest =
+                  inTypeContext: bool; scope: int;
+                  useSuppliedInfo = false): Suggest =
   new(result)
   result.section = section
   result.quality = quality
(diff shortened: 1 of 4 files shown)


compiler/suggest.nim
```diff
@@ -463,11 +468,14 @@ when defined(nimsuggest):
...
   if s.isNil: return
-  if isTracked(info, conf.m.trackPos, s.name.s.len):
-    suggestResult(conf, symToSuggest(conf, s, isLocal=false, ideDef, info, 100, PrefixMatch.None, false, 0))
-    suggestQuit()
+  if isTracked(info, conf.m.trackPos, s.name.s.len) or s == usageSym:
+    if sfForward notin s.flags:
+      suggestResult(conf, symToSuggest(conf, s, isLocal=false, ideDef, info, 100, PrefixMatch.None, false, 0, useSuppliedInfo = s == usageSym))
+      suggestQuit()
+    else:
+      usageSym = s
```
Clyybber: I think `if sfForward notin s.flags: suggestQuit()` should give us the forward decl too.

alaviss: My concerns are with 3rd parties editor plugins. I probably can just issue a changelog for this and everyone should be happy though.

Clyybber (approved):

RSDuck:
is it intentional that both declarations are only returned when the location from which the definition is asked from is before the definition?

```nim
proc test(): int

proc test(): int = discard

discard test() # def from here only returns the body
```

alaviss:
@RSDuck

Yea, it's a small quirk in the way nimsuggest perform its analysis. To make it work we would need to rethink how nimsuggest functions, and I'm not up for that atm :P

PR #15557, opened 2020-10-12
compiler/ast: make copyTree and derivatives iterative

alaviss:
When building nimsuggest in debug mode, certain projects like niminst may cause
the compiler to crash due to the debug call depth limit.

This commit modifies copyTree() and friends into interative procedures to
avoid deep recursion.

### DIFF
--- a/compiler/ast.nim
+++ b/compiler/ast.nim
@@ -1602,22 +1602,22 @@ proc transitionToLet*(s: PSym) =
   s.alignment = obj.alignment
 
 template copyNodeImpl(dst, src, processSonsStmt) =
-  if src == nil: return
-  dst = newNode(src.kind)
-  dst.info = src.info
-  dst.typ = src.typ
-  dst.flags = src.flags * PersistentNodeFlags
-  dst.comment = src.comment
-  when defined(useNodeIds):
-    if dst.id == nodeIdToDebug:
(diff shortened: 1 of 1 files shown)


compiler/ast.nim
```diff
@@ -1626,17 +1626,24 @@ proc shallowCopy*(src: PNode): PNode =
...
-    newSeq(result.sons, src.len)
-    for i in 0..<src.len:
-      result[i] = copyTree(src[i])
+  var stack = @[(addr result, src)]
+  while stack.len > 0:
+    let (target, source) = stack.pop()
+    copyNodeImpl(target[], source):
+      newSeq(target.sons, source.len)
+      for i in 0..<source.len:
+        stack.add((addr target.sons[i], source[i]))
```
alaviss, line 1635: We will probably need a way to add capacity into the seq or this `.add` will slow everything down :(

Clyybber:
This is a lost cause IMO, most parts of the compiler are recursive and it isn't a problem in practice.
You can just set `-d:nimCallDepthLimit` to something higher for debugging.

(if we were to ever hit the real stack limit we could use cps to lift our variables to the heap and use trampolines)

Araq:
Sorry, but I think this is worse than rare stack problems.

PR #8024, opened 2018-06-12
All modules should be importable without the std/ prefix.

dom96:
Refs #7938

PR #15564, opened 2020-10-13
Make IntSet a generic ordinal set PackedSet[A] + new module packedsets

landerlo:
This makes intsets generic for ordinal types like enums or distinct of ints. For cases when multiple intsets with complex logic making them generic boosts their usability and type safety.
E.g. iqqn my use case I need to distinguish different ids making them distinct and get compilation errors if
I do any set operations with incompatible ids.

The PR achieves this maintaining full backwards compatibility and same performance.
As a drop in replacement, the current api is unchanged and IntSet is aliased to OrdSet[int].
OrdSet is for ordinal set. I realise there might be a confusion with ordered sets, so naming suggestions welcome.

Better error messages on compilation errors will be achieved when the concepts are ready, but I didn't want to tie it to a future roadmap, and the current solution achieves the stated goals satisfactorily IMHO.


lib/pure/collections/intsets.nim
```diff
@@ -42,13 +52,17 @@ type
     bits: array[0..IntsPerTrunk - 1, BitScalar] # a bit vector
 
   TrunkSeq = seq[PTrunk]
-  IntSet* = object       ## An efficient set of `int` implemented as a sparse bit set.
+
+  ## A must be an ordinal equivalent type: int | enum | distinct of an int
+  OrdSet*[A] = object    ## An efficient set of `int` implemented as a sparse bit set.
```
timotheecour: shouldn't it be: `OrdSet*[A: Ordinal]` or something similar?

lib/pure/collections/intsets.nim
```diff
@@ -42,13 +52,17 @@ type
...
     elems: int           # only valid for small numbers
     counter, max: int
```
timotheecour: where is `A` used in type fields?

lib/pure/collections/intsets.nim
```diff
@@ -253,35 +279,37 @@ proc incl*(s: var IntSet, other: IntSet) =
...
     assert len(b) == 2
 
-  result = initIntSet()
-  for item in items(x):
+  result = initOrdSet[A]()
+  for item in x:
     result.incl(item)
 
-proc containsOrIncl*(s: var IntSet, key: int): bool =
+proc toIntSet*(x: openArray[int]): IntSet {.since: (1, 3).} = toOrdSet[int](x)
```
timotheecour: not `since`, `toIntSet` symbol already existed prior your PR (albeit with different implementation)

ditto wherever relevant

lib/pure/collections/intsets.nim
```diff
@@ -7,13 +7,23 @@
 #    distribution, for details about the copyright.
 #
 
-## The ``intsets`` module implements an efficient `int` set implemented as a
+## The ``intsets`` module implements an efficient ordinal set implemented as a
```
timotheecour: what about instead adding `lib/pure/collections/ordsets.nim` (with no mention of intsets there) and then making  `lib/pure/collections/intsets.nim` a thin wrapper over `OrdSet[int]` that imports ordsets ?

lib/pure/collections/intsets.nim
landerlo: Thank you, fixed

lib/pure/collections/intsets.nim
landerlo: I kept what it was in devel, toIntSet already has since 1.3

lib/pure/collections/intsets.nim
timotheecour: if an item is fixed (as opposed to pushback), you can click "resolve conversation" ;-)

lib/pure/collections/intsets.nim
landerlo: I thought about that, but I wanted to keep it as simple as possible to spot the differences, does github detect moved to show the diff?

landerlo: Now that I think of it, with the move in a separate commit the diff is still seen in the first commit, so I did the move to ordsets and kept intsets as an alias to ordsets. Not sure if that causes any issue with backwards compatibility. This is my first nim code,

landerlo: Still, there are lots of imports in the compiler. is there any way of importing transitively in nim? i.e. I leave intset just with the IntSet type alias and IntSet specific procs and import from ord. It seems that if some old user code imports intset it won't get the other procs transitively even if intset imports from ordset, is this correct?

disruptek:
I liked it better before it touched the compiler, but whatever.  I think we have to wait until `Ordinal` is a concept before it will work for `distinct` types.

lib/pure/collections/intsets.nim
timotheecour: did you try this?
```nim
# intsets.nim
import std/ordsets
export ordsets
```

timotheecour:
> I liked it better before it touched the compiler, but whatever. 

/cc @landerl
I agree with @disruptek , I don't think there's a need to change client code, see my answer here: https://github.com/nim-lang/Nim/pull/15564#discussion_r504416600


> I think we have to wait until Ordinal is a concept before it will work for distinct types.

what do you mean? this passes:
```nim
type int2 = distinct int
doAssert int2 is Ordinal
```

lib/pure/collections/intsets.nim
landerlo: Ok, I didn't know about exports. However it says it 'Error: cannot export: ordsets'

landerlo:
> > I liked it better before it touched the compiler, but whatever.
> 
> /cc @landerl
> I agree with @disruptek , I don't think there's a need to change client code, see my answer here: [#15564 (comment)](https://github.com/nim-lang/Nim/pull/15564#discussion_r504416600)
> 
> > I think we have to wait until Ordinal is a concept before it will work for distinct types.
> 
> what do you mean? this passes:
> 
> ```nim
> type int2 = distinct int
> doAssert int2 is Ordinal
> ```
See my comment above, making all the type params A: Ordinal didn't work for me, but just making OrdSet[A: Ordinal] and the api procs [A] seems to work and has now better compilation errors, so yes, concepts are not necessary for this.

lib/pure/collections/intsets.nim
timotheecour: why? do you have a minimal example (or simply a temporary branch in your nim fork ) where it gives this error so we can look into it?
for eg this works and is a common pattern:
```nim
# in foo.nim
import std/tables
export tables
```

lib/pure/collections/intsets.nim
landerlo: I hadn't been able to make [A: Ordinal] work and now I see that apparently you cannot have both the type and the procs in the api constrained to ordinal. Nevertheless unconstraining the Api and making the type constrained works and it's better for the type safety. Changed that.

disruptek:
I mean that `Ordinal` doesn’t capture a type than can nominally behave as an ordinal because it’s magic and not a concept.

landerlo:
> I mean that `Ordinal` doesn’t capture a type than can nominally behave as an ordinal because it’s magic and not a concept.

At the moment is magic but there could be a concept modeling Ordinal  with `ord(a: A): A`, `fromOrd(o: int): A`, and an implicit assumption that the constraint: `forall x: int ord(fromOrd(int)) == x && forall a: A fromOrd(ord(a)) == a`, that's left to an axiom for ints, enums and distincts of an ordinal. But maybe user could define other ordinal types to benefit from things like ordsets.

That Ordinal concept could be further generalised to Isomorphic. so concept `Ordinal[A] = Isomorphic[A, int]` with the above signature.

tests/stdlib/tordsets.nim
```diff
@@ -0,0 +1,97 @@
+import ordsets
```
timotheecour: new modules should not be importable without `std/` prefix to avoid name clashes with future nimble packages; maybe ordsets should be under `std/ordsets` even though it's a collection

lib/pure/collections/ordsets.nim
```diff
@@ -0,0 +1,784 @@
...
+
+  ## An efficient set of `Ordinal` implemented as a sparse bit set.
+  OrdSet*[A: Ordinal] = object
+    elems: int           # only valid for small numbers
+    counter, max: int
+    head: PTrunk
+    data: TrunkSeq
+    a: array[0..33, int] # profiling shows that 34 elements are enough
+
+  IntSet* = OrdSet[int]
```
timotheecour: is this used at all here? why not move to intsets.nim
(make sure tordsets.nim sill works, if needed you can also import std/intsets from there)

tests/stdlib/tordsets.nim
landerlo: What does this entail? moving the ordsets.nim to Nim/lib/std/ folder?

timotheecour: it's marked as resolved but was it resolved? ;-)

timotheecour: nevermind, i hadn't clicked refresh!

tests/stdlib/tordsets.nim
```diff
@@ -0,0 +1,97 @@
+import std/ordsets
```
timotheecour: you've changed this to `import std/ordsets` (good) but the problem remains:
`import ordsets` should not compile IMO, there's a policy (i forgot where) that new modules show require std prefix; that's to avoid clashes with nimble packages

hence my last remark.
can you move ordsets to: `lib/std/ordsets` ?

yes I know it's a collection but that's a less important point than module name global scope pollution IMO

feel free to check with @araq before doing this change

landerlo: I've pushed the move to lib/std, I'm switching off now, but I can check with him tomorrow. Thank you for the review, this is literally my first nim code so I'm learning along.

landerlo: I've checked RFCs about nimble and naming and other that the std/ prefix as you suggested I couldn't find any clear criteria about whether it should go on folder lib/std or lib/pure/collections. What's your executive decision on this?
I've amended the commits to be clearer, 1. Full change over intset file for an easier diff of the change. 2. Rename file to ordsets and factor out intsets with the ordsets exports. 3. Just the tentative file move to to /lib/std

timotheecour: see https://github.com/nim-lang/Nim/pull/8024#issuecomment-396943421
> There is no reason whatsoever we can't have a long deprecation period for import strutils and the like, we can even decide the old stdlib modules should be seen as "keyword-like" and do not have to deprecate it at all (!) but all the new stuff ends up in std.

I'm adding this convention in this PR: https://github.com/nim-lang/Nim/pull/15590

landerlo: The last commit moves `ordsets.nim` to `lib/std`. Is there anything else that needs doing?

landerlo:
@timotheecour is there anything else needed for the PR? ordsets is in lib/std folder as per your request.

disruptek:
I feel bad for how a simple PR became such a large bikeshed.  Can this go in?  Please?

tests/stdlib/tordsets.nim
```diff
@@ -0,0 +1,97 @@
...
+
+      template excl2(i) =
+        t.excl i
+        t0.excl i
+
+      var expected: seq[`A`]
+      let n = 100
+      let n2 = n*2
+      for i in 0..<n:
+        incl2(`A`(i))
```
timotheecour: nit: why the quoting of A?
ditto elsewhere

tests/stdlib/tordsets.nim
```diff
@@ -0,0 +1,97 @@
...
+
+  # Test distincts
+  var distT: OrdSet[id]
+  var distT0: HashSet[id]
+  testDel(id, distT, distT0)
+
+  assert union(distT, initOrdSet[id]()) == distT
+
+  # mixing sets of different types shouldn't compile
+  #assert union(distT, initOrdSet[int]()) == distT # KO! typesafe
```
timotheecour: instead of "should fail" (etc), test for it using:
`doAssert not compiles(union(distT, initOrdSet[int]()))` or similar

ditto elsewhere

tests/stdlib/tordsets.nim
```diff
@@ -0,0 +1,97 @@
...
+
+  #Should fail
+  type nonIntDistinct = distinct string
```
timotheecour: style: user types start with capital

ditto elsewhere eg `type id = distinct int`

tests/stdlib/tordsets.nim
```diff
@@ -0,0 +1,97 @@
...
+  var t: OrdSet[int]
+  var t0: HashSet[int]
+  testDel(int, t, t0)
+
+  # Test distincts
+  var distT: OrdSet[id]
+  var distT0: HashSet[id]
+  testDel(id, distT, distT0)
+
+  assert union(distT, initOrdSet[id]()) == distT
```
timotheecour: use `doAssert` in tests, not `assert`

lib/std/ordsets.nim
```diff
@@ -0,0 +1,782 @@
...
+
+proc `$`*[A](s: OrdSet[A]): string =
+  ## The `$` operator for int sets.
+  ##
+  ## Converts the set `s` to a string, mostly for logging and printing purposes.
+  dollarImpl()
+
+
+
+when isMainModule:
```
timotheecour: (could be done in cleanup post-merge) tests must be merged with tests/stdlib/tordsets.nim, not  in a isMainModule block; and possibly reduced in case some of it is redundant with tests/stdlib/tordsets.nim

lib/std/ordsets.nim
```diff
@@ -0,0 +1,782 @@
...
+    if t != nil:
+      var u = key and TrunkMask
+      t.bits[u shr IntShift] = t.bits[u shr IntShift] and
+          not(BitScalar(1) shl (u and IntMask))
+
+template dollarImpl(): untyped =
+  result = "{"
+  for key in items(s):
+    if result.len > 1: result.add(", ")
+    result.add($(ord(key)))
```
timotheecour: this is the only real blocker.
```nim
proc main() =
  type Foo = enum k0 k1 k2
  var a = initOrdSet[Foo]()
  a.incl k0
  a.incl k2
  echo $a # got {0, 2}, expected {k0, k2}
main()
```

lib/std/ordsets.nim
```diff
@@ -0,0 +1,782 @@
...
+#        (c) Copyright 2012 Andreas Rumpf
+#
+#    See the file "copying.txt", included in this
+#    distribution, for details about the copyright.
+#
+
+## The ``ordsets`` module implements an efficient `Ordinal` set implemented as a
+## `sparse bit set`:idx:.
+##
+## Supported `A: Ordinal` types include enums and distincts with int base.
```
timotheecour: true but incomplete (excludes char, bool etc).
instead, maybe just say: Supports any `Ordinal` type.

landerlo:
> I feel bad for how a simple PR became such a large bikeshed. Can this go in? Please?

It replaces a collection used in the compiler and I imagine quite popular, so I'm glad it's been properly scrutinized, but I think after the iterations it's ready to be merged. I also checked there is no performance overhead so the zero cost abstraction works as expected.

lib/std/ordsets.nim
```diff
@@ -0,0 +1,782 @@
...
+  assert length > t.counter
+  result = (length * 2 < t.counter * 3) or (length - t.counter < 4)
+
+proc nextTry(h, maxHash: Hash, perturb: var Hash): Hash {.inline.} =
+  const PERTURB_SHIFT = 5
+  var perturb2 = cast[uint](perturb) shr PERTURB_SHIFT
+  perturb = cast[Hash](perturb2)
+  result = ((5*h) + 1 + perturb) and maxHash
+
+proc intSetGet[A](t: OrdSet[A], key: int): PTrunk =
```
timotheecour: nit (can be made post merge in a cleanup PR) all these private procs could be renamed eg `ordSetGet` etc

timotheecour (approved):
LGTM, fixing `$` is the main thing missing, the rest are nits that can be addressed in future cleanup PRs

note to other reviewers:
github and git diff aren't smart enough here, to view the diff use:
```
git diff 675cf57:lib/pure/collections/intsets.nim lib/std/ordsets.nim
git diff 675cf57:tests/stdlib/tintsets.nim tests/stdlib/tordsets.nim
```
(adapt `675cf57` as needed)

lib/std/ordsets.nim
landerlo: So the change is to supports distincts with any ordinal base? I didn't check any other cases other than int but that makes sense.

timotheecour: I mean that `Ordinal` is already defined elsewhere so there's no need to specify again what it means (and excluding some members from its definition eg `char`, `bool` etc). I just checked locally that `bool` and `enum` do work with ordsets.

lib/std/ordsets.nim
landerlo: There is an ordSetGet already, the old method is kept as it seems it's necessary to keep compatibility with tests for version 1.0, tests would fail otherwise If I remember correctly, but I can double-check.

landerlo:
> LGTM, fixing `$` is the main thing missing, the rest are nits that can be addressed in future cleanup PRs
> 
> note to other reviewers:
> github and git diff aren't smart enough here, to view the diff use:
> 
> ```
> git diff 675cf57:lib/pure/collections/intsets.nim lib/std/ordsets.nim
> git diff 675cf57:tests/stdlib/tintsets.nim tests/stdlib/tordsets.nim
> ```
> 
> (adapt `675cf57` as needed)

All the changes are in the first commit to be more visible. Then there is one renaming it.

lib/std/ordsets.nim
landerlo: Ok, so implementing proc $, is Nim's string representation? I'll change that to get the ordinal name for the ints we store in the bitset

tests/stdlib/tordsets.nim
landerlo: It's a typedesc arg for the template, isn't the quoting required for the typecast?

lib/std/ordsets.nim
landerlo: I carried over the main tests existing from intset but I'll migrate to tordset.nim if there is anything not covered already in those tests

tests/stdlib/tordsets.nim
timotheecour: no, this works:
```nim
  template fn(A: typedesc) = echo A(1)
  fn(float)
```

lib/std/ordsets.nim
timotheecour: all that's needed is:
```nim
# in dollarImpl
result.add($(ord(key)))
=>
result.add $key
```

note for distinct:
```nim
when true:
  import std/ordsets
  type Foo = distinct int16
  proc `$`(a: Foo): string {.borrow.} # `echo a` below won't work without `$` defined, as expected
  var a = initOrdSet[Foo]()
  a.incl Foo(3)
  a.incl Foo(4)
  echo a
```

lib/std/ordsets.nim
timotheecour: > There is an ordSetGet already

I don't see it?

> keep compatibility with tests for version 1.0

that doesn't make sense, this is private proc

in any case, feel free to delay this to a future cleanup PR since it's not client facing

lib/pure/collections/intsets.nim
```diff
@@ -7,698 +7,18 @@
...
-const
-  InitIntSetSize = 8              # must be a power of two!
-  TrunkShift = 9
-  BitsPerTrunk = 1 shl TrunkShift # needs to be a power of 2 and
-                                  # divisible by 64
-  TrunkMask = BitsPerTrunk - 1
-  IntsPerTrunk = BitsPerTrunk div (sizeof(BitScalar) * 8)
-  IntShift = 5 + ord(sizeof(BitScalar) == 8) # 5 or 6, depending on int width
-  IntMask = 1 shl IntShift - 1
+proc toIntSet*(x: openArray[int]): IntSet {.since: (1, 3), inline.} = toOrdSet[int](x)
```
timotheecour: `(1,5,1)` now

lib/std/ordsets.nim
landerlo: My mistake I was thinking about the public methods toIntSet/toOrdSet. I'll rename intSetGet

lib/pure/collections/intsets.nim
landerlo: toIntset was public and had the since 1.3 from before, so shouldn't we keep that version info?

timotheecour: misread

timotheecour (approved):

lib/pure/collections/intsets.nim
```diff
@@ -7,698 +7,18 @@
...
 
-## The ``intsets`` module implements an efficient `int` set implemented as a
-## `sparse bit set`:idx:.
-##
-## **Note**: Currently the assignment operator ``=`` for ``IntSet``
-## performs some rather meaningless shallow copy. Since Nim currently does
-## not allow the assignment operator to be overloaded, use `assign proc
-## <#assign,IntSet,IntSet>`_ to get a deep copy.
-##
+## Deprecated by the generic `OrdSet` for ordinal sparse sets.
```
timotheecour: CI fails:

Error: undeclared field: 'id' for type system.int [type declared in /Users/runner/work/Nim/Nim/lib/system/basic_types.nim(2, 3)]

timotheecour:
I'd like to merge this soon unless there are further objections
/cc @Araq

Araq:
Sorry I'm late to the party, but is ``OrdSet`` really the best name we could come up with? How about a ``BitmaskSet``, ``BitSet``, ``HashBitSet``?

landerlo:
> Sorry I'm late to the party, but is `OrdSet` really the best name we could come up with? How about a `BitmaskSet`, `BitSet`, `HashBitSet`?

There is the `IntSet` alias, which is basically what people would expect when they read `BitSet`. But if we underlie that it's about Ordinals, this showcases Nim's generics and Ordinals features.
In my use case I have sets of different id types (distincts of int). The generic Ordinal mask set gets you  type safety when you insert and combine different sets with the performance and memory footprint of a bitset. I cannot think of any other language I know where this can be provided this seamlessly.

That said, I have some doubts about `OrdSet` as the Ord could be thought as Ordered, rather than Ordinal. Please advice a name and I'll rename this.

timotheecour:
`BitSet` (in `std/bitsets`) would be the best IMO (short and self documenting) but there's already `compiler/bitsets.nim`; this could be renamed to avoid confusion but with small caveat that compiler is a nimble package too so could potentially break client code relying on it (but compiler code should be changeable IMO, outside of a stable subset `compilerapi`).

Next best would be BitMaskSet (in std/bitmasksets) but it's rather lengthy

Araq:
``BucketSet`` ? Sorry, I'm still looking for a good name. Or maybe ``ClusterSet`` as it's a set of bit clusters.

timotheecour:
`PackedSet` or `SparseSet` also is possible; anything with 2 words is fine (`HashBitSet` is a bit long)

disruptek:
`OrdinalSet`

Araq:
Let's go with ``PackedSet``.

timotheecour:
@landerlo ok so PackedSet + std/packedsets + tpackedsets.nim; hopefully it won't be too painful to do this renaming; thanks!

landerlo:
No problem, renamed to PackedSet/packedsets

timotheecour (approved):

timotheecour:
LGTM; good to merge?

timotheecour:
> landerlo requested a review from timotheecour 16 minutes ago

I've already did a full review and gave my LGTM ...

@araq or other nim member can i merge this?

Araq:
Sorry, I thought it already was merged.

landerlo:
I do realise now that sed doesn't go through the commit messages, so the commit log still talks about OrdSet and not  the renamed PackedSet. Nevermind.

disruptek:
Thanks for all your work on this, @landerlo -- I know it was a slog but this is a nice contribution. ❤️

disruptek:
I hate to say it, but this is causing the `items` iterator against an `IntSet` to fail:

https://github.com/disruptek/gram/runs/1419861309?check_suite_focus=true

```nim
/home/adavidoff/nims/lib/std/packedsets.nim(370, 14) Error: type mismatch: got <PackedSet[system.int]>
but expected one of: 
iterator items(a: cstring): char
  required type for a: cstring
  but expression 's' is of type: PackedSet[system.int]
iterator items(a: string): char
  required type for a: string
  but expression 's' is of type: PackedSet[system.int]
iterator items(n: NimNode): NimNode
  required type for n: NimNode
  but expression 's' is of type: PackedSet[system.int]
iterator items[IX, T](a: array[IX, T]): T
  required type for a: array[IX, T]
  but expression 's' is of type: PackedSet[system.int]
iterator items[N, E; F: static[GraphFlags]](graph: Graph[N, E, F]): N
  required type for graph: Graph[items.N, items.E, items.F]
  but expression 's' is of type: PackedSet[system.int]
iterator items[T: char](a: openArray[T]): T
  required type for a: openArray[T: char]
  but expression 's' is of type: PackedSet[system.int]
iterator items[T: enum](E: typedesc[T]): T
  required type for E: type T: enum
  but expression 's' is of type: PackedSet[system.int]
iterator items[T: not char](a: openArray[T]): lent2 T
  required type for a: openArray[T: not char]
  but expression 's' is of type: PackedSet[system.int]
iterator items[T](a: seq[T]): lent2 T
  required type for a: seq[T]
  but expression 's' is of type: PackedSet[system.int]
iterator items[T](a: set[T]): T
  required type for a: set[T]
  but expression 's' is of type: PackedSet[system.int]
iterator items[T](s: HSlice[T, T]): T
  required type for s: HSlice[items.T, items.T]
  but expression 's' is of type: PackedSet[system.int]

expression: items(s)
```

timotheecour:
@disruptek can you provide a minimized reproducible example?

disruptek:
Is gram too complex for you?

timotheecour:
that's basically what everyone has to go through when submitting bug reports, providing minimized reproducing examples. More work for the reporter, less work for everyone else looking at it.

disruptek:
Yeah, but only if they didn't manage to get their package into `important_packages` first.  If they were able to do that, then for some reason the developers are willing to do the same work themselves.  Weird.

timotheecour:
* if its's a compiler bug: minimize with 0 stdlib module dependencies, when possible
* if it's a stdlib bug: minimize with as few stdlib module dependencies, ideally just the module in question
it's about minimizing total work = 1 * work(bug reporter) + N * work(people looking at issue)

disruptek:
Let's agree that you'll be the `1` to do the work and all `N` of us will reap the benefits.

Araq:
As a compromise, both modules could expand a common template implementation -- then there cannot be regressions like this.

landerlo:
@disruptek Sorry for the regression. Tests test int iterators but there must be some issue why the iterator is not properly inferred.
I've created a PR that simply reverts IntSet to previous impl rather than as an alias for PackedSet[int], that should do the trick for now. I want to dig deeper into the issue anyway at some point.
Rollbak IntSet as before:
https://github.com/nim-lang/Nim/pull/16059

disruptek:
Can't we just fix it instead?

timotheecour:
> As a compromise, both modules could expand a common template implementation -- then there cannot be regressions like this.

=> https://github.com/nim-lang/Nim/pull/16060, much cleaner

disruptek:
Nice work!  Another good reason to be explicit with iterators, at least until this is fixed.

PR #15567, opened 2020-10-14
Fix #13609

### LINKED ISSUE #13609: Templates can't be made discardable
The Nim manual contains the following snippet under [push and pop pragmas](https://nim-lang.github.io/Nim/manual.html#pragmas-push-and-pop-pragmas):

```nim
{.push discardable, boundChecks: off, compileTime, noSideEffect, experimental.}
template example(): string = "https://nim-lang.org"
{.pop.}
```

But if you try to use this template like so:

```nim
example()
```

You get:
```
Error: expression '"https://nim-lang.org"' is of type 'string' and has to be discarded
```

I'm not sure how many of the demonstrated pragmas are supposed to be meaningful (I can see that it's just a loose example after all), but this one stuck out to me because I've previously run into a situation where I've wished discardable would work with templates. My workaround was to wrap the result in a helper function like so:

```nim
func makeDiscardable[T](a: T): T {.discardable, inline.} = a
template example(): string = makeDiscardable("https://nim-lang.org")
example()
```

But regardless, I think either {.discardable.} for templates should be a real thing, or the manual should be changed so that people don't think the code in the snippet would actually work.

me7:
fixt  #13609 
This pull request add documenation in nim manual that {.discardable.} pragma not effective in template.


doc/manual.rst
```diff
@@ -2592,6 +2592,17 @@ been declared with the `discardable`:idx: pragma:
 
   p(3, 4) # now valid
 
+however discardable pragma not working on template because template do AST substitute in place, for example
```
Araq: Typos, grammar.

euantorano:
```suggestion
however the discardable pragma does not work on templates as templates substitute the AST in place. For example:
```

doc/manual.rst
```diff
@@ -2592,6 +2592,17 @@ been declared with the `discardable`:idx: pragma:
...
+
+.. code-block:: nim
+  {.push discardable .}
+  template example(): string = "https://nim-lang.org"
+  {.pop.}
+
+  example()
+
+This template will resolve into "https://nim-lang.org" which is string literal type, and since {.discardable.} don't apply to literals, the compiler will error.
```
euantorano:
```suggestion
This template will resolve into "https://nim-lang.org" which is a string literal and since {.discardable.} doesn't apply to literals, the compiler will error.
```

me7: Thank you for help fix typo and grammar. Apologize for the mistake, english is not my native language.

PR #15570, opened 2020-10-14
Make useVersion:1.0 disable the proc arg sym change

Clyybber:


Araq:
This should be active both for  ``--useVersion:1.2`` and ``--useVersion:1.0``.

PR #15577, opened 2020-10-14
[backport: 1.2] Rename `=` to `=copy`

Clyybber:


compiler/sempass2.nim
```diff
@@ -809,7 +809,8 @@ proc trackCall(tracked: PEffects; n: PNode) =
 
   if a.kind == nkSym and a.sym.name.s.len > 0 and a.sym.name.s[0] == '=' and
         tracked.owner.kind != skMacro:
-    let opKind = find(AttachedOpToStr, a.sym.name.s.normalize)
+    var opKind = find(AttachedOpToStr, a.sym.name.s.normalize)
+    if a.sym.name.s.normalize == "=": opKind = attachedAsgn.int
```
Araq, line 813: No need for ``normalize`` here.

Araq:
Why do we need this commit:

Duplicate tests for =copy hook

alaviss:
Isn't this a breaking change? Can we have some coverage on why is this change being made and why will it be backported to 1.2? I have projects that use the custom `=` hook.

ghost:
@alaviss 
> Isn't this a breaking change? Can we have some coverage on why is this change being made and why will it be backported to 1.2? I have projects that use the custom `=` hook.

It's not a breaking change, old code with = will still work, but the new proposed way is =copy.

IIRC the justification for that change is that = name for a hook might be misleading, =copy is much cleaner

alaviss:
That's good to know, but can we have this in the changelog? I don't think this rename has been covered much other than a few mentions here and there.

Araq:
I took over, sorry. I really want to get 1.4 out.

PR #15604, opened 2020-10-16
add criterion to important packages

disruptek:
It is expected not to pass tests at the moment.


disruptek:
It's expected to pass tests and has a custom test for compiler ci so we can tweak it later if it proves to be too lengthy; I expect this one to take a more reasonable 5mins or so.

Araq:
5 mins is quite much. Can you do fewer iterations in your tests instead?

disruptek:
I don't know how criterion works but I'll see what I can do.

PR #12779, opened 2019-11-30
nimgrep improvements

a-mr:
Those are a few changes to nimgrep:
- add context printing (lines after and before a match)
- add exclude/include options for files and exclude for directories
- add a `--newLine` style option for starting matching/context
      lines from a new line
- add color themes: 3 new themes (ack, gnu, bnw) besides default `simple`
- enable printing of multi-line matches with line numbers
- proper display of replace when there was another match replaced at
      the same line / context block 
- improve cmdline arguments error reporting
- improve error printing
- improve symlink handling
- context printing: handle Posix file endings with native appearance
- rename dangerous `-r` argument. Having -r as --replace is dangerous: it can be easily confused with --recursive, and indeed a few popular Unix tools like GNU grep, ack, freebsd grep, etc have -r as recursive which makes it especially unfortunate.

Here is an example of invocation:
```
nimgrep -y --recursive --color --excludeDir:"\.git$" --excludeDir:"nimcache$" \
  --noExt --newLine --colortheme:ack --afterContext:1 --beforeContext:3 \
  linesBefore .|less
```

image

Also a minor bug with terminal.writeStyled/styledWrite for Posix was found, see fix and the (visual) test in the 1st commit.

One minor thing: I also observed that there is no way to detect for Posix style file ending at the end of a file in Nim, that is when '\l' at the end of a file is considered as end of previous line and does not generate next line according to Posix standard. 2nd commit here is to be able to handle this correctly in splitLines/countLines (it does not change the default behavior). The test is updated.

PR #15612, opened 2020-10-17
Nimgrep improvements 2

a-mr:
I'm using nimgrep a lot in my daily job, and I added a few useful things.

* Speed up search by using multiple threads (options --nWorkers:N, -n:N)
* UPDATED: implement pipe mode `-`, which allows for using nimgrep in Unix filters.
* Ability to limit output by cropping lines at a fixed number of 8-bit characters (--limit:N, -m:N) or by terminal width (--fit). See example on picture below. It's useful when dealing with binary files with long lines or when printing on small screens, since line breaks mess up visual perception.
* Better displaying of non-printable characters by substituted ASCII characters with colors (--onlyAscii, -o). It's useful for displaying binary files, especially with the limit options above.
* Ability to detect binary files and exclude them from the search (--bin:no = --text, -t) or include only them (--bin:only)
* Add --sortTime option to be able e.g. search starting from recently changed files — when you remember that what you are looking for is recent
* The order in which files appear is fixed now: it's alphabetic sorting by default (unless --sortTime is specified). I want the ability to expect that files will appear in stable order any time to memorize where should I look to; while Linux directory walking yields you files and directories non-deterministically. It's especially important with threads that would mess up the order if they were given the chance. (However this increases memory consumption in multi-thread mode since search results need to be stored in memory until contiguous sequence of files is received from thread workers)
* Add --match and --noMatch option. Useful when you want to search only specific files based on presence of string/match in them, e.g. an appropriate import statement, or copyright year, or language extensions/features/versions for such languages...
* Proper printing of TABs — useful e.g. when searching through some C code bases like linux kernel
* Add --count option for displaying only number of matches in files
* add --includeDir option for limiting search only to specific directories (e.g. '.*tests' if your project have such a convention for unit tests)
* UPDATED: (probably breaking) replacement mode now requires specifying all 3 positional arguments, since it seems too dangerous that something will be omitted and works not as expecting, especially considering that --confirm is not default. Once playing with --replace I made a mistake of forgetting specify any replacement, and nimgrep accepted it as an empty string "", basically deleting all patterns. Targets also should be specified explicitly: current dir "." is too dangerous for a default setting.
* (possibly a breaking change?) if no directory is specified then "." is used instead of the fullpath, which was redundant IMHO
* Peg can be used for all patterns now (including --includeFile, --match, etc)
* fix performance regression introduced in nimgrep improvements #12779, which slowed down search by 10—15 %.
* switch to --gc:orc
* miscellaneous: better error reporting. #13950 is fixed in another way (CC @genotrance). Fix #15318 (CC @juancarlospaco).
* UPDATED: --help is reworked and better structured.

image

A note on thread performance: the speed up with regex is not very great, it's limited to factor 2.5—3. Basically after reaching -n:3 or -n:4 threads performance stops to increase. I debugged that with gdb, it seems that the program with large number of threads (e.g. -n:20) spends almost all its time in syscalls read, that is reading the files into the process memory. I don't know whether it's gotten memory-bound or Linux is unable to deliver file contents because of some internal locks. When using (slow) Peg pattern, performance increases nearly linearly: I got speed-up 5.5—5.8 on my 6-core Ryzen 3600 with -n:6 and -n:12.

With --gc:orc speed is about the same as with --gc:markandsweep. But with orc nimgrep consumes ~2 times less memory in multi-threaded mode: 170 MB with -n:6 on Nim repository, while it is 370 MB with --gc:markandsweep. (Measure by Linux utility /usr/bin/time --verbose).

### DIFF
--- a/testament/categories.nim
+++ b/testament/categories.nim
@@ -262,7 +262,8 @@ proc debuggerTests(r: var TResults, cat: Category, options: string) =
   if fileExists("tools/nimgrep.nim"):
     var t = makeTest("tools/nimgrep", options & " --debugger:on", cat)
     t.spec.action = actionCompile
-    # force target to C because of MacOS 10.15 clang++ bug (see #15612)
+    # force target to C because of MacOS 10.15 SDK headers bug
+    # https://github.com/nim-lang/Nim/pull/15612#issuecomment-712471879
     t.spec.targets = { targetC }
     testSpec r, t
 
--- a/testament/categories.nim
+++ b/testament/categories.nim
@@ -262,6 +262,8 @@ proc debuggerTests(r: var TResults, cat: Category, options: string) =
   if fileExists("tools/nimgrep.nim"):
     var t = makeTest("tools/nimgrep", options & " --debugger:on", cat)
     t.spec.action = actionCompile
+    # force target to C because of MacOS 10.15 clang++ bug (see #15612)
+    t.spec.targets = { targetC }
     testSpec r, t
 
 # ------------------------- JS tests ------------------------------------------
--- a/tools/nimgrep.nim
+++ b/tools/nimgrep.nim
@@ -621,7 +621,7 @@ proc printOutput(filename: string, output: Output, curCol: var Column) =
 iterator searchFile(pattern: Pattern; filename: string;
                     buffer: string): Output =
   var prevMi, curMi: MatchInfo
-  curMi.lineEnd = 1
+  prevMi.lineEnd = 1
   var i = 0
   var matches: array[0..re.MaxSubpatterns-1, string]
   for j in 0..high(matches): matches[j] = ""
@@ -634,32 +634,32 @@ iterator searchFile(pattern: Pattern; filename: string;
                      firstLine: prevMi.lineEnd)
       break
 
-    let lineBeg = curMi.lineEnd + countLineBreaks(buffer, i, t.first-1)
+    let lineBeg = prevMi.lineEnd + countLineBreaks(buffer, i, t.first-1)
     curMi = (first: t.first,
              last: t.last,
              lineBeg: lineBeg,
              lineEnd: lineBeg + countLineBreaks(buffer, t.first, t.last),
              match: buffer.substr(t.first, t.last))
     if prevMi.lineBeg == 0: # no prev. match, so no prev. block to finalize
-      yield Output(kind: BlockFirstMatch,
-                   pre: getSubLinesBefore(buffer, curMi),
-                   match: curMi)
+      let pre = getSubLinesBefore(buffer, curMi)
+      prevMi = curMi
+      yield Output(kind: BlockFirstMatch, pre: pre, match: move(curMi))
     else:
       let nLinesBetween = curMi.lineBeg - prevMi.lineEnd
       if nLinesBetween <= linesAfter + linesBefore + 1: # print as 1 block
-        yield Output(kind: BlockNextMatch,
-                     pre: getSubLinesBetween(buffer, prevMi, curMi),
-                     match: curMi)
+        let pre =  getSubLinesBetween(buffer, prevMi, curMi)
+        prevMi = curMi
+        yield Output(kind: BlockNextMatch, pre: pre, match: move(curMi))
       else: # finalize previous block and then print next block
-        yield Output(kind: BlockEnd,
-                     blockEnding: getSubLinesAfter(buffer, prevMi),
+        let after = getSubLinesAfter(buffer, prevMi)
+        yield Output(kind: BlockEnd, blockEnding: after,
                      firstLine: prevMi.lineEnd)
+        let pre = getSubLinesBefore(buffer, curMi)
+        prevMi = curMi
         yield Output(kind: BlockFirstMatch,
-                     pre: getSubLinesBefore(buffer, curMi),
-                     match: curMi)
-
+                     pre: pre,
+                     match: move(curMi))
     i = t.last+1
-    prevMi = curMi
 
 func detectBin(buffer: string): bool =
   for i in 0 ..< min(1024, buffer.len):
@@ -779,7 +779,7 @@ iterator processFile(searchOptC: SearchOptComp[Pattern], filename: string,
       reason = "contains a forbidden match"
 
   if reject:
-    yield Output(kind: Rejected, reason: reason)
+    yield Output(kind: Rejected, reason: move(reason))
   else:
     var found = false
     var cnt = 0
@@ -793,7 +793,7 @@ iterator processFile(searchOptC: SearchOptComp[Pattern], filename: string,
     if optCount in options and cnt > 0:
       yield Output(kind: JustCount, matches: cnt)
     if yieldContents and found and optCount notin options:
-      yield Output(kind: FileContents, buffer: buffer)
+      yield Output(kind: FileContents, buffer: move(buffer))
 
 proc hasRightFileName(path: string, walkOptC: WalkOptComp[Pattern]): bool =
   let filename = path.lastPathPart
--- a/tools/nimgrep.nim
+++ b/tools/nimgrep.nim
@@ -155,6 +155,10 @@ type
     checkNoMatch: Pat
   SinglePattern[PAT] = tuple  # compile single pattern for replacef
     pattern: PAT
+  Column = tuple  # current column info for the cropping (--limit) feature
+    terminal: int
+    file: int
+    overflowMatches: int
 
 var
   paths: seq[string] = @[]
@@ -328,17 +332,13 @@ proc blockHeader(filename: string, line: int|string, replMode=false) =
       printBlockLineN($line.`$`.align(alignment) & ":")
     stdout.write("\n")
 
-type Column = tuple  # current column info for the cropping (--limit) feature
-  terminal: int
-  file: int
-  overflowMatches: int
-
 proc newLn(curCol: var Column) =
   stdout.write("\n")
   curCol.file = 0
   curcol.terminal = 0
 
-proc lineHeader(filename: string, line: int|string, isMatch: bool, curCol: var Column) =
+proc lineHeader(filename: string, line: int|string, isMatch: bool,
+                curCol: var Column) =
   let lineSym =
     if isMatch: $line & ":"
     else: $line & " "
@@ -457,7 +457,8 @@ proc getSubLinesBefore(buf: string, curMi: MatchInfo): string =
   result = substr(buf, first, curMi.first-1)
 
 proc printSubLinesBefore(filename: string, beforeMatch: string, lineBeg: int,
-                         curCol: var Column, reserveChars: int, replMode=false) =
+                         curCol: var Column, reserveChars: int,
+                         replMode=false) =
   # start block: print 'linesBefore' lines before current match `curMi`
   let sLines = splitLines(beforeMatch)
   let startLine = lineBeg - sLines.len + 1
@@ -1168,6 +1169,8 @@ for kind, key, val in getopt():
     of "fit":
       incl(options, optFit)
       limitChar = terminalWidth()
+      when defined(windows):  # Windows cmd&powershell add an empty line when
+        limitChar -= 1        # printing '\n' right after the last column
     of "onlyascii", "only-ascii", "o": optOnlyAscii = true
     of "verbose": incl(options, optVerbose)
     of "filenames": incl(options, optFilenames)
--- a/tools/nimgrep.nim
+++ b/tools/nimgrep.nim
@@ -11,12 +11,12 @@ import
   os, strutils, parseopt, pegs, re, terminal, osproc, tables, algorithm, times
 
 const
-  Version = "1.5"
+  Version = "1.6"
   Usage = "nimgrep - Nim Grep Utility Version " & Version & """
 
   (c) 2012 Andreas Rumpf
 Usage:
-  nimgrep [options] [pattern] [replacement] (file/directory)*
+  nimgrep [options] pattern [replacement] (file/directory)*
 Options:
   --find, -f          find the pattern (default)
   --replace, -!       replace the pattern
@@ -38,9 +38,11 @@ Options:
                       empty one ("--ext") means files with missing extension
   --noExt:EX1|...     exclude files having given extension(s), use empty one to
                       skip files with no extension (like some binary files are)
-  --includeFile:PAT   include only files whose names match the given regex PAT
-  --excludeFile:PAT   skip files whose names match the given regex pattern PAT
-  --excludeDir:PAT    skip directories whose names match the given regex PAT
+  --includeFile:PAT   include only files whose names match the given PATttern
+  --excludeFile:PAT   skip files whose names match the given pattern PAT
+  --excludeDir:PAT    skip directories whose names match the given pattern PAT
+  --match:PAT, -m:PAT select files containing a (not displayed) match of PAT
+  --noMatch:PAT       select files not containing any match of PAT
   --bin:yes|no|only   process binary files? (detected by \0 in first 1K bytes)
   --text, -t          process only text files, the same as --bin:no
   --count             just count number of matches
@@ -54,8 +56,8 @@ Options:
                -b:N   print N lines of leading context before every match
   --context:N, -c:N   print N lines of leading context before every match and
                       N lines of trailing context after it
-  --sortTime[:desc|asc],
-      -s[:desc|asc]   order files by modification time descending or ascending
+  --sortTime          order files by the last modification time -
+       -s[:desc|asc]  - descending (default) or ascending
   --group, -g         group matches by file
   --newLine, -l       display every matching line starting from a new line
   --verbose           be verbose: list every processed file
@@ -98,29 +100,32 @@ type
   Trequest = (int, string)
   Tresult = tuple[finished: bool, fileNo: int,
                   filename: string, fileResult: seq[Output]]
-  WalkOpt = tuple
+  WalkOpt = tuple  # used for walking directories/producing paths
     extensions: seq[string]
     skipExtensions: seq[string]
     excludeFile: seq[string]
     includeFile: seq[string]
     excludeDir : seq[string]
-  WalkOptPat = tuple
-    excludeFileP: seq[Peg]
-    includeFileP: seq[Peg]
-    excludeDirP : seq[Peg]
-    excludeFileR: seq[Regex]
-    includeFileR: seq[Regex]
-    excludeDirR : seq[Regex]
-
-using pattern: Pattern
+  WalkOptComp[Pat] = tuple  # a compiled version of the previous
+    excludeFile: seq[Pat]
+    includeFile: seq[Pat]
+    excludeDir : seq[Pat]
+  SearchOpt = tuple  # used for searching inside a file
+    pattern: string
+    checkMatch: string
+    checkNoMatch: string
+    checkBin: Bin
+  SearchOptComp[Pat] = tuple  # a compiled version of the previous
+    pattern: Pat
+    checkMatch: Pat
+    checkNoMatch: Pat
 
 var
   paths: seq[string] = @[]
-  pattern = ""
   replacement = ""
   options: TOptions = {optRegex}
-  opt {.threadvar.}: WalkOpt
-  checkBin = biYes
+  walkOpt {.threadvar.}: WalkOpt
+  searchOpt {.threadvar.}: SearchOpt
   justCount = false
   sortTime = false
   sortTimeOrder = SortOrder.Descending
@@ -136,6 +141,8 @@ var
   searchRequestsChan: Channel[Trequest]
   resultsChan: Channel[Tresult]
   colorTheme: string = "simple"
+
+searchOpt.checkBin = biYes
 
 proc ask(msg: string): string =
   stdout.write(msg)
@@ -423,7 +430,7 @@ proc printOutput(filename: string, output: Output) =
   of GroupEnd:
     printLinesAfter(filename, output.groupEnding, output.firstLine)
 
-iterator searchFile(pattern; filename: string; buffer: string): Output =
+iterator searchFile(pattern: Pattern; filename: string; buffer: string): Output =
   let si: SearchInfo = (buf: buffer, filename: filename)
   var prevMi, curMi: MatchInfo
   curMi.lineEnd = 1
@@ -471,75 +478,15 @@ func detectBin(buffer: string): bool =
     if buffer[i] == '\0':
       return true
 
-iterator processFile(pattern; filename: string, yieldContents=false): Output =
-  var buffer: string
-
-  if optFilenames in options:
-    buffer = filename
-  else:
-    try:
-      buffer = system.readFile(filename)
-    except IOError as e:
-      yield Output(kind: OpenError, msg: e.msg)
-
-  var reject = false
-  if checkBin in {biNo, biOnly}:
-    let isBin = detectBin(buffer)
-    if isBin and checkBin == biNo:
-      reject = true
-    if (not isBin) and checkBin == biOnly:
-      reject = true
-
-  if reject:
-    yield Output(kind: Rejected)
-  else:
-    var found = false
-    var cnt = 0
-    for output in searchFile(pattern, filename, buffer):
-      found = true
-      if not justCount:
-        yield output
-      else:
-        if output.kind in {GroupFirstMatch, GroupNextMatch}:
-          inc(cnt)
-    if justCount and cnt > 0:
-      yield Output(kind: JustCount, matches: cnt)
-    if yieldContents and found and not justCount:
-      yield Output(kind: FileContents, buffer: buffer)
-
-proc hasRightFileName(path: string, wopt: WalkOptPat): bool =
-  let filename = path.lastPathPart
-  let ex = filename.splitFile.ext.substr(1) # skip leading '.'
-  if opt.extensions.len != 0:
-    var matched = false
-    for x in opt.extensions:
-      if os.cmpPaths(x, ex) == 0:
-        matched = true
-        break
-    if not matched: return false
-  for x in opt.skipExtensions:
-    if os.cmpPaths(x, ex) == 0: return false
-  template checkFileName(patInclude: untyped, patExclude: untyped) =
-    if patInclude.len != 0:
-      var matched = false
-      for pat in patInclude:
-        if filename.match(pat):
-          matched = true
-          break
-      if not matched: return false
-    for pat in patExclude:
-      if filename.match(pat): return false
-  checkFileName(wopt.includeFileR, wopt.excludeFileR)
-  checkFileName(wopt.includeFileP, wopt.excludeFileP)
-  result = true
-
-proc hasRightDirectory(path: string, wopt: WalkOptPat): bool =
-  let dirname = path.lastPathPart
-  for pat in wopt.excludeDirR:
-    if dirname.match(pat): return false
-  for pat in wopt.excludeDirP:
-    if dirname.match(pat): return false
-  result = true
+proc compilePeg(initPattern: string): Peg =
+  var pattern = initPattern
+  if optWord in options:
+    pattern = r"(^ / !\letter)(" & pattern & r") !\letter"
+  if optIgnoreStyle in options:
+    pattern = "\\y " & pattern
+  elif optIgnoreCase in options:
+    pattern = "\\i " & pattern
+  result = peg(pattern)
 
 proc styleInsensitive(s: string): string =
   template addx =
@@ -575,7 +522,111 @@ proc styleInsensitive(s: string): string =
         addx()
     else: addx()
 
-iterator walkDirBasic(dir: string, wopt: WalkOptPat): string =
+proc compileRegex(initPattern: string): Regex =
+  var pattern = initPattern
+  var reflags = {reStudy}
+  if optIgnoreStyle in options:
+    pattern = styleInsensitive(pattern)
+  if optWord in options:
+    # see https://github.com/nim-lang/Nim/issues/13528#issuecomment-592786443
+    pattern = r"(^|\W)(:?" & pattern & r")($|\W)"
+  if {optIgnoreCase, optIgnoreStyle} * options != {}:
+    reflags.incl reIgnoreCase
+  result = if optRex in options: rex(pattern, reflags)
+           else: re(pattern, reflags)
+
+template declareCompiledPatterns(compiledStruct: untyped,
+                                 StructType: untyped,
+                                 body: untyped) =
+  if optRegex notin options:
+    var compiledStruct: StructType[Peg]
+    proc compile(p: string): Peg = p.compilePeg()
+    proc compileArray(initPattern: seq[string]): seq[Peg] =
+      for pat in initPattern:
+        result.add pat.compilePeg()
+    body
+  else:
+    var compiledStruct: StructType[Regex]
+    proc compile(p: string): Regex = p.compileRegex()
+    proc compileArray(initPattern: seq[string]): seq[Regex] =
+      for pat in initPattern:
+        result.add pat.compileRegex()
+    body
+
+iterator processFile(searchOptC: SearchOptComp[Pattern], filename: string, yieldContents=false): Output =
+  var buffer: string
+
+  if optFilenames in options:
+    buffer = filename
+  else:
+    try:
+      buffer = system.readFile(filename)
+    except IOError as e:
+      yield Output(kind: OpenError, msg: e.msg)
+
+  var reject = false
+  if searchOpt.checkBin in {biNo, biOnly}:
+    let isBin = detectBin(buffer)
+    if isBin and searchOpt.checkBin == biNo:
+      reject = true
+    if (not isBin) and searchOpt.checkBin == biOnly:
+      reject = true
+
+  if not reject:
+    if searchOpt.checkMatch != "":
+      reject = not contains(buffer, searchOptC.checkMatch, 0)
+
+  if not reject:
+    if searchOpt.checkNoMatch != "":
+      reject = contains(buffer, searchOptC.checkNoMatch, 0)
+
+  if reject:
+    yield Output(kind: Rejected)
+  else:
+    var found = false
+    var cnt = 0
+    for output in searchFile(searchOptC.pattern, filename, buffer):
+      found = true
+      if not justCount:
+        yield output
+      else:
+        if output.kind in {GroupFirstMatch, GroupNextMatch}:
+          inc(cnt)
+    if justCount and cnt > 0:
+      yield Output(kind: JustCount, matches: cnt)
+    if yieldContents and found and not justCount:
+      yield Output(kind: FileContents, buffer: buffer)
+
+proc hasRightFileName(path: string, walkOptC: WalkOptComp[Pattern]): bool =
+  let filename = path.lastPathPart
+  let ex = filename.splitFile.ext.substr(1) # skip leading '.'
+  if walkOpt.extensions.len != 0:
+    var matched = false
+    for x in walkOpt.extensions:
+      if os.cmpPaths(x, ex) == 0:
+        matched = true
+        break
+    if not matched: return false
+  for x in walkOpt.skipExtensions:
+    if os.cmpPaths(x, ex) == 0: return false
+  if walkOptC.includeFile.len != 0:
+    var matched = false
+    for pat in walkOptC.includeFile:
+      if filename.match(pat):
+        matched = true
+        break
+    if not matched: return false
+  for pat in walkOptC.excludeFile:
+    if filename.match(pat): return false
+  result = true
+
+proc hasRightDirectory(path: string, walkOptC: WalkOptComp[Pattern]): bool =
+  let dirname = path.lastPathPart
+  for pat in walkOptC.excludeDir:
+    if dirname.match(pat): return false
+  result = true
+
+iterator walkDirBasic(dir: string, walkOptC: WalkOptComp[Pattern]): string =
   var dirStack = @[dir]  # stack of directories
   var timeFiles = newSeq[(times.Time, string)]()
   while dirStack.len > 0:
@@ -585,17 +636,17 @@ iterator walkDirBasic(dir: string, wopt: WalkOptPat): string =
     for kind, path in walkDir(d):
       case kind
       of pcFile:
-        if path.hasRightFileName(wopt):
+        if path.hasRightFileName(walkOptC):
           files.add(path)
       of pcLinkToFile:
-        if optFollow in options and path.hasRightFileName(wopt):
+        if optFollow in options and path.hasRightFileName(walkOptC):
           files.add(path)
       of pcDir:
-        if optRecursive in options and path.hasRightDirectory(wopt):
+        if optRecursive in options and path.hasRightDirectory(walkOptC):
           dirs.add path
       of pcLinkToDir:
         if optFollow in options and optRecursive in options and
-           path.hasRightDirectory(wopt):
+           path.hasRightDirectory(walkOptC):
           dirs.add path
     if sortTime:  # sort by time - collect files before yielding
       for file in files:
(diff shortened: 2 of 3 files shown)


a-mr:
The compilation failure with nim cpp test on MacOS is related to using terminalWidth:
https://dev.azure.com/nim-lang/Nim/_build/results?buildId=8706&view=logs&j=a0440cd6-2060-5545-8b53-639e777de0c6&t=e8126762-32e4-52a6-97de-95cf0bedbe3d&l=2144
```
Undefined symbols for architecture x86_64:
  "ctermid(char*)", referenced from:
      terminalWidth__s9bdJLFVciGfQ0i2sQaMF1g() in stdlib_terminal.nim.cpp.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
Error: execution of an external program failed: 'clang++   -o /Users/runner/work/1/s/tools/nimgrep  /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_assertions.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_locks.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_sharedlist.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_io.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_system.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_parseutils.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_math.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_algorithm.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_unicode.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_strutils.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_pathnorm.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_posix.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_times.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_os.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_parseopt.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_pegs.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_pcre.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_rtarrays.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_re.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_strformat.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_terminal.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_cpuinfo.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_osproc.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/stdlib_tables.nim.cpp.o /Users/runner/work/1/s/nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d/@mnimgrep.nim.cpp.o  -lm   -ldl'
```
I cannot reproduce this problem in MacOS Mojave Virtualbox VM. When I run `bin/nim cpp --hints:on -d:testing --nimblePath:tests/deps --debugger:on --nimCache:nimcache/tools/nimgrep_50b398663f4fa8506176c89182a8ee4d --stdout '--hint[Path]:off' '--hint[Processing]:off' tools/nimgrep`, it compiles successfully.

a-mr:
No, adding -lc did not help :-(

a-mr:
reproduced with mac OS X 10.15 Catalina.

a-mr:
I checked the generated code, stdlib_terminal.cpp, and managed to boil the failure down to just 2 lines.

It's either a bug in clang++ compiler or in the headers, the bug appears only when unistd.h is included BEFORE stdio.h (where ctermid is defined)

This compiles:
```cpp
#include <stdio.h>
#include <unistd.h>  // <- may comment this line, it doesn't matter
int main() {
    ctermid(NULL);
}
```

This does not:
```cpp
#include <unistd.h>
#include <stdio.h>
int main() {
    ctermid(NULL);
}
```

image

If I compile with -c and then look into the *.o file by `nm`, then symbol _ctermid is not present indeed, so it's not a linker problem.

tools/nimgrep.nim.cfg
```diff
@@ -1,5 +1,2 @@
-# The GC is stable enough now:
-
-#--gc:none
-
-
+# using markandsweep because of bug https://github.com/nim-lang/Nim/issues/14138
```
Araq: Fine with me but we should really use ``--gc:orc``, let's eat our own dogfood.

Araq:
I only skimmed this patch and hope @timotheecour does reivew it with more care than I did.

Oh and thanks for the awesome work!

testament/categories.nim
```diff
@@ -262,6 +262,8 @@ proc debuggerTests(r: var TResults, cat: Category, options: string) =
   if fileExists("tools/nimgrep.nim"):
     var t = makeTest("tools/nimgrep", options & " --debugger:on", cat)
     t.spec.action = actionCompile
+    # force target to C because of MacOS 10.15 clang++ bug (see #15612)
```
timotheecour: make it more precise using comment link https://github.com/nim-lang/Nim/pull/15612#issuecomment-712471879 ; or even better, file a new bug and refer to it

tools/nimgrep.nim
```diff
@@ -8,15 +8,15 @@
 #
 
 import
-  os, strutils, parseopt, pegs, re, terminal
+  os, strutils, parseopt, pegs, re, terminal, osproc, tables, algorithm, times
 
 const
-  Version = "1.5"
+  Version = "1.6"
```
timotheecour: `1.6.0` (more standard semver)

tools/nimgrep.nim
```diff
@@ -8,15 +8,15 @@
...
   Usage = "nimgrep - Nim Grep Utility Version " & Version & """
 
   (c) 2012 Andreas Rumpf
 Usage:
-  nimgrep [options] [pattern] [replacement] (file/directory)*
+  nimgrep [options] pattern [replacement] (file/directory)*
```
timotheecour: note that pattern would still be optional in some cases eg `nimgrep -v` or `nimgrep -h`

tools/nimgrep.nim
```diff
@@ -50,44 +58,132 @@ Options:
                -b:N   print N lines of leading context before every match
   --context:N, -c:N   print N lines of leading context before every match and
                       N lines of trailing context after it
+  --sortTime          order files by the last modification time -
+       -s[:asc|desc]  - ascending (default: recent files go last) or descending
   --group, -g         group matches by file
   --newLine, -l       display every matching line starting from a new line
+  --limit[:N], -m[:N] limit max width of lines from files by N characters (80)
```
timotheecour: use instead the more intuitive: `--cols:N or by terminal width (--cols:auto).`

tools/nimgrep.nim
```diff
@@ -50,44 +58,132 @@ Options:
...
+  --fit               calculate --limit from terminal width for every line
```
timotheecour: remove in favor of --cols:auto (see https://github.com/nim-lang/Nim/pull/15612/files#r509810340)

tools/nimgrep.nim
```diff
@@ -33,13 +33,21 @@ Options:
   --word, -w          the match should have word boundaries (buggy for pegs!)
   --ignoreCase, -i    be case insensitive
   --ignoreStyle, -y   be style insensitive
+  --nWorkers:N, -n:N  speed up search by N additional workers (threads)
```
timotheecour: instead: `--threads:N, -t:N  speed up search by N additional threads`
(if -t is for --text, then maybe just remove shortcut `-t:N`, users can write their own alias anyways)

tools/nimgrep.nim
```diff
@@ -569,45 +1108,69 @@ for kind, key, val in getopt():
...
     of "newline", "l": newLine = true
     of "oneline": oneline = true
     of "group", "g": oneline = false
+    of "limit", "m":
+      incl(options, optLimitChars)
+      if val == "":
+        limitChar = 80
+      else:
+        limitChar = parseNonNegative(val, key)
+    of "fit":
```
timotheecour: `--fit` doesn't work reliably, eg go to root of nim repo, type:
`$ nimgrep -r -t --fit PSym`

image

that said, it "almost works" so if it's hard to fix in this PR, just file an issue so it can be fixed later.

**EDIT** can `wcwidth` be used and is that the problem?
ideally `wcwidth` should be added to `std/terminal`, see https://man7.org/linux/man-pages/man3/wcwidth.3.html, with best effort (even incorrect) fallback for OS without it

tools/nimgrep.nim
```diff
@@ -50,44 +58,132 @@ Options:
                -b:N   print N lines of leading context before every match
   --context:N, -c:N   print N lines of leading context before every match and
                       N lines of trailing context after it
+  --sortTime          order files by the last modification time -
+       -s[:asc|desc]  - ascending (default: recent files go last) or descending
   --group, -g         group matches by file
   --newLine, -l       display every matching line starting from a new line
+  --limit[:N], -m[:N] limit max width of lines from files by N characters (80)
+  --fit               calculate --limit from terminal width for every line
+  --onlyAscii, -o     use only printable ASCII Latin characters 0x20-0x7E
```
timotheecour: remove `-o` shortcut, it's too often associated with output file (again, users can alias their favorite commands in the shell)

tools/nimgrep.nim
```diff
@@ -50,44 +58,132 @@ Options:
...
+                      (substitutions: 0 -> @, 1-0x1F -> A-_, 0x7F-0xFF -> !)
```
timotheecour: somewhat serious bug: `$ nimgrep x bin/nim` hangs

tools/nimgrep.nim
```diff
@@ -33,13 +33,21 @@ Options:
...
-  --excludeFile:PAT   skip files whose names match the given regex pattern PAT
-  --excludeDir:PAT    skip directories whose names match the given regex PAT
+  --includeFile:PAT   search only files whose names match the given PATttern
+  --excludeFile:PAT   skip files whose names match the given pattern PAT
+  --includeDir:PAT    search only files with full directory name matching PAT
+  --excludeDir:PAT    skip directories whose names match the given pattern PAT
+  --if,--ef,--id,--ed abbreviations of 4 options above
+  --match:PAT         select files containing a (not displayed) match of PAT
+  --noMatch:PAT       select files not containing any match of PAT
+  --bin:yes|no|only   process binary files? (detected by \0 in first 1K bytes)
```
timotheecour: how about an option `--strings` with following behavior:
when current searched file is a binary, pipe it to `strings` (or equivalent) on systems that have it (likely all have it) before searching into it.

this is almost always what you want when searching in binary files; eg: try: `strings bin/nim` and look at the output.

tools/nimgrep.nim
```diff
@@ -223,242 +324,357 @@ proc writeArrow(s: string) =
...
+proc printRaw(c: char, curCol: var Column, allowTabs = true) =
+  # print taking into account tabs and optOnlyAscii
+  if c == '\t':
+    if allowTabs:
+      let spaces = 8 - (curCol.file mod 8)
+      curCol.file += spaces
+      curCol.terminal += spaces
+      if optOnlyAscii:
+        printSpecial " "
+        stdout.write " ".repeat(spaces-1)
```
timotheecour: `stdout.write` (which results in a syscall) should be much slower than using your own line buffer (a single `buffer: string` that you reuse and reset) that you flush after each new line

tools/nimgrep.nim
```diff
@@ -223,242 +324,357 @@ proc writeArrow(s: string) =
...
+      if optOnlyAscii:
+        printSpecial " "
+      else:
+        stdout.write " "
+  elif not optOnlyAscii or (0x20 <= int(c) and int(c) <= 0x7e):
+    stdout.write c
+    curCol.file += 1
+    curCol.terminal += 1
+  else:  # substitute characters that are not ACSII Latin
+    let substitute =
```
timotheecour: I don't understand this, it results in hard to interpret ascii being written; why not simply use `repr` for those characters?

eg look at output of:
`for i in 0..<255: echo i, ":", repr(i.char)`

tools/nimgrep.nim
timotheecour: use `on|off|only` simply for consistency with nim command line options and standard nim tools
(and update comment in `process only text files, the same as --bin:no`)

tools/nimgrep.nim
```diff
@@ -494,28 +710,342 @@ proc styleInsensitive(s: string): string =
...
+      files.sort()
+      for file in files:
+        yield file
+      dirs.sort(order = SortOrder.Descending)
+    for dir in dirs:
+      dirStack.add(dir)
+  if sortTime:
+    timeFiles.sort(sortTimeOrder)
+    for (_, file) in timeFiles:
+      yield file
```
timotheecour, line 1041: multiple yield in an inline iterator causes code bloat, see https://nim-lang.github.io/Nim/manual.html#iterators-and-the-for-statement-first-class-iterators

This would be avoided (and this code would be simplified a lot) using glob PR https://github.com/nim-lang/Nim/pull/15598 which takes care of those details in a generic way /cc @araq

tools/nimgrep.nim
```diff
@@ -494,28 +710,342 @@ proc styleInsensitive(s: string): string =
...
+iterator walkRec(paths: seq[string]): (string, string) =
+  declareCompiledPatterns(walkOptC, WalkOptComp):
+    walkOptC.excludeFile.add walkOpt.excludeFile.compileArray()
+    walkOptC.includeFile.add walkOpt.includeFile.compileArray()
+    walkOptC.includeDir.add  walkOpt.includeDir.compileArray()
+    walkOptC.excludeDir.add  walkOpt.excludeDir.compileArray()
+    for path in paths:
+      if dirExists(path):
+        for p in walkDirBasic(path, walkOptC):
+          yield ("", p)
```
timotheecour, line 1052: same comment as https://github.com/nim-lang/Nim/pull/15612/files#r509826384

tools/nimgrep.nim
```diff
@@ -494,28 +710,342 @@ proc styleInsensitive(s: string): string =
...
+      for file in files:
+        yield file
+      dirs.sort(order = SortOrder.Descending)
+    for dir in dirs:
+      dirStack.add(dir)
+  if sortTime:
+    timeFiles.sort(sortTimeOrder)
+    for (_, file) in timeFiles:
+      yield file
+
```
timotheecour, line 1042: performance seems bad, not exactly sure whether:
with -d:danger --gc:arc:
nimgrep -r --filenames . | wc -l
(63723 files)
5.5 seconds

with https://github.com/nim-lang/Nim/pull/15598:
```
import std/[globs,os,sugar]
for ai in glob(".", relative = true, sortCmp = (a,b) => cmp(a.path, b.path)):
  if ai.kind == pcFile: echo ai.path
```
0.22 seconds

echo 'echo 5.5 / 0.22' | nim r -
=> 25 slower

tools/nimgrep.nim
```diff
@@ -569,45 +1108,69 @@ for kind, key, val in getopt():
...
+      else:
+        nWorkers = parseNonNegative(val, key)
+    of "ext": walkOpt.extensions.add val.split('|')
+    of "noext", "no-ext": walkOpt.skipExtensions.add val.split('|')
+    of "excludedir", "exclude-dir",   "ed": walkOpt.excludeDir.add val
+    of "includedir", "include-dir",   "id": walkOpt.includeDir.add val
+    of "includefile", "include-file", "if": walkOpt.includeFile.add val
+    of "excludefile", "exclude-file", "ef": walkOpt.excludeFile.add val
+    of "match": searchOpt.checkMatch = val
+    of "nomatch", "notmatch", "not-match", "no-match":
```
timotheecour: we don't need all those aliases. Just handle `--nomatch:pattern` and `--match:pattern`

tools/nimgrep.nim
```diff
@@ -50,44 +58,132 @@ Options:
                -b:N   print N lines of leading context before every match
   --context:N, -c:N   print N lines of leading context before every match and
                       N lines of trailing context after it
+  --sortTime          order files by the last modification time -
```
timotheecour: This needs more thought.

I can see how we'll quickly end up with a combinatorial explosion of flags and increased complexity / bugs. Instead we should use separation of concern (file listing vs file content matching) to reduce flags and complexity and increase flexibility as follows:

* support reading input files from stdin

This is an essential feature of `ripgrep` (arguably the best file searcher available eg see https://blog.burntsushi.net/ripgrep/ https://github.com/BurntSushi/ripgrep), eg:

```
find . | rg foo
find . -depth 2 | sort | rg foo
```
etc.

This allows pipelining and enables more flexibility than if you were to cram everything in a single program invocation

tools/nimgrep.nim
a-mr: I believe it's OK, it does not result in a syscall. write(File, string) in io.nim:244 call writeBuffer on :199, which calls fwrite. fwrite from C is already buffered. To flush one needs to call stdout.flushFile

tools/nimgrep.nim
a-mr: it's more concise, just one char => also more space will be available in --cols:auto. repr(i.char) contains 1,2, or 3 chars.

I think there is no actually demand to interpret this, rather just highlight the presence of a non-printable symbol.

BTW vim also print such characters in this form, though it adds ^, so '\12' in it is represented as ^L (in blue).

tools/nimgrep.nim
a-mr: I cannot reproduce :-/

What's your OS? Could you compile with --gc:markandsweep (I don't quite trust my recent change to --gc:orc)?

tools/nimgrep.nim
a-mr: Yes, its presence is questionable, I agree. And it's not a Unix style.

But telling people to learn Shell is also controversial advice.

Sorting by time from Shell is not really that easy:
`find . -type f -printf "%T@ %p\n" | sort -n | cut -f2 -d' '`
I think that support reading from stdin is  a really good idea, but having --sortTime at hand would still be good. I do not agree that our instruments necessarily have to be "orthogonal" or minimal.

timotheecour: > Sorting by time from Shell is not really that easy:

I agree, all I'm arguing for here is composability via pipeline `|`

I'd be happy if you support this:

`nimgrep -r --sortTime --filenames . | nimgrep mypattern`

(the first part of the pipeline can be anything of course)
that way, we can make common useful things (eg Sorting by time) easy while still retaining full flexibility of pipelines

tools/nimgrep.nim
timotheecour: repro:
I'm on OSX, using iterm
```
XDG_CONFIG_HOME= $nim_140_X c tools/nimgrep.nim # (XDG_CONFIG_HOME to isolate from my environment)
tools/nimgrep x $nim_140_X
```

adding `--gc:markandsweep` or `--gc:arc` makes no difference

* this hangs: `tools/nimgrep x $nim_140_X`
* this does not hang: `tools/nimgrep x $nim_140_X | wc -l`
* this does not hang: `tools/nimgrep --onlyAscii $nim_140_X | wc -l`

this suggests that the culprit is binary data being printed to the terminal are being interpreted by the terminal, causing arbitrary behavior including hanging.

This should not be the default behavior (printing binary data to terminal can have potentially dangerous / malicious consequences), so I suggest making `--onlyAscii:on` the default, and users that know what they're doing can us explicit `--onlyAscii:off` to disable

We can bikeshed the `--onlyAscii` name separately

tools/nimgrep.nim
timotheecour: fwrite still incurs overhead IIRC; could this explain (parts of) the 25x slowdown in https://github.com/nim-lang/Nim/pull/15612#discussion_r509846344 ?

tools/nimgrep.nim
timotheecour: > it's more concise, just one char

but you're collapsing a range of chars to just `!`, that's not good; if I want to print binary as ascii, I want to be able to interpret it, otherwise I'd just ask to skip.

see also how less renders binary data: (consecutive char's):
```
^@^A^B^C^D^E^F^G^H
^K^L^M^N^O^P^Q^R^S^T^U^V^W^X^Y^Z\^]^^^_ !"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~^?<80><81><82><83><84><85><86><87><88><89><8A><8B><8C><8D><8E><8F><90><91><92><93><94><95><96><97><98><99><9A><9B><9C><9D><9E><9F>
<A0><A1><A2><A3><A4><A5><A6><A7><A8><A9><AA><AB><AC><AD><AE><AF><B0><B1><B2><B3><B4><B5><B6><B7><B8><B9><BA><BB><BC><BD><BE><BF><C0><C1><C2><C3><C4><C5><C6><C7><C8><C9><CA><CB><CC><CD><CE><CF><D0><D1><D2><D3><D4><D5><D6><D7><D8><D9><DA><DB><DC><DD><DE><DF><E0><E1>
<E2><E3><E4><E5><E6><E7><E8><E9><EA><EB><EC><ED><EE><EF><F0><F1><F2><F3><F4><F5><F6><F7><F8><F9><FA><FB><FC><FD><FE><FF>^@^A^B^
```
more readable/interpretable than your current scheme.

 and shorter than `repr`


how about following:
```nim
import sugar, strutils
proc render(result: var string, c: char) =
  result.add '\\' & c.BiggestUInt.toHex(2)
doAssert "".dup(render(243.char)) == r"\F3"
```
which is 1 char shorter that repr (`\F3` instead of `\243`)

tools/nimgrep.nim
timotheecour: EDIT: indeed, see the stackoverflow answer I just posted in https://stackoverflow.com/a/64492448/1426932 which demonstrates this

tools/nimgrep.nim
a-mr: You'd better use command `bin/nimgrep -r --filenames '.*' | wc -l`, in my case it improved speed by factor **11x** w.r.t. to matching by '.'

The problem here is that the first positional argument that you pass to nimgrep is actually your pattern (not a directory to search in).
So you are essentially matching any char in a file name and then passing every char one by one to the printing proc.

You may have been confused because it prints every filename on one line. It's because (starting from nimgrep improvements 1.5) nimgrep prints matches in a smart way, displaying all matches in one line. For example, when you match file 'abc' by '.' in old nimgrep 1.4 it will show:
<ins>a</ins>bc
a<ins>b</ins>c
ab<ins>c</ins>
While nimgrep 1.5/1.6 shows:
<ins>abc</ins>
in both cases of matching by '.' and '.*'
But in the first case the chars are passed one-by-one.

tools/nimgrep.nim
```diff
@@ -33,13 +33,21 @@ Options:
...
   --ignoreStyle, -y   be style insensitive
+  --nWorkers:N, -n:N  speed up search by N additional workers (threads)
   --ext:EX1|EX2|...   only search the files with the given extension(s),
                       empty one ("--ext") means files with missing extension
   --noExt:EX1|...     exclude files having given extension(s), use empty one to
                       skip files with no extension (like some binary files are)
-  --includeFile:PAT   include only files whose names match the given regex PAT
-  --excludeFile:PAT   skip files whose names match the given regex pattern PAT
-  --excludeDir:PAT    skip directories whose names match the given regex PAT
+  --includeFile:PAT   search only files whose names match the given PATttern
```
timotheecour: you changed regex PAT to PATttern:
* shouldn't that be PATtern ?
* not clear what kind of patterns are accepted; regex? peg? unix glob?

testament/categories.nim
a-mr: Site https://bugreport.apple.com/ does not open for me.
I left feedback in https://feedbackassistant.apple.com/feedback/8825785, but I'm not sure whether it's publicly available or not?
So I will insert the comment link.

tools/nimgrep.nim
a-mr: I've run your test on my Linux Debian 10 and I see completely different results:
case 1 — 3.4 s
case 2 — 5.4 s
So using in-app buffering actually decreased speed of the program significantly!
My first guess is that fwrite is implemented inefficiently in OSX or may be buffer size is too small.

tools/nimgrep.nim
a-mr: yes, limiting was not implemented for matches itself, only for surrounding context. I'm working on it...

> `wcwidth`

If you mean unicode support, then currently it is not implemented, but I'm trying to do it by using nim's `unicode` module. AFAIK `wcwidth` is for wide 16-bit/32-bit characters that are not recommended even on Windows now, so currently the reasonable option for all the platforms is UTF-8, which exactly what `unicode` does. I think `wcwidth` can be regarded as de-facto deprecated (may be it'll revive in future if fashion changes).

timotheecour: see https://github.com/nim-lang/RFCs/issues/279

> I'm trying to do it by using nim's unicode module
> the reasonable option for all the platforms is UTF-8, which exactly what unicode does

I don't see anything in std/unicode that does this

tools/nimgrep.nim
a-mr: Done.

tools/nimgrep.nim
a-mr: I chose --threads and -j, as in ripgrep

tools/nimgrep.nim
a-mr: makes sense, thank you.

tools/nimgrep.nim
a-mr: yes, I've finally understood what you mean.
However, fully supporting unicode is a bit too difficult then, I gave up the idea of that at the moment.
Adding unicode will be left for future.

In the current version cropping Ascii matches is supported well.

tools/nimgrep.nim
a-mr: In updated version version in-app buffering is implemented (though mainly for different purpose), it may increase the speed in your case.

tools/nimgrep.nim
a-mr: OK, I chose representation like ^@ and 'FF for 0xFF. I hope it doesn't seem too strange.

tools/nimgrep.nim
a-mr: Now both regex and peg patterns are accepted.
I'm not sure what's the right place to describe this, place for --help should be restricted. Manual page then? I'm pondering about adding additional documentation switch like --tutorial or --examples where some additional information would be provided, especially examples of usage.

tools/nimgrep.nim
a-mr: Agree. However I really need such shortcuts, so I added -@. I hope it's not too weird.

tools/nimgrep.nim
timotheecour: > Now both regex and peg patterns are accepted.

great.
Adding just line in the `--help` would remove a lot of ambiguities as to what PAT represents.

> Manual page then? 

Further details can be found in `nim doc nimgrep` output, as is standard.
We can in a future PR add a link to `nim doc nimgrep` output alongside `fusion docs`, `compiler docs`, maybe inside a new `tools docs`, in https://nim-lang.github.io/Nim/lib.html

tools/nimgrep.nim
```diff
@@ -62,10 +62,11 @@ Options:
        -s[:asc|desc]  - ascending (default: recent files go last) or descending
   --group, -g         group matches by file
   --newLine, -l       display every matching line starting from a new line
-  --limit[:N], -m[:N] limit max width of lines from files by N characters (80)
-  --fit               calculate --limit from terminal width for every line
-  --onlyAscii, -o     use only printable ASCII Latin characters 0x20-0x7E
-                      (substitutions: 0 -> @, 1-0x1F -> A-_, 0x7F-0xFF -> !)
+  --cols[:N]          limit max width of lines from files by N characters (off)
+  --cols:auto, -%     calculate columns from terminal width for every line
```
timotheecour, line 107: the problem with abbreviations like `-%` and `-@` is they're cryptic, hard to search, non standard/self documenting (and -%` might cause escaping problems on some shells, not sure)

same comment as araq's comment in my PR: https://github.com/nim-lang/Nim/pull/15687#discussion_r511829251
> Please only introduce the long version, I cannot remember all these single char abbreviations anymore.

It's easy enough to alias your own commands privately in your shell with your favorite options.

That said, this isn't a blocker in case you insist on keeping those.

tools/nimgrep.nim
```diff
@@ -62,10 +62,11 @@ Options:
...
+  --onlyAscii, -@     use only printable ASCII Latin characters 0x20-0x7E
```
timotheecour: optional nit: rename to `--ascii` ?

tools/nimgrep.nim
```diff
@@ -50,44 +58,135 @@ Options:
                -b:N   print N lines of leading context before every match
   --context:N, -c:N   print N lines of leading context before every match and
                       N lines of trailing context after it
+  --sortTime          order files by the last modification time -
+       -s[:asc|desc]  - ascending (default: recent files go last) or descending
   --group, -g         group matches by file
   --newLine, -l       display every matching line starting from a new line
+  --cols[:N]          limit max width of lines from files by N characters (off)
```
timotheecour: can be done in future work but `ripgrep` can also show column (of match), not just line; it makes it easier when jump to line+col in an editor or on terminal

tools/nimgrep.nim
a-mr: Reading from stdin/pipe is implemented.
Now it's possible to do almost as you expected:
```
nimgrep -r --sortTime --filenames | nimgrep mypattern -
```
The only difference is that nimgrep reads files from current directory by default (depth 1). So to avoid a breaking change, reading from pipe requires `-`.

tools/nimgrep.nim
a-mr: ok, I left only --nomatch.

tools/nimgrep.nim
a-mr: I expanded --help and structured it better. Hopefully it's much clearer now.

tools/nimgrep.nim
a-mr: those cases are too marginal and obvious to mention them in main Usage. In new expanded version of --help Usage is split to 3 basic variants (search, replace, filenames). Hopefully what's required and what's not is clearer now.

tools/nimgrep.nim
```diff
@@ -12,64 +12,106 @@ import
...
+Positional arguments, from left to right:
+* PATERN is either Regex (default) or Peg if --peg is specified.
+  PATTERN and REPLACEMENT should be skipped when --stdin is specified.
+* REPLACEMENT supports $1, $# notations for captured groups in PATTERN.
+  Note: --replace mode DOES NOT ask confirmation unless --confirm is specified!
+* Final arguments are a list of paths (FILE/DIRECTORY) or a standalone
+  minus '-' (pipe) or not specified (empty). Note for the empty case: when
+  no FILE/DIRECTORY/- is specified nimgrep DOES NOT read the pipe, but
+  searches files in the current dir instead!
+    -                   read buffer once from stdin: pipe or terminal input;
```
timotheecour: sounds like you're reading all of stdin before starting to process? if so that doesn't seem good, the whole point is streaming

tools/nimgrep.nim
```diff
@@ -12,64 +12,106 @@ import
...
+* To list file names:
+  nimgrep [options] --filenames [PATTERN] [(FILE/DIRECTORY)*]
+
+Positional arguments, from left to right:
+* PATERN is either Regex (default) or Peg if --peg is specified.
+  PATTERN and REPLACEMENT should be skipped when --stdin is specified.
+* REPLACEMENT supports $1, $# notations for captured groups in PATTERN.
+  Note: --replace mode DOES NOT ask confirmation unless --confirm is specified!
+* Final arguments are a list of paths (FILE/DIRECTORY) or a standalone
+  minus '-' (pipe) or not specified (empty). Note for the empty case: when
```
timotheecour, line 34: add note: incompatible with `--stdin` (I'm assuming?)

tools/nimgrep.nim
```diff
@@ -12,64 +12,106 @@ import
...
+  no FILE/DIRECTORY/- is specified nimgrep DOES NOT read the pipe, but
+  searches files in the current dir instead!
+    -                   read buffer once from stdin: pipe or terminal input;
+                        in --replace mode the result is directed to stdout.
+    (empty)             current directory '.' is assumed
```
timotheecour: seems dangerous:

`nimgrep foo --replace bar **/*.js` => when glob `**/*.js` returns nothing by the shell, it'd behave the same as `nimgrep foo --replace bar .` which has very different semantics?

tools/nimgrep.nim
```diff
@@ -12,64 +12,106 @@ import
...
+* Filter file content:
   --match:PAT         select files containing a (not displayed) match of PAT
   --noMatch:PAT       select files not containing any match of PAT
   --bin:on|off|only   process binary files? (detected by \0 in first 1K bytes)
+                      (default: on - binary and text files treated the same way)
   --text, -t          process only text files, the same as --bin:off
-  --count             only print counts of matches for files that matched
+
+* Represent results:
   --nocolor           output will be given without any colours
```
timotheecour: s/colours/colors/ (cf recent forum post on preferring US english)

tools/nimgrep.nim
```diff
@@ -115,20 +157,23 @@ type
                     lineBeg: int, lineEnd: int, match: string]
   outputKind = enum
     OpenError, Rejected, JustCount,
-    BlockFirstMatch, BlockNextMatch, BlockEnd, FileContents
+    BlockFirstMatch, BlockNextMatch, BlockEnd, FileContents, FileName
```
timotheecour: leave as a TODO for future PR but style is wrong, should be: `OutputKind = enum openError, ...`

tools/nimgrep.nim
```diff
@@ -115,20 +157,23 @@ type
...
+    of JustCount: matches: int          # the only output for option --count
+    of BlockFirstMatch, BlockNextMatch: # the normal case: match itself
       pre: string
       match: MatchInfo
-    of BlockEnd:
+    of BlockEnd:                        # block ending right after prev. match
       blockEnding: string
-      firstLine: int       # = last lineNo of last match
-    of FileContents:
+      firstLine: int
```
timotheecour, line 185: 0 or 1 based indexing?

tools/nimgrep.nim
```diff
@@ -851,6 +897,8 @@ iterator processFile(searchOptC: SearchOptComp[Pattern], filename: string,
   var error = false
   if optFilenames in options:
     buffer = filename
+  elif optPipe in options:
+    buffer = stdin.readAll()
```
timotheecour, line 914: that's the thing I was complaining about above; one of the main points of `stdin` is to allow streaming, eg so that
* you don't need to hold in memory the whole input (eg with `find / | nimgrep pat`)
* you can start processing right away

`ripgrep` does the right thing here, see `find / | rg foobar`

tools/nimgrep.nim
```diff
@@ -1298,28 +1361,42 @@ checkOptions({optCount, optReplace}, "count", "replace")
...
   if gVar.errors != 0:
     printError $gVar.errors & " errors"
-  printBold($gVar.matches & " matches")
-  stdout.write("\n")
+  if searchOpt.pattern != "":
+    # PATTERN allowed to be empty if --filenames is given
+    printBold($gVar.matches & " matches")
+    stdout.write("\n")
   if gVar.errors != 0:
     quit(1)
```
timotheecour, line 1416: unrelated TODO for future PR: `reportError`?

tools/nimgrep.nim
timotheecour: it is, thanks!

tools/nimgrep.nim
a-mr: yes, that's why replace made a special case: it requires explicit specification of all 3 positional arguments (and that is written right 3 lines below), so in this case nimgrep will report an error.

tools/nimgrep.nim
a-mr: It's incompatible with multi-line patterns.
So we need to add option for switching to one-line patterns to go to streaming mode.
And it's incompatible with options like context.
BTW when GNU grep switches to PCRE (for multi-line patterns), it also starts to read buffer by whole.

What's so great about the streaming mode? Memory consumption, more parallel execution, ...?

tools/nimgrep.nim
a-mr: AFAIK line numbers are always numbered from 1 in all text editors, so I assumed that it goes without saying

tools/nimgrep.nim
a-mr: I'm going to leave -@ and -% for brevity. I really use these ones lately. There are often cases when you need to work with nimgrep almost interactively: when previous search gave too huge results, you go to previous cmdline and add -@% option in 4 key presses.

tools/nimgrep.nim
a-mr: yes, added to limitations section

tools/nimgrep.nim
a-mr: done in this PR, thank you!

tools/nimgrep.nim
a-mr: added to limitations.

tools/nimgrep.nim
a-mr: it reports error 4 lines above. It's done this way for matches to always be last:
```
<M> errors 
<N> matches
```

tools/nimgrep.nim
a-mr: right, added this to --help

tools/nimgrep.nim
a-mr: added to TODO

tools/nimgrep.nim
a-mr: all fair, but I left them, see my comment below.

tools/nimgrep.nim
a-mr: added to TODO

a-mr:
@timotheecour , thank you for great review!

I think it's finished.
There are 2 problems left unaddressed:
* hang with MacOS iterm. I guess it's because of iterm. Unfortunately I did not have time to look into that.
* code bloating because of iterators with multiple yields. I tried to tackle that one, but mostly failed because rewriting in one-yield style requires to obfuscate the logic to some extent. (I would rather change to 1st-class iterators, but attempt to do so caused compiler's internal errors and segfaults - I will try to make a minimal reproducer and report a bug.) For now I consider the problem as not critical: nimgrep size increases from 500KB to 800KB. However there is a more annoying problem of drastic increase in compilation time. Let us address this problem in future work.

tools/nimgrep.nim
timotheecour: `reading from pipe requires -` => good, explicit is better than implicit

timotheecour:
LGTM, thanks for your patience!

> code bloating because of iterators with multiple yields

ya, leave it as is; hopefully https://github.com/nim-lang/fusion/pull/32 gets merged soon and then you can reuse it, simplifying code and avoiding multiple yield in the process, among other benefits.

> hang with MacOS iterm

ok to leave as TODO; for future reference: https://github.com/nim-lang/Nim/pull/15612#discussion_r509814737

I'd also like something like https://github.com/BurntSushi/ripgrep/issues/1727 but that can be done in future work

timotheecour (approved):

PR #15625, opened 2020-10-18
Tables, use sink val arguments more actively

cooldome:
Make tables module more usable for non-copyable types


cooldome:
Checked, test failure of glob package is unrelated. glob package doesn't use tables module.
sempass2 changesa are fixing that notNil checks are not handling tyVar, tySink

PR #15626, opened 2020-10-18
Fixes compilation for --os:any + --exception:setjmp

dom96:


Araq:
This is "nim cpp" specific code. Do you use "nim cpp --os:any --exceptions:setjmp" ?

dom96:
yep

PR #15632, opened 2020-10-19
fix #15631

### LINKED ISSUE #15631: Testament does not check memory leaks with Valgrind
[Testament docs](https://github.com/nim-lang/Nim/blob/b16b2eb06fa09b22ab50b3bd2703fe79c0ca4fc6/doc/testament.rst) claim the following:
```nim
valgrind: false   # Can use Valgrind to check for memory leaks, or not (Linux 64Bit only).
```

This is incorrect. Testament tests do not check for memory leaks even when `valgrind: true` is set.

### Example
```nim
# tests/leak.nim
discard """
  valgrind: true
  cmd: "nim $target --gc:arc -d:useMalloc $options $file"
"""
var mem: pointer = cast[pointer](alloc(sizeof(int)))
mem = nil
```

### Current Output
`leak.nim` definitely leaks memory, but `testament` does not catch it.

```console
user@user:~$ testament run leak.nim
PASS: tests/leak.nim C                                             ( 0.61 sec)
```

### Possible Solution

The reason is that `valgrind` returns an exit code of 0 even if memory leaks are detected unless `--leak-check=yes|full` is set. Evidence:
```console
user@user:~$ nim compile --gc:arc -d:useMalloc leak.nim 
Hint: used config file '/home/user/nim-1.4.0/config/nim.cfg' [Conf]
Hint: used config file '/home/user/nim-1.4.0/config/config.nims' [Conf]
....
Hint:  [Link]
Hint: 19076 lines; 0.165s; 16.574MiB peakmem; Debug build; proj: /home/user/leak.nim; out: /home/user/leak [SuccessX]
user@user:~$ valgrind --error-exitcode=1 ./leak
==3525== Memcheck, a memory error detector
==3525== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==3525== Using Valgrind-3.14.0 and LibVEX; rerun with -h for copyright info
==3525== Command: ./leak
==3525== 
==3525== 
==3525== HEAP SUMMARY:
==3525==     in use at exit: 8 bytes in 1 blocks
==3525==   total heap usage: 1 allocs, 0 frees, 8 bytes allocated
==3525== 
==3525== LEAK SUMMARY:
==3525==    definitely lost: 8 bytes in 1 blocks
==3525==    indirectly lost: 0 bytes in 0 blocks
==3525==      possibly lost: 0 bytes in 0 blocks
==3525==    still reachable: 0 bytes in 0 blocks
==3525==         suppressed: 0 bytes in 0 blocks
==3525== Rerun with --leak-check=full to see details of leaked memory
==3525== 
==3525== For counts of detected and suppressed errors, rerun with: -v
==3525== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
user@user:~$ echo $?
0
```

With `--leak-check=yes|full` set, Valgrind returns a non-zero exit code.
```console
user@user:~$ valgrind --error-exitcode=1 --leak-check=yes ./leak
==5155== Memcheck, a memory error detector
==5155== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==5155== Using Valgrind-3.14.0 and LibVEX; rerun with -h for copyright info
==5155== Command: ./leak
==5155== 
==5155== 
==5155== HEAP SUMMARY:
==5155==     in use at exit: 8 bytes in 1 blocks
==5155==   total heap usage: 1 allocs, 0 frees, 8 bytes allocated
==5155== 
==5155== 8 bytes in 1 blocks are definitely lost in loss record 1 of 1
==5155==    at 0x483577F: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==5155==    by 0x10D04B: allocImpl__KzdpcuLT9aef9bsiSHlIu9aFg_3 (in /home/user/leak)
==5155==    by 0x10D8BD: NimMainModule (in /home/user/leak)
==5155==    by 0x10D7F0: NimMainInner (in /home/user/leak)
==5155==    by 0x10D81D: NimMain (in /home/user/leak)
==5155==    by 0x10D857: main (in /home/user/leak)
==5155== 
==5155== LEAK SUMMARY:
==5155==    definitely lost: 8 bytes in 1 blocks
==5155==    indirectly lost: 0 bytes in 0 blocks
==5155==      possibly lost: 0 bytes in 0 blocks
==5155==    still reachable: 0 bytes in 0 blocks
==5155==         suppressed: 0 bytes in 0 blocks
==5155== 
==5155== For counts of detected and suppressed errors, rerun with: -v
==5155== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 0 from 0)
user@user:~$ echo $?
1
```

Currently, Testament only sets [`--error-exitcode=1`](https://github.com/nim-lang/Nim/blob/78cbb52fa01d6c73307fe276bdf7d4c34d7a556c/testament/testament.nim#L505). 

### Additional Information
```
user@user:~$ nim -v
Compiled at 2020-10-16

```

n5m:
fix #15631


Clyybber (commented):
The tests without `-d:useMalloc` should be removed as they only really test that we *don't* use malloc or that valgrind can only detect leaks via malloc (not something we should test or rely on); we also don't need seperate tests for orc/arc in this case IMO.

n5m:
What is the best way to get the Valgrind tests to pass on non-Linux hosts in the pipelines?

One way could be to rewrite the tests to look like
```nim
discard """
   # Testament configs
   # Testament configs
   # Testament configs
"""
when defined(linux) and sizeof(int) == 8:
  # original test code
  discard alloc(1)
else:
  # simulate the expected failed valgrind output
  echo "   definitely lost: 7 bytes in 2 blocks"
  # exit with the same code as --error-exitcode=1
  quit(1)
```
which follows the example set here:
https://github.com/nim-lang/Nim/blob/d6160ef9e4e43a53c54b21d62f9b3115b87cf610/testament/specs.nim#L308
and fakes the output expected in test cases like
https://github.com/nim-lang/Nim/blob/d6160ef9e4e43a53c54b21d62f9b3115b87cf610/tests/valgrind/tleak_arc_useMalloc.nim#L5
but this does not seem like a great solution.

n5m:
Oh, this looks like it will be much better:
https://github.com/nim-lang/Nim/blob/d6160ef9e4e43a53c54b21d62f9b3115b87cf610/testament/specs.nim#L317-L325
https://github.com/nim-lang/Nim/blob/d6160ef9e4e43a53c54b21d62f9b3115b87cf610/testament/specs.nim#L339-L341

tests/valgrind/tleak_arc.nim
```diff
@@ -0,0 +1,14 @@
+discard """
+valgrind: true
+cmd: "nim $target --gc:arc -d:useMalloc $options $file"
+exitcode: 1
+outputsub: "   definitely lost: 7 bytes in 2 blocks"
+disabled: "freebsd"
+disabled: "macosx"
+disabled: "openbsd"
+disabled: "windows"
+disabled: "32bit"
```
alaviss, line 10: You don't need that:

https://github.com/nim-lang/Nim/blob/9a110047cbe2826b1d4afe63e3a1f5a08422b73f/testament/specs.nim#L307-L316

Clyybber: We actually do in this case, since otherwise the test will be run without valgrind on these platforms and thus we won't get the error message/exit code.

Clyybber:
`tests/arc/tasyncorc.nim` and `tests/arc/thavlak_orc_stress.nim` leak :)

tests/valgrind/tleak_arc.nim
alaviss: I think we should add a "valgrind: leak" attribute.

Clyybber: Doesn't help here, but had to be done for the failing tests anyways.

Clyybber:
Sorry for force-pushing to your branch @n5m , I thought that a rebase would solve the errors.
I now added a `leaks` switch to the valgrind option to accomodate for the two failing tests.
(Test failures unrelated)

n5m:
Quite all right, @Clyybber. Happy to have helped with this PR.

I am interested, however, in seeing [tvalgrind.nim](https://github.com/nim-lang/Nim/pull/15632/commits/a2d4245d1494e65707926ea922e6cfd2f1539920#diff-7fbf4dcb13a5825032238f284b2242dbd7ffb2c15ac708683c4ade9403e9c633L1) added back into the codebase. My rationale is that while [tleak_arc.nim](https://github.com/n5m/Nim/blob/13b9699a501f7c94f8b2e9317c9d6d687266196d/tests/valgrind/tleak_arc.nim) tests that memory leaks are found, tvalgrind.nim tests Testament, *itself*. It also nicely parallels the remaining tests in `testament/tests/shouldfail/`, which test Testament's other configuration options.

Would you be opposed to a tvalgrind.nim like this?
```nim
discard """
valgrind: enabled
cmd: "nim $target --gc:arc -d:useMalloc $options $file"
disabled: "freebsd"
disabled: "macosx"
disabled: "openbsd"
disabled: "windows"
disabled: "32bit"
"""

when defined(linux) and sizeof(int) == 8:
  # test that a memory leak is caught by valgrind, but valgrind only
  # runs on 64-bit Linux machines...
  discard alloc(1)
else:
  # ...so on all other OS/architectures, simulate the same exit code
  # that valgrind would have thrown for this test case
  quit(1) # chosen from --error-exit=1
```

Clyybber:
@n5m Oh, yeah that looks fine.

PR #15633, opened 2020-10-19
Add tests to #15363

### LINKED ISSUE #15363: map, mapIt, filter cause a segfault at compile time in a static block when passed an inline function
### Example
```nim
import sequtils

func identity(a: bool): bool = a
var a: seq[bool] = static:
    newSeq[bool](0).mapIt(it) # segfaults
var a: seq[bool] = static:
    newSeq[bool](0).filterIt(it) # does not segfault
var b: seq[bool] = static:
    newSeq[bool](0).map(identity) # does not segfault
var c: seq[bool] = static:
    newSeq[bool](0).map(proc (a: bool): bool = false) # segfaults
var d: seq[bool] = static:
    newSeq[bool](0).filter(identity) # does not segfault
var e: seq[bool] = static:
    newSeq[bool](0).filter(proc (a: bool): bool = false) # segfaults
echo k
```

### Current Output
```
```
at compile time, in cases with the `#segfaults` comment

### Expected Output
```
@[]
```
at runtime
## Addtional Information
 filterIt does not seem to have the problem
## Version
```
PS> nim -v
Nim Compiler Version 1.2.6 [Windows: amd64]
Compiled at 2020-07-29

git hash: bf320ed172f74f60fd274338e82bdc9ce3520dd9
```

ghost:
Closes #15363

### DIFF
--- a/tests/vm/tvmmisc.nim
+++ b/tests/vm/tvmmisc.nim
@@ -204,3 +204,29 @@ block: # bug #13081
   doAssert j1.x1 == 12
   static:
     doAssert j1.x1 == 12
+
+# bug #15363
+import sequtils
+
+block:
+  func identity(a: bool): bool = a
+
+  var a: seq[bool] = static:
+      newSeq[bool](0).mapIt(it) # segfaults
+  var b: seq[bool] = static:
+      newSeq[bool](0).filterIt(it) # does not segfault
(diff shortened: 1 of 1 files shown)


ghost:
Failure not related to this PR:
image

ghost:
Also `glob` package from Nimble failed

PR #15660, opened 2020-10-20
fixes bootstrapping for any machine that has a Nim already installed …

Araq:
…[backport:1.4]


lib/packages/docutils/rstgen.nim
```diff
@@ -27,8 +27,7 @@
 
 import strutils, os, hashes, strtabs, rstast, rst, highlite, tables, sequtils,
   algorithm, parseutils
-import "$lib/../compiler/nimpaths"
-import "$lib/../compiler/pathutils"
+
```
Araq, line 30: See? This is completely broken.

PR #15675, opened 2020-10-22
Add C function dup and dup2 posix to system/io

Clonkk:
Posix functions (and their windows equivalent) added as duplicateFileHandle and assignFileHandle.

This is a small addition, mainly used for redirecting stdout properly motivated by these recent questions:
* https://forum.nim-lang.org/t/6909#43303 
* https://stackoverflow.com/questions/64026829/how-to-temporarily-capture-stdout/64032172#64032172


lib/system/io.nim
```diff
@@ -318,6 +328,40 @@ proc getFileHandle*(f: File): FileHandle =
...
+      # Raise
+      var e: ref Exception
+      new(e)
+      e.msg = $strerror(errno)
+      raise e
+    else:
+      # Same as checkErr
+      quit(1)
+
+proc assignFileHandle*(oldfd: FileHandle, newfd: FileHandle) =
```
alaviss: I'd call this: `duplicateTo`. We don't need the `FileHandle` within the name since `FileHandle` is a distinct type already.

alaviss (commented):
We would need some tests and change log entries for this.

If possible I'd like `duplicate` to get an `inheritable` parameter (defaults to `true`), which controls whether the resulting `fd` is inheritable, similar to how `io` works.

A note saying that FD from network sockets should not be duplicated with these functions would be helpful too (on Windows they has to be duplicated via the WSA equivalent of `dup`).

lib/system/io.nim
Clonkk: Wouldn't `duplicate` create confusion since `dup` already exists in the language ?

timotheecour:
`io` is imported by default for all nim programs (via system), these procs aren't essential enough to justify being defined for "all programs" so shouldn't be added to std/io but to another module, we can debate which one.
Modules are cheap.

lib/system/io.nim
timotheecour: `duplicateTo` is different enough from `dup`, nim doesn't define gratuitous aliases

Clonkk:
> We would need some tests and change log entries for this.

Will do.

 > If possible I'd like duplicate to get an inheritable parameter (defaults to true), which controls whether the resulting fd is inheritable, similar to how io works.

To clarify, do you mean an`inheritable` pragma or using `setInheritable` ?

 > io is imported by default for all nim programs (via system), these procs aren't essential enough to justify being defined for "all programs" so shouldn't be added to std/io but to another module, we can debate which one.
Modules are cheap.

This makes a lot of sense. I initially regrouped it with system/io because it was related to file system.
Should we have an io_extras or posix_extras ? What is the policy for naming such things ?

timotheecour:
> posix_extras

it's also defined for windows so posix is too restrictive

utils is the most common suffix in stdlib (13 modules), all attached. So `std/ioutils` sounds good (to me).

alaviss:
Is there any practical reason preventing this to be in `io`? Don't we have plans to remove `io` from being imported by default?

alaviss:
> To clarify, do you mean an`inheritable` pragma or using `setInheritable` ?

Just use a default parameter, there aren't any pragmas here :P

You can set the resulting fd inheritable state with `setInheritable`, but I'd recommend looking into native APIs to see if you can turn inheritable on/off at duplication time (for *nix there's `dup3` for example).

Araq:
Breaks on Windows. Also: Should be in its own module, unfortunately io.nim is still part of system.

timotheecour:
so, ok for std/ioutils.nim ?

Clonkk:
I marked this a WIP while I'm working on the modification you asked.

timotheecour:
you can use "convert to draft"

image

(might depend on your access level to this repo though)

tests/stdlib/tioutils.nim
```diff
@@ -0,0 +1,51 @@
...
+  echo msg & "1"
+  var s: string
+
+  captureStdout(s):
+    echo msg & "2"
+    msg = "ciao"
+
+  doAssert s == "hello2\n"
+
+when isMainModule:
```
timotheecour: why `when isMainModule:` ? makes test non-joinable (see `megatest`)

tests/stdlib/tioutils.nim
```diff
@@ -0,0 +1,51 @@
...
+let tmpFileName = "./tmpFile.txt"
+discard tryRemoveFile(tmpFileName)
+
+template captureStdout*(ident: untyped, body: untyped) =
+  var stdout_fileno = stdout.getFileHandle()
+  # Duplicate stoud_fileno
+  var stdout_dupfd = duplicate(stdout_fileno)
+  # Create a new file
+  # You can use append strategy if you'd like
+  var tmp_file: File = open(tmpFileName, fmWrite)
```
timotheecour: camelCase (+ elsewhere)

tests/stdlib/tioutils.nim
```diff
@@ -0,0 +1,51 @@
...
+import std/ioutils
+import os
+
+# Dummy filename
+let tmpFileName = "./tmpFile.txt"
+discard tryRemoveFile(tmpFileName)
+
+template captureStdout*(ident: untyped, body: untyped) =
+  var stdout_fileno = stdout.getFileHandle()
+  # Duplicate stoud_fileno
```
timotheecour: typo

tests/stdlib/tioutils.nim
```diff
@@ -0,0 +1,51 @@
...
+"""
+
+import std/ioutils
+import os
+
+# Dummy filename
+let tmpFileName = "./tmpFile.txt"
+discard tryRemoveFile(tmpFileName)
+
+template captureStdout*(ident: untyped, body: untyped) =
```
timotheecour: `template captureStdout*(ident, body) =`

tests/stdlib/tioutils.nim
```diff
@@ -0,0 +1,51 @@
...
+  output: '''
+hello1
+hello1
+  '''
+"""
+
+import std/ioutils
+import os
+
+# Dummy filename
```
timotheecour: nest all below under a `block` so you each test can be scoped when more procs are added eg
```nim
block: # duplicateTo, duplicate
  ...

block: # other procs in the future
```

tests/stdlib/tioutils.nim
```diff
@@ -0,0 +1,49 @@
...
+  output: '''
+hello1
+hello1
+  '''
+"""
+
+import std/ioutils
+import os
+
+block:
```
timotheecour: block: # duplicateTo, duplicate

tests/stdlib/tioutils.nim
timotheecour: you can click "resolve conversation" when comment addressed ;-)

tests/stdlib/tioutils.nim
```diff
@@ -0,0 +1,49 @@
...
+  let tmpFileName = "./tmpFile.txt"
```
timotheecour: * too generic, could potentially clash with other temp files from other tests in the future (tests can run in parallel), either use a uuid or name it `tioutils.txt`, which also makes its association with this file more greppable.
* nit: in some tests I'm using `from stdtest/specialpaths import buildDir` (which avoids polluting sources with tmp files, eg in case of a crash before tmp file collection which makes git status dirty), but feel free to ignore since it's not (yet?) common practice

tests/stdlib/tioutils.nim
```diff
@@ -0,0 +1,49 @@
...
+  template captureStdout*(ident, body) =
+    var stdoutFileno = stdout.getFileHandle()
+    # Duplicate stoudFileno
+    var stdout_dupfd = duplicate(stdoutFileno)
```
timotheecour: s/var/let/ (+ everywhere relevant)

timotheecour: stdout_dupfd => camlCase

tests/stdlib/tioutils.nim
```diff
@@ -0,0 +1,49 @@
...
+    ident = readFile(tmpFileName)
+    # Restore stdout
+    duplicateTo(stdout_dupfd, stdoutFileno)
+
+  proc duplicateStdout() =
+    var msg = "hello"
+    echo msg & "1"
+    var s: string
+
+    captureStdout(s):
```
timotheecour: why not the simpler:
```nim
let s = captureStdout:
  ...
```

and adapting `captureStdout` as needed, eg:
```nim
template captureStdout*(body): untyped =
...
let ret = readFile(tmpFileName)
...
ret
```

tests/stdlib/tioutils.nim
```diff
@@ -0,0 +1,49 @@
...
+      echo msg & "2"
+      msg = "ciao"
```
timotheecour: why?

tests/stdlib/tioutils.nim
```diff
@@ -0,0 +1,49 @@
...
+
+    doAssert s == "hello2\n"
+
+  discard tryRemoveFile(tmpFileName)
+  duplicateStdout()
+  # Check it works twice
+  duplicateStdout()
+  discard tryRemoveFile(tmpFileName)
```
timotheecour: `doAssert tryRemoveFile(tmpFileName)` for this one (it should succeed here)

lib/std/ioutils.nim
```diff
@@ -0,0 +1,62 @@
...
+else:
+  proc c_dup(oldfd: FileHandle): FileHandle{.
+    importc: "dup", header: "<unistd.h>".}
+  proc c_dup2(oldfd: FileHandle, newfd: FileHandle): cint {.
+    importc: "dup2", header: "<unistd.h>".}
+
+const SupportIoctlInheritCtl = (defined(linux) or defined(bsd)) and
+                              not defined(nimscript)
+when SupportIoctlInheritCtl:
+  var
```
timotheecour: * s/var/let/ since it's a C constant (it's a new-ish feature that allows importc let without a value since it's defined in C)
* it's neither used not exported, either use it or export it disable with `when false` for future un-disabling

lib/std/ioutils.nim
```diff
@@ -0,0 +1,62 @@
...
+    FIOCLEX {.importc, header: "<sys/ioctl.h>".}: cint
+    FIONCLEX {.importc, header: "<sys/ioctl.h>".}: cint
+
+
+proc strerror(errnum: cint): cstring {.importc, header: "<string.h>".}
+
+when not defined(NimScript):
+  var
+    errno {.importc, header: "<errno.h>".}: cint ## error variable
```
timotheecour: add comment: `also defined in std/posix`

lib/std/ioutils.nim
```diff
@@ -0,0 +1,62 @@
...
+
+proc duplicate*(oldfd: FileHandle): FileHandle =
+  ##[
+    Return a copy of the file handle oldfd.
```
timotheecour: * dedent:
```nim
##[
Return a copy of the file handle oldfd. ...
]##
```
* refer to all symbols with single backticks:
```nim
Return a copy of the file handle `oldfd`.
```
(ditto with `FileHandle` etc)

lib/std/ioutils.nim
```diff
@@ -0,0 +1,62 @@
...
+    They refer to the same open file description and share file offset and status flags.
+    Calls POSIX function `dup` on Linux and  `_dup` on Windows
+  ]##
+  let retValue = c_dup(oldfd)
+  if retValue != -1:
+    retValue
+  else:
+    when not defined(NimScript):
+      # Raise
+      var e: ref Exception
```
timotheecour: this is not good.

* see how other code raises:
```
raise newException(IOError, "...")
```
* you're essentially raising a `Exception` instead of appropriate subclass `IOError`.
* DRY violation: refactor common error handling from here and duplicateTo into a common non-exported proc:

something like:
```nim
template checkCode(ret: cint) =
  ...
```
it's a common pattern (similar exists in os and io) that should be factored out (and IMO one day it should be exported in some module so all code can reuse it instead of duplicating it but not in this PR)

lib/std/ioutils.nim
```diff
@@ -0,0 +1,50 @@
...
+
+when false:
+  const SupportIoctlInheritCtl = (defined(linux) or defined(bsd)) and
+                                not defined(nimscript)
+  when SupportIoctlInheritCtl:
+    var
+      FIOCLEX {.importc, header: "<sys/ioctl.h>".}: cint
+      FIONCLEX {.importc, header: "<sys/ioctl.h>".}: cint
+
+## Also defined in std/posix and system/io
```
timotheecour, line 20: sorry I meant regular comment, otherwise it will end in docs (check output of `nim doc`)

lib/std/ioutils.nim
```diff
@@ -0,0 +1,50 @@
...
+proc strerror(errnum: cint): cstring {.importc, header: "<string.h>".}
+when not defined(NimScript):
+  var errno {.importc, header: "<errno.h>".}: cint ## error variable
+
+template checkError(ret: cint)=
+  if ret == -1:
+    when not defined(NimScript):
+      raise newException(IOError, $strerror(errno))
+    else:
+      quit(1)
```
timotheecour: I know quit(1) is used in some other related places but maybe `doAssert false` is better refs https://forum.nim-lang.org/t/4042

lib/std/ioutils.nim
```diff
@@ -0,0 +1,50 @@
...
+
+proc duplicate*(oldfd: FileHandle): FileHandle =
+  ##[
+  Return a copy of the file handle `oldfd`.
+  After a successful return, both `FileHandle` may be used interchangeably.
+  They refer to the same open file description and share file offset and status flags.
+  Calls POSIX function `dup` on Linux and  `_dup` on Windows
+  ]##
+  let retValue = c_dup(oldfd)
```
timotheecour: simpler + more idiomatic:
```nim
result = c_dup(oldfd)
checkError(retValue)
```

lib/std/ioutils.nim
```diff
@@ -0,0 +1,50 @@
...
+  checkError(retValue)
+  retValue
+
+proc duplicateTo*(oldfd: FileHandle, newfd: FileHandle) =
+  ##[
+  Perform the same task a `duplicate` but instead of using the lowest unused file descriptor
+  it uses the FileHandle` specified by `newfd`.
+  Calls POSIX function `dup2` on Linux and `_dup2` on Windows.
```
timotheecour: s/on Linux/on posix/

tests/stdlib/tioutils.nim
```diff
@@ -0,0 +1,50 @@
+discard """
+  output: '''
+hello1
+hello1
+  '''
```
timotheecour: seems dubious because string ends in "  " (2 spaces)

instead:
```nim
discard """
  output: '''
hello1
hello1
'''
"""
```

tests/stdlib/tioutils.nim
```diff
@@ -0,0 +1,50 @@
...
+"""
+
+# import std/ioutils
+import ../../lib/std/ioutils
```
timotheecour: why? `import std/ioutils` should work.

tests/stdlib/tioutils.nim
```diff
@@ -0,0 +1,50 @@
...
+import os
+from stdtest/specialpaths import buildDir
+
+block: #duplicate, duplicateTo
```
timotheecour: sorry for the nitpick, it's `block: # duplicate, duplicateTo`

lib/std/ioutils.nim
```diff
@@ -0,0 +1,50 @@
...
+  ]##
```
timotheecour, line 61: can you add a very simple 2 or 3 line `runnableExamples:` to illustrate usage un-ambiguously (especially wrt argument ordering which is easy to forget / get wrong).

If not, then forget it but it would be nice. Ditto with `duplicate`.

Most new procs have `runnableExamples:` and it really improves docs.

lib/std/ioutils.nim
```diff
@@ -0,0 +1,50 @@
...
+      FIONCLEX {.importc, header: "<sys/ioctl.h>".}: cint
+
+## Also defined in std/posix and system/io
+proc strerror(errnum: cint): cstring {.importc, header: "<string.h>".}
+when not defined(NimScript):
+  var errno {.importc, header: "<errno.h>".}: cint ## error variable
+
+template checkError(ret: cint)=
+  if ret == -1:
+    when not defined(NimScript):
```
timotheecour: nit:

```
rg -F 'defined(NimScript)' | wc -l
       3
rg -F 'defined(nimscript)'|wc -l
      62
```

timotheecour:
@Clonkk thanks for your patience, almost there... please click "resolve conversation" on each resolved comment, thanks!

lib/std/ioutils.nim
Clonkk: If we don't want the inheritable flag handling right now (and do it later in another PR), we can remove this part entierely.

tests/stdlib/tioutils.nim
Clonkk: Typo during localtest

lib/std/ioutils.nim
```diff
@@ -0,0 +1,79 @@
...
+#     var
+#       FIOCLEX {.importc, header: "<sys/ioctl.h>".}: cint
+#       FIONCLEX {.importc, header: "<sys/ioctl.h>".}: cint
+
+## Also defined in std/posix and system/io
+proc strerror(errnum: cint): cstring {.importc, header: "<string.h>".}
+when not defined(nimScript):
+  var errno {.importc, header: "<errno.h>".}: cint ## error variable
+
+template checkError(ret: cint)=
```
timotheecour: `s/=/= ` (space)

lib/std/ioutils.nim
```diff
@@ -0,0 +1,79 @@
...
+#   const SupportIoctlInheritCtl = (defined(linux) or defined(bsd)) and
+#                                 not defined(nimscript)
+#   when SupportIoctlInheritCtl:
+#     var
+#       FIOCLEX {.importc, header: "<sys/ioctl.h>".}: cint
+#       FIONCLEX {.importc, header: "<sys/ioctl.h>".}: cint
+
+## Also defined in std/posix and system/io
+proc strerror(errnum: cint): cstring {.importc, header: "<string.h>".}
+when not defined(nimScript):
```
timotheecour: s/nimScript/nimscript/ (+elsewhere)
style insensitivity means it's the same but almost all other code uses `nimscript`

lib/std/ioutils.nim
```diff
@@ -0,0 +1,79 @@
...
+      raise newException(IOError, $strerror(errno))
+    else:
+      doAssert(false)
+
+proc duplicate*(oldfd: FileHandle): FileHandle =
+  ##[
+  Return a copy of the file handle `oldfd`.
+  After a successful return, both `FileHandle` may be used interchangeably.
+  They refer to the same open file description and share file offset and status flags.
+  Calls POSIX function `dup` on Posix platform and  `_dup` on Windows
```
timotheecour: typo: double space after and was used

lib/std/ioutils.nim
```diff
@@ -0,0 +1,79 @@
...
+
+proc duplicateTo*(oldfd: FileHandle, newfd: FileHandle) =
+  ##[
+  Perform the same task a `duplicate` but instead of using the lowest unused file descriptor
+  it uses the FileHandle` specified by `newfd`.
+  Calls POSIX function `dup2` on Posix platform and `_dup2` on Windows.
+  ]##
+  runnableExamples:
+    # Redirect stdout to a file temporarily
+    let tmpFileName = "./hidden_output.txt"
```
timotheecour: excellent runnableExamples; but can you use `import os` and `getTempDir() / "tduplicateTo.txt"` to avoid side effects?

timotheecour (approved):
these are my last comments, LGTM after they're addressed, thanks!

Clonkk:
Windows doc generation failed because runnable example fails for windows. 

It appears that 
```
proc open*(f: var File, filehandle: FileHandle,
           mode: FileMode = fmRead): bool {.tags: [], raises: [], benign.}
```
does not work on windows but does on Linux. 

Test that fails on Windows to reproduce the problem : 
```nim
  let file1 = open("dummy.txt", mode=fmWrite)
  let dupfile1fd = duplicate(file1.getFileHandle())
  var f : File
  doAssert(isNil(f))
  let res = open(f, dupfile1fd, mode=fmWrite)
  # Both fails
  doAssert(not isNil(f))
  doAssert(res)
  let msg = "This is a test message that will be displayed ! \n"
  f.write(msg)
  f.close()
```

Is it acceptable to rework the runnable like this  ?
 
```nim
  runnableExamples:
    # stdoutDuplicate is a copy of stdout FileHandle that points to STDOUT
    let stdoutDuplicate = duplicate(stdout.getFileHandle())
    # Writing to stdoutDuplicate will write to stdout
    doAssert(stdoutDuplicate != stdout.getFileHandle())
    # On windows, opening a file from a FileHandle does not work
    when not defined(windows):
      var f : File
      let res = open(f, stdoutHolder, mode=fmWrite)
      let msg = "This is a test message that will be displayed ! \n"
      f.write(msg)
      # Output "Test"
      f.close()
```

timotheecour:
> Is it acceptable to rework the runnable like this ?

yes, we can fix/revisit later

timotheecour:
@Clonkk 
> Is it acceptable to rework the runnable like this ?

also, could you open a bug for that?

Clonkk:
Yeah, I'll take care of it 

Opened https://github.com/nim-lang/Nim/issues/15736

timotheecour:
@Clonkk merci pour ta patience, looking forward for more contributions!

PR #15678, opened 2020-10-22
fix #15662

### LINKED ISSUE #15662: NimVM generic procs that take anonymous tuples gives nil access error.
When using `T: typedesc` and passing a anonymous tuple the NimVM is incapable of running the code
### Example
```nim
static:
  proc a(T: typedesc)= echo T.type
  a(float)
  a((int, int))
```

### Current Output
```
float
/usercode/in.nim(4, 6)   in
/usercode/in.nim(4, 6) Error: attempt to access a nil address
```

### Expected Output
```
float
(int, int)
```
### Nim version
```
Compiled at 2020-10-16

git hash: 018ae963ba83934a68d815c3c1c44c06e8ec6178
```

cooldome:
fix #15662. 
Skip typedesc arguments, as far I cal tell they are used only to instantiate the proc

### DIFF
--- a/compiler/vmgen.nim
+++ b/compiler/vmgen.nim
@@ -586,8 +586,6 @@ proc genCall(c: PCtx; n: PNode; dest: var TDest) =
   # varargs need 'opcSetType' for the FFI support:
   let fntyp = skipTypes(n[0].typ, abstractInst)
   for i in 0..<n.len:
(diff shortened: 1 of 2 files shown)


cooldome:
test failures are unrelated. glob package again.

PR #15681, opened 2020-10-22
Handle BLOB column type in SQLite as binary data

Clonkk:
Insertion of binary data can be done using prepared statement but retrieval cannot even on BLOB type.

This avoids issues of trying to retrieve binary data that contains characters \0, \n from truncating data without having to devle into the low level wrapper of sqlite3. 

Casting TEXT or VARCHAR to cstring is fine, but BLOB is explicit binary data and shouldn't but submitted to cstring limitation.

### DIFF
--- a/lib/impure/db_sqlite.nim
+++ b/lib/impure/db_sqlite.nim
@@ -94,6 +94,57 @@
 ##
 ##    db.close()
 ##
+## Storing binary data example
+##----------------------------
+##
+## .. code-block:: nim
+##
+##   import random
+##
+##   ## Generate random float datas
+##   var orig = newSeq[float64](150)
+##   randomize()
+##   for x in orig.mitems:
+##     x = rand(1.0)/10.0
+##
+##   let db = open("mysqlite.db", "", "", "")
+##   block: ## Create database
+##     ## Binary datas needs to be of type BLOB in SQLite
+##     let createTableStr = sql"""CREATE TABLE test(
+##       id INTEGER NOT NULL PRIMARY KEY,
+##       data BLOB
+##     )
+##     """
+##     db.exec(createTableStr)
+##
+##   block: ## Insert data
+##     var id = 1
+##     ## Data needs to be converted to seq[byte] to be interpreted as binary by bindParams
+##     var dbuf = newSeq[byte](orig.len*sizeof(float64))
+##     copyMem(unsafeAddr(dbuf[0]), unsafeAddr(orig[0]), dbuf.len)
+##
+##     ## Use prepared statement to insert binary data into database
+##     var insertStmt = db.prepare("INSERT INTO test (id, data) VALUES (?, ?)")
+##     insertStmt.bindParams(id, dbuf)
+##     let bres = db.tryExec(insertStmt)
+##     ## Check insert
+##     doAssert(bres)
+##     # Destroy statement
+##     finalize(insertStmt)
+##
+##   block: ## Use getValue to select data
+##     var dataTest = db.getValue(sql"SELECT data FROM test WHERE id = ?", 1)
+##     ## Calculate sequence size from buffer size
+##     let seqSize = int(dataTest.len*sizeof(byte)/sizeof(float64))
+##     ## Copy binary string data in dataTest into a seq
+##     var res: seq[float64] = newSeq[float64](seqSize)
+##     copyMem(unsafeAddr(res[0]), addr(dataTest[0]), dataTest.len)
+##
+##     ## Check datas obtained is identical
+##     doAssert res == orig
+##
+##   db.close()
+##
 ##
 ## Note
 ## ====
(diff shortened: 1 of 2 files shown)


Clonkk:
I realized I didn't provided an example as motivation for this PR.

This is useful to store Sequence of float (or any binary data) in SQLite Database and retrieving it easily by using ``BLOB`` type.

```nim
import random
## Prepare data
const dbName = "./local.db"
var orig = newSeq[float64](150)
randomize()
for x in orig.mitems:
  x = rand(1.0)/10.0

## reset database
proc resetDb()=
  let db = open(dbName, "", "", "")
  db.exec(sql"DROP TABLE IF EXISTS test")
  let createTableStr = sql"""CREATE TABLE test(
    id INTEGER,
    name VARCHAR(50) unique,
    data BLOB
  )
  """
  db.exec(createTableStr)
  db.close()

## Insert data
proc insertData()=
  let db = open(dbName, "", "", "")
  ## Dummy data
  var id = 12
  var name = "BobbyTable"
  ## Data needs to be seq[byte] for prepare to handle it as binary
  var dbuf = newSeq[byte](orig.len*sizeof(float64))
  copyMem(unsafeAddr(dbuf[0]), unsafeAddr(orig[0]), dbuf.len)

  # Use prepared statement to insert binary data
  var insertStmt = db.prepare("INSERT INTO test (id, name, data) VALUES (?, ?, ?)")
  insertStmt.bindParams(id, name, dbuf)
  let bres = db.tryExec(insertStmt)
  doAssert(bres)
  # Destroy statement
  finalize(insertStmt)

  db.close()

proc getValueData()=
  let db = open(dbName, "", "", "")
  ## getValue return string
  ## With theses modifications, since ``data`` is a BLOB field it won't be truncated
  var x = db.getValue(sql"SELECT data FROM test")
  db.close()

  ## Calculate sequence size from buffer size
  let seqSize = int(x.len*sizeof(byte)/sizeof(float64))
  ## Copy binary string data into seq
  var res: seq[float64] = newSeq[float64](seqSize)
  copyMem(unsafeAddr(res[0]), addr(x[0]), x.len)

  ## Sanity check
  ## length check currently fails
  doAssert res.len == orig.len
  doAssert res == orig

block: # SQLite binary data
  resetDb()
  insertData()
  getValueData()
```

Currently, this would not work. 
This PR aims at making ``BLOB`` field that actually contians binary data work with ``getValue`` and ``getRow``. The API doesn't change, ``Row`` are still ``seq[string]``, ``getValue`` still returns a string; it's just that the string is not constructed from ``column_text`` (which return ``cstring``) but from ``column_byte`` whenever ``column_type == SQLITE_BLOB``.

Araq:
Great PR but something is messed up:

FAIL: tests/niminaction/Chapter7/Tweeter/tests/database_test.nim C  --nilseqs:on
Test "tests/niminaction/Chapter7/Tweeter/tests/database_test.nim" in category "niminaction"
Failure: reExitcodesDiffer
Expected:
exitcode: 0

Gotten:
exitcode: 1

Output:
database_test.nim(24)    database_test
database.nim(56)         follow
db_sqlite.nim(227)       exec
db_sqlite.nim(147)       dbError
Error: unhandled exception: unrecognized token: "'" [DbError]

Clonkk:
I added this use case to the documentation.
I can adapt the example into a test as well .
Does ``tests/stdlib/sqlitebindatas.nim`` sounds good or is there a better place (or are sqlite tests not wanted) ?

It would prob. be something close to this :

```nim
discard """
  action: "run"
  exitcode: 0
"""
import db_sqlite
import random
import os
from stdtest/specialpaths import buildDir

block tsqlitebindatas: ## db_sqlite binary data
  const dbName = buildDir / "tsqlitebindatas.db"

  let origName = "Bobby"
  var orig = newSeq[float64](150)
  randomize()
  for x in orig.mitems:
    x = rand(1.0)/10.0

  discard tryRemoveFile(dbName)
  let db = open(dbName, "", "", "")
  let createTableStr = sql"""CREATE TABLE test(
    id INTEGER NOT NULL PRIMARY KEY,
    name TEXT,
    data BLOB
  )
  """
  db.exec(createTableStr)

  var dbuf = newSeq[byte](orig.len*sizeof(float64))
  copyMem(unsafeAddr(dbuf[0]), unsafeAddr(orig[0]), dbuf.len)

  var insertStmt = db.prepare("INSERT INTO test (id, name, data) VALUES (?, ?, ?)")
  insertStmt.bindParams(1, origName, dbuf)
  let bres = db.tryExec(insertStmt)
  doAssert(bres)

  finalize(insertStmt)

  var nameTest = db.getValue(sql"SELECT name FROM test WHERE id = ?", 1)
  doAssert nameTest == origName

  var dataTest = db.getValue(sql"SELECT data FROM test WHERE id = ?", 1)
  let seqSize = int(dataTest.len*sizeof(byte)/sizeof(float64))
  var res: seq[float64] = newSeq[float64](seqSize)
  copyMem(unsafeAddr(res[0]), addr(dataTest[0]), dataTest.len)
  doAssert res.len == orig.len
  doAssert res == orig

  db.close()
  doAssert tryRemoveFile(dbName)
```

Clonkk:
The CI fails for openbsd on SSL tests 
```
FAIL: tests/stdlib/tssl.nim C
Test "tests/stdlib/tssl.nim" in category "stdlib"
Failure: reExitcodesDiffer
Expected:
```
I assume this is known ?

Issue #7865, opened 2018-05-22
Memory corruption on newSeq[MyAlignedType](1)


cooldome:
Compile and run to reproduce a crash:
```nim
{.passC: "-march=native".}

type
  m256d {.importc: "__m256d", header: "immintrin.h".} = object

proc set1(x: float): m256d {.importc: "_mm256_set1_pd", header: "immintrin.h".}

for _ in 1..1000:
  var x = newSeq[m256d](1)
  x[0] = set1(1.0)
```

The reason is clear, Nim incorrectly computed the amount bytes that needs to be allocated:
From lib/system/gc.nim:
```nim
proc newSeq(typ: PNimType, len: int): pointer {.compilerRtl.} =
  let size = addInt(mulInt(len, typ.base.size), GenericSeqSize)
```
GenericSeqSize is 16 bytes and  typ.base.size is 32 bytes. 48 bytes are allocated. 
Due alignment and padding it is not enough. It should 64 bytes instead, there is a padding between seq header and data.

krux02:
I tried to compile it with the ``--newruntime`` switch, and the code block does not crash. But I can't explain why this is the case. According to my understanding of newruntime, it should crash for exactly the same reason.

narimiran:
FWIW, it still crashes with the default GC, but it works with `--gc:arc`.

Araq:
Well it works with --gc:arc so that's good enough.

PR #15697, opened 2020-10-23
ARC now capable of custom extra alignment. Ref, closure and seq support.

cooldome:
~~Max Alignment is 256 bytes. Let me know if you want me to increase it.~~
Max Alignment is  32768 bytes.

For `ref T` I had to do the following trick, suppose the alignment is 64 bytes and we are overallocating then memory layout is:  padding first then `RefHeader` then the object. Hence there is a padding at start such that `head(myref)` stil  works and it does `cast[int](myref) - sizeof(RefHeader)`.

New functions  `alignedAlloc, alignedRealloc, alignedDealloc` mimic what the functions `posix_memalign`, `aligned_alloc` do but in the platform independent way. These functions are overallocating on windows and linux.


cooldome:
ready for review, test failures seem unrelated

cooldome:
found the bug with useMalloc. Investigating

cooldome:
Ready for review. Increased max alignment to 32768.

Araq:
> New functions alignedAlloc, alignedRealloc, alignedDealloc mimic what the functions posix_memalign, aligned_alloc do but in the platform independent way. These functions are overallocating on windows and linux.

Does this mean that we also overallocate with our default native Nim allocator? When exactly do we overallocate?

cooldome:
Here are my findings. Functions like  posix_memalign, aligned_alloc are always overallocating because they need to store alignment offset somewhere hence using them in all cases is not an option as Nim's memory footprint would increase significantly. There is no standard way to reallocate aligned memory while this op is used by seqv2 implementation. Hence I implemented my own portable 
version of these functions.

In order to keep overhead low I am checking if extra alignment is required at all. If not I revert back to standard allocation (either nim allocator or c_malloc depending on useMalloc). 
The downside is that alignment needs to be passed as the argument in all cases even for dealloc. Such that `if` statement could be done:

```nim
 proc alignedDealloc(p: pointer, align: int) {.compilerproc.} = 
    if align <= MemAlign:
      existingDealloc(p)
    else:      
      let offset = cast[ptr uint16](p -! sizeof(uint16))[]
      existingDealloc(p -! offset)
```

Summary:
Alignment works both with useMalloc and Nim allocator. For types that doesn't need extra alignment the only overhead is extra if statement `if align <= MemAlign:`. Compared to allocation itself is a small overhead.

cooldome:
Also it is a good time to reduce default `MemAlign` to reduce memory footprint for apps.

cooldome:
fixes #7865

lib/system/memalloc.nim
```diff
@@ -304,6 +304,80 @@ when hasAlloc and not defined(js):
...
+
+  proc alignedAlloc(size, align: Natural): pointer =
+    if align <= MemAlign:
+      when compileOption("threads"):
+        result = allocShared(size)
+      else:
+        result = alloc(size)
+    else: 
+      when compileOption("threads"):
+        let base = allocShared(size + align - 1 + sizeof(uint16))
```
Araq, line 325: Write a comment that explains this logic. And also why it doesn't produce **un**aligned stores (which can be performance desasters on non-x86 CPUs).

cooldome: I have added the comments. I have carefully reviewed the code for unaligned reads or stores. I dont' see any.

cooldome:
ready for review

planetis-m:
@cooldome Do you plan to make doc comments, export aligned* and add changelog entry in a follow up or it wasn't exported on purpose?

PR #15700, opened 2020-10-23
Allow return in expressions again

### LINKED ISSUE #251: Allow return in expressions again
This is a feature that was introduced in https://github.com/nim-lang/Nim/pull/13520 and removed again in https://github.com/nim-lang/Nim/pull/15281.
You can argue that for consistency with the fact that raise/break/continue are allowed too return should be allowed too, but there are also practical valid usecases of that feature, like allowing one to implement a custom exception mechanism, using something like this
```nim
template customRaise =
  setException()
  return
```
which could then be used in the same manner like a builtin `raise` statement would. This ceases to work after https://github.com/nim-lang/Nim/pull/15281

Clyybber:
Reverts nim-lang/Nim#15281
Closes https://github.com/nim-lang/RFCs/issues/251
[backport: 1.2] | [backport: 1.4]


Clyybber:
CI failure unrelated.

PR #15701, opened 2020-10-23
Try to fix CI failures

Clyybber:


testament/important_packages.nim
```diff
@@ -1,134 +1,136 @@
...
+template pkg2(name: string; cmd = "nimble test"; url = "", useHead = true): untyped =
+  packages2.add((name, cmd, url, useHead))
 
-var packages1*: seq[tuple[name, cmd: string; hasDeps: bool; url: string, useHead: bool]] = @[]
-var packages2*: seq[tuple[name, cmd: string; hasDeps: bool; url: string, useHead: bool]] = @[]
+var packages1*: seq[tuple[name, cmd: string; url: string, useHead: bool]] = @[]
+var packages2*: seq[tuple[name, cmd: string; url: string, useHead: bool]] = @[]
 
+# If url == "" useHead indicates wether nimble will install the latest head or release version
+# If url != "" useHead indicates wether nimble will be involved at all (useHead = false: nimble won't be involved)
```
timotheecour: that sounds weird, why can't we do:

`nimble develop "https://github.com/jackmott/easygl"`

testament/categories.nim
```diff
@@ -501,37 +480,36 @@ proc testNimblePackages(r: var TResults; cat: Category; packageFilter: string, p
   let packagesDir = "pkgstemp"
   var errors = 0
   try:
-    for name, url, cmd, hasDep, useHead in listPackages(part):
+    let importantList =
```
timotheecour: PR title "Use cligens release version" needs to be updated to reflect those bigger changes

timotheecour:
relevant links for future reference:
* https://github.com/c-blake/cligen/issues/170#issuecomment-715635442
* https://github.com/timotheecour/Nim/issues/167

testament/important_packages.nim
Clyybber: Because nimble doesn't support that.

timotheecour: how so? I just tried it and it seems to work

Clyybber: Oh, nice, I tried with "https://krux02@bitbucket.org/krux02/tensordslnim.git" and that didn't work, so I assumed it didn't work at all, but I just had to remove the "krux02@" login prefix; will adapt accordingly.

Clyybber: Oh, I think I found the culprit :D

timotheecour: > Btw @timotheecour do you know what might cause these weird CI failures? It can't find nimble, I can reproduce locally, but afaict execCmdEx2 uses poUsePath so that shouldn't be the issue (using an absolute path fails too)

```
/Users/runner/work/Nim/Nim/testament/testament.nim(825) testament
  /Users/runner/work/Nim/Nim/testament/testament.nim(781) main
  /Users/runner/work/Nim/Nim/testament/categories.nim(715) processCategory
  /Users/runner/work/Nim/Nim/testament/categories.nim(517) testNimblePackages
  /Users/runner/work/Nim/Nim/testament/testament.nim(110) execCmdEx2
  /Users/runner/work/Nim/Nim/lib/pure/osproc.nim(990) startProcess
  /Users/runner/work/Nim/Nim/lib/pure/osproc.nim(1058) startProcessAuxSpawn
  /Users/runner/work/Nim/Nim/lib/pure/os.nim(1390) setCurrentDir
  /Users/runner/work/Nim/Nim/lib/pure/includes/oserr.nim(94) raiseOSError
  Error: unhandled exception: No such file or directory
  Additional info: "pkgstemp/nake" [OSError]
```

let (installCmdLine, installOutput, installStatus) = execCmdEx2("nimble", ["develop", if useHead: name & "@#head" else: name, "-y"])
would call
`nimble develop nake -y`

this creates nake and pkgs (not sure exactly why, that might've changed from before where IIRC it was only installing package, not creating pkgs; not 100% sure)


then 
`let (buildCmdLine, buildOutput, buildStatus) = execCmdEx2(cmdArgs[0], cmdArgs[1..^1], workingDir=buildPath)`

would fail since `workingDir=buildPath` doesn't exist (since let packagesDir = "pkgstemp")

which should explain why CI fails

Clyybber: Hmm, it doesn't seem to be that fleshed out yet (it clones into "jackmott/easygl" for example, not into "easygl" directly, and in general the ability to test packages that lack a .nimble file seems useful)

Clyybber: Oh, and then theres the issue that you can't develop binary only packages :/

Clyybber: And another issue, nimble develop will always use head, so that won't work either. (And never did)

timotheecour: ok; so how about `nimble install -y --depsOnly` as follows:
```
export NIMBLE_DIR=/tmp/nim_ci_nimbledir # avoid messing up user's local machine
# for each pkg:
git clone url destdir
cd destdir
git checkout HEAD or latest release tag (or commit specified in important_package, would be nice to add)
if a binary package:
  nimble build
else:
  nimble install -y --depsOnly
run test cmd (nimble test or the one in important_packages)
```

Clyybber: Using git clone now, (useHead = false will use the latest tag) and nimble install --depsOnly

testament/categories.nim
```diff
@@ -479,16 +479,6 @@ proc makeSupTest(test, options: string, cat: Category): TTest =
   result.options = options
   result.startTime = epochTime()
 
-proc actionRetry(maxRetry: int, backoffDuration: float, action: proc: bool): bool =
```
timotheecour, line 482: @Clyybber why did you remove this? this was there for a good reason, reducing spurious git clone failures which used to happen a lot before this was added

timotheecour: ping @Clyybber , I'm now getting 503/443 errors again after your removal of `actionRetry` which was introduced for this very purpose in https://github.com/nim-lang/Nim/pull/14547

eg: in https://github.com/nim-lang/Nim/pull/15765/checks?check_run_id=1323632253 i just got:

```
Installing recaptcha@>= 1.0.3
  Downloading euantorano/recaptcha.nim using git
  download.nim(93)         getTagsListRemote
  Error: unhandled exception: Unable to query remote tags for euantorano/recaptcha.nim. Git returned: fatal: unable to access 'euantorano/recaptcha.nim': Failed to connect to github.com port 443: Connection refused
   [OSError]
```

timotheecour: ping @Clyybber this type of failure keeps happening and was mostly solved with `actionRetry`

Araq: Yeah...

Clyybber: Sorry, this slipped my notifications. PR incoming EDIT: #16262

PR #15715, opened 2020-10-25
[backport] Add support to the latest LibreSSL version

ghost:
Found that when I was trying to run my program on Void Linux :)


ghost:
@narimiran I think this should be backported for older nim versions to support latest LibreSSL

ghost:
test failure unrelated - https://github.com/nim-lang/Nim/issues/15713

PR #15729, opened 2020-10-26
Implements streams for sockets

PMunch:


juancarlospaco:
Needs changelog, since

ghost:
@juancarlospaco does a new module really need "since"? hm, maybe it does, I don't remember

juancarlospaco:
But changelog would be nice.

Clyybber:
It doesn't need `since`.

dom96:
What is the motivation behind this? I can't think of a reason why i'd want this abstraction.

disruptek:
I like the concept, but why do we need the `rs` and `ws` prefixes on the procedures?  Don't you want these to be as similar as possible to the `streams` API?

Araq:
> What is the motivation behind this? 

I didn't look at the code, but in theory: So that you can use Nim's stdlib parsers directly when data comes through the wire?

PMunch:
> I like the concept, but why do we need the `rs` and `ws` prefixes on the procedures? Don't you want these to be as similar as possible to the `streams` API?

That's just the prefix of the implementation, if you look in streams string stream has the `ss` prefix and file stream has the `fs` prefix. The names that the user will use are the ones defined in the `Stream` object.

> > What is the motivation behind this?
> 
> I didn't look at the code, but in theory: So that you can use Nim's stdlib parsers directly when data comes through the wire?

That's pretty much it, yeah. I had [this issue](https://github.com/PMunch/binaryparse/issues/6) opened in `binaryparse`, and I've been asked about it multiple times on IRC as well. And now I needed it for some parsing. My use-case was parsing DNS messages, and you don't know the length of those before you have read various fields in the structure itself, so you need to parse it to know how much data you need. This was hard to implement without a `SocketStream`.

cooldome:
Looks neat, could you please add the test. Create a local socket and write from one end and read from another end, writes and reads are through the stream.

disruptek:
> That's just the prefix of the implementation, if you look in streams string stream has the ss prefix and file stream has the fs prefix. The names that the user will use are the ones defined in the Stream object.

Great, but copying bad design is not good design.  What if I want to document these procedures?  Why do we need this troublesome abstraction?

PMunch:
Well, that is how the streams "interface" is implemented in Nim.. Don't really know what to tell you. But they're not directly exported so they wouldn't even show up in documentation.

disruptek:
Man, that sucks.  What can we do?

Araq:
> Man, that sucks. What can we do?

Oh shut up. Or use an ``Impl`` suffix instead, who cares.

Araq:
This needs tests and a changelog entry. But apart from that, seems solid.

lib/pure/socketstreams.nim
```diff
@@ -0,0 +1,174 @@
...
+  var s = WriteSocketStream(s)
+  discard s.data.send(s.buf[s.lastFlush].addr, s.buf.len - s.lastFlush)
+  s.lastFlush = s.buf.len
+
+proc rsClose(s: Stream) =
+  var s = ReadSocketStream(s)
+  s.data.close()
+
+proc newReadSocketStream*(s: Socket): owned ReadSocketStream =
+  new(result)
```
Araq: Use a constructor here please, 

```nim

result = ReadSocketStream(data: s, ...)

```

lib/pure/socketstreams.nim
```diff
@@ -0,0 +1,174 @@
...
+  result.readDataImpl = rsReadData
+  result.peekDataImpl = rsPeekData
+  result.readDataStrImpl = rsReadDataStr
+
+proc resetStream*(s: ReadSocketStream) =
+  s.buf = @[]
+  s.pos = 0
+
+proc newWriteSocketStream*(s: Socket): owned WriteSocketStream =
+  new(result)
```
Araq: Use a constructor here please.

lib/pure/socketstreams.nim
PMunch: I was just following the other Streams as a guide, should those be changed as well?

PMunch:
> This needs tests and a changelog entry. But apart from that, seems solid.

Oops, thought I had pushed that already

dom96:
Is there any reason why we can't have a single stream type for this instead of the read/write split?

dom96:
Also, this should be put in the ``std`` directory. It would also be nice if this could be put inside a ``streams`` directory and named ``sockets`` (so ``import std/streams/sockets`` would work) but I think Nim will not like that :/

PMunch:
> Is there any reason why we can't have a single stream type for this instead of the read/write split?

Short answer: Yes.
Long answer: With the `StringStream` and `FileStream` you expect this to work:
``` Nim
var s = newStringStream()
s.writeLine("Hello world")
s.setPosition(0)
assert s.readLine() == "Hello world"
```
If these were both a single stream then `writeLine` for a socket stream would send data over the socket, and `readLine` would read data from the socket, while `setPosition` wouldn't make any sense at all. By splitting it like this the `WriteSocketStream` essentially behaves exactly like a string stream and a file stream with the only difference being that you can't write to something that has already been flushed. The `ReadSocketStream` on the other hand doesn't support writing, but otherwise it will behave exactly like the other streams, only that it reads from the socket if you try to read past the end of the current buffer. Mixing these two would just give you a confusing mess that won't work with anything that uses streams and expects them to work like string and file streams.

PMunch:
> Also, this should be put in the `std` directory. It would also be nice if this could be put inside a `streams` directory and named `sockets` (so `import std/streams/sockets` would work) but I think Nim will not like that :/

Why wouldn't Nim like that? Because we already have a module named sockets? I can move it no problem, just figure out where.

lib/pure/socketstreams.nim
Araq: Yes but can be done in a follow-up PR.

Araq:
Move it to ``std`` please but not to ``std/streams``, we don't use hierarchies as they don't work. (Why std/streams/socket, why not std/sockets/stream? We avoid these questions.)

dom96:
> If these were both a single stream then writeLine for a socket stream would send data over the socket, and readLine would read data from the socket, while setPosition wouldn't make any sense at all. By splitting it like this the WriteSocketStream essentially behaves exactly like a string stream and a file stream with the only difference being that you can't write to something that has already been flushed. The ReadSocketStream on the other hand doesn't support writing, but otherwise it will behave exactly like the other streams, only that it reads from the socket if you try to read past the end of the current buffer. Mixing these two would just give you a confusing mess that won't work with anything that uses streams and expects them to work like string and file streams.

To me this already hints that this isn't a great abstraction to be implementing. You're breaking a lot of the functionality that are expected of streams already, in particular the `setPosition` features.

I would honestly prefer for this to be in a Nimble package first before being accepted into the stdlib so it can mature and we can figure out its limitations and whether it actually is useful enough to be adopted into the stdlib.

PMunch:
What is `std` anyways? 

> To me this already hints that this isn't a great abstraction to be implementing. You're breaking a lot of the functionality that are expected of streams already, in particular the `setPosition` features.

Well we don't have any other abstraction for streaming data to/from arbitrary sources. In fact I'd argue that it would be better to create the abstractions of `ReadStream` and `WriteStream`. I tried by very best to not break any functionality, not sure how you mean I break `setPosition`, it works just fine and exactly as you expect (with the only limitation that you can't write into an already sent buffer, but since you typically have control over `flush` this isn't that big of an issue). In fact it doesn't behave much different than a `FileStream` that is opened with `fmRead` or `fmWrite`. `binaryparse`, which is what I used this for, uses `setPosition` and it worked without issue with this implementation. DNS messages also have a simple compression algorithm that requires a lot of jumping long distance within the data, and it also worked without issue.

> I would honestly prefer for this to be in a Nimble package first before being accepted into the stdlib so it can mature and we can figure out its limitations and whether it actually is useful enough to be adopted into the stdlib.

Can't speak for its maturity, but it is useful. I implemented it because I needed it, and as I mentioned I've gotten questions multiple times (both on IRC and in the `binaryparse` repo) on how to use `Streams` for this purpose.

Araq:
``std`` is the namespace of our stdlib modules. Your modules belong to std or to fusion, still making up my mind. :-)

planetis-m:
`ReadStream` could also be used to improve `httpclient`, correct?

Araq:
> ReadStream could also be used to improve httpclient, correct?

Probably.

narimiran:
@PMunch Conflicts with the latest devel are now resolved, together with a fix for mismatching tags for `close`.

Now the question remains: should this be a part of stdlib, or should we add it to `fusion`?

lib/pure/socketstreams.nim
```diff
@@ -0,0 +1,171 @@
...
+    closeImpl: rsClose,
+    atEndImpl: wsAtEnd,
+    setPositionImpl: rsSetPosition,
+    getPositionImpl: rsGetPosition,
+    writeDataImpl: wsWriteData,
+    readDataImpl: wsReadData,
+    peekDataImpl: wsPeekData,
+    flushImpl: wsFlush)
+
+proc resetStream*(s: WriteSocketStream) =
```
planetis-m: Minor: `resetStream` are not part of the Streams interface.

planetis-m:
It doesn't really work for me, here is an example:

```nim
import net, streams, fusion/socketstreams

var
  client = newSocket()
  clientRequestStream = newWriteSocketStream(client)
  clientResponseStream = newReadSocketStream(client)

client.connect("data.pr4e.org", Port(80))
clientRequestStream.write("GET http://data.pr4e.org/page1.htm HTTP/1.0\c\L\c\L")
clientRequestStream.flush()

var buffer = newString(512)
while true:
  clientResponseStream.readStr(buffer.len, buffer)
  if buffer.len == 0:
    break
  stdout.write buffer
```
Edit: Looks like recv is to blame

lib/pure/socketstreams.nim
```diff
@@ -0,0 +1,171 @@
...
+
+proc rsSetPosition(s: Stream, pos: int) =
+  var s = ReadSocketStream(s)
+  s.pos = pos
+
+proc rsGetPosition(s: Stream): int =
+  var s = ReadSocketStream(s)
+  return s.pos
+
+proc rsPeekData(s: Stream, buffer: pointer, bufLen: int): int =
```
planetis-m: change with this:
```nim
proc rsPeekData(s: Stream, buffer: pointer, bufLen: int): int =
  let s = ReadSocketStream(s)
  if bufLen > 0:
    let oldLen = s.buf.len
    s.buf.setLen(max(s.pos + bufLen, s.buf.len))
    if s.pos > oldLen:
      result = s.data.recv(s.buf[s.pos].addr, s.pos - s.buf.len + bufLen)
    elif s.pos == oldLen:
      result = s.data.recv(s.buf[s.pos].addr, bufLen)
    elif s.pos + bufLen > oldLen:
      result = s.data.recv(s.buf[oldLen].addr, s.buf.len - oldLen)
    s.atEnd = result == 0
    if result > 0: copyMem(buffer, s.buf[s.pos].addr, result)
```

planetis-m: Hey, might still be wrong, its hard to reason with so many levels of abstraction

narimiran: > `s.atEnd = result == 0`

gives "'atEnd(s)' cannot be assigned to", i.e. "attempting to call undeclared routine: 'atEnd='"

planetis-m: yeah, I made `atEnd` a field of `ReadSocketStreamObj` and `rsAtEnd` read that value, it worked but had other bugs, I didn't understood at the time, so I gave up

planetis-m: Edit: should be good now, but needs a test

narimiran: Fixed!

Now it works for your example, as well as for the examples where you manually set position (as mentioned on IRC).

planetis-m:
One question only, why `WriteSocketStream` is both read and write? It seems to me it complicates the implementation. Why reading from `ReadSocketStream` what a write from `WriteSocketStream` is allowed?

PMunch:
Reading from a write socket will only return what you have written to it. This was done to assure the highest possible comparability with other streams.

planetis-m:
I get it now, thanks

Issue #12928, opened 2019-12-19
filterIt().sortedByIt() result with following Error: internal error: environment misses: result`gensym340031


jringuette:
Doing a filterIt(...).sortedByIt(...) crash with the following: `Error: internal error: environment misses: result`gensym340031`

### Example
```nim
import sequtils, algorithm, unittest

type
    Student* = tuple[name: string, grade: int]
    School* = object
        students*: seq[Student]

func grade*(school: School, grade: int): seq[string] =
    return school.students.filterIt(it.grade == grade).sortedByIt(
            it.name).mapIt(it.name)

test "grade returns the students in that grade in alphabetical order":
  let school = School(students: @[("Franklin", 5), ("Bradley", 5), ("Jeff", 1)])
  check school.grade(5) == @["Bradley", "Franklin"]
```
https://play.nim-lang.org/#ix=24Pp

### Current Output
```
/playground/nim/lib/pure/collections/sequtils.nim(495, 7) Error: internal error: environment misses: result`gensym340031
```

### Expected Output
```
[OK] grade returns the students in that grade in alphabetical order
```

### Possible Solution

Use a temp variable to split the filter from the sort like this

```nim
func grade*(school: School, grade: int): seq[string] =
    let tmp = school.students.filterIt(it.grade == grade)
    return tmp.sortedByIt(it.name).mapIt(it.name)
```
https://play.nim-lang.org/#ix=24Pq

### Additional Information

```
$ nim -v
Nim Compiler Version 1.0.4 [Linux: amd64]
Compiled at 2019-11-27
Copyright (c) 2006-2019 by Andreas Rumpf

git hash: c8998c498f5e2a0874846eb31309e1d1630faca6
active boot switches: -d:release
```

ghost:
Closing as a  duplicate of https://github.com/nim-lang/Nim/issues/10456

jpleboeuf:
I encountered the same exact internal error working on this Exercism problem a few days ago, which I reported on the Nim-lang forum there: "[Why is the chaining order of It templates a problem in this example?](https://forum.nim-lang.org/t/6983#43773)".

@shirleyquirk suggested in a reply that in [`sortedByIt`](https://github.com/nim-lang/Nim/blob/226595515c25785eaf078834dfb6f0ac337a5278/lib/pure/algorithm.nim#L504) changing the type of `x` and `y` from `typeof(seq1[0])` to `auto` seems to fix the error.

shirleyquirk:
[...]

PR #15734, opened 2020-10-26
fixes #10456,#12928 issues when chaining templates to sortedByIt

### LINKED ISSUE #10456: ICE with sequtils + algorithm (mapIt, sortedByIt) + json
When running that code snippet:
```nim
import sequtils, strutils, algorithm, json


proc test() = 
  try: 
    let info = parseJson("""
    {"a": ["1", "2", "3"]}
    """)
    let prefixes = info["a"].getElems().mapIt(it.getStr()).sortedByIt(it).reversed()
    echo prefixes
  except:
    discard
  
test()
```
(sorry, I tried to make it smaller, but couldn't)

the compiler crashes with an internal error:
```
/home/dian/Stuff/Nim/lib/pure/collections/sequtils.nim(98, 17) Error: internal error: environment misses: :tmp
```

Used latest Nim compiler as of writing this (commit https://github.com/nim-lang/Nim/commit/268197add8e9fbcbf15d6a042c8095828c29869f)

shirleyquirk:
I've no clue what causes the error, it's over my head whether this is at heart an issue with template expansion or lambda lifting or what, but this at least fixes the stdlib to work with chaining templates.

I'm also not sure where the testcase should best go, hopefully i guessed appropriately.

solutions that also work, kinda:
define the types as auto: (works for this particular case but auto has many restrictions)
```nim
  var result = sorted(seq1, proc(x,y: auto): int =
```

alias the sequence (like evalOnceAs does for many templates in sequtils)
```nim
  template seq2:untyped = seq1
  var result = sorted(seq2, proc(x, y: typeof(seq2[0])): int =
```
or

```nim
  let seq2 = seq1
  var result = sorted(seq2, proc(x, y: typeof(seq2[0])): int =
```


cooldome:
Thank you.
Test failure is unrelated, merging.

PR #15744, opened 2020-10-27
refactor(nimdoc): rewrite docs JS in Nim

jonahsnider:
As discussed on the Nim Discord yesterday this rewrites the JS on the generated docs to use a script rewritten in Nim.
It also fixes a bug where the preferred theme would not be applied on `theindex.html`.
Unfortunately I couldn't figure out how to fully integrate the new script into the build system.

I dropped the Nim file I wrote in `config/doc.nim` and am requesting that someone with more knowledge of the build system move it somewhere appropriate.
Looking at how `dochack.js` is used this seems like a relatively complex task to do, but I'd be happy to help if given proper guidance.

- [ ] Move the new Nim file somewhere appropriate
- [ ] Compile it just like dochack during doc generation
- [ ] Update tests for the new output
- [ ] Maybe add tests for the new Nim file?

### DIFF
--- a/config/doc.nim
+++ b/config/doc.nim
@@ -1,0 +1,41 @@
+import dom
+import jsffi
+
+let localStorage {.importc, nodecl.}: LocalStorage
+let documentDocumentElement {.importc: "document.documentElement",
+    nodecl.}: Element
+let documentDocumentElementDataset {.importc: "document.documentElement.dataset",
+    nodecl.}: JSObject
+
+proc main* {.exportc.} =
+  let toggleSwitch = document.querySelector(""".theme-switch input[type="checkbox"]""")
+  let currentTheme = localStorage.getItem("theme")
+
+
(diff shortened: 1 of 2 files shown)


Varriount:
Hm, this is an interesting problem. Either the HTML needs to make an assumption on where the generated JavaScript file will be, relative to the document, or it needs to be included inline.

 If we go for the first option, then there will need to be a way to set the relative path to something else (for custom deployments) and the build process will need to run both compilation tasks. 

If we go with the second, then the documentation generator will need to compile the new Nim code to JS and incorporate it at compile time.

jonahsnider:
The current system for dochack.js uses a replacement for the URL of the compiled JS.

https://github.com/nim-lang/Nim/blob/12143d90c828bb4142751a6511ce790aac8aca51/config/nimdoc.cfg#L235

I added one for `$docjs`, which is missing causing the builds to fail.

PR #15755, opened 2020-10-27
Spawn in arc doesn't need gcsafe proc

cooldome:
While I was removing {.gcsafe.} sprinkle from the code I have found spawn still does gcsafe analysis which no longer mandatory.
Which this change I can remove all {.gcsafe.} annotations in my project.


Araq:
Are you sure about that? Global GC'ed memory is as unsafe as before with our non-atomic refs. Sad, I know.

dom96:
Someone ran into this and tried to use `guard` today, can we disable the gc safety analysis for guarded global vars for arc/orc? (https://irclogs.nim-lang.org/28-10-2020.html#19:28:59)

cooldome:
Incorrect PR

PR #15759, opened 2020-10-28
Fix #15750

### LINKED ISSUE #15750: OrderedTable.== throws error on empty table: unhandled exception: index out of bounds...
OrderedTable.== throws error on empty table: unhandled exception: index out of bounds, the container is empty [IndexDefect]

### Example
```nim
import tables

var t1 = OrderedTable[int, string]()
var t2 = OrderedTable[int, string]()

echo t1 == t2
```

### Current Output
Error: unhandled exception: index out of bounds, the container is empty [IndexDefect]

### Expected Output
true

$ nim -v
Nim Compiler Version 1.4.0 [MacOSX: amd64]
Compiled at 2020-10-27

narimiran:
The fix is in the first commit. The second commit deals with the deprecated `add`.

### DIFF
--- a/lib/pure/collections/tables.nim
+++ b/lib/pure/collections/tables.nim
@@ -1619,6 +1619,8 @@ proc `==`*[A, B](s, t: OrderedTable[A, B]): bool =
 
   if s.counter != t.counter:
     return false
+  if s.counter == 0 and t.counter == 0:
+    return true
   var ht = t.first
   var hs = s.first
   while ht >= 0 and hs >= 0:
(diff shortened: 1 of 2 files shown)


lib/pure/collections/tables.nim
```diff
@@ -1619,6 +1619,8 @@ proc `==`*[A, B](s, t: OrderedTable[A, B]): bool =
 
   if s.counter != t.counter:
     return false
+  if s.counter == 0 and t.counter == 0:
```
ringabout, line 1622:
```suggestion
  if s.counter == 0:
```

narimiran: Yes, I'm aware of that :)

But I wanted to be more explicit (than necessary) just to convey what is the purpose of that part of the code (if you don't read the above lines).

Araq:
This needs to be backported btw.

PR #15762, opened 2020-10-28
IPv6 text representation according to RFC 5952

rockcavera:
Currently, the textual representation of IPv6 in Nim shortens a single group of zeros within the IPv6 address with "::". Let's see:
```
::2:3:4:5:6:7:8
1::3:4:5:6:7:8
1:2::4:5:6:7:8
1:2:3::5:6:7:8
1:2:3:4::6:7:8
1:2:3:4:5::7:8
1:2:3:4:5:6::8
1:2:3:4:5:6:7::
```

Yes, this is in accordance with [RFC 4291](https://tools.ietf.org/html/rfc4291#section-2.2). However, there is [RFC 5952](https://tools.ietf.org/html/rfc5952#section-4) which recommends not to shorten a single group of zeros with "::". The above examples would look like this:
```
0:2:3:4:5:6:7:8
1:0:3:4:5:6:7:8
1:2:0:4:5:6:7:8
1:2:3:0:5:6:7:8
1:2:3:4:0:6:7:8
1:2:3:4:5:0:7:8
1:2:3:4:5:6:0:8
1:2:3:4:5:6:7:0
```

Therefore, I made this pull request to discuss which representation to adopt.

### DIFF
--- a/lib/pure/net.nim
+++ b/lib/pure/net.nim
@@ -1807,7 +1807,7 @@ proc `$`*(address: IpAddress): string =
         var word: uint16 = (cast[uint16](address.address_v6[i*2])) shl 8
         word = word or cast[uint16](address.address_v6[i*2+1])
 
-        if biggestZeroCount != 0 and # Check if group is in skip group
+        if biggestZeroCount > 1 and # Check if group is in skip group
           (i >= biggestZeroStart and i < (biggestZeroStart + biggestZeroCount)):
           if i == biggestZeroStart: # skip start
             result.add("::")


rockcavera:
I believe @dom96 has not seen the request for review.

dom96:
I have little to no context on this :)

rockcavera:
I believe that this PR is more a matter of style than error correction, since the most current recommendation is to represent IPv6 verbatim without shortening a single group of 0 with "::".

I am in favor of adopting the new recommendation, as it is more beautiful and uniform.

Varriount:
I'm in favor of adopting this logic as well. Though I admit I don't have any experience with IPv6 specifically, my experience with data value parsing in general leads me to believe that, given 100 implementations of a parser for X, at least half of those will cover only the bare minimum required to accept the most common inputs.

rockcavera:
> I'm in favor of adopting this logic as well. Though I admit I don't have any experience with IPv6 specifically, my experience with data value parsing in general leads me to believe that, given 100 implementations of a parser for X, at least half of those will cover only the bare minimum required to accept the most common inputs.

Just to be sure, this does not involve parses in itself, but representation. The parser must be able to handle both situations without acknowledging an error. On the other hand, representation, according to the most current norms, should be the proposal in this PR.

PR #15763, opened 2020-10-28
Make default state public

planetis-m:
It was previously [discussed](https://github.com/nim-lang/Nim/pull/14725#issuecomment-646959447). that exporting state is needed for writing 3rd party modules (like in `fusion`) that add new random functions.


Araq:
Needs a changelog entry and a ``.since`` annotation.

PR #15773, opened 2020-10-29
canAlias improvement

cooldome:
improvement discussed offline. Arguments without var, ptr, ref don't come with alias risk.

### DIFF
--- a/compiler/isolation_check.nim
+++ b/compiler/isolation_check.nim
@@ -64,9 +64,18 @@ proc canAlias(arg, ret: PType; marker: var IntSet): bool =
   else:
     result = false
 
+proc isValueOnlyType(t: PType): bool = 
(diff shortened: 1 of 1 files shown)


compiler/isolation_check.nim
```diff
@@ -64,9 +64,18 @@ proc canAlias(arg, ret: PType; marker: var IntSet): bool =
   else:
     result = false
 
+proc isValueOnlyType(t: PType): bool = 
+  # t doesn't contain pointers and references
+  proc wrap(t: PType): bool {.nimcall.} = t.kind in {tyRef, tyPtr, tyVar}
```
Araq: Also add ``tyLent`` here.

cooldome: done

PR #15777, opened 2020-10-29
EnumUtils, speed up findStr in compiler

cooldome:
New functions in strutils parseEnumRange.
I plan to use them in compiler itself, likely useful in other places too.


Araq:
One idea: Move ``strutils.parseEnum`` to ``std / enumutils`` and then add your ``parseEnumRange`` to ``std / enumutils`` and add an ``export`` statement to strutils for backwards compat.

cooldome:
@Araq, I have tried moving to enumutils. Unfortunately, it creates the circular dependency problem. nimIdentNormalize from strutils is used in enumutils.

Test failure is unrelated.

Clyybber:
Can't we simply use
```nim
proc parseEnumRange*[T: enum](s: string, a, b: static[T]): T =
  result = parseEnum[T](s)
  if result < a or result > b: raise newException(ValueError)
```
in the compiler instead?
It seems much simpler and doesn't require changes to the stdlib.

cooldome:
Tried it, doesn't work. ParseEnum complains at compile time:
```
 Error: Ambiguous enums cannot be parsed, field User appears multiple times!
```

It is really ambiogius, some warnings and hints have the same string representation in the compiler's enum.
It does work with parseEnumRange as expected since ambiguities are chopped off, hints are parsed separately from warnings.

Seems parseEnumRange is also useful tool to deal with ambiguities

Araq:
> It is really ambiogius, some warnings and hints have the same string representation in the compiler's enum.

Ok, but I think the way this message enum is used inside the Nim compiler is really hardly idiomatic and so almost nobody (except from us inside the Nim compiler) needs it.

cooldome:
parseEnumRange doesn't work either because it is using `nimIdentNormalize` instead of `normalize`.
I have not a bad idea how to organize everything. `genEnumStmt` macro goes into `std / enumutils`.
Macro accepts normalization function as argument this helps with breaking circular dependency.

`genEnumStmt` is used directly by `findStr` in `wordrecg`. `parseEnumRange` is no longer required.


Does it sound good?

juancarlospaco:
Needs changelog.

cooldome:
Added changelog entry. Also found a way how to pass normalizer to macro only once

PR #15783, opened 2020-10-30
fix adding empty sequence to HTTP headers

ringabout:
I lost the information by error due to forcing push, so I resubmmit a PR for this.

https://github.com/nim-lang/Nim/pull/15467


dom96 (approved):
Nice! looks good, but can you also add a test for `hasKey` when the key is expected to be there? i.e. it returns `true`?

PR #15784, opened 2020-10-30
closes #7658

### LINKED ISSUE #7658: Union types in javascript
Not sure if it's possible or planned to make union types work in Javascript, currently it compiles but as separate fields.

```Nim
type Foo{.union.} = object
  as_bytes: array[8, int8]
  data: int64


let a = Foo(data: 12345)

echo a.as_bytes # [0, 0, 0, 0, 0, 0, 0, 0]
echo a.data     # 12345

a.as_bytes[0] = 1

echo a.data     # 12345
echo a.as_bytes # [1, 0, 0, 0, 0, 0, 0, 0]
```

I suppose unions in JS should be disallowed like `casts` are.

ringabout:


Araq:
Ideally you do this in the JS backend so that ``.union`` remains to be allowed in compile-time code. But it's ok this way since the VM deosn't support .union either.

PR #15788, opened 2020-10-30
promote `collect` macro as a map+filter replacement

narimiran:


lib/pure/collections/sequtils.nim
```diff
@@ -361,7 +368,11 @@ proc map*[T, S](s: openArray[T], op: proc (x: T): S {.closure.}):
   ## Since the input is not modified you can use it to
   ## transform the type of the elements in the input container.
   ##
+  ## Instead of using `map` and `filter`, consider using the `collect` macro
+  ## from `sugar` module.
```
Araq: "from THE `sugar` module"

narimiran:
```suggestion
  ## from the `sugar` module.
```

bluenote10:
This is not "functional style programming".

Araq:
> This is not "functional style programming".

Yeah, it's better as it doesn't confuse (lazy) functions with control flow aspects.

bluenote10:
Just pointing out that if the docs list this `collect` under "functional programming" you can easily disappoint functional programmers. It may be "better", but this ugly `untyped -> untyped` "function" doesn't get the idea of functional programming, exactly because it introduces imperative style control flow.

In functional programming `map` + `filter` is `flatMap`.

Araq:
Well, there is nothing wrong with teaching functional programmers.

Issue #9091, opened 2018-09-27
Regression: overriding streams.write(T) does not work on 0.19 anymore


niv:
```nim
import streams

type Mine = ref object
  a: int

proc write*(io: Stream, t: Mine) =
  io.write("sure")

let str = newStringStream()
let mi = new Mine

str.write(mi)
```
On 0.18, this works as expected (it prints "sure").
0.19 fails with:

```
test.nim(12, 11) Error: type mismatch: got <Mine>
but expected one of:
proc `$`[T: tuple |
    object](x: T): string
proc `$`(x: bool): string
proc `$`[T](x: set[T]): string
proc `$`(x: uint64): string
proc `$`(x: float): string
proc `$`(x: char): string
proc `$`[T](x: openArray[T]): string
proc `$`[Enum: enum](x: Enum): string
proc `$`(x: int): string
proc `$`(w: WideCString; estimate: int; replacement: int = 0x0000FFFD): string
proc `$`(s: WideCString): string
proc `$`[T, IDX](x: array[IDX, T]): string
proc `$`(x: string): string
proc `$`(x: cstring): string
proc `$`[T](x: seq[T]): string
proc `$`(x: int64): string
```

LemonBoy:
Quoting the manual here:

> An empty subscript [] notation can be used to derefer a reference, the addr procedure returns the address of an item. An address is always an untraced reference. Thus the usage of addr is an unsafe feature.

And:

> The . (access a tuple/object field operator) and [] (array/string/sequence index operator) operators perform implicit dereferencing operations for reference types:

Moreover:

> Automatic dereferencing is also performed for the first argument of a routine call. But currently this feature has to be only enabled via {.experimental: "implicitDeref".}:

And since:

```nim
type Mine = ref object
  a: int
```

This means that

```nim
str.write(mi)
```

won't work and you have to:
- Introduce a deference `mi[]`
- Add `"implicitDeref"` to your file

The title of the ticket is completely off the track though.

demotomohiro:
I Added proc ``$`` for type ``Mine`` to that code, and that was compiled without err.
But proc ``$`` for ``Mine`` was never called.

```nim
  import streams

  type Mine = ref object
    a: int

  proc `$`(t: Mine): string = "Not sure"

  proc write*(io: Stream, t: Mine) =
    io.write("sure")

  let str = newStringStream("      ")
  let mi = new Mine

  str.write(mi)
  str.setPosition 0
  echo str.readAll
```

[...]

PR #15791, opened 2020-10-30
add testcase for #9091

ringabout:


narimiran:
@xflywind when writing a commit, please do something like "closes #9091 by adding a test case", because the `closes #9091` part automatically then closes the issue. Otherwise, we need to do it manually.

PR #15795, opened 2020-10-30
[closes #13549]add testcase for #13549

### LINKED ISSUE #13549: Compiler wrongly thinks generic tuple/object is ambiguous when mixing implicit/explicit generics
### Example
```nim
proc foo[T](x: T) = echo "a"

proc foo[T: tuple](x: T) = echo "b"

foo((1, 2, 3))

type Obj = object
proc foo(x: object) = echo "c"
foo(Obj())
```

### Current Output
```
Error: ambiguous call; both in.foo(x: T) and in.foo(x: T: tuple) match for: (tuple of (int, int, int))
Error: ambiguous call; both in.foo(x: T) and in.foo(x: object) match for: (Obj)
```

### Expected Output
```
b
c
```

### Additional Information

* Before 1.4.0, even if they were both implicit/both explicit, both of these calls would still fail as ambiguous.
* This is only for tuples and objects.

```
$ nim -v
Nim Compiler Version 1.4.0
```

ringabout:


tests/overload/t13549.nim
```diff
@@ -0,0 +1,16 @@
+discard """
+  output: '''b
+c'''
+"""
+
+proc foo[T](x: T) = echo "a"
+
+
+proc foo[T: tuple](x: T) = echo "b"
+proc foo(x: tuple) = echo "b"
```
Clyybber: Please only add one of these to the testcase, since technically both of these together *should* give an ambiguous call error I think.
```suggestion
proc foo[T: tuple](x: T) = echo "b"

```

Clyybber, line 10:
```suggestion
proc foo[T: tuple](x: T) = echo "b"
```

Clyybber, line 10:
```suggestion
proc foo[T: tuple](x: T) = echo "b"
```

ringabout:
It doesn't work.

PR #15801, opened 2020-10-31
fix #15618

### LINKED ISSUE #15618: unittest.check rejects with a type error, code that Nim otherwise accepts
### Example
```nim
func double*(n: float): float = n * 2

doAssert(double(-2) == -4)

when defined(good):
  import unittest
  test "double -2": check(double(-2) == (-4))

when defined(bad):
  import unittest
  test "double -2": check(double(-2) == -4)
```

### Current Output -d:good
```
[OK] double -2
```
### Current Output -d:bad
```
/home/jfondren/nim/learn/double3.nim(12, 8) template/generic instantiation of `test` from here
/home/jfondren/.choosenim/toolchains/nim-1.4.0/lib/pure/unittest.nim(669, 14) Error: type mismatch: got <float, int>
but expected one of:
proc `==`(x, y: float): bool
  first type mismatch at position: 2
  required type for y: float
  but expression ':c2' is of type: int
proc `==`(x, y: float32): bool
  first type mismatch at position: 2
  required type for y: float32
  but expression ':c2' is of type: int
26 other mismatching symbols have been suppressed; compile with --showAllMismatches:on to see them

expression: :c1 == :c2
```

As you can see, the 'bad' expression is doAsserted in both cases without error.

### Additional Information

```
$ nim -v
Compiled at 2020-10-16

```

ringabout:


ringabout:
If function has sideEffect, it will be called twice.

PR #15809, opened 2020-11-01
fix #8821

### LINKED ISSUE #8821: JS codegen can produce extreme switch statements with `case a of range`
then try to compile the following silly int32 tester for the js backend:
```nim
# isInt32.nim
proc isInt32(i: int): bool =
  case i 
  of low(int32)..high(int32):
    return true
  else:
    return false

echo isInt32(1)
```
Unless you have lots and lots of memory and disk space, running `nim js isInt32.nim` will make your system unusable. So if you are adventurous enough to try it, save all relevant work before the experiment. Most likely you will have to perform a hard reset...
Nim should refuse to compile harmless looking stupidities like this one. For this code on the js backend Nim tries to produce a switch statement with 4294967295 case labels. This is a little bit too much ...

ringabout:
Added a limitation for total length of range.

Choose 65535 because it generates 1 MB JS file for simple program
```nim
proc isInt32(i: int): bool =
  case i 
  of 1 .. 65536:
    return true
  else:
    return false

echo isInt32(1)
```

### DIFF
--- a/compiler/jsgen.nim
+++ b/compiler/jsgen.nim
@@ -865,6 +865,7 @@ proc genRaiseStmt(p: PProc, n: PNode) =
 proc genCaseJS(p: PProc, n: PNode, r: var TCompRes) =
   var
     cond, stmt: TCompRes
+    totalRange = 0
   genLineDir(p, n)
   gen(p, n[0], cond)
   let stringSwitch = skipTypes(n[0].typ, abstractVar).kind == tyString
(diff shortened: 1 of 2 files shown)


Varriount:
I read the original issue, but it wasn't very clear - why do "extreme" ranges cause JS memory issues?

ringabout:
It just causes compilation program crash when compiling Nim to JS. Because it generates too many labels for JS program. It's hard for me to optimize range branches. We should set a limitation in case that compilation program(NIm JS) crash.

bung87:
Your reason to choose `65535` doesn't make sense, you will gonna explain why " 1 MB" and why "simple program"

ringabout:
It's a temporary solution and the issue is still open(The right solution is to change `tranf.nim`). If users have some problems with them, it's easy to change the number. This is not backported and won't be shipped until Nim 1.6.0 (namely many months).
See https://github.com/nim-lang/Nim/issues/8821 for details

narimiran:
> This is not backported

I did backport it, because based on the PR title I thought it is a simple non-breaking fix. I can revert the backport if necessary.

ringabout:
Though this is not a proper fix. But I think it doesn't cause terrible issues. If users complain this, then we could revert this change or increase or decrease threshold. Anyway it prevents compilation program from crashing when compiling to JS.

PR #15819, opened 2020-11-02
fix #15594

### LINKED ISSUE #15594: cannot `capture result`, produces unhelpful eror
[`capture result`](https://nim-lang.org/docs/sugar.html#capture.m%2Cvarargs%5Btyped%5D%2Cuntyped) is impossible and the error message does not make it clear that `result` cannot captured.

### Example
```nim
import sugar

proc begin(): int =
  capture result:
    echo 1+1
  result
```

### Current Output
Compiling with 
```shell
$ nim compile anony.nim 
```

Produced with Nim 1.2.6
```
Hint: used config file '/home/user/nim-1.2.6/config/nim.cfg' [Conf]
Hint: system [Processing]
Hint: widestrs [Processing]
Hint: io [Processing]
Hint: anony [Processing]
Hint: sugar [Processing]
Hint: macros [Processing]
Hint: typetraits [Processing]
Hint: underscored_calls [Processing]
/home/user/anony.nim(4, 11) template/generic instantiation of `capture` from here
/home/user/nim-1.2.6/lib/pure/sugar.nim(189, 76) Error: redefinition of 'result'; previous declaration here: /home/user/nim-1.2.6/lib/pure/sugar.nim(189, 34)
```

Produced with Nim 1.3.7 nightly from nim-lang/Nim@b50a0ee
```
Hint: used config file '/home/user/nim-1.3.7/config/nim.cfg' [Conf]
Hint: used config file '/home/user/nim-1.3.7/config/config.nims' [Conf]
........
/home/user/anony.nim(4, 11) template/generic instantiation of `capture` from here
/home/user/nim-1.3.7/lib/pure/sugar.nim(210, 76) Error: redefinition of 'result'; previous declaration here: /home/user/nim-1.3.7/lib/pure/sugar.nim(210, 34)
```

### Expected Output
A successful compile, or a helpful error message explaining that `result` is special and cannot be captured.

ringabout:


lib/pure/sugar.nim
```diff
@@ -207,6 +207,8 @@ macro capture*(locals: varargs[typed], body: untyped): untyped {.since: (1, 1).}
   let locals = if locals.len == 1 and locals[0].kind == nnkBracket: locals[0]
                else: locals
   for arg in locals:
+    if arg.strVal == "result":
```
Araq, line 210: But this is not correct:

```nim

import sugar

proc begin() =
  var result = 0
  capture result:
    echo 1+1

```

Is fine. You need to check for ``arg.kind == nnkSym and arg.symKind == nskResult``. (Or similar.)

ringabout: This code doesn't compile too for the same reason.

ringabout: And so does this:
```nim
import sugar


var result = 0
capture result:
  echo 1+1
```

Araq: For what reason? The point is that "result" is only special for the ``nskResult`` symbol kind.

ringabout: `capture` rewrites the variable name as the parameter name of the functions with `auto` return type. And `result`  can't be used as parameters name with `auto` return type.

```nim
# redefinition of 'result';
proc hello(result: int): auto = 
  discard
```

"result" this name causes redefinition errors.

Clyybber: That should be fixed instead IMO

ringabout: I think it is reasonable for function with auto type can't use `result` as parameter name. 

We could change `capture` if people have a better solution.

PR #15828, opened 2020-11-03
Update iup

juancarlospaco:
- Update `iup` module from IUP `3.0` (Deprecated, year 2010) to IUP `3.30` (Latest, year 2020).
- Add 10 examples to the documentation.
- Changelog, `since`, style checked, examples with links and screenshots.
- Use `{.push.}` and `{.pop.}` for smaller code base DRY.
- Still a direct wrapper with just proc definitions, no logic to maintain, all names are same as IUP.

:)

PR #15830, opened 2020-11-03
remove iup from stdlib in contrast to #15828

ringabout:
Since it is deprecated for five years, outdated compared to latest official iup version and already in nimble, I think we should remove it from stdlib.
It could move to fusion if it is important.

https://github.com/nim-lang/iup

from iup readme

> The wrapper was originally part of the standard library. This project was created to remove it from the standard library, update it to the latest version of IUP and integrate it with nimble.

Related: #15828


changelog.md
```diff
@@ -17,6 +17,8 @@
 
 - Added `randState` template that exposes the default random number generator. Useful for library authors.
 
+- Removed deprecated `iup` module from stdlib, they have already moved to [nimble](https://github.com/nim-lang/iup).
```
ringabout: ``
```suggestion
- Removed deprecated `iup` module from stdlib, it has already moved to [nimble](https://github.com/nim-lang/iup).
```

PR #15831, opened 2020-11-03
fix #15463

### LINKED ISSUE #15463: VM segmentation faults with swap
Found through trying general examples from #15461. I swear I checked if this is a duplicate, I couldn't find anything.

### Example
```nim
static:
  var s = @[1, 2, 3]
  swap(s[0], s[2])
```

### Current Output
```
```

### Expected Output
s becomes @[3, 2, 1]

### Additional Information

* In versions before 0.19, this says `cannot evaluate at compile time: :tmp`, and ever since 0.19 this segmentation faults.

```
$ nim -v
(latest devel)
```

ringabout:


compiler/lowerings.nim
```diff
@@ -135,9 +135,14 @@ proc lowerTupleUnpackingForAsgn*(g: ModuleGraph; n: PNode; idgen: IdGenerator; o
 proc lowerSwap*(g: ModuleGraph; n: PNode; idgen: IdGenerator; owner: PSym): PNode =
   result = newNodeI(nkStmtList, n.info)
   # note: cannot use 'skTemp' here cause we really need the copy for the VM :-(
-  var temp = newSym(skVar, getIdent(g.cache, genPrefix), nextId(idgen), owner, n.info, owner.options)
+  let temp =
+    if owner == nil:
```
Araq: Owner shoulnd't be nil.

ringabout: I will look into this tomorrow.

ringabout: `c.prc` != nil but `c.prc.sym` == nil

ringabout: I think `c.prc` is initialized in `refresh` proc, so it isn't nil, but `c.prc.sym` is nil.
```nim
proc refresh*(c: PCtx, module: PSym; idgen: IdGenerator) =
  c.module = module
  echo "refresh: ", module
  c.prc = PProc(blocks: @[])
  c.loopIterations = c.config.maxLoopIterationsVM
  c.idgen = idgen
```



```
.refresh: test@369098752
mSwap: test@369098752
```

PR #15832, opened 2020-11-03
fix deprecated messages regarding high(value)

ringabout:


ghost:
maybe just high(int) ?

PR #15834, opened 2020-11-03
Clarify the sense in which Nim supports recursive iterators in docs

c-blake:
Fix manual, tutorial, and `tbintree` test comments to not mislead so much about lack of recursive iterator support.  This has confused quite  few people over the years.

See cblake & others comments in (at least!):
  https://forum.nim-lang.org/t/7020
  https://forum.nim-lang.org/t/6649
  https://forum.nim-lang.org/t/6393
  https://forum.nim-lang.org/t/5047
but probably more and probably a lot in the IRC as well.

### DIFF
--- a/doc/manual.rst
+++ b/doc/manual.rst
@@ -4031,9 +4031,12 @@ Closure iterators and inline iterators have some restrictions:
 1. For now, a closure iterator cannot be executed at compile time.
 2. ``return`` is allowed in a closure iterator but not in an inline iterator
    (but rarely useful) and ends the iteration.
-3. Neither inline nor closure iterators can be recursive.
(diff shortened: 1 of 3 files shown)


c-blake:
Tagging @mratsim who might have a critique on language used.  The simplest possible language may be "but Nim supports recursive iterator factories which are good enough/better via explicitness", but that only makes sense after someone understands iterator factories.

Also, the CI failure is just some `tests/vm/tslow_tables.nim` which cannot possibly be related to the comment-only/doc only changes.

c-blake:
Also, if you want, we could make the `toItr` example [more general](https://forum.nim-lang.org/t/7020#44192) with a probable trade-off for ease of understanding.

Araq:
Good enough for me.

PR #15839, opened 2020-11-04
fix #15663

### LINKED ISSUE #15663: `$`(s: WideCString) changes result to repr on --gc:arc
### Example
```nim
let ws = newWideCString("Test")
echo ws
```

### Current Output
Tested with Nim 1.4 and 1.5.1 git hash: `319aac4d43b04113831b529f8003e82f4af6a4a5`
```
(bytes: 10, data: Test)
```

### Expected Output
```
Test
```

### A Hint to a Possible Solution

At the exit point of ```proc `$`*(s: WideCString): string``` the `result` variable is as expected, but immediately after exit of the proc the result appears to be transformed into the `repr` of the `WideCStringObj` object.

### Additional Information
The result is as expected without `--gc:arc`/`--gc:orc` in the latest version 1.5.1 mentioned above.

ringabout:


lib/system/widestrs.nim
```diff
@@ -214,3 +214,10 @@ proc `$`*(w: WideCString, estimate: int, replacement: int = 0xFFFD): string =
 
 proc `$`*(s: WideCString): string =
   result = s $ 80
+
+when defined(nimv2):
+  proc `$`*(s: WideCStringObj, estimate: int, replacement: int = 0xFFFD): string =
```
Araq, line 219: Shouldn't ``replacement`` be used?

ringabout: my mistake

PR #15846, opened 2020-11-05
Fix #15706

### LINKED ISSUE #15706: Underflow not detected when using dec on distinct ranges

### Example
```nim
var x: distinct range[0..5]
dec(x)
echo int(x)
```

### Current Output
```
-1
```

### Expected Output
```
/usercode/in.nim(4)      in
Error: unhandled exception: over- or underflow [OverflowDefect]
```

### Additional Information
The following code works as expected: 
```
var x: range[0..5]
dec(x)
echo int(x)
```

Nim version is latest release installed through scoop
```
$ nim -v
Nim Compiler Version 1.4.0 [Windows: amd64]
Compiled at 2020-10-18

```

bung87:


cooldome (changes requested):
I have reviewed the changes. Actually approach needs to be a little bit different.
Find invocation of `binaryArithOverflowRaw` from `genMagicExpr`. `binaryArithOverflowRaw` is called with type `ranged`. Like this:
```nim
      let ranged = skipTypes(e[1].typ, {tyGenericInst, tyAlias, tySink, tyVar, tyLent})
      let res = binaryArithOverflowRaw(p, ranged, ....)
```
It needs to be
```nim
      let ranged = skipTypes(e[1].typ, {tyGenericInst, tyAlias, tySink, tyVar, tyLent, tyDistinct})
      let res = binaryArithOverflowRaw(p, ranged, ....)
```

cooldome:
I confirm that have reviewed all the places where overflow checks actually made, distinct types are handled everywhere including jsgen except one spot. This PR fixes this single spot.

tests/overflw/tdistinct_range.nim
```diff
@@ -0,0 +1,6 @@
+discard """
+  outputsub: "Error: unhandled exception: over- or underflow [OverflowDefect]"
```
cooldome, line 2: better to use errormsg: "unhandled exception: over- or underflow"

bung87:
Thanks for review my PR, am not sure am properly to change the error message, I checked the jsgen it doesn't even care of range, it only handle destinate type range.

PR #15860, opened 2020-11-06
follow #15357 and move decodeQuery

ringabout:
`decodeData` is very common to use and should work in JS backend(`cgi` doesn't work in JS backend).

So I move `decodeData` to `std/uri` and rename it to `decodeQuery`. Move `CgiError` etc to `std/uri` to keep backwards compatibility.

Or we should continue #15357 and change `proc` to `iterator`.


timotheecour:
Change it to iterator and make cgi call this

lib/pure/uri.nim
```diff
@@ -153,6 +163,41 @@ func encodeQuery*(query: openArray[(string, string)], usePlus = true,
...
+    setLen(name, 0) # reuse memory
+    i = parseData(data, i, name)
+    setLen(value, 0) # reuse memory
+    if i < data.len and data[i] == '=':
+      inc(i) # skip '='
+      i = parseData(data, i, value)
+    yield (name.TaintedString, value.TaintedString)
+    if i < data.len:
+      if data[i] == '&': inc(i)
+      else: cgiError("'&' expected")
```
timotheecour: * inline the code: `else: raise newException(CgiError, "'&' expected")` and remove `cgiError`
* IMO CgiError should be `ValueError` but it's tricky if it's used by cgi.nim; maybe it can all work though, by making cgi.nim use: `type CgiError = ValueError` ?

Clyybber: `cgi.decodeData` could also just catch ValueError and reraise CgiError.

ringabout: I think the `reraise` idea is better.

lib/pure/cgi.nim
```diff
@@ -88,31 +80,8 @@ proc getEncodedData(allowedMethods: set[RequestMethod]): string =
 iterator decodeData*(data: string): tuple[key, value: TaintedString] =
```
timotheecour, line 87: `template decodeData*(a: varargs[untyped]): untyped {.deprecated: "use uri.decodeQuery".}= uri.decodeQuery(a)`

or even better:
```
{.deprecated: [decodeData: uri.decodeQuery].}
```
(unless it hits https://github.com/nim-lang/Nim/issues/14819 in important_packages)

lib/pure/uri.nim
timotheecour: why though? what's the downside of the alias ?

lib/pure/cgi.nim
Clyybber: There is no reason to deprecate `cgi.decodeData`

lib/pure/uri.nim
ringabout: Original `CgiError` is object of `IOError`. It may break backwards compatibility a bit.

```nim
type
  CgiError* = object of IOError
  
  
try:
  raise newException(CgiError, "vgi")
except IOError:
  echo "True"
```

lib/pure/cgi.nim
timotheecour: having 2 ways to do the same thing is not good; you can keep a long deprecation cycle if needed.

lib/pure/uri.nim
timotheecour: reason more to deprecate cgi.decodeData (refs https://github.com/nim-lang/Nim/pull/15860#discussion_r518780631); if we deprecate `cgi.decodeData` I'm fine with this

lib/pure/cgi.nim
ringabout: I agree on deprecating `cgi.decodeData`. But I think we should keep a long deprecation cycle. Because many nim repos use `cgi.decodeData`.

https://github.com/dom96/jester/search?q=decodedata
https://github.com/planety/prologue/search?q=decodedata
https://github.com/itsumura-h/nim-basolato/search?q=decodedata

`cgi.decodeData` is very common in net library.

lib/pure/cgi.nim
```diff
@@ -52,11 +52,18 @@ proc xmlEncode*(s: string): string =
...
 type
+  CgiError* = object of IOError ## Exception that is raised if a CGI error occurs
   RequestMethod* = enum ## the used request method
     methodNone,         ## no REQUEST_METHOD environment variable
     methodPost,         ## query uses the POST method
     methodGet           ## query uses the GET method
 
+proc cgiError*(msg: string) {.noreturn.} =
+  ## Raises an ECgi exception with message `msg`.
+  var e: ref CgiError
```
timotheecour: raise newException(CgiError, msg)

lib/pure/uri.nim
```diff
@@ -59,6 +59,9 @@ type
     opaque*: bool
     isIpv6: bool # not expose it for compatibility.
 
+  UriParseError* = object of ValueError
```
timotheecour, line 62: no strong opinion, but isn't ValueError good enough instead of deep exception hierarchies?

lib/pure/uri.nim
```diff
@@ -153,6 +156,42 @@ func encodeQuery*(query: openArray[(string, string)], usePlus = true,
...
+    i = parseData(data, i, name)
+    setLen(value, 0) # reuse memory
+    if i < data.len and data[i] == '=':
+      inc(i) # skip '='
+      i = parseData(data, i, value)
+    yield (name.TaintedString, value.TaintedString)
+    if i < data.len:
+      if data[i] == '&': inc(i)
+      else:
+        raise newException(UriParseError, "'&' expected")
```
timotheecour: newException(UriParseError, "'&' expected at " & $i)

lib/pure/uri.nim
ringabout: But I need reraise this exception, So I want to choose an unique exception. I don't want to catch the wrong ValueError caused by `decodeQuery`. Backwards compatibility should be kept.

lib/pure/cgi.nim
timotheecour: that's fine, we don't have to decide now how long deprecation cycle will be, what matters is to deprecate it at same time as the duplication is introduced

lib/pure/cgi.nim
```diff
@@ -96,8 +96,11 @@ iterator decodeData*(allowedMethods: set[RequestMethod] =
   ## data consists of. If the client does not use a method listed in the
   ## `allowedMethods` set, a ``CgiError`` exception is raised.
   let data = getEncodedData(allowedMethods)
-  for key, value in decodeData(data):
-    yield (key, value)
+  try:
+    for (key, value) in uri.decodeQuery(data):
+      yield (key, value)
+  except UriParseError as e:
+    cgiError(e.msg)
```
timotheecour, line 106: I think this should be `reraise` otherwise you loose the stacktrace context where it occurred; can you please check?

ringabout: I think It has the similar error messages like before.

image


Before:
a
1
b
2c
/playground/nim/lib/pure/cgi.nim(115) in
/playground/nim/lib/pure/cgi.nim(67) cgiError
Error: unhandled exception: '&' expected [CgiError]

timotheecour: > I think It has the similar error messages like before.


no, before your PR you were getting more useful stacktrace, pointing to that line:
```nim
if data[i] == '&': inc(i)
else: cgiError("'&' expected")
```
whereas after your PR, you lost where it was raised in 1st place.


this is fixable, by using this:
```nim
raise newException(CgiError, e.msg, e)
```

this attaches `e` to parent of raised exception, so that client code can get the context:
```nim
import std/cgi
import std/strformat

proc exceptionToString(e: ref Exception): string =
  var e = e
  while e!=nil:
    result.add fmt"name: {e.name} msg: {e.msg} stacktrace: {e.getStackTrace}"
    e = e.parent

proc fn1()=
  for ai in decodeData("asdf=bar=baz"):
    echo ai

proc main() =
  try:
    fn1()
  except Exception as e:
    echo e.exceptionToString

main()
```

prints:
```
name: CgiError msg: '&' expected at 8 stacktrace: /Users/timothee/git_clone/nim/timn/tests/nim/all/t11246.nim(26) t11246
/Users/timothee/git_clone/nim/timn/tests/nim/all/t11246.nim(21) main
/Users/timothee/git_clone/nim/Nim_prs/lib/pure/cgi.nim(91) fn1
name: UriParseError msg: '&' expected at 8 stacktrace: /Users/timothee/git_clone/nim/timn/tests/nim/all/t11246.nim(26) t11246
/Users/timothee/git_clone/nim/timn/tests/nim/all/t11246.nim(21) main
/Users/timothee/git_clone/nim/Nim_prs/lib/pure/uri.nim(193) fn1

```


=> otherwise you loose where the exception is raised from (/Users/timothee/git_clone/nim/Nim_prs/lib/pure/uri.nim(193) fn1)


note that this reveals a nim bug (unless I'm missing something) which we should file, which is that parent exceptions aren't shown by default; nim should use something like `exceptionToString` to show parents => EDIT: https://github.com/nim-lang/Nim/issues/15868

ringabout: Since it is a bug, let's solve it later after this has been merged?

timotheecour: there are other ways to show error messages, https://github.com/nim-lang/Nim/issues/15868 just shows the default catch-all doesn't show parent currently; including `parent` is harmless.

but if for some reason you don't want to do it in this PR, not a blocker

lib/pure/uri.nim
```diff
@@ -153,6 +156,42 @@ func encodeQuery*(query: openArray[(string, string)], usePlus = true,
...
+    i = parseData(data, i, name)
+    setLen(value, 0) # reuse memory
+    if i < data.len and data[i] == '=':
+      inc(i) # skip '='
+      i = parseData(data, i, value)
+    yield (name.TaintedString, value.TaintedString)
+    if i < data.len:
+      if data[i] == '&': inc(i)
+      else:
+        raise newException(UriParseError, "'&' expected at " & $i)
```
Araq: This used to be proc call. Details like that are not mistakes as the code generator produces bloated code for ``raise``.

Araq:
Sorry but this seems to be a bad idea.

> "Added `decodeQuery` to `std/uri` and deprecated `decodeData` in `std/cgi`."

Omg why? It's ``std/cgi``, don't deprecate things in there. Stop messing around with the details -- some people here try to write code against the std libraries...

lib/pure/cgi.nim
ringabout: Deprecating can be done after this PR is merged if someone is willing to deprecate it in the future.

ringabout:
So we tend not to deprecate duplicated things in stdlib and only deprecate wrong/outdated things in stdlib?

Araq:
Your code introduced the duplication as far as I can tell. We should avoid both the duplication and the deprecation process.

timotheecour:
> Your code introduced the duplication as far as I can tell. We should avoid both the duplication and the deprecation process.

as of right now, only the API declaration is duplicated between cgi and uri, not the implementation, that's an acceptable tradeoff vs the benefits it brings:

> decodeData is very common to use and should work in JS backend(cgi doesn't work in JS backend).

plus discoverability, `decodeQuery` really does belong in std/uri. And this isn't the 1st PR attempting to address this.

So the current version of this PR is fine IMO, not deprecating cgi.decodeData (but uri.decodeQuery could be mentioned in docs for cgi.decodeData) and introducing uri.decodeQuery

timotheecour (approved):

ringabout:
```nim
ECgi* = object of EIO  ## the exception that is raised, if a CGI error occurs
```
`ECgi` has changed to `CgiError`, but some docs didn't change. I will clean them after this PR.

Does `decodeQuery` need `since`? I try to add `since(1, 5, 1)` but failed in CI. It told me that
```
/home/runner/work/Nim/Nim/lib/pure/cgi.nim(88, 29) Error: undeclared identifier: 'decodeQuery'
FAILURE
```

timotheecour:
> Does decodeQuery need since? I try to add since(1, 5, 1) but failed in CI. It told me that

probably bootstrap but can't reproduce locally (`bin/nim_csources c koch` works for me); can you provide a full link to CI logs? I can see several ways to handle this, including `when Nim >= (1,5,1) or Nim == (0,20,0)` (in pseudocode), depending on the CI log error

ringabout:
This one:
https://github.com/nim-lang/Nim/pull/15860/commits/b43dbc9b6262e9316a4c5b0e54db46d5c06a288b
And:
https://github.com/nim-lang/Nim/runs/1374923254

timotheecour:
ok, I can reproduce (locally too, via `rm bin/nim` and booting from csources); this would be easily fixed via `{.privateImport.}` (see https://github.com/nim-lang/Nim/pull/11865):
```nim
# uri.nim:
iterator decodeQuery(data: string): tuple[key, value: TaintedString] = ...
since (1,5,1): export decodeQuery

# cgi.nim
from uri {.privateImport.} import decodeQuery
```
(or equivalently, via `iterator decodeQueryImpl(data: string): tuple[key, value: TaintedString] = ...`; `template decodeQuery*(data: string) {.since: (1,5,1).}: untyped = decodeQueryImpl(data)`
and in cgi.nim: `from uri {.privateImport.} import decodeQueryImpl`)

but until then the simplest IMO is to just leave out `since` in this case.
EDIT: I've rebased https://github.com/nim-lang/Nim/pull/11865 so this option can still be on the table if needed (now or later)

timotheecour:
needs rebase against devel for conflict

lib/pure/uri.nim
```diff
@@ -153,6 +160,42 @@ func encodeQuery*(query: openArray[(string, string)], usePlus = true,
       result.add('=')
       result.add(encodeUrl(val, usePlus))
 
+iterator decodeQuery*(data: string): tuple[key, value: TaintedString] =
+  ## Reads and decodes query string ``data`` and yields the (name, value) pairs the
```
timotheecour: optional nit: `(name, value)` => `(key, value)` ?

tests/stdlib/tdecodequery.nim
```diff
@@ -0,0 +1,10 @@
+import uri, sequtils
```
timotheecour: move content to inside existing `tests/stdlib/turi.nim` since this tests std/uri

lib/pure/uri.nim
ringabout: Is it better to change `tuple[key, value: TaintedString]` to `tuple[name, value: TaintedString]` ?

tests/stdlib/tdecodequery.nim
```diff
@@ -0,0 +1,10 @@
...
+
+
+block:
+  const queryString = "a=1&b=0"
+  const wrongQueryString = "a=1&b=2c=6"
+
+  doAssert toSeq(decodeQuery(queryString)) == @[("a", "1"), ("b", "0")]
```
timotheecour: you're only using `queryString` and `wrongQueryString`, so it's better to inline them: more readble.

```nim
doAssert toSeq(decodeQuery("a=1&b=0")) == @[("a", "1"), ("b", "0")]
doAssertRaises(UriParseError):
  discard toSeq(decodeQuery("a=1&b=2c=6"))
```

tests/stdlib/tdecodequery.nim
ringabout: - [x] I could make a follow up PR since I also need to clean `cgi` module(`Egi` exception for example).

tests/stdlib/tdecodequery.nim
ringabout: same as above

lib/pure/uri.nim
```diff
@@ -153,6 +160,42 @@ func encodeQuery*(query: openArray[(string, string)], usePlus = true,
...
+    i = parseData(data, i, name)
+    setLen(value, 0) # reuse memory
+    if i < data.len and data[i] == '=':
+      inc(i) # skip '='
+      i = parseData(data, i, value)
+    yield (name.TaintedString, value.TaintedString)
+    if i < data.len:
+      if data[i] == '&': inc(i)
+      else:
+        uriParseError("'&' expected at " & $i)
```
timotheecour: this gives a not very helpful error. instead use this:
```nim
uriParseError("'&' expected at index '$#' for '$#'" % [$i, data])
```

lib/pure/uri.nim
```diff
@@ -153,6 +160,42 @@ func encodeQuery*(query: openArray[(string, string)], usePlus = true,
       result.add('=')
       result.add(encodeUrl(val, usePlus))
 
+iterator decodeQuery*(data: string): tuple[key, value: TaintedString] =
+  ## Reads and decodes query string ``data`` and yields the (name, value) pairs the
+  ## data consists of.
+  runnableExamples:
+    var queryData: seq[(string, string)]
```
timotheecour: let's promote stdlib:
```nim
  import std/sugar
  let s = collect(newSeq):
    for k,v in decodeQuery("foo=1&bar=2"): (k, v)
  doAssert s == @[("foo", "1"), ("bar", "2")]
```

lib/pure/uri.nim
timotheecour: why name?
IMO, it's best to be consistent with std/cgi (and key/value is more common in stdlib, eg tables etc)

tests/stdlib/tdecodequery.nim
timotheecour: => https://github.com/nim-lang/Nim/pull/16500

PR #15864, opened 2020-11-06
fix #12558

### LINKED ISSUE #12558: `mapIt` from `sequtils` not working in `{.push compile_time.}` context
### Example
```nim
import sequtils

{.push compile_time.}

proc foo =
  discard map_it([1, 2, 3], $it)

{.pop.}
```

### Current Output
```
/usercode/in.nim(6, 17) template/generic instantiation of `mapIt` from here
/playground/nim/lib/pure/collections/sequtils.nim(824, 13) Error: cannot evaluate at compile time: it
```

### Additional Information

It works with macros, which are also compile-time only
```nim
import sequtils

macro foo =
  discard map_it([1, 2, 3], $it)

foo
```

```
$ nim -v
Nim Compiler Version 1.0.99 [Windows: amd64]
Compiled at 2019-10-10
```

ringabout:


compiler/pragmas.nim
```diff
@@ -896,7 +896,11 @@ proc singlePragma(c: PContext, sym: PSym, n: PNode, i: var int,
       of wMagic: processMagic(c, it, sym)
       of wCompileTime:
         noVal(c, it)
-        incl(sym.flags, sfCompileTime)
+        if comesFromPush:
+          if sym.kind == skProc:
```
cooldome: needs to be at least `sym.kind in {skProc, skFunc}`

ringabout:
```suggestion
          if sym.kind in {skProc, skFunc}:
```

PR #15865, opened 2020-11-06
export PrettyOptions,prettyPrint from nimpretty

bung87:
not sure it will work for external usage, but export the type and proc first.

### DIFF
--- a/nimpretty/nimpretty.nim
+++ b/nimpretty/nimpretty.nim
@@ -44,11 +44,11 @@ proc writeVersion() =
   quit(0)
 
 type
-  PrettyOptions = object
-    indWidth: Natural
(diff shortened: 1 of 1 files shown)


cooldome:
It is not difficult to export when it becomes needed, please find a real world usage of these types/procs and then you can export no problem.

bung87:
I'v create a PR of nimlsp see https://github.com/PMunch/nimlsp/pull/59/files  

I modified nimpretty locally, but this PR will make nimlsp cause error when initNimSuggest, it will always get invalid `conf` no outDir , no outFile and other errors.

Araq:
Ok for now but nimpretty doesn't really expose an API, it's only a tool.

PR #15869, opened 2020-11-06
Make `{.requiresInit.}` to work for distinct types

bobeff:
Make `requiresInit` pragma to work for distinct types in addition to objects. Tagging of distinct types with `requiresInit` pragma was already supported, but its impact wasn't applied. Now its behavior when applied on distinct types is as follows.

Given the following distinct type definitions:

  ```nim
  type
    DistinctObject {.requiresInit, borrow: `.`.} = distinct MyObject
    DistinctString {.requiresInit.} = distinct string
  ```

The following code blocks will fail to compile:

  ```nim
  var foo: DistinctFoo
  foo.x = "test"
  doAssert foo.x == "test"
  ```

  ```nim
  var s: DistinctString
  s = "test"
  doAssert s == "test"
  ```

But these ones will compile successfully:

  ```nim
  let foo = DistinctFoo(Foo(x: "test"))
  doAssert foo.x == "test"
  ```

  ```nim
  let s = "test"
  doAssert s == "test"
  ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,8 @@
 
 ## Standard library additions and changes
 
+- Make `{.requiresInit.}` pragma to work for `distinct` types.
+
 - `prelude` now works with the JavaScript target.
 
 - Added `ioutils` module containing `duplicate` and `duplicateTo` to duplicate `FileHandle` using C function `dup` and `dup2`.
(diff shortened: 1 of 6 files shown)


zah (approved):

cooldome:
Thank you, fantastic

PR #15886, opened 2020-11-08
include example of error-marked copy proc

n5m:
Will help guide users that 
```nim
proc `=copy`(dest: var T; source: T) = {.error.}
```
(with an equals sign) is incorrect.

### DIFF
--- a/doc/destructors.rst
+++ b/doc/destructors.rst
@@ -191,7 +191,7 @@ that otherwise would lead to a copy is prevented at compile-time. This looks lik
   proc `=copy`(dest: var T; source: T) {.error.}
 
 but a custom error message (e.g., ``{.error: "custom error".}``) will not be emitted
-by the compiler.
+by the compiler. Notice that there is no ``=`` before the ``{.error.}`` pragma.
 
 Move semantics
 ==============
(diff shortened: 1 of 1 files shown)


ghost:
Should this really be documented? This is usual Nim's pragma syntax :thinking:

n5m:
@Yardanico to answer your question, here is why I would like it included. Consider this program:
```nim
# example.nim
type Wrapper = object
  discard

### Insert one of the =copy implementations here ###

proc main() =
  var a: Wrapper
  var b: Wrapper = a

main()
```

If you were to insert
```nim
proc `=copy`(dest: var Wrapper, source: Wrapper) {.error: "custom".}
```
before `proc main()` (notice no `=`), when compiling you would see this error message:
```console
user@user:~$ nim c --hints:off example.nim 
/home/user/example.nim(8, 20) Error: '=copy' is not available for type <Wrapper>; requires a copy because it's not the last read of 'a'; routine: main
```

On the other hand, if you were to insert
```nim
proc `=copy`(dest: var Wrapper, source: Wrapper) = {.error: "custom".}
```
before `proc main()` (notice the `=`), when compiling you would see this error message:
```console
user@user:~$ nim c --hints:off example.nim 
/home/user/example.nim(4, 59) Error: custom
```

PR #15888, opened 2020-11-08
Fix 15629

### LINKED ISSUE #15629: Destructor not called for ref member of object in specific circumstances with ARC
The destructor is not being called for 'inner' in the example code.

### Example
```nim
type inner = object
type outer = ref inner

proc `=destroy`(b: var inner) =
  echo "destroy"

proc newOuter(): outer =
  echo "create"
  new(result)

type holder = object
  contents: outer

proc main() = 
  # Something about this specific case causes the destructor to not be called.
  var t: holder
  t.contents = newOuter()
  # Putting the above two lines outside of the proc also causes the destructor
  # to be called correctly.
  # new(t.contents) # This works correctly
  # var u = holder(contents: newOuter()) # So does this

main()
```

### Current Output
```
create
```

### Expected Output
```
create
destroy
```

### Additional Information

```
Compiled at 2020-10-18

git hash: 78cbb52fa01d6c73307fe276bdf7d4c34d7a556c
```

This problem occurs at least in 1.2.6 and 1.4.0 too, with both the C and C++ backends. Both ARC and ORC are affected, the other GC strategies don't seem to reliably call destructors anyway(?).

cooldome:
Fix #15629

### DIFF
--- a/compiler/semtypes.nim
+++ b/compiler/semtypes.nim
@@ -285,7 +285,7 @@ proc semRange(c: PContext, n: PNode, prev: PType): PType =
     localError(c.config, n.info, errXExpectsOneTypeParam % "range")
     result = newOrPrevType(tyError, prev, c)
 
-proc semArrayIndex(c: PContext, n: PNode): PType =
+proc semArrayIndex(c: PContext, n: PNode): PType =  
   if isRange(n):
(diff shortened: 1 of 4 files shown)


Araq:
@narimiran this needs to be backported to 1.4

PR #15906, opened 2020-11-10
Correct all eggs

narimiran:
@timotheecour as the author of most of these "egs", please write "e.g." in the future.

### DIFF
--- a/build_all.sh
+++ b/build_all.sh
@@ -1,7 +1,7 @@
 #! /bin/sh
 
 # build development version of the compiler; can be rerun safely.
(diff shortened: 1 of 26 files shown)


juancarlospaco:
They are abbreviations by definition, if you are taking the time, would it be better to change them by the full word `for example` ?. :thinking:

narimiran:
> They are abbreviations by definition, if you are taking the time ...

`e.g.` is an abbreviation. `eg` is nothing more than a misspelled `egg`.

PR #10670, opened 2019-02-14
Httpclient improvements

LemonBoy:
I really hate github.

PR #10668, opened 2019-02-13
Httpclient improvements

LemonBoy:
- Teach httpclient that `308` is just another `302`
- Let me pass a damn `Uri` without having to convert it into a string only to convert it back to a `Uri`
- Use `HttpMethod` enum in more places
- Secure by default, do not send `Authorization` on redirection to a new domain

PR #15919, opened 2020-11-11
Httpclient improvements

planetis-m:
Rebase old pr #10670 by @LemonBoy.


planetis-m:
I think it's also worth redoing this [commit](https://github.com/nim-lang/Nim/pull/10668/commits/7ead616d4c02733bd9a728ffa03e00c0c1d9e69c) that uses HttpMethod instead of string.

planetis-m:
Ready for review @Araq, @dom96

planetis-m:
This needs a review @dom96 it's ready

lib/pure/httpclient.nim
```diff
@@ -325,9 +325,12 @@ proc getDefaultSSL(): SslContext =
       result = defaultSslContext
       doAssert result != nil, "failure to initialize the SSL context"
 
-proc newProxy*(url: string, auth = ""): Proxy =
+proc newProxy*(url: Uri | string, auth = ""): Proxy =
```
dom96: Instead of doing this, overload the procs.

lib/pure/httpclient.nim
```diff
@@ -457,29 +460,23 @@ proc sendFile(socket: Socket | AsyncSocket,
...
+    result.anchor = parsedLocation.anchor
+  else:
+    result = parsedLocation
+
+proc generateHeaders(requestUrl: Uri, httpMethod: HttpMethod, headers: HttpHeaders,
                      proxy: Proxy): string =
   # GET
-  let upperMethod = httpMethod.toUpperAscii()
-  result = upperMethod
+  result = $httpMethod
```
dom96, line 480: Why this change? Is the output the same? (uppercase?)

lib/pure/httpclient.nim
```diff
@@ -1037,33 +1033,86 @@ proc request*(client: HttpClient | AsyncHttpClient, url: string,
   ##
   ## You need to make sure that the ``url`` doesn't contain any newline
   ## characters. Failing to do so will raise ``AssertionDefect``.
-  doAssert(not url.contains({'\c', '\L'}), "url shouldn't contain any newline characters")
+  ## **Deprecated since v1.5**: use HttpMethod enum instead; string parameter httpMethod is deprecated
```
dom96, line 1041:
```suggestion
  ## 
  ## **Deprecated since v1.5**: use HttpMethod enum instead; string parameter httpMethod is deprecated
```

lib/pure/httpclient.nim
```diff
@@ -1037,33 +1033,86 @@ proc request*(client: HttpClient | AsyncHttpClient, url: string,
...
+  when url is string:
+    doAssert(not url.contains({'\c', '\L'}), "url shouldn't contain any newline characters")
```
dom96, line 1043: This seems like it should be thrown by `parseUri` below rather than handled here.

lib/pure/httpclient.nim
```diff
@@ -1037,33 +1033,86 @@ proc request*(client: HttpClient | AsyncHttpClient, url: string,
...
+    of Http301, Http302, Http303:
+      # The method is changed to GET unless it is GET or HEAD (RFC2616)
+      if httpMethod notin {HttpGet, HttpHead}:
+        redirectMethod = HttpGet
+      else:
+        redirectMethod = httpMethod
+      # The body is stripped away
+      redirectBody = ""
+      # Delete any header value associated with the body
+      if headers != nil:
```
dom96:
```suggestion
      if not headers.isNil():
```

lib/pure/httpclient.nim
```diff
@@ -1037,33 +1033,86 @@ proc request*(client: HttpClient | AsyncHttpClient, url: string,
...
+    # Check if the redirection is to the same domain or a sub-domain (foo.com
+    # -> sub.foo.com)
+    if redirectTo.hostname != lastURL.hostname and
+      not redirectTo.hostname.endsWith("." & lastURL.hostname):
+      # Perform some cleanup of the header values
+      if headers != nil:
+        # Delete the Host header
+        headers.del("Host")
+        # Do not send any sensitive info to a unknown host
+        headers.del("Authorization")
```
dom96, line 1116: Is this behaviour part of the HTTP spec?

tests/stdlib/thttpclient.nim
```diff
@@ -39,7 +39,7 @@ proc makeIPv6HttpServer(hostname: string, port: Port,
 
 proc asyncTest() {.async.} =
   var client = newAsyncHttpClient()
-  var resp = await client.request("http://example.com/")
+  var resp = await client.request("http://example.com/", "GET")
```
dom96: Our tests shouldn't have deprecations warnings ideally.

dom96 (approved):
Overall looks good. Please resolve my comments though.

Also, would be nice if we could get test coverage for this redirect logic.

lib/pure/httpclient.nim
planetis-m: yes, variable `httpMethod` changed from `string` to `enum` and the `HttpMethod` enum has all uppercase string values (defined in httpcore.nim)

lib/pure/httpclient.nim
planetis-m: Agreed, should be handled by uri. Maybe #15028 has to be done differently or I can put the condition elsewere.

planetis-m: Well I don't really understand the original PR intentions (nor the Uri spec)  should be done in a seperate PR on uri module, for now I leave it as is.

lib/pure/httpclient.nim
planetis-m: I don't think it's specified (in HTTP/1.0), but seems like a good security measure.

lib/pure/httpclient.nim
```diff
@@ -1087,7 +1089,7 @@ proc request*(client: HttpClient | AsyncHttpClient, url: Uri | string,
       # The body is stripped away
       redirectBody = ""
       # Delete any header value associated with the body
-      if headers != nil:
+      if not headers.isNil():
```
xflywind, line 1092: Why use `isNil`? AFAIR  `!= nil` style may be preferred.

planetis-m: :shrug:

xflywind: https://discord.com/channels/371759389889003530/371759389889003532/696718991019999252

lib/pure/httpclient.nim
dom96: If it's not part of the spec then we shouldn't introduce it. HTTP is ultimately responsible for security in this context AFAIK and this may break existing applications.

planetis-m: @dom96 Have you checked it's not? Im not the original author of this commit (as it can be seen from the git log)

lib/pure/httpclient.nim
dom96: Instead of guessing intentions let's do what we deem right. Otherwise we will end up with code that nobody understands except the original person who is no longer around to answer for it :)

lib/pure/httpclient.nim
dom96: Okay, I just looked into it. There is some evidence that it is done by other HTTP clients at least: https://stackoverflow.com/a/28671822/492186

lib/pure/httpclient.nim
dom96: Since this is the only thing I think needs to be changed though, can you create a separate PR for this? I will merge this.

dom96 (approved):

planetis-m:
Thanks for merging this!

PR #15926, opened 2020-11-11
Add activeElement and documentElement to document.

treeform:
See:
https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/activeElement
https://developer.mozilla.org/en-US/docs/Web/API/Document/documentElement


ringabout:
@treeform 
wrong branch, should be devel instead

PR #15927, opened 2020-11-11
Add support for FreeBSD/PowerPC64 Little Endian

lenoil98:


tools/niminst/buildsh.nimf
```diff
@@ -159,7 +159,10 @@ case $ucpu in
       COMP_FLAGS="$COMP_FLAGS -m64"
       LINK_FLAGS="$LINK_FLAGS -m64"
       mycpu=`uname -p`
-	else
+      if [ "$(mycpu)" = "powerpc64le"] ; then
+	mycpu="powerpc64el"
```
ghost, line 163: is this really needed? doesn't uname -p return that exact value (you can see mycpu above defined as uname -p) ?

ghost: ah wait this is fully redundant :)

tools/niminst/makefile.nimf
```diff
@@ -124,6 +124,9 @@ ifeq ($(ucpu),powerpc)
     mycpu = $(shell sh -c 'uname -p | tr "[:upper:]" "[:lower:]"')
     CFLAGS += -m64
     LDFLAGS += -m64
+    ifeq ($(mycpu),powerpc64le)
+      mycpu = powerpc64el
```
ghost, line 128: this is redundant as well, why are you comparing to powerpc64le and then assigning mycpu to its exact value? :D

ghost:
hmm, I get the mycpu assignments now, but should we really have them here in 3 different places?

ghost:
I guess we do

PR #15929, opened 2020-11-12
Add Trash

juancarlospaco:
- Add OS Trash support, as alternative to `removeFile`, interesting for library authors.
- Allows to trash several files with the same name.
- Documentation with links, `runnableExamples`, `since`, changelog, etc.

```nim
import os
writeFile("example.txt", "")
echo moveFileToTrash("example.txt")
moveFileFromTrash(getCurrentDir() / "example.txt")
```

```nim
import os
echo moveFileToTrash("example.txt", trashPath = "/my/custom/trash/")
moveFileFromTrash(getCurrentDir() / "example.txt", trashPath = "/my/custom/trash/")
```

### DIFF
--- a/lib/pure/os.nim
+++ b/lib/pure/os.nim
@@ -3342,9 +3342,10 @@ since (1, 5):
     ## * `tryRemoveFile proc <#tryRemoveFile,string>`_
     ## * `removeFile proc <#removeFile,string>`_
     runnableExamples:
-      writeFile("example.txt", "")
-      moveFileToTrash("example.txt")
-      moveFileFromTrash(getCurrentDir() / "example.txt")
+      when not defined(js) and not defined(nimscript):
+        writeFile("example.txt", "")
+        moveFileToTrash("example.txt")
+        moveFileFromTrash(getCurrentDir() / "example.txt")
 
     assert filename.len > 0, "filename must not be empty string"
     if dirExists(trashPath):
--- a/lib/pure/os.nim
+++ b/lib/pure/os.nim
@@ -3303,7 +3303,7 @@ func isValidFilename*(filename: string, maxLen = 259.Positive): bool {.since: (1
 
 
 since (1, 5):
-  let osTrash*: string =
+  let osTrash* {.noWeirdTarget.}: string =
     when defined(linux) or defined(bsd):
       getEnv("XDG_DATA_HOME", getHomeDir()) / ".local/share/Trash"
     elif defined(osx):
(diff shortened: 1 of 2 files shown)


lib/pure/os.nim
```diff
@@ -3300,3 +3300,58 @@ func isValidFilename*(filename: string, maxLen = 259.Positive): bool {.since: (1
     find(f.name, invalidFilenameChars) != -1): return false
   for invalid in invalidFilenames:
     if cmpIgnoreCase(f.name, invalid) == 0: return false
+
+
+since (1, 5):
+  const osTrash*: string =
```
ghost: uhh, this shouldn't be a const, since you're getting environment variables

lib/pure/os.nim
```diff
@@ -3300,3 +3300,58 @@ func isValidFilename*(filename: string, maxLen = 259.Positive): bool {.since: (1
...
+    ## * http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html
+    ##
+    ## See also:
+    ## * `tryRemoveFile proc <#tryRemoveFile,string>`_
+    ## * `removeFile proc <#removeFile,string>`_
+    runnableExamples:
+      when not defined(js) and not defined(nimscript):
+        writeFile("example.txt", "")
+        moveFileToTrash("example.txt")
+        moveFileFromTrash(getCurrentDir() / "example.txt")
```
ghost: what if you move two example.txt to trash and want to restore them both? I think you need to append a unique prefix, and in ToTrash remove the actual path where the file was moved to

lib/pure/os.nim
```diff
@@ -3300,3 +3300,58 @@ func isValidFilename*(filename: string, maxLen = 259.Positive): bool {.since: (1
...
+    ## * `removeFile proc <#removeFile,string>`_
+    assert filename.len > 0, "filename must not be empty string"
+    discard existsOrCreateDir(trashPath)
+    when defined(linux) or defined(bsd):
+      let fullPath = expandFilename(filename)
+      let fname = extractFilename(fullPath)
+      discard existsOrCreateDir(trashPath / "files")
+      discard existsOrCreateDir(trashPath / "info")
+      moveFile(fullPath, trashPath / "files" / fname)
+      writeFile(trashPath / "info" / fname & ".trashinfo",
```
ghost: For better efficiency, you should allocate a string and do `add` on it later so you don't have a lot of concatenations

lib/pure/os.nim
```diff
@@ -3300,3 +3300,89 @@ func isValidFilename*(filename: string, maxLen = 259.Positive): bool {.since: (1
     find(f.name, invalidFilenameChars) != -1): return false
   for invalid in invalidFilenames:
     if cmpIgnoreCase(f.name, invalid) == 0: return false
+
+
+since (1, 5):
+  let osTrash* {.noWeirdTarget.}: string =
```
metagn: This is a top level let with side effects, in a very common stdlib module. Shouldn't this be a proc? You could have a top level `{.threadvar.}` var and assign to it every time the proc is called like [`defaultSslContext` in `httpclient`](https://github.com/nim-lang/Nim/blob/19a11bc8a66509c8f23cf315070b633ec228fa1a/lib/pure/httpclient.nim#L316-L326), but in httpclient defaultSslContext never needs a different value, and here I'm not sure what happens to the value if you call it again later.

disruptek:
Please, a library first.  This has limited appeal and it’s in prime stdlib real estate.  :-1:

Araq:
I like it but it needs to be in fusion.

PR #15932, opened 2020-11-12
close #2771(add testcase for #2771)

### LINKED ISSUE #2771: Embedded templates with iterators fails to compile
The following code fails to compile. I'm wrapping templates with templates, and if the passed value is an iterator, then I get **Error: illformed AST** message.

``` nimrod
# this compiles:
template t1(i: int): int=
  i+1
template t2(i: int): int=
  i+1

echo t1(10).t2()


# this doesn't:
template it1(i: int): iterator(): int =
  iterator result(): int {.closure, gensym.} =
    yield i+1
  result

template it2(iter: iterator(): int): iterator(): int =
  iterator result(): int {.closure, gensym.} =
    yield iter()+1
  result

let x = it1(10).it2()
```

If I save the intermediate result to a variable, then it compiles and runs fine:

``` nimrod
let x = it1(10)
let y = x.it2()
```

ringabout:


ringabout:
I will merge this?

PR #15943, opened 2020-11-12
Add documentElement to Document.

treeform:
See: https://developer.mozilla.org/en-US/docs/Web/API/Document/documentElement


treeform:
From https://github.com/nim-lang/Nim/pull/15926

PR #12325, opened 2019-10-01
expose processClient

disruptek:
This allows a server to be written that can capture exceptions yielded by this procedure without leaving its event loop.

Issue #3324, opened 2015-09-14
Could asynchttpserver export processClient?


tulayang:
Hi, I am writting something to transfer client sockets between multiprocess via Unix domain IPC. I found `asynchttpserver.processClient` very useful, but `asynchttpserver.serve` is not. I need a custom `serve`.

Could `asynchttpserver` export `processClient` via `*`?

dom96:
Why do you need to process the clients manually?

tulayang:
@ dom96 

I need to call processClient, my code like this:

```
if existsEnv("ASYNC_CLUSTER"):  ## multi instances mod
    var pipefd = paramStr(1).parseInt().AsyncFD()  ## a socketpair fd from argv
    var data = $port.int() & "\r\n" & address & "\r\n"  ## pass port and adress to master process
    await pipefd.send(data, {}) 
    recvAlways(oneCallback)  ## "accept" fd from master process, call  processClient                                
else:  ## single instance mod
    bindAddr()
    listen()
    appcetAddr()
    ...
```

When used  multi instances (processes) mod, I need to pass client descroptor to child process, then the child will call `processClient`.

dom96:
I still don't understand why ``serve`` doesn't suffice.

Araq:
This is finally possible and will also be backported.

PR #15957, opened 2020-11-13
attempt to make asynchttpserver better; fixes #15925; [backport:1.0]

### LINKED ISSUE #15925: The "try except" not work when the "OSError: Too many open files" error occurs!
In certain circumstances the "try except" not work when the "OSError: Too many open files" error occurs!
This issue causes the program to die. when there are many connections.

### Example
```nim
import asynchttpserver, asyncdispatch, asyncnet, posix

var fdLim: RLimit
discard getrlimit(RLIMIT_NOFILE, fdLim)
fdLim.rlim_cur = 5 # fdLim.rlim_max
echo "max-max fd = ", fdLim.rlim_max, " soft = ", fdLim.rlim_cur
discard setrlimit(RLIMIT_NOFILE, fdLim)

proc fileServer(req: Request, staticDir="") {.async.} =
  await req.client.send(
    "HTTP/1.1 200\c\Lcontent-type: text/html\c\Lcontent-length: 3\c\L\c\Lhi\n")

  try:
    await req.fileServer("static")
  except:
    echo "Happen a Error!"
    await req.respond(Http200, "Happen a Error!")

waitFor server.serve(Port(8080), cb)
```
To test:
```
terminal1$ ./bug4
terminal2$ curl http://localhost:8080/
terminal1 output:
  max-max fd = 4096 soft = 5
<waiting; then after curl immediate program exit with:>
  bug4.nim(21) bug4
  /usr/lib/nim/lib/pure/asyncdispatch.nim(1934) waitFor
  /usr/lib/nim/lib/pure/asyncdispatch.nim(1626) poll
  /usr/lib/nim/lib/pure/asyncdispatch.nim(1367) runOnce
  /usr/lib/nim/lib/pure/asyncdispatch.nim(208) processPendingCallbacks
  /usr/lib/nim/lib/pure/asyncmacro.nim(22) serveNimAsyncContinue
  /usr/lib/nim/lib/pure/asyncmacro.nim(139) serveIter
  /usr/lib/nim/lib/pure/asyncfutures.nim(372) read
  [[reraised from:
  bug4.nim(21) bug4
  /usr/lib/nim/lib/pure/asyncdispatch.nim(1936) waitFor
  /usr/lib/nim/lib/pure/asyncfutures.nim(372) read
  ]]
  Error: unhandled exception: Too many open files
  Async traceback:
    bug4.nim(21)            bug4
    /usr/lib/nim/lib/pure/asyncdispatch.nim(1934) waitFor
    /usr/lib/nim/lib/pure/asyncdispatch.nim(1626) poll
    /usr/lib/nim/lib/pure/asyncdispatch.nim(1367) runOnce
    /usr/lib/nim/lib/pure/asyncdispatch.nim(208)  processPendingCallbacks
    /usr/lib/nim/lib/pure/asyncmacro.nim(22)      serveNimAsyncContinue
    /usr/lib/nim/lib/pure/asyncmacro.nim(139)     serveIter
    /usr/lib/nim/lib/pure/asyncfutures.nim(372)   read
  Exception message: Too many open files
  Exception type: [OSError]
```

### Additional Information
https://forum.nim-lang.org/t/7064
```
$ nim -v
Nim Compiler Version 1.4.0
```

Araq:


disruptek:
And just like that, #12325 was fixed.  👍

Araq:
@disruptek Yeah, I had a vague memory of your PR. Good we're finally sorting it out.

ringabout:
There is an issue releated to `processClient`: https://github.com/nim-lang/Nim/issues/3324

lib/pure/asynchttpserver.nim
```diff
@@ -21,14 +21,27 @@
...
+##           "Content-type": "text/plain; charset=utf-8"}
+##       await req.respond(Http200, "Hello World", headers.newHttpHeaders())
+##
+##     server.listen Port(5555)
+##     while true:
+##       if server.shouldAcceptRequest(5):
+##         var (address, client) = await server.socket.acceptAddr()
+##         asyncCheck processClient(server, client, address, cb)
+##       else:
+##         poll()
```
dom96, line 41: This example is under "basic usage", you shouldn't complicate it.

lib/pure/asynchttpserver.nim
```diff
@@ -309,9 +325,44 @@ proc serve*(server: AsyncHttpServer, port: Port,
...
+            callback: proc (request: Request): Future[void] {.closure, gcsafe.}) {.async.} =
+  ## Accepts a single request. Write an explicit loop around this proc so that
+  ## errors can be handled properly.
+  var (address, client) = await server.socket.acceptAddr()
+  asyncCheck processClient(server, client, address, callback)
+
+proc serve*(server: AsyncHttpServer, port: Port,
+            callback: proc (request: Request): Future[void] {.closure, gcsafe.},
+            address = "";
+            assumedDescriptorsPerRequest = 5) {.async.} =
```
dom96, line 346: This is very arbitrary and therefore likely error prone.

lib/pure/asynchttpserver.nim
```diff
@@ -309,9 +325,44 @@ proc serve*(server: AsyncHttpServer, port: Port,
...
+  ## When a request is made by a client the specified callback will be called.
+  ##
+  ## If `assumedDescriptorsPerRequest` is 0 or greater the server cares about
+  ## the process's maximum file descriptor limit. It then ensures that the
+  ## process still has the resources for `assumedDescriptorsPerRequest`
+  ## file descriptors before accepting a connection.
+  ##
+  ## You should prefer to call `acceptRequest` instead with a custom server
+  ## loop so that you're in control over the error handling and logging.
+  listen server, port, address
```
dom96, line 359: This shouldn't be here, unless I'm missing something.

dom96 (changes requested):
If I understand the problem this is solving then I think you can simplify this and make it clearer to the users:

* ``maxClients`` argument to AsyncHttpServer
* increment a `clientsCount` counter in `processClient`
* decrement after connection is closed
* don't ``accept`` unless ``clientsCount`` < ``maxClients``.

lib/pure/asynchttpserver.nim
Araq: People copy&paste this section, it needs to be reasonably complete rather than simple. In fact, I should probably add some basic error handling there too.

lib/pure/asynchttpserver.nim
Araq: It's **much** better than no limit and limits are always arbitrary.

lib/pure/asynchttpserver.nim
Araq: It has to be there thanks to bad API design (it should be part of ``newAsyncHttpServer`` IMHO but this would be a breaking change).

Araq:
> If I understand the problem this is solving then I think you can simplify this and make it clearer to the users:

This wouldn't work as well because serving a single async request can require multiple different FDs, say if you load data from a DB before serving the request. It's much more robust to ask the event loop instead.

c-blake:
I think I agree with @Araq here.  At first my reaction was kind of like @dom96, but then I thought about it a little more.  The current way or its mathematical complement (clients limit & `maxDescriptors()` subtraction) way are basically equivalent control, but a user is probably more likely to be able to estimate/bound how many new descriptors a worst case request might need.  So, if it were the max clients way, most users wanting to be safe would be doing that same subtraction to go back to the complementary sense, anyway.  It is also easier to default "spare fds"/show in example code this way.

There might be a "silent underprovisioning" argument to be made.  The least max fds I've ever seen in 30 years of using Unix is 64, HP-UX in the 1990s, IIRC.  So, std(in|out|err)+cpl overhead + 5 headroom still leaves (probably) ~54 concurrent connections in the crazy worst case.  That's actually still a lot of concurrent connections, TBH.  Except for debugging/testing like for this problem, people never lower such small fd limits.  So the "spare fds way" doesn't really risk underprovision, and the complex guesstimation above kind of exhibits what happens to careful users in the "complementary coordinates".  Such guestimation seems ugly & more error prone..You may be a forked process inheriting a bunch of fds, too.  So, I think specifying from the spares/headroom side is best.

I do think having a 5 line API call to crank up the soft max to the hard max (somewhere) would be helpful..basically just:
```Nim
import posix # add more error checks, of course
var fdLim: RLimit
discard getrlimit(RLIMIT_NOFILE, fdLim)
fdLim.rlim_cur = fdLim.rlim_max
discard setrlimit(RLIMIT_NOFILE, fdLim)
```
Maybe call it `softToHardDescriptorLimit` or something like that.  (In truth the only errors possible here are `EFAULT` for a bad `fdLim` addr and `EINVAL` for a bad `RLIMIT_NOFILE`, but it's good to check all one's errors).  May even make sense to layer it, like `softToHardLimit(resource = RLIMIT_NOFILE)` so it could work for any of the RLIMIT_*.

lib/pure/asyncdispatch.nim
```diff
@@ -1934,3 +1934,27 @@ proc waitFor*[T](fut: Future[T]): T =
...
+when defined(posix):
+  import posix
+
+when defined(linux) or defined(windows) or defined(macosx) or defined(bsd):
+  proc maxDescriptors*(): int {.raises: OSError.} =
+    ## Returns the maximum number of active file descriptors for the current
+    ## process. This involves a system call. For now `maxDescriptors` is
+    ## supported on the following OSes: Windows, Linux, OSX, BSD.
+    when defined(windows):
+      result = 16_700_000
```
timotheecour, line 1955: factor with hardcoded value in
```nim
nimMaxDescriptorsFallback* {.intdefine.} = 16_000 
```
(unless meant to be different?)

lib/pure/asynchttpserver.nim
timotheecour, line 346: factor `assumedDescriptorsPerRequest = 5` with the constant used in 
```nim
proc shouldAcceptRequest*(server: AsyncHttpServer;
                          assumedDescriptorsPerRequest = 5): bool {.inline.} =
```

timotheecour:
@araq good PR but please always squash and merge, this PR will break `git bisect` workflows on which I and others rely a lot to find regressions/accidental bugfixes etc, because it keeps intermediate broken commits

image


can we do the proposal in https://github.com/nim-lang/Nim/issues/8664 would prevent such issues in future:
> in nim-lang/Nim/settings, disable "Allow merge commits", as follows: [...]

(and even force squash and merge); in the rare cases where a squash and merge isn't desirable, the other merge strategies could be enabled temporarily

lib/pure/asynchttpserver.nim
dom96: Won't this literally break code? You'll have code in the wild calling `listen` twice now.

Edit: oh, I see, didn't realise you introduced the `listen`.

dom96 (commented):
> This wouldn't work as well because serving a single async request can require multiple different FDs, say if you load data from a DB before serving the request. It's much more robust to ask the event loop instead.

Then perhaps it should be part of asyncdispatch itself?

lib/pure/asyncdispatch.nim
Araq: It's the right, Windows-specific value. Meant to be different, yes.

Araq:
> Won't this literally break code? You'll have code in the wild calling listen twice now.

But previously ``listen`` wasn't exported so I don't see how this could happen.

Araq:
> Then perhaps it should be part of asyncdispatch itself?

The APIs that are added are part of asyncdispatch, yes.

c-blake:
The best place to limit fd creation for a network service is just before `accept(2)`.  Then the usual `listen(2)` backlog will just keep the requester clients in limbo until fds have been freed up and then `accept(2)` as usual.  I believe that's what @Araq did.

Araq:
> (and even force squash and merge); in the rare cases where a squash and merge isn't desirable, the other merge strategies could be enabled temporarily

I know, it was a simple mistake. Sometimes github defaults to the wrong button for some reason.

dom96:
> The APIs that are added are part of asyncdispatch, yes.

No, I mean the handling of FDs. Wouldn't it be better to have it handled in ``accept`` automatically?

c-blake:
There could be a `safeAccept`, I suppose, that no-ops if there aren't enough spare fds, but probably the event loop then also needs to be aware of a no-op return, as opposed to just successful or an error case to raise.

Araq:
Maybe, but this would be an even more invasive change. Also: People requested an API like mine for asynchttpserver before, nobody requested this feature for ``accept``.

c-blake:
Technically, this same problem exists for literally every system call returning a file descriptor.  So, some kind of `safeOpen` template-y/-ish system might be warranted someday/somehow, but I think that could live side-by-side with @Araq's API.

dom96:
>  Then the usual listen(2) backlog will just keep the requester clients in limbo until fds have been freed up and then accept(2) as usual.

Huh, this implies that the OS should handle this for us then. If so why do we need a separate mechanism in the stdlib?

c-blake:
The OS still limits your open fds.  What I meant is that the clients will not see an ECONNREFUSED-type situation until the backlog buffer gets filled up.  So, just not doing the accept syscall is more graceful, deferring failure if there is just a very transient spike in connections.

Araq:
> Technically, this same problem exists for literally every system call returning a file descriptor. So, some kind of safeOpen template-y/-ish system might be warranted someday/somehow, but I think that could live side-by-side with @Araq's API.

That's true but please consider:

This PR with an API change will be backported to version 1.0.x. I consider it a critical omission -- more critical than other bugfixes that we backported -- but at the same time I tried my best to keep the changes to a minimum.

c-blake:
Oh, sure.  I was just responding to the @dom96 idea of `safeAccept`.  Re: the listen backlog/overall dynamics, it is also true that few logical situations allow you to just punt to a no-op and try again later.  So, utility of `safeOpen` may be pretty limited.

dom96:
Does anyone have a real repro for issue #15925? All I see are synthetic code samples that don't seem realistic to me. I'm quite confused about how this would crop up in real code without it being due to a real FD leak.

Araq:
> Does anyone have a real repro for issue #15925? 

As far as I can know this affects our very own forum software. (Still investigating though.)

c-blake:
The origin case was in the Forum thread.  The `wrk` http demon workhorse tester with 1024 parallel connections (-c 1024).  That is a lot more than is typical organically, unless you are under a DoS attack or a Big Tech company. (EDIT: but our stuff should still handle this scenario.)

timotheecour:
[How to prevent a program from crashing when the "OSError: Too many open files" error occurs? - Nim forum](https://forum.nim-lang.org/t/7064)

dom96:
Well... we just tested his exploit and it doesn't do anything. Furthermore, I checked the server and it has been running for 2 months without crashes.

dom96:
So my recommendation is to change this slightly:

* It shouldn't be the default
* The docs should be reverted to the basic example
* Docs should be provided showing how to use this new API

lib/pure/asynchttpserver.nim
```diff
@@ -309,9 +325,44 @@ proc serve*(server: AsyncHttpServer, port: Port,
...
+proc shouldAcceptRequest*(server: AsyncHttpServer;
+                          assumedDescriptorsPerRequest = 5): bool {.inline.} =
+  ## Returns true if the process's current number of opened file
+  ## descriptors is still within the maximum limit and so it's reasonable to
+  ## accept yet another request.
+  result = assumedDescriptorsPerRequest < 0 or
+    (activeDescriptors() + assumedDescriptorsPerRequest < server.maxFDs)
+
+proc acceptRequest*(server: AsyncHttpServer, port: Port,
+            callback: proc (request: Request): Future[void] {.closure, gcsafe.}) {.async.} =
```
dom96, line 337: also, the `port` here is unused

Araq: Ah, good one. Will do a follow-up PR later.

c-blake:
I'm not sure how the Forum server process never died, but you also said server not server process.  So, maybe ambiguous?  I didn't look at the Forum code, anyway.

I do think it is not great example code/library behavior to just exhaust file descriptors.  Reserving at least a few for the user code just makes very basic sense to me.  I'm not sure why anyone would be against it.  You need at least 1 free for the `accept` itself to even work.

My simpler test https://github.com/nim-lang/Nim/issues/15925#issuecomment-726031666 behaved as one would expect with that `wrk` http test prog from `strace` s that I did.  I haven't yet tried it with the recent @Araq change.  Could be wrinkles.

Araq:
> I'm not sure how the Forum server process never died, but you also said server not server process. So, maybe ambiguous? I didn't look at the Forum code, anyway.

The Forum *process* itself never died.

c-blake:
I don't know what's going on with the Forum recovery from overload, but this simple program:
```Nim
import asynchttpserver, asyncdispatch, asyncnet, posix

var fdLim = RLimit(rlim_cur: 7, rlim_max: 1024)
discard setrlimit(RLIMIT_NOFILE, fdLim)
const
  s = "HTTP/1.1 200\r\ncontent-type: text/html\r\ncontent-length: 3\r\n\r\nHi\n"
proc svc(req: Request, staticDir="") {.async.} = await req.client.send(s)
proc cb(req: Request) {.async.} = await req.svc("static")

waitFor server.serve(Port(8080), cb)
```
dies immediately with the old code even for `wrk --latency -d 1s -t 1 -c 3 http://localhost:8080/` yet runs fine with `wrk --latency -d 1s -t 1 -c 1024 http://localhost:8080/` with the new code. Raising an avoidable exception under overload (without an easy plan to clear the condition) seems bad while gracefully handling massive overload seems good.  So, I think the new code is both more robust & a better example of managing a scarce (on Unix) resource.  Like all code, it may be imperfect.

PR #15969, opened 2020-11-14
Documentation improve search

juancarlospaco:
- https://caniuse.com/input-search


timotheecour:
@juancarlospaco you can use `-d:fixup` to fix the test (need to carefully manually check the new output is correct though)
```
# To run this, cd to the git repo root, and run "nim c -r nimdoc/tester.nim".
# to change expected results (after carefully verifying everything), use -d:fixup
```

juancarlospaco:
"Works on my machine" :shrug:

timotheecour:
> "Works on my machine" 🤷

could you be using a different nim (eg PATH issue)?

it fails on mine; all i did is: 
* checkout your PR
* compile nim to bin/nim_15969
* bin/nim_15969 c -r nimdoc/tester.nim

the fix, again, is simply to run with -d:fixup, checking that new files are good, then re-running `c -r nimdoc/tester.nim`

juancarlospaco:
That test makes evolving the html of docs unnecessarily complicated for no gain.

timotheecour:
> That test makes evolving the html of docs unnecessarily complicated for no gain.

@juancarlospaco it's a bit rigid but it helps prevents regressions. If you have a better idea for that, happy to hear them.

I tried your PR, it's a reasonable improvement.

image

image

re-opening is welcome

Using `-d:fixup` has always worked for me everytime i changed something in docs, why would it not work here?

PR #15976, opened 2020-11-15
remove conf.writelnHook assignment inside runCmd

bung87:
conf.writelnHook should controlled before runCmd, since we already have nimsuggest instance.

### DIFF
--- a/nimsuggest/nimsuggest.nim
+++ b/nimsuggest/nimsuggest.nim
@@ -736,12 +736,8 @@ else:
     var retval: seq[Suggest] = @[]
     let conf = nimsuggest.graph.config
     conf.ideCmd = cmd
-    conf.writelnHook = proc (line: string) =
-      retval.add(Suggest(section: ideMsg, doc: line))
     conf.suggestionResultHook = proc (s: Suggest) =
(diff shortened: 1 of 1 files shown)


nimsuggest/nimsuggest.nim
```diff
@@ -736,12 +736,8 @@ else:
     var retval: seq[Suggest] = @[]
     let conf = nimsuggest.graph.config
     conf.ideCmd = cmd
-    conf.writelnHook = proc (line: string) =
-      retval.add(Suggest(section: ideMsg, doc: line))
     conf.suggestionResultHook = proc (s: Suggest) =
       retval.add(s)
-    conf.writelnHook = proc (s: string) =
-      stderr.write s & "\n"
```
Araq, line 744: But this changes the behavior, doesn't it?

bung87: yeah, maybe  let initNimsuggest take a conf arg to resolve this

bung87: the bad part is current implementation will print duplicated messages to stderr that cant be stoped.

PR #15998, opened 2020-11-16
Clean parsesql

bung87:


lib/pure/parsesql.nim
```diff
@@ -1424,7 +1424,7 @@ proc ra(n: SqlNode, s: var SqlWriter) =
     for i in 1..n.len-1:
       if i > 1: s.add(',')
       ra(n.sons[i], s)
-    s.add(");")
```
Araq, line 1427: This changes the existing behaviour.

lib/pure/parsesql.nim
```diff
@@ -1443,7 +1443,7 @@ proc ra(n: SqlNode, s: var SqlWriter) =
     for i in 2..n.len-1:
       if i > 2: s.add(", ")
       ra(n.sons[i], s)
-    s.add(");")
```
Araq, line 1446: Same.

lib/pure/parsesql.nim
bung87: the extro semicolon for what , I dont like double semicolon at statement end.

PR #16021, opened 2020-11-17
Add parseFileUri function.

moigagoo:


moigagoo:
While fixing a bug in nimlsp [1], I've noticed that there is no obvious way to convert a file URI into a path in Nim. This is useful, for example, when working with the Language Server Protocol, which uses file URI spec to communicate the file paths to the LSP server [2].

I've added a simple function that basically cuts the necessary platform-specific prefix (7 chars on Posix, 8 on Windows) and normalizes the path. This should not break anything.

It is a trivial addition but I think it's useful to have such a function in the stdlib, so that people wouldn't have to figure out how to parse those file URIs like I had to.

1: https://github.com/PMunch/nimlsp/issues/38
2: https://microsoft.github.io/language-server-protocol/specifications/specification-current/

lib/pure/os.nim
```diff
@@ -3301,3 +3301,23 @@ func isValidFilename*(filename: string, maxLen = 259.Positive): bool {.since: (1
...
+  ##     winUri = "file:///C:/foo/bar.nim"
+  ##     unixUri = "file:///etc/foo/bar.nim"
+  ##
+  ##   when defined(windows):
+  ##     assert parseFileUri(winUri) == "C:\\foo\\bar.nim"
+  ##     assert parseFileUri(unixUri) == "etc\\foo\\bar.nim"
+  ##   when defined(posix):
+  ##     assert parseFileUri(unixUri) == "/etc/foo/bar.nim"
+  ##     assert parseFileUri(winUri) == "/C:/foo/bar.nim"
+  # https://tools.ietf.org/html/rfc8089
```
juancarlospaco:
```suggestion
  ##
  ## See also:
  ## * https://tools.ietf.org/html/rfc8089
```

lib/pure/os.nim
```diff
@@ -3301,3 +3301,23 @@ func isValidFilename*(filename: string, maxLen = 259.Positive): bool {.since: (1
...
+
+  let startIdx = when defined(windows): 8 else: 7
+
+  result = normalizedPath(uri[startIdx..^1])
```
juancarlospaco, line 3325:
```suggestion
  result = normalizedPath(uri[(when defined(windows): 8 else: 7) .. ^1])
```

dom96 (commented):
Maybe we could implement this on the `Uri` object defined in the `uri` module? It should be able to parse these `file://` schemes, maybe we could define a `toFilePath` proc on it?

lib/pure/os.nim
moigagoo: @juancarlospaco I do not agree with this suggestion. It makes the code less readable and generally feels like shortness for the sake of shortness.

SolitudeSF:
you refer to 8089 rfc, but this implements almost nothing from the rfc.

moigagoo:
> Maybe we could implement this on the Uri object defined in the uri module? It should be able to parse these file:// schemes, maybe we could define a toFilePath proc on it?

@dom96 This was my initial intention. But three reasons made me choose otherwise:
1. This function requires normalizedPath from os, I didn't want to introduce a new import in uri.
2. If it were in uri, it had to deal with other kinds of URIs, whereas file URI is in fact a very specific kind of URI. Other URI don't necessarily resolve to paths. I mean, if we had a func called toPath in uri, you'd expect it to work with something like http://foo.com/bar, which it can't do anything with.
3. os module deals with paths, and the goal of this function is to produce a path. Yes, the input is a URI, but we're calling it for its output, which has nothing to do with URIs.

I understand that it's more logical to look for a function like this in uri, and this is what I did when I was looking for an existing function for my task. But the second best bet is os. In fact, I just use global search. My point is the the function is nearly as discoverable in os as it would be in uri.

moigagoo:
@SolitudeSF 

> you refer to 8089 rfc, but this implements almost nothing from the rfc.

Well, yes, the goal was not to implement the RFC. The link is there to describe the input string format. Maybe, a Wikipedia link would be a better reference: https://en.wikipedia.org/wiki/File_URI_scheme

Araq:
Add it to fusion instead please and also add some checking, don't blindly strip away the prefixes.

PR #16028, opened 2020-11-18
follow #15994 for block _

ringabout:
before this PR compiles
```nim
block _:
  echo "123"
  break _
```
after this PR
```
"undeclared identifier: '_'"
```


cooldome:
Not sure about this one. You can have a block without a label if you need anonymous block. Also `_` is a valid variable name, it is used in many DSLs written in Nim so it will stay like this. My opinion `_` is special only in tuple unpacking and for loops where variable is not used. 
@Araq,
What is your opinion?

Araq:
Seems slightly worse than today's solution. Plus consistency is overrated.

PR #16029, opened 2020-11-18
almostEqual: fix for Inf values and add an option for NaN

lbartoletti:
almostEqual contains a bug with Inf values, where `Inf == Inf` returns true, `almostEqual(Inf, Inf)` returns false.

BTW, I propose to add an option to handle NaN case when we hope that NaN == NaN returns true; which is not normally the case, but can sometimes be useful. This option exists in numpy for example.

### DIFF
--- a/lib/pure/math.nim
+++ b/lib/pure/math.nim
@@ -158,7 +158,7 @@ proc classify*(x: float): FloatClass =
     return fcSubnormal
   return fcNormal
 
-proc almostEqual*[T: SomeFloat](x, y: T; unitsInLastPlace: Natural = 4): bool {.
+proc almostEqual*[T: SomeFloat](x, y: T; unitsInLastPlace: Natural = 4, equalNan:  bool = false): bool {.
       since: (1, 5), inline, noSideEffect.} =
   ## Checks if two float values are almost equal, using
   ## `machine epsilon <https://en.wikipedia.org/wiki/Machine_epsilon>`_.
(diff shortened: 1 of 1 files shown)


lib/pure/math.nim
```diff
@@ -173,10 +173,27 @@ proc almostEqual*[T: SomeFloat](x, y: T; unitsInLastPlace: Natural = 4): bool {.
...
+    doAssert almostEqual(Inf, Inf)
+    doAssert almostEqual(-Inf, -Inf)
+    doAssert not almostEqual(Inf, -Inf)
+    doAssert not almostEqual(-Inf, Inf)
+    doAssert not (NaN == NaN)
+    doAssert not almostEqual(NaN, NaN)
+    doAssert almostEqual(NaN, NaN, equalNan = true)
+
+  if (equalNan == true) and (classify(x) == fcNan) and (classify(y) == fcNan):
+      return true
```
Araq:
```suggestion
    return true
```

lib/pure/math.nim
```diff
@@ -158,7 +158,7 @@ proc classify*(x: float): FloatClass =
     return fcSubnormal
   return fcNormal
 
-proc almostEqual*[T: SomeFloat](x, y: T; unitsInLastPlace: Natural = 4): bool {.
+proc almostEqual*[T: SomeFloat](x, y: T; unitsInLastPlace: Natural = 4, equalNan:  bool = false): bool {.
```
Araq:
```suggestion
proc almostEqual*[T: SomeFloat](x, y: T; unitsInLastPlace: Natural = 4, equalNan = false): bool {.
```

planetis-m:
python.isClose doesn't have the same option, math docs mention:
```
The IEEE 754 special values of NaN, inf, and -inf will be handled according to IEEE rules. Specifically, NaN is not considered close to any other value, including NaN. inf and -inf are only considered close to themselves.
```

Don't know why we want to mimic numpy, there is already `classify` proc available, but whatever others say.

lbartoletti:
> Don't know why we want to mimic numpy,  there is already `classify` proc available

You're right, and I use it. Sometimes, you want to check if a value is NaN so isClose can also be a shortcut with this option (and we don't respect the IEEE754 rule).

It's a just a suggestion ;), we can delete it or do another function if you find it relevant (it also removes an if). Let me know

planetis-m:
It's totally fine by me, I was just voicing my concerns.

lib/pure/math.nim
```diff
@@ -173,10 +173,27 @@ proc almostEqual*[T: SomeFloat](x, y: T; unitsInLastPlace: Natural = 4): bool {.
   ## and multiplied by the desired precision in ULPs unless the difference is
   ## subnormal.
   ##
+  ## `equalNan` Whether to compare NaN’s as equal. If True, NaN’s in a will be
```
ringabout, line 176:
```suggestion
  ## `equalNan` Whether to compare NaN's as equal. If True, NaN's in a will be
```

lib/pure/math.nim
```diff
@@ -173,10 +173,27 @@ proc almostEqual*[T: SomeFloat](x, y: T; unitsInLastPlace: Natural = 4): bool {.
...
+  ## considered equal to NaN’s in b. Same as `Numpy
```
ringabout, line 177:
```suggestion
  ## considered equal to NaN's in b. Same as `Numpy
```

lib/pure/math.nim
ringabout: Maybe you could also rephrase the comments

Araq:
Please extract the bugfix from the feature addition because I don't like the feature addition. If you know NaNs are a thing it's easy enough to call classify on your own first.

planetis-m:
Done #16122

PR #16030, opened 2020-11-18
Semfold for nil cast

cooldome:
In commit e6e1e9574de8c090cb9de7a809064037e2af39f7 semfold for nil was removed, because it caused issues for codegen.
I am fixing codegen in this PR, hence semfold for nil can be done again.


pb-cdunn:
This commit broke my codebase, which now fails to compile. It's a bit large, and I don't have any info yet on what fails. But I narrowed it down to this commit by bisecting in git and re-running `koch boot`.

See #16148

pb-cdunn:
Btw, e6e1e9574 works fine. The first commit that breaks me is this one, 87d3e53

PR #16040, opened 2020-11-18
RST tables: fix latex col number; allow <= 3 of =

a-mr:
Fix 2 minor errors in RST tables:
* some Latex tables are incorrectly generated because number of columns is confused with number of rows (e.g. long LaTeX table with > 40 rows when provided columns are > 40 Latex got finally confused and messed up generation). See incorrect generation of tables in pdf version of `doc/gc.rst` and `doc/intern.rst`
* tables with short number (<= 3) of equals signs in horizontal borders/adornments lead to an `Error: new section expected`. Fixed by adding the exception: an adornment starting with `=` is never turned into `tkPunct` even when it's short


Araq:
I wrote this code against the RST spec so please point out where in the spec the special case is made for "=".

a-mr:
Yes, there is no mentioning of such a special case.
But there is also no requirement for adornment to have more than 3 punctuation character.
You must have confused something or used an old version of the spec (that code line with "<= 3" is very ancient, it dates back to 2009-12-07).

There is a requirement for [transitions](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#id48) to have more than 3 punctuation characters.

Regarding chapters it's ok to be short, e.g. such chapter is correct:
```rst

Ch
--

```
and indeed `rst2html` from docutils parses it as a chapter, while nim does not.

So it seems that the real problem here is that 3 different objects
* adornments
* transitions
* simple table horizontal borders

use the same basic token tkAdornment.
It will be more difficult to fix it properly than I thought.

Araq:
> that code line with "<= 3" is very ancient, it dates back to 2009-12-07.

Indeed, we've only recently begun to modernize it. :-)

a-mr:
Another situation that rst.nim appeared to be slightly over-restrictive:
```rst
Fails to compile:
:v: ab .... cd
```
It's obvious that rst.nim sees all long punctuation rows (like .... in the example) as adornments and tries to make sense of them and fails.

I added a commit above that adds a context variable `adornmentLines` to lexer. When rst.nim encounters punctuation character at the beginning of a line, it checks that all the line consists from the same characters (or whitespaces for tables) to be accepted as an adornment.

Of course, that required other adjustments: now literal block markers `::` and markdown code block \`\`\` can go either into category of adornment or punctuation, depending on what line they were found on.

Also it has become obvious that rst.nim did not check for adornment length before accepting it as a legal underline for the section.
[While according to the spec](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#id47) section is delimited by an underline, which is:

> single repeated punctuation character that begins in column 1 and forms a line extending at least as far as the right edge of the title text

This became more noticeable in situations like this:
```
Some chapter name
-
```
Without the length check that would be accepted as a section.

a-mr:
@narimiran , @Araq PTAL

lib/packages/docutils/rst.nim
```diff
@@ -1067,35 +1117,29 @@ proc whichSection(p: RstParser): RstNodeKind =
...
-      rstMessage(p, meGridTableNotImplemented)
     elif isDefList(p):
       result = rnDefList
     elif isOptionList(p):
       result = rnOptionList
     else:
       result = rnParagraph
   of tkWord, tkOther, tkWhite:
-    if match(p, tokenAfterNewline(p), "ai"): result = rnHeadline
+    let tokIdx =  tokenAfterNewline(p)
```
Araq:
```suggestion
    let tokIdx = tokenAfterNewline(p)
```

Araq:
Looks ok to me but @narimiran should review it too.

a-mr:
@narimiran please review

lib/packages/docutils/rst.nim
```diff
@@ -1052,13 +1098,26 @@ proc findPipe(p: RstParser, start: int): bool =
...
+    elif match(p, p.idx, "+a+"):
+      result = rnGridTable
+      rstMessage(p, meGridTableNotImplemented)
     elif match(p, p.idx + 1, " a"): result = rnTable
-    elif match(p, p.idx + 1, "i"): result = rnOverline
-    elif isMarkdownHeadline(p):
-      result = rnHeadline
+    elif currentTok(p).symbol == "|" and isLineBlock(p):
+      result = rnLineBlock
+    elif match(p, p.idx + 1, "i") and isAdornmentHeadline(p, p.idx):
```
narimiran, line 1119: Please double check:

Should this be `isAdornmentHeadline(p, p.idx + 1):`?

a-mr: No, second parameter to `isAdornmentHeadline` is the index of adornment itself. So for overline adornment it's p.idx in this `tkAdornment` branch. Note that cases of underline and overline are processed differently in `isAdornmentHeadline`:
* underline case is assumed when second argument > current p.idx and headline length is counted from p.idx till newline
* overline case — when second argument = p.idx and headline is assumed to start from p.idx+2

PR #16058, opened 2020-11-19
fixes #15836; WIP, [backport:1.4]

### LINKED ISSUE #15836: Dangerous implicit type conversion from auto + generics
A combination of generics and the `auto` type can trigger dangerous implicit type conversions.

### Example
```nim
proc takesProc[T](x: T, f: proc(x: T): int) =
    echo f(x) + 2

takesProc(1, proc (a: int): int = 2) # ok, prints 4
takesProc(1, proc (a: auto): auto = 2) # ok, prints 4
takesProc(1, proc (a: auto): auto = "uh uh") # prints garbage
```

### Expected Output

Final call to `takesProc` should not compile.

### Additional Information

A close to real world example:
```nim
type
    Object = object
        field: int64

var values = @[
    Object(field: -1688270553),
    Object(field: -1592610305),
    Object(field: 1301173200),
    Object(field: 1414267200)
]

values.sort((a, b) => a.field < b.field) # Wrong! Comparer function should return `int` and not `bool`.
                                         # Compilation succeeds anyway but with unexpected runtime results
```

Araq:


Araq:
@mratsim This makes arraymancer tests fail but it might be that the compiler now rightly complains about one of your tests cases. Please investigate.

mratsim:
Looking in the error: https://github.com/nim-lang/Nim/pull/16058/checks?check_run_id=1430415082#step:11:1064

image

And it's location: https://github.com/mratsim/Arraymancer/blob/4ae9b81180ac7d6e5361baa9655f4f8d8100b074/tests/nn_primitives/test_nnp_convolution.nim#L153

It's a compile-time error and I think it's likely that sugar `=>` has an `auto` regression

Araq:
Ok, so in other words, you're confirming that your test cases are correct. Thanks, will investigate further.

Araq:
Refresh.

PR #16063, opened 2020-11-20
fix #16022 Serializing and deserializing a  SIGSEV's

### LINKED ISSUE #16022: Serializing and deserializing a `proc` SIGSEV's
```nim
import marshal
let p : proc () = proc () = echo "hello world"
(to[typeof(p)]($$p))()
```

### Current Output
```
/home/deech/Nim/test.nim(3)
Error: execution of an external program failed: '/home/deech/Nim/test.nim '
```

### Expected Output
```
hello world
```

bung87:


ringabout:
see alternative fix #16064

Araq:
The alternative was preferable because the code used to work. (Even though it's foolish that it ever worked.)

PR #16064, opened 2020-11-20
alternative way to fix #16022

### LINKED ISSUE #16022: Serializing and deserializing a `proc` SIGSEV's

ringabout:
Let's see whether this could cause memory pollution
https://github.com/nim-lang/Nim/issues/9754

The test for #9754
https://github.com/nim-lang/Nim/blob/devel/tests/stdlib/t9754.nim

`genicAssign` will copy the memory of pointer.

### DIFF
--- a/lib/core/typeinfo.nim
+++ b/lib/core/typeinfo.nim
@@ -327,10 +327,10 @@ proc setPointer*(x: Any, y: pointer) =
   ## ``akString``, ``akCString``, ``akProc``, ``akRef``, ``akPtr``,
   ## ``akPointer``, ``akSequence``.
   assert x.rawType.kind in pointerLike
-  if y != nil:
+  if y != nil and x.rawType.kind != tyPointer:
     genericAssign(x.value, y, x.rawType)
   else:
-    cast[ppointer](x.value)[] = nil
+    cast[ppointer](x.value)[] = y
 
 proc fieldsAux(p: pointer, n: ptr TNimNode,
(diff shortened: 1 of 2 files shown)


bung87:
well, vmmarshal will not handle this case , json not, please see the marshal description "This module contains procs for serialization and deserialization of arbitrary Nim data structures. The serialization format uses JSON."

ringabout:
It's a regression. It works before Nim v1.2.8 and Nim v 1.0.10.

> please see the marshal description "This module contains procs for serialization and deserialization of  arbitrary Nim data structures. The serialization format uses JSON."

The JSON is 
```nim
import marshal
let p: proc () = proc () = echo "hello world"
let poc = (to[typeof(p)]($$p))

echo $$p
```
output:
```
{"Field0": 93875474016942, "Field1": null}
```
If you think `marshal` shouldn't support serializing and deserializing `proc`, please also document this at `marshal` docs in your PR

Araq:
If this is the "alternative way to fix" what was the other way?

Araq:
Never mind.

bung87:
I dont get this idea, JSON dont have function, the code make it work at certain runtime point, if you store your marshal data to somewhere then load it , you get what ?

PR #16070, opened 2020-11-20
ref #16054 remove useless stuff in JS backend

ringabout:
I think it is related to dead code elimination regarding var stmt in JS backend, is it expected or bug or known issue in JS backend?

BTW could someone give me some hints where dead code elimination  is executed?

### DIFF
--- a/lib/system.nim
+++ b/lib/system.nim
@@ -1897,53 +1897,54 @@ else:
 # however, stack-traces are available for most parts
 # of the code
 
-var
-  globalRaiseHook*: proc (e: ref Exception): bool {.nimcall, benign.}
-    ## With this hook you can influence exception handling on a global level.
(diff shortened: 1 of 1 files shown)


ringabout:
```nim
echo ":("
```

Before this PR:
```js
var globalRaiseHook_33557677 = [null];
if (globalThis.localRaiseHook_33557682 === undefined) {
  localRaiseHook_33557682 = [null];
}
var outOfMemHook_33557685 = [null];
var unhandledExceptionHook_33557690 = [null];
```
After this PR:
these codes are removed

Araq:
> BTW could someone give me some hints where dead code elimination is executed?

It's not an optimization at all, instead the codegen only emits code for top level statements (and exportc'ed procs) and then recursively for what they called.

PR #16071, opened 2020-11-20
fix #14056

### LINKED ISSUE #14056: xmlparser removes significant white space
If the parseXml procedure is used to create an XmlNode object from the following XML/HTML:

`<p>one <em>two</em> three</p>`

and the XmlNode object is then immediately serialized again, the result is:

`<p>one <em>two</em>three</p>`

That is, the white space preceding "three" (which is clearly significant) is missing from the output version. It gets lost during parsing, not during serialization as far as I can see.

I haven't tested it, but I suspect that the following input:

`<p>one <em>two</em> <strong>three</strong> four</p>`

would be output as:

`<p>one <em>two</em><strong>three</strong>four</p>`

That is, the space between "two" and "three" would be lost as well as the space before "four".

kemifl:
I have added a boolean whitespace parameter (default false) to all the parsexml() procedures, If it's set to true, then the xnWhitespace option is set on the underlying parser, and white space is not ignored by the parser.

I also added a test for this option to one of the test files.

This is my first nim PR, so please let me know if there are things I have done wrong.

### DIFF
--- a/tests/stdlib/tmitems.nim
+++ b/tests/stdlib/tmitems.nim
@@ -16,7 +16,18 @@ fpqeew
 <Students>
   <Student Name="Aprilfoo" />
   <Student Name="bar" />
-</Students>'''
+</Students>
+<chapter>
+    <title>This is a Docbook title</title>
+    <para>
+        This is a Docbook paragraph containing <emphasis>emphasized</emphasis>,
+        <literal>literal</literal> and <replaceable>replaceable</replaceable>
+        text. Sometimes scrunched together like this:
+        <literal>literal</literal><replaceable>replaceable</replaceable>
+        and sometimes not:
+        <literal>literal</literal> <replaceable>replaceable</replaceable>
+    </para>
+</chapter>'''
 """
 
 block:
@@ -123,7 +134,7 @@ block:
     x.num += 10
   echo j
 
-import xmltree, xmlparser, streams, strtabs
+import xmltree, xmlparser, parsexml, streams, strtabs
 
 block:
   var d = parseXml(newStringStream """<Students>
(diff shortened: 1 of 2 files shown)


cooldome:
I would rather say pass extra argument of type `XmlParseOption` and have default value for it

kemifl:
Ah, of course, that makes sense.

I'll update the branch.

cooldome:
Great, and of course I meant `set[XmlParseOption]`

tests/stdlib/tmitems.nim
```diff
@@ -134,3 +145,17 @@ block:
...
+    <para>
+        This is a Docbook paragraph containing <emphasis>emphasized</emphasis>,
+        <literal>literal</literal> and <replaceable>replaceable</replaceable>
+        text. Sometimes scrunched together like this:
+        <literal>literal</literal><replaceable>replaceable</replaceable>
+        and sometimes not:
+        <literal>literal</literal> <replaceable>replaceable</replaceable>
+    </para>
+</chapter>""",{reportComments, reportWhitespace})
+  echo d
```
cooldome, line 161: I am afraid `echo d` doesn't test if behavior is correct. 
We need something like this to test:
```nim

block:
  const para = """
        This is a Docbook paragraph containing <emphasis>emphasized</emphasis>,
        <literal>literal</literal> and <replaceable>replaceable</replaceable>
        text. Sometimes scrunched together like this:
        <literal>literal</literal><replaceable>replaceable</replaceable>
        and sometimes not:
        <literal>literal</literal> <replaceable>replaceable</replaceable>
"""
  var d = parseXml(newStringStream """<chapter>
    <title>This is a Docbook title</title>
    <para>""" & para &  """</para>
</chapter>""",{reportComments, reportWhitespace})

  doAssert(d.selectParaSomeHow == para)
```

Code above completely untested. Idea that we need `doAssert` checks to confirm spaces are preserved

kemifl:
I understand what you are saying, but the test I created follows the
pattern of all the other tests in the same file. The output from the echo
is matched against a string defined in a discard block at the top of the
file. I checked that it passes, and that it doesn't pass if the output
doesn't match the string at the top of the file.

Maybe I put the test in the wrong place?

kemifl:
I understand what you are saying, but the test I created follows the
pattern of all the other tests in the same file. The output from the echo
is matched against a string defined in a discard block at the top of the
file. I checked that it passes, and that it doesn't pass if the output
doesn't match the string at the top of the file.

Maybe I put the test in the wrong place?

PR #16097, opened 2020-11-22
add sugar.exported for exporting macro outputs

metagn:
Limitations:

* While it's probably possible, I don't think it would be wise to indiscriminately export all object fields. This is meant to be a unified way of exporting, but you should be able to choose which fields you export.
* Type and variable macro pragmas cannot be used as `typed` arguments. For typedefs I tried doing `typed{~nkTypeDef}` and overloading for `untyped{nkTypeDef}`, and it still chose the typed overload. I also could not get it to work for variable macro pragmas, but even if I did the signature for the overload would make it hard to distinguish as being specifically for variable macro pragmas. (also you can't know if it's let or var)
* This still outputs `x is declared but not used` because the `typed` argument has to be checked. I don't know how to work around this.

The utopic best case scenario would be modifying the `export` statement syntax and behavior to make it possible to do something like:

```nim
export proc foo(x: int) = echo x
```

which you could then use macro calls on instead of `proc foo = ...`. Many people have complained about postfix `*` and nothing's changed yet so I'm sure there's a reason this wouldn't be optimal.

A question you might have is, why can't we just do `export foo` afterward? You can do this for types and variables but not for routines because it would export every single overload of the symbol that you are exporting.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -47,6 +47,9 @@
 - `repr` now doesn't insert trailing newline; previous behavior was very inconsistent,
   see #16034. Use `-d:nimLegacyReprWithNewline` for previous behavior.
 
+- Added `sugar.exported` for converting unexported statements to exported ones.
+  Can be used for macro outputs.
+
 ## Language changes
 
 - `nimscript` now handles `except Exception as e`
--- a/lib/pure/sugar.nim
+++ b/lib/pure/sugar.nim
@@ -364,6 +364,44 @@ macro collect*(init, body: untyped): untyped {.since: (1, 1).} =
       call.add init[i]
   result = newTree(nnkStmtListExpr, newVarStmt(res, call), resBody, res)
 
+macro exported*(decl: typed) {.since: (1, 5).} =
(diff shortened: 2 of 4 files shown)


metagn:
> * This still outputs `x is declared but not used` because the `typed` argument has to be checked. I don't know how to work around this.

Would like help on this.

planetis-m:
You can use `freshIdentNodes`, I think it might help.

Araq:
Looks ok but it doesn't come up often enough for ``sugar.nim``, this should be its own module and it should be added to fusion instead. (Fusion ships with Nim too.)

metagn:
I made it a nimble package at first but then realized that having to do `nimble install exported` then `import exported` everywhere you need it would be weird so I decided to put it here. Someone else can PR it to fusion without credit to me if they want because the "is not used" warning is not good at all and would probably be best solved by changing the compiler in some arcane way that I could never come up with, so I'm not bothering. It's super small code anyway I'm sure something could be done better in it

Issue #5617, opened 2017-03-27
Ability to set a NimNode's lineinfo


dom96:
The only way to set this information is via ``newNimNode``, but a lot of the time I have a node that was created via ``newCall`` or some other convenience function. These functions don't allow me to specify the line info information so it would be nice to be able to assign it after the fact.

dom96:
Reminder for myself (or others if they wish to take this on): go through ``asyncmacro`` and give every new node line info, especially [here](https://github.com/nim-lang/Nim/commit/e0bb65e45c8dec9683740f1a656f72f7eaa4d32f#diff-5933d70ab845f7daf7c238319f0b55b6R133).

dom96:
Another request:

Please expose something that will allow me to make these unwieldy async errors clearer. Take a look at this:

```
Traceback (most recent call last)
server.nim(213)          server
asyncdispatch.nim(1081)  waitFor
asyncdispatch.nim(1072)  poll
asyncdispatch.nim(1242)  cb
asyncfutures.nim(118)    complete
asyncdispatch.nim(1387)  :anonymous
asyncfutures.nim(118)    complete
asyncmacro.nim(34)       cb
asyncmacro.nim(334)      sendIter
asyncfutures.nim(118)    complete
asyncmacro.nim(34)       cb
asyncmacro.nim(334)      verifyWebsocketRequestIter
asyncfutures.nim(109)    complete
asyncmacro.nim(50)       cb
asyncfutures.nim(156)    fail
asyncmacro.nim(50)       cb
asyncfutures.nim(156)    fail
asyncmacro.nim(50)       cb
asyncfutures.nim(156)    fail
asyncfutures.nim(322)    :anonymous
Error: unhandled exception: index out of bounds
  onRequest's lead up to read of failed Future:
    Traceback (most recent call last)
    server.nim(213)          server
    asyncdispatch.nim(1081)  waitFor
    asyncdispatch.nim(1072)  poll
    asyncdispatch.nim(1242)  cb
    asyncfutures.nim(118)    complete
    asyncdispatch.nim(1387)  :anonymous
    asyncfutures.nim(118)    complete
    asyncmacro.nim(34)       cb
    asyncmacro.nim(334)      sendIter
    asyncfutures.nim(118)    complete
    asyncmacro.nim(34)       cb
    asyncmacro.nim(334)      verifyWebsocketRequestIter
    asyncfutures.nim(109)    complete
    asyncmacro.nim(34)       cb
[...]

PR #16102, opened 2020-11-23
ref #5617 add lineinfo to complete

ringabout:


cooldome:
I am happy to accept PR as is, but we need to think of ways to test lineinfo.
I generally find MOST of established macros (for example in sugar module) need to care about line info we need regression tests to make sure it stays correct.

PR #16111, opened 2020-11-24
improve the documentation of ropes

ringabout:
image


lib/pure/ropes.nim
```diff
@@ -171,28 +220,73 @@ proc `&`*(a, b: Rope): Rope {.rtl, extern: "nroConcRopeRope".} =
...
 
 proc `&`*(a: openArray[Rope]): Rope {.rtl, extern: "nroConcOpenArray".} =
-  ## the concatenation operator for an openarray of ropes.
-  for i in countup(0, high(a)): result = result & a[i]
+  ## The concatenation operator for an openarray of ropes.
+  runnableExamples:
+    let s = @[rope("Hello, "), rope("Nim"), rope("!")]
+    let r = &s
+    doAssert $r == "Hello, Nim!"
+  for item in a: result = result & item
```
xflywind, line 250: small improvement here

ringabout, line 217: small improvement here
`for i in countup(0, high(a)): result = result & a[i]` > `for item in a: result = result & item`

lib/pure/ropes.nim
```diff
@@ -15,6 +15,39 @@
...
+##|"this string"    con
+##|                /   \
+##|               /     \
+##|              /       \
+##|             /         \
+##|            /           \
+##|" is internally "        "represented as"
+##|
+##|  The 'con' operator is associative! This does not matter however for
+##|  the algorithms we use for ropes.
```
Araq: This ASCII art is excessive and doesn't help much. Please remove it again. Also: Ropes are slower than mutable strings and we don't encourage anybody to use them. (In the Nim compiler they were a mistake.)

Issue #16119, opened 2020-11-24
Copying of a sink parameter into a variable not allowed


planetis-m:
### Example
module A
```nim
type
  Foo*[T] = object
    v*: T
proc `=destroy`*[T](f: var Foo[T]) =
  echo "destroyed ", f.v
proc `=copy`*[T](f: var Foo[T], s: Foo[T]) =
  echo "copied ", f.v

proc foo*[T](t: sink Foo[T]) =
  var t = t #same for let
  t.v = T(2)
```
module B
```nim
import a

var a = Foo[int](v: 1)
foo(a)
echo a
```
### Current Output
`b.nim(4, 4) template/generic instantiation of `foo` from here`
`a.nim(10, 7) Error: invalid type: 'sink Foo[system.int]' for var`

### Additional Information

Introduced in #15690 

```
$ nim -v
Nim Compiler Version 1.5.1 [Linux: amd64]
Compiled at 2020-11-24
git hash: d306a04466b7f1129620dc3ab35443119ed4c867
```

PR #16121, opened 2020-11-24
Add package manu to tested packages

planetis-m:
While not important it uses destructors and can help catch regressions early. Blocked by #16120 and #16119


Araq:
Please rebase, should be green now.

PR #16122, opened 2020-11-24
fix for comparing infinities

planetis-m:
Alternative to #16029


lib/pure/math.nim
```diff
@@ -177,6 +177,17 @@ proc almostEqual*[T: SomeFloat](x, y: T; unitsInLastPlace: Natural = 4): bool {.
...
+    doAssert almostEqual(-Inf, -Inf)
+    doAssert almostEqual(Inf, -Inf) == false
+    doAssert almostEqual(-Inf, Inf) == false
+    doAssert almostEqual(Inf, NaN) == false
+    doAssert almostEqual(NaN, NaN) == false
+
+  if x == y:
+    # short circuit exact equality -- needed to catch two infinities of
+    # the same sign. And perhaps speeds things up a bit sometimes.
+    return true
```
Araq, line 190: Please use ``result = true`` and an ``else``.

Issue #16110, opened 2020-11-24
`macros.getImpl` still return incorrect AST for Sym "XXX:ObjectType" 


jangko:
First of all, I would like to thank you for all improvements since nim-lang/RFCs#176.
Instead of returning `nilLit`, `macros.getImpl` now gives us some AST. But unfortunately, the AST is still incorrect.

```Nim
import macros

macro see(x: type): untyped =
  let z = getType(x)[1]
  let y = getImpl(z)
  debugEcho getType(x).treeRepr
  debugEcho "---"
  debugEcho y.treeRepr
```

We will use the same macro code above for all examples below.

I'll start with ordinary object:
```Nim
type
  Tire = object
    code: int

see(Tire)
```

The result is:
```text
BracketExpr
  Sym "typeDesc"
  Sym "Tire"
---
TypeDef
  Sym "Tire"
  Empty
  ObjectTy
    Empty
    Empty
    RecList
      IdentDefs
        Ident "code"
        Sym "int"
        Empty
```

Now, we will use ref object:
```Nim
type
  TireRef = ref Tire

var x: TireRef
type XX = type x[]  # should be equal to `Tire`
see(XX)
```

output still ok:
```text
BracketExpr
  Sym "typeDesc"
  Sym "Tire"
---
TypeDef
  Sym "Tire"
  Empty
  ObjectTy
    Empty
    Empty
    RecList
      IdentDefs
        Ident "code"
        Sym "int"
        Empty
```

Last, we will try it with ref/ptr shortcut:
```Nim
type
  TirePtr = ptr object
    code: int

var z: TirePtr
type ZZ = type z[] # the deref works but the internal representation is somewhat weird
see(ZZ)
```

the result is still `PtrTy`, not `ObjectTy`, and the Sym is incorrect:
```text
BracketExpr
  Sym "typeDesc"
  Sym "TirePtr:ObjectType"  # this already correct
---
TypeDef
  Sym "TirePtr"   # how could Sym "TirePtr:ObjectType" become Sym "TirePtr"?
  Empty
  PtrTy                # `PtrTy` should be removed
    ObjectTy        # we only need this `ObjectTy` part
      Empty
      Empty
      RecList
        IdentDefs
          Ident "code"
          Sym "int"
          Empty
```

## Expected output

```text
BracketExpr
  Sym "typeDesc"
  Sym "TirePtr:ObjectType"
---
TypeDef
  Sym "TirePtr:ObjectType"
  Empty
[...]

PR #16130, opened 2020-11-25
More on #16110

cooldome:
My previous fix for #16110 caused pragmas to dissappear from AST.
More tests, preserve pragmas.


Clyybber:
This doesn't fix `PtrTy` still existing in the ast

cooldome:
Good point, fixed it. Test added

PR #16145, opened 2020-11-26
fix #16120

### LINKED ISSUE #16120: Const seq into sink arg: =copy operator not found

cooldome:
The problem was that constant args never went through createTypeBoundOps, due to this line:
```nim
if getConstExpr(tracked.ownerModule, n, tracked.c.idgen, tracked.graph) != nil:
    return
```

Tha change is minimal, replaced:
```nim
if getConstExpr(tracked.ownerModule, n, tracked.c.idgen, tracked.graph) != nil:
  return
```
with
```nim
if getConstExpr(tracked.ownerModule, n, tracked.c.idgen, tracked.graph) == nil:
  ...
```    
and moved ops in or out from this if statement, no changes to ops themselves.

### DIFF
--- a/compiler/sempass2.nim
+++ b/compiler/sempass2.nim
@@ -755,63 +755,66 @@ proc trackCall(tracked: PEffects; n: PNode) =
   if n.typ != nil:
     if tracked.owner.kind != skMacro and n.typ.skipTypes(abstractVar).kind != tyOpenArray:
       createTypeBoundOps(tracked, n.typ, n.info)
-  if getConstExpr(tracked.ownerModule, n, tracked.c.idgen, tracked.graph) != nil:
-    return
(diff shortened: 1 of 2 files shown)


Araq:
@narimiran Backport to 1.4 please.

PR #16171, opened 2020-11-28
Update Documentation db_mysql (devel)

shayanhabibi:
Adding notation to help new users use impure libraries that does not have the dependency included with nim.

Was running into some errors trying to update my previous fork to devel so just reforked and pulling again.


Araq:
Nim doesn't ship other db specific libraries either. And why would it, you're supposed to install MySQL if you need MySQL. The remarks add more confusion than they help, sorry.

PR #16190, opened 2020-11-30
fix broken => for typed procs with pragmas

metagn:


lib/pure/sugar.nim
```diff
@@ -66,12 +63,12 @@ macro `=>`*(p, b: untyped): untyped =
 
     type
       Bot = object
-        call: proc (): string {.nosideEffect.}
+        call: proc (name: string): string {.noSideEffect.}
 
     var myBot = Bot()
 
-    myBot.call = () {.nosideEffect.} => "I'm a bot."
```
ringabout, line 73: Please move original example(or both) to tests.

Clyybber: Why?

ringabout: The PR removes the original example, add that to test in case of regression.
And it seems not stable enough considering causing this issue. So it needs more tests,

Clyybber: Ah ok, thanks for clarifying.

metagn:
To give some more info in case this looks sketchy (I wrote this a little quick):

I introduced problems with syntax handling in #14200. I wrote very minimal tests and this feature was pretty broken the way I implemented it.

Before, I had made it so if the left hand side was a single expression wrapped in parentheses it would expand it work on that. I did this so `((x, y: int) -> int) {.noSideEffect.}` would work, which I think is too ugly for it to be allowed to work now, and it was not documented before. It broke expressions like `(x: int)`. I removed this handling, now you have to do `(x, y: int) {.noSideEffect.} -> int`.

It was also broken for expressions like `foo(x, y: int)` because this was an `ObjConstr` expression and not a `Call` expression. This is now supported.

ringabout:
@hlaaftana 
Good job!

PR #16195, opened 2020-11-30
fix #16185

### LINKED ISSUE #16185: [ARC] Memory zeroed after `=sink` even for self assignment
Working on https://github.com/mratsim/Arraymancer/pull/477 I've chased a weird segfault originating for self assignments of tensors, which only happens for tensors stored in objects. 
I've reduced it to a relatively simple piece of code now, which mostly reproduces the bug. However, in my actual code the issue also happens if compiled without `--gc:arc`(sink is used in that case even for default GC. Does that make sense? Or should the `=sink` procs always be guarded by `defined(gcDestructors)`?).

Or maybe I still misunderstand how to use ARC?

### Example
```nim
type
  CpuStorage*[T] = ref CpuStorageObj[T]
  CpuStorageObj[T] = object
    size*: int
    raw_buffer*: ptr UncheckedArray[T]
  Tensor[T] = object
    buf*: CpuStorage[T]
  TestObject = object
    x: Tensor[float]

proc `=destroy`[T](s: var CpuStorageObj[T]) =
  s.raw_buffer.deallocShared()
  s.size = 0

proc `=destroy`[T](t: var Tensor[T]) =
  `=destroy`(t.buf)

proc `=destroy`(t: var TestObject) =
  `=destroy`(t.x)

proc `=`[T](a: var CpuStorageObj[T]; b: CpuStorageObj[T]) {.error.}

#proc `=sink`[T](a: var CpuStorageObj[T], b: CpuStorageObj[T]) =
#  if a.raw_buffer != b.raw_buffer:
#    `=destroy`(a)
#    wasMoved(a)
#    a.raw_buffer = b.raw_buffer
#    a.size = b.size
#
#proc `=sink`[T](a: var Tensor[T], b: Tensor[T]) =
#  if a.buf != b.buf:
#    `=destroy`(a)
#    wasMoved(a)
#    `=sink`(a.buf, b.buf)

proc allocCpuStorage[T](s: var CpuStorage[T], size: int) =
  new(s)
  s.raw_buffer = cast[ptr UncheckedArray[T]](allocShared0(sizeof(T) * size))
  s.size = size

proc newTensor[T](size: int): Tensor[T] =
  allocCpuStorage(result.buf, size)

proc `[]`[T](t: Tensor[T], idx: int): T = t.buf.raw_buffer[idx]
proc `[]=`[T](t: Tensor[T], idx: int, val: T) = t.buf.raw_buffer[idx] = val

proc toTensor[T](s: seq[T]): Tensor[T] =
  result = newTensor[T](s.len)
  for i, x in s:
    result[i] = x

proc main() =
  var t: TestObject
  t.x = toTensor(@[1.0, 2, 3, 4])
  t.x = t.x
  # the following line holds true even if it shouldn't of course
  doAssert t.x.buf.isNil

main()

# does ``not`` happen if the code is at top level!
#var t: TestObject
#t.x = toTensor(@[1.0, 2, 3, 4])
#t.x = t.x
#doAssert t.x.buf.isNil
```

### Current Output

Nothing, because the `doAssert` falsely holds.

### Expected Output

`doAssert` should not hold, thus program should fail.

### Additional Information

It seems to be a more complex version of https://github.com/nim-lang/Nim/issues/10689 to me.

The generated C code (using `--gc:arc -d:danger`):

```c
	nimZeroMem((void*)(&T1_), sizeof(tyObject_Tensor__IQAwMnmu4uBKMjqXsM9adCA));
	toTensor__UpFt1bYA1ehyF1G4AabWcQ(colontmpD_, (&T1_));
	if (NIM_UNLIKELY(*nimErr_)) goto BeforeRet_;
	eqsink___UU2NWj0I19chxYozquBBXMA((&t.x), T1_);
	eqsink___UU2NWj0I19chxYozquBBXMA((&t.x), t.x);
	nimZeroMem((void*)(&t.x), sizeof(tyObject_Tensor__IQAwMnmu4uBKMjqXsM9adCA));
	{
		if (!!((t.x.buf == 0))) goto LA4_;
		failedAssertImpl__W9cjVocn1tjhW7p7xohJj6A(TM__zIQ3yoETjuVGvXrPQFKeSQ_3);
		if (NIM_UNLIKELY(*nimErr_)) goto BeforeRet_;
	}
```

The latter `nimZeroMem` shouldn't be there, because the `eqsink` before is a self assignment. It doesn't matter if I define the `=sink` in the code or not. 

I just pulled the newest devel:

```
Compiled at 2020-11-29

git hash: d29eddf92a1b8147c4709fd828f77152c1f3aab2
```

cooldome:
fix #16185.

To simplify review:
The change made is the following piece of code:
```nim
 if dest.kind == nkSym and dest.sym == ri.sym:
    # rule (self-assignment-removal):
    result = newNodeI(nkEmpty, dest.info)
```
is replaced with 
```nim
if exprStructuralEquivalent(dest, ri, strictSymEquality = true):
  # rule (self-assignment-removal):
  result = newNodeI(nkEmpty, dest.info)
```
at the top of `moveOrCopy`. No other changes

### DIFF
--- a/compiler/injectdestructors.nim
+++ b/compiler/injectdestructors.nim
@@ -969,67 +969,68 @@ proc p(n: PNode; c: var Con; s: var Scope; mode: ProcessMode): PNode =
       internalError(c.graph.config, n.info, "cannot inject destructors to node kind: " & $n.kind)
 
 proc moveOrCopy(dest, ri: PNode; c: var Con; s: var Scope, isDecl = false): PNode =
-  case ri.kind
-  of nkCallKinds:
-    result = c.genSink(dest, p(ri, c, s, consumed), isDecl)
-  of nkBracketExpr:
-    if isUnpackedTuple(ri[0]):
-      # unpacking of tuple: take over the elements
-      result = c.genSink(dest, p(ri, c, s, consumed), isDecl)
-    elif isAnalysableFieldAccess(ri, c.owner) and isLastRead(ri, c) and
-        not aliases(dest, ri):
-      # Rule 3: `=sink`(x, z); wasMoved(z)
-      var snk = c.genSink(dest, ri, isDecl)
-      result = newTree(nkStmtList, snk, c.genWasMoved(ri))
-    else:
-      result = c.genCopy(dest, ri)
-      result.add p(ri, c, s, consumed)
-      c.finishCopy(result, dest, isFromSink = false)
-  of nkBracket:
-    # array constructor
-    if ri.len > 0 and isDangerousSeq(ri.typ):
-      result = c.genCopy(dest, ri)
-      result.add p(ri, c, s, consumed)
-      c.finishCopy(result, dest, isFromSink = false)
-    else:
-      result = c.genSink(dest, p(ri, c, s, consumed), isDecl)
-  of nkObjConstr, nkTupleConstr, nkClosure, nkCharLit..nkNilLit:
-    result = c.genSink(dest, p(ri, c, s, consumed), isDecl)
-  of nkSym:
-    if dest.kind == nkSym and dest.sym == ri.sym:
-      # rule (self-assignment-removal):
-      result = newNodeI(nkEmpty, dest.info)
-    elif isSinkParam(ri.sym) and isLastRead(ri, c):
-      # Rule 3: `=sink`(x, z); wasMoved(z)
-      let snk = c.genSink(dest, ri, isDecl)
-      result = newTree(nkStmtList, snk, c.genWasMoved(ri))
-    elif ri.sym.kind != skParam and ri.sym.owner == c.owner and
-        isLastRead(ri, c) and canBeMoved(c, dest.typ) and not isCursor(ri, c):
-      # Rule 3: `=sink`(x, z); wasMoved(z)
-      let snk = c.genSink(dest, ri, isDecl)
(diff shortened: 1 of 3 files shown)


Clyybber:
This is not correct, consider
```nim
import random
proc main =
  var s = @[@[1], @[2], @[3]]
  var i = 0
  s[rand(2)] = s[rand(2)]
  echo s

main()
```

cooldome:
@Clyybber,
Agree but what would be the correct fix:
```nim
s[rand(2)] = s[rand(2)]
```
can produce self-assingment as well.
Should we exclude only guaranted the same location and produce if statement for anything that MAY be a self-assignment.

cooldome:
Added a fix for @Clyybber comment

cooldome:
Added a fix for @Clyybber comment. I have reverted part.
```nim
if dest.kind == nkSym and dest.sym == ri.sym:
    # rule (self-assignment-removal):
    result = newNodeI(nkEmpty, dest.info)
```
because `isAnalysableFieldAccess` doesn't covert thread local and global

Clyybber:
>  Should we exclude only guaranted the same location and produce if statement for anything that MAY be a self-assignment.

`=copy` hook handles that for us. So only eliminate the self assignment if we can guarantee it.

Clyybber (approved):

PR #7417, opened 2018-03-26
Add RISC-V "riscv64" support

FedericoCeratto:
Tested by crosscompiling using GCC 8.0.1 from 20180321 (devel version) running koch tests under qemu-riscv64 version 2.11.50 (v2.11.0-2256-g819fd4699c-dirty).

Some tests failed due to missing libraries; The threadpool module failed to compile due to an unrecognized opcode `pause'.

This PR could be merged as a first step to add RISC-V support and more testing can be performed once real hardware is available.

Closes #7147

PR #16231, opened 2020-12-02
fix #16184 add 32-bit RISC-V support

### LINKED ISSUE #16184: RISC-V 32 support
Nim supports riscv64 since 2018.
Now 32 bit microcontrollers are becoming more available, so it would be very nice to support rv32 as a cross-compilation target.

fwsGonzo:
I have a need for 32-bit RISC-V due to memory consumption issues. I followed the conversation in https://github.com/nim-lang/Nim/pull/7417 and added riscv32 platform that way.

Please tell me if there is any missing parts.

I have tested this locally, and I have built a few 32-bit RISC-V Nim programs.
https://gist.github.com/fwsGonzo/cad2c071036be3d6a3038bf93da85a55
-->
```
$ file nim
nim: ELF 32-bit LSB executable, UCB RISC-V, version 1 (SYSV), statically linked, not stripped
```

### DIFF
--- a/compiler/installer.ini
+++ b/compiler/installer.ini
@@ -6,7 +6,7 @@ Name: "Nim"
 Version: "$version"
 Platforms: """
   windows: i386;amd64
-  linux: i386;hppa;ia64;alpha;amd64;powerpc64;arm;sparc;sparc64;m68k;mips;mipsel;mips64;mips64el;powerpc;powerpc64el;arm64;riscv64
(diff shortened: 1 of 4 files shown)


ringabout:
Related: https://github.com/nim-lang/Nim/issues/16184

You could add `fix #16184` in header/description to close this issue automatically.

FedericoCeratto:
Please also update: https://github.com/nim-lang/Nim/blob/devel/lib/system.nim#L1065

PR #16240, opened 2020-12-03
add support for parsing chars in `scanf` macro

narimiran:


lib/pure/parseutils.nim
```diff
@@ -241,6 +241,10 @@ proc parseIdent*(s: string, start = 0): string =
     while i < s.len and s[i] in IdentChars: inc(i)
     result = substr(s, start, i-1)
 
+proc parseChar*(s: string, ident: var char, start = 0): int =
```
xflywind, line 244: need `{.since: (1, 5, 1).}`?

lib/pure/parseutils.nim
```diff
@@ -241,6 +241,10 @@ proc parseIdent*(s: string, start = 0): string =
...
+  ident = s[start]
```
ringabout, line 245: check index(not empty)?

lib/pure/parseutils.nim
ringabout, line 244: need {.since: (1, 5, 1).}?
Maybe this API can not be marked as {.since.}, because it is relied by other module.
See #15860 (comment)

lib/pure/parseutils.nim
Araq: Done in scanf though.

ringabout: But other parse* function did check the index? Since this proc is exported, I'm not sure the proper behaviour.

pietroppeter: I agree with @xflywind. Not that this is any big deal since this proc is much more unlikely - than e.g. parseIdent - to be called independently from scanf. Still it would be better for consistency with the other parseX procs to have the same behaviour. I could do a small PR to fix this since this is merged with this type of content (adding also docstring, a test and maybe updating docstring of parseIdent for consistency): https://play.nim-lang.org/#ix=2Go3

pietroppeter: done: https://github.com/nim-lang/Nim/pull/16245

PR #16245, opened 2020-12-03
fix parseChar see #16240

pietroppeter:
fix minor issue in recent PR by @narimiran (#16240), see:
https://github.com/nim-lang/Nim/pull/16240/files#r535452099

### DIFF
--- a/lib/pure/parseutils.nim
+++ b/lib/pure/parseutils.nim
@@ -211,6 +211,7 @@ proc parseHex*[T: SomeInteger](s: string, number: var T, start = 0,
 proc parseIdent*(s: string, ident: var string, start = 0): int =
   ## Parses an identifier and stores it in ``ident``. Returns
   ## the number of the parsed characters or 0 in case of an error.
+  ## If error, the value of ``ident`` is not changed.
   runnableExamples:
     var res: string
     doAssert parseIdent("Hello World", res, 0) == 5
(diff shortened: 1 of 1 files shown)


lib/pure/parseutils.nim
```diff
@@ -211,6 +211,7 @@ proc parseHex*[T: SomeInteger](s: string, number: var T, start = 0,
 proc parseIdent*(s: string, ident: var string, start = 0): int =
   ## Parses an identifier and stores it in ``ident``. Returns
   ## the number of the parsed characters or 0 in case of an error.
+  ## If error, the value of ``ident`` is not changed.
```
ringabout: Maybe use single quote which is more simple.
```nim
If error, the value of `ident` is not changed.
```

lib/pure/parseutils.nim
```diff
@@ -241,9 +242,19 @@ proc parseIdent*(s: string, start = 0): string =
     while i < s.len and s[i] in IdentChars: inc(i)
     result = substr(s, start, i-1)
 
-proc parseChar*(s: string, ident: var char, start = 0): int =
-  ident = s[start]
-  result = 1
+proc parseChar*(s: string, c: var char, start = 0): int =
+  ## Parses a single character, stores it in ``c`` and returns 1.
+  ## In case of error (if start >= s.len) it returns 0
+  ## and the value of ``c`` is unchanged.
```
ringabout: same as above: use single quote?

lib/pure/parseutils.nim
pietroppeter: I can definitely do that, but then I would change it throughout the document (lots of other places where double backticks appear) to keep it consistent. should I do that?

ringabout: Yeah, two way exists. You don't have to clean them in this PR, just not introducing them is enough.
See https://github.com/timotheecour/Nim/issues/415

pietroppeter: done

PR #16247, opened 2020-12-03
nimdoc: Initialize theme switch and pragma dots on DOMContentLoaded

sebbert:
The default HTML template for nimdoc currently initializes the dark mode switch and pragma dots when the onload event is fired. But since the onload event does not fire until all external resources (images, stylesheets, fonts, etc.) have been loaded, the light theme is shown for a brief moment before the document is fully loaded, and it switches to the dark theme. This is quite jarring, especially on slower internet connections. So let's instead initialize these things on the DOMContentLoaded event, which is fired right after the document has been parsed and the initial DOM structure is ready. This means that we now display the dark mode immediately, without having to wait for external resources to load first.

For reference, see:
- https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event
- https://developer.mozilla.org/en-US/docs/Web/API/Window/DOMContentLoaded_event

I have updated the snapshot tests in `nimdoc/`, and done some manual testing of both the theme switch and the pragma dots, to confirm that this does not break anything. Please let me know if I've missed anything.

### DIFF
--- a/config/nimdoc.cfg
+++ b/config/nimdoc.cfg
@@ -268,10 +268,12 @@ function main() {
     }
   }
 }
+
+window.addEventListener('DOMContentLoaded', main);
 </script>
 
 </head>
-<body onload="main()">
+<body>
 <div class="document" id="documentId">
   <div class="container">
     <h1 class="title">$title</h1>
--- a/nimdoc/rst2html/expected/rst_examples.html
+++ b/nimdoc/rst2html/expected/rst_examples.html
@@ -56,10 +56,12 @@ function main() {
     }
   }
 }
+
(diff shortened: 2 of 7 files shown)


sebbert:
Agh, seems like this fails due to some problems connecting to GitHub to clone repositories for the package tests.
Doesn't seem like I am allowed to re-run jobs from the UI, but I'll try to amend this commit to force a rebuild.

sebbert:
Nope, they're still flaking :( For what it's worth, these changes should have no effect on the package tests, and the docs tests are succeeding.

ringabout:
See https://github.com/nim-lang/Nim/pull/15701

PR #16275, opened 2020-12-06
[JS] Resolved variable's being considered indirect, making interop easier.

beef331:
Prior to this change the JS emitted code for variables would always be considered indirect resulting in the following Nim code emitting the following JS. Which I believe to be wrong as it made the interoping require accessing `[0]` of an array if passed to native js.
```nim
import jsffi
var t = jsNull
t = newJsObject()
```
```js
var t_2616014 = [null];
t_2616014[0] = {};
```


beef331:
Apparently I cannot test well, as it seems to work fine without this change.

PR #16283, opened 2020-12-07
use generics in random

ringabout:
Ref https://github.com/timotheecour/Nim/issues/408

### DIFF
--- a/lib/pure/random.nim
+++ b/lib/pure/random.nim
@@ -217,8 +217,7 @@ proc rand*(r: var Rand; max: Natural): int {.benign.} =
   ## Returns a random integer in the range `0..max` using the given state.
   ##
   ## See also:
-  ## * `rand proc<#rand,int>`_ that returns an integer using the default
-  ##   random number generator
+  ## * `rand proc<#rand,T>`_ `T` are integers, floats, and enums without holes.
   ## * `rand proc<#rand,Rand,range[]>`_ that returns a float
   ## * `rand proc<#rand,Rand,HSlice[T,T]>`_ that accepts a slice
   ## * `rand proc<#rand,typedesc[T]>`_ that accepts an integer or range type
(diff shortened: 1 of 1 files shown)


lib/pure/random.nim
```diff
@@ -233,28 +232,6 @@ proc rand*(r: var Rand; max: Natural): int {.benign.} =
...
       return int(x mod (uint64(max)+1u64))
 
-proc rand*(max: int): int {.benign.} =
-  ## Returns a random integer in the range `0..max`.
-  ##
-  ## If `randomize<#randomize>`_ has not been called, the sequence of random
-  ## numbers returned from this proc will always be the same.
-  ##
-  ## This proc uses the default random number generator. Thus, it is **not**
-  ## thread-safe.
```
Araq, line 243: What about this remark? Shouldn't we keep it?

ringabout: It is already kept
https://github.com/nim-lang/Nim/blob/de29bed82e40597a5d6450e697f06a9f9f87d16a/lib/pure/random.nim#L264

`rand(max: float)` already has the same comments and this generic procs reuse that commentbefore.

lib/pure/random.nim
```diff
@@ -276,8 +253,10 @@ proc rand*(r: var Rand; max: range[0.0 .. high(float)]): float {.benign.} =
     let u = (0x3FFu64 shl 52u64) or (x shr 12u64)
     result = (cast[float](u) - 1.0) * max
 
-proc rand*(max: float): float {.benign.} =
-  ## Returns a random floating point number in the range `0.0..max`.
+proc rand*[T: Ordinal or SomeFloat](max: T): T {.benign.} =
+  ## Returns a random floating point number in the range `T(0) .. max`.
+  ## 
+  ## Allowed types for `T` are integers, floats, and enums without holes.
```
timotheecour, line 259: this comment is redundant (and only approximately correct) and should be removed, it's already specified what T is in the generated docs:

image

furthermore, `Ordinal` is clickable (thanks to docgen) so users who see `Ordinal` for the 1st time can just click to see the definition, which is explained in 1 place, where it belongs (https://nim-lang.github.io/Nim/system.html#Ordinal) => "1 definition rule" for docs

note that the definition for https://nim-lang.github.io/Nim/system.html#Ordinal should be fixed via `s/enums/enums without holes/` but that's a separate issue

lib/pure/random.nim
```diff
@@ -276,8 +253,10 @@ proc rand*(r: var Rand; max: range[0.0 .. high(float)]): float {.benign.} =
     let u = (0x3FFu64 shl 52u64) or (x shr 12u64)
     result = (cast[float](u) - 1.0) * max
 
-proc rand*(max: float): float {.benign.} =
-  ## Returns a random floating point number in the range `0.0..max`.
+proc rand*[T: Ordinal or SomeFloat](max: T): T {.benign.} =
+  ## Returns a random floating point number in the range `T(0) .. max`.
```
timotheecour, line 257: s/random floating point number/random number/

timotheecour:
@xflywind thanks for tackling this; but this needs more care:
```nim
when defined case1:
  import std/random
  when defined case1a:
    # Error: type mismatch: got <Rand, Foo>
    type Fooa = enum k0,k1,k2
    echo rand(Fooa.high)
  when defined case1b:
    # Error: type mismatch: got <Rand, Dollar>
    type Dollar = distinct int
    echo rand(int.high.Dollar)
  when defined case1c: # ok
    echo rand(int64.high)
  when defined case1d:
    # Error: unhandled exception: value out of range [RangeDefect]
    echo rand(uint64.high)
  when defined case1e:
    # Error: type mismatch: got <Rand, char>
    echo (rand(char.high),)
```

in particular this should also be updated to be made generic:
```nim
proc rand*(r: var Rand; max: Natural): int {.benign.} =
```
and you can move the old underlying `proc rand*(r: var Rand; max: Natural): int {.benign.} =` implementation to some 
```nim
proc randImpl(r: var Rand; max: Natural): int {.benign.} = ...
```

lib/pure/random.nim
```diff
@@ -286,16 +265,21 @@ proc rand*(max: float): float {.benign.} =
...
-    randomize(234)
+    randomize(123)
+    doAssert rand(100) == 0
+    doAssert rand(100) == 96
+    doAssert rand(100) == 66
+
     let f = rand(1.0)
     ## f = 8.717181376738381e-07
-  rand(state, max)
+  result = T(rand(state, max))
```
timotheecour, line 282: cast to avoid overflow errors

PR #16284, opened 2020-12-07
clean up old codes

ringabout:
Ref https://github.com/timotheecour/Nim/issues/389


planetis-m:
Fyi, you can also do:
```nim
let d = collect(for i in ...) # notice no `newSeq`
```

PR #16286, opened 2020-12-07
Added Python to highlite module.

xbello:
* Copy-pasted the Nim Tokenizer to remove the non-existent multiline comment
operator, and added a 'keywords' parameter (like in cLikeNextToken).

* Added the keywords for Python

### DIFF
--- a/lib/packages/docutils/highlite.nim
+++ b/lib/packages/docutils/highlite.nim
@@ -62,11 +62,11 @@ type
 
   SourceLanguage* = enum
     langNone, langNim, langCpp, langCsharp, langC, langJava,
-    langYaml
+    langYaml, langPython
 
 const
   sourceLanguageToStr*: array[SourceLanguage, string] = ["none",
-    "Nim", "C++", "C#", "C", "Java", "Yaml"]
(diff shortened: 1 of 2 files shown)


Araq:
> Copy-pasted the Nim Tokenizer to remove the non-existent multiline comment
operator, and added a 'keywords' parameter (like in cLikeNextToken).

Instead patch the Nim Tokenizer and disable the multiline comment for ``g.lang == langPython`` (if ``g.lang`` doesn't yet exist, introduce it).

lib/packages/docutils/highlite.nim
```diff
@@ -886,6 +1042,16 @@ proc yamlNextToken(g: var GeneralTokenizer) =
   g.length = pos - g.pos
   g.pos = pos
 
+proc pythonNextToken(g: var GeneralTokenizer) =
+  const
+    keywords: array[0..34, string] = [
+      "False", "None", "True", "and", "as", "assert", "async", "await",
```
Araq, line 1048: Iirc, False, None and True are not keywords.

xbello (commented):
Had to made a couple of additional changes to pass the tests (tests/stdlib/trstgen.nim).

Araq:
Excellent work!

PR #16288, opened 2020-12-08
make rand consistent with other overloads

ringabout:
Make it consistent with other overloads
```nim
proc rand*[T: Ordinal or SomeFloat](r: var Rand; x: HSlice[T, T]): T =
  when T is SomeFloat:
    result = rand(r, x.b - x.a) + x.a
  else: # Integers and Enum types
    result = T(rand(r, int(x.b) - int(x.a)) + int(x.a))
```

### DIFF
--- a/lib/pure/random.nim
+++ b/lib/pure/random.nim
@@ -217,8 +217,7 @@ proc rand*(r: var Rand; max: Natural): int {.benign.} =
   ## Returns a random integer in the range `0..max` using the given state.
   ##
   ## See also:
-  ## * `rand proc<#rand,int>`_ that returns an integer using the default
-  ##   random number generator
+  ## * `rand proc<#rand,T>`_ `T` are integers, floats, and enums without holes.
   ## * `rand proc<#rand,Rand,range[]>`_ that returns a float
   ## * `rand proc<#rand,Rand,HSlice[T,T]>`_ that accepts a slice
   ## * `rand proc<#rand,typedesc[T]>`_ that accepts an integer or range type
@@ -232,28 +231,6 @@ proc rand*(r: var Rand; max: Natural): int {.benign.} =
     let x = next(r)
     if x <= randMax - (randMax mod Ui(max)):
       return int(x mod (uint64(max)+1u64))
-
-proc rand*(max: int): int {.benign.} =
-  ## Returns a random integer in the range `0..max`.
-  ##
-  ## If `randomize<#randomize>`_ has not been called, the sequence of random
-  ## numbers returned from this proc will always be the same.
-  ##
-  ## This proc uses the default random number generator. Thus, it is **not**
-  ## thread-safe.
-  ##
-  ## See also:
-  ## * `rand proc<#rand,Rand,Natural>`_ that returns an integer using a
-  ##   provided state
-  ## * `rand proc<#rand,float>`_ that returns a float
-  ## * `rand proc<#rand,HSlice[T,T]>`_ that accepts a slice
-  ## * `rand proc<#rand,typedesc[T]>`_ that accepts an integer or range type
-  runnableExamples:
-    randomize(123)
-    doAssert rand(100) == 0
-    doAssert rand(100) == 96
-    doAssert rand(100) == 66
-  rand(state, max)
 
 proc rand*(r: var Rand; max: range[0.0 .. high(float)]): float {.benign.} =
   ## Returns a random floating point number in the range `0.0..max`
@@ -276,8 +253,10 @@ proc rand*(r: var Rand; max: range[0.0 .. high(float)]): float {.benign.} =
     let u = (0x3FFu64 shl 52u64) or (x shr 12u64)
     result = (cast[float](u) - 1.0) * max
 
-proc rand*(max: float): float {.benign.} =
-  ## Returns a random floating point number in the range `0.0..max`.
+proc rand*[T: Ordinal or SomeFloat](max: T): T {.benign.} =
+  ## Returns a random floating point number in the range `T(0) .. max`.
+  ## 
+  ## Allowed types for `T` are integers, floats, and enums without holes.
   ##
   ## If `randomize<#randomize>`_ has not been called, the sequence of random
   ## numbers returned from this proc will always be the same.
@@ -286,13 +265,18 @@ proc rand*(max: float): float {.benign.} =
   ## thread-safe.
   ##
   ## See also:
+  ## * `rand proc<#rand,Rand,Natural>`_ that returns an integer using a
+  ##   provided state
   ## * `rand proc<#rand,Rand,range[]>`_ that returns a float using a
   ##   provided state
-  ## * `rand proc<#rand,int>`_ that returns an integer
   ## * `rand proc<#rand,HSlice[T,T]>`_ that accepts a slice
   ## * `rand proc<#rand,typedesc[T]>`_ that accepts an integer or range type
   runnableExamples:
-    randomize(234)
+    randomize(123)
+    doAssert rand(100) == 0
+    doAssert rand(100) == 96
+    doAssert rand(100) == 66
+
     let f = rand(1.0)
     ## f = 8.717181376738381e-07
   rand(state, max)
@@ -335,8 +319,7 @@ proc rand*[T: Ordinal or SomeFloat](x: HSlice[T, T]): T =
   ## See also:
   ## * `rand proc<#rand,Rand,HSlice[T,T]>`_ that accepts a slice and uses
   ##   a provided state
-  ## * `rand proc<#rand,int>`_ that returns an integer
-  ## * `rand proc<#rand,float>`_ that returns a floating point number
+  ## * `rand proc<#rand,T>`_ `T` are integers, floats, and enums without holes.
   ## * `rand proc<#rand,typedesc[T]>`_ that accepts an integer or range type
   runnableExamples:
     randomize(345)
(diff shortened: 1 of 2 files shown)


tests/stdlib/trandom.nim
```diff
@@ -57,3 +57,15 @@ randomize(223)
 
 for i in 0 .. 10:
   main()
+
+
+block:
```
timotheecour, line 41: We should really remove
```
discard """
  action: compile
"""
```
before anything else; 

in fact if I simply comment out
```
  randomize(223)
  for i in 0 .. 10:
    main()
```
in this test, `nim r tests/stdlib/trandom.nim` fails with `RangeDefect`

so I recommend following:
change this:
```nim
discard """
  action: compile
"""
```
to this:
```nim
discard """
  targets: "c cpp js"
  joinable: false
"""
```
(joinable: false so that the global random state doesn't affect other megatest joinable tests)

* then disable calling `main()` with a comment: `# TODO: fix properly so it can be run`

* then make sure the tests you've added work (right now they fail)

xflywind:
`distinct float` doesn't work before, also fix this issue
```nim
  type Test = distinct float
  echo rand(Test(float(3.45) .. float(4.85)))
```

lib/pure/random.nim
```diff
@@ -279,7 +277,11 @@ proc rand*[T: Ordinal or SomeFloat](max: T): T {.benign.} =
 
     let f = rand(1.0)
     ## f = 8.717181376738381e-07
```
timotheecour: pre-existing but how about the more informative: `doAssert x >= 0.0 and x <= 1.0`

tests/stdlib/trandom.nim
```diff
@@ -36,24 +33,33 @@ proc main =
   doAssert rand(0) == 0
   doAssert sample("a") == 'a'
 
-  when compileOption("rangeChecks"):
```
timotheecour, line 39: `doAssertRaises` is good but keep `when compileOption("rangeChecks"):`, otherwise you get `SIGFPE: Arithmetic error.` when running with -d:danger for example

xflywind:
Notes: current `rand[T](max: T)` and `rand[T](HSlice[T, T])` doesn't handle uint64 properly
They could be fixed in the following PR.

This PR makes `rand[T](max: T)` consistent with `rand[T](HSlice[T, T])`.

timotheecour:
- [x] How about statically disallow (CT error) enum and range types for which low != 0, otherwise it's error prone:

```nim
when defined case2:
  # RangeDefect
  import std/random
  type DiceRoll = range[3..6]
  for i in 0..<100000:
    echo rand(5.DiceRoll)

when defined case3:
  # RangeDefect
  import std/random
  type Foob = enum g0=1,g1,g2
  for i in 0..<100:
    echo i
    echo rand(Foob.high)
```

the rationale is that `0` is meaningless for those cases (ie returning 0..max doesnt' make sense).
(Note that signed integer types are fine though; `0` is still meaningful enough)

- [x] slices are fine though

lib/pure/random.nim
```diff
@@ -254,9 +255,7 @@ proc rand*(r: var Rand; max: range[0.0 .. high(float)]): float {.benign.} =
     result = (cast[float](u) - 1.0) * max
 
 proc rand*[T: Ordinal or SomeFloat](max: T): T {.benign.} =
-  ## Returns a random floating point number in the range `T(0) .. max`.
-  ## 
-  ## Allowed types for `T` are integers, floats, and enums without holes.
+  ## Returns a random number in the range `T(0) .. max`.
```
timotheecour: add:
```
`max` must be >= T(0).
```

in doc comment and add `assert max >= T(0)` in code otherwise you get bad error messages


```nim
  import std/random
  for i in 0..<100:
    echo rand(-2i8)
```
fails with confusing error:
```
random.nim(285)
 -2 notin 0 .. 9223372036854775807 [RangeDefect]
```

xflywind: But `0 .. -1` is not correct too.

ringabout: The issue exists before.

ringabout:
```nim
import random


echo rand(-1)
echo rand(12 .. -12)

var r = initRand(12)

echo r.rand(12 .. -12)
```
https://play.nim-lang.org/#ix=2H7X

tests/stdlib/trandom.nim
```diff
@@ -1,21 +1,18 @@
...
+randomize(233)
+
 proc main =
   var occur: array[1000, int]
 
   var x = 8234
   for i in 0..100_000:
     x = rand(high(occur))
     inc occur[x]
-  for i, oc in occur:
```
timotheecour, line 14: we should keep this test and adjust it to make it pass, it prevents regressions in case distribution becomes non-uniform. This works:
```nim
  var occur: array[1000, int]
  for i in 0..<100_000:
    var x = rand(high(occur))
    inc occur[x]
  # echo (min(occur), max(occur))
  doAssert max(occur) <= 140 and min(occur) >= 60 # gives some slack
```

(`var x = 8234` was useless)

ringabout:
Notes: current rand[T](max: T) and rand[T](HSlice[T, T]) doesn't handle uint64 properly
They could be fixed in the following PR.

This PR makes rand[T](max: T) consistent with rand[T](HSlice[T, T]).

~~assert max >= T(0) in code otherwise you get bad error messages (existing issue before)
https://github.com/nim-lang/Nim/pull/16288#discussion_r538147839~~

tests/stdlib/trandom.nim
```diff
@@ -1,21 +1,18 @@
...
-    if oc < 69:
-      doAssert false, "too few occurrences of " & $i
-    elif oc > 150:
-      doAssert false, "too many occurrences of " & $i
```
timotheecour, line 18: we should bring back this test, it's a good one, to prevent regressions.
this works:

```nim
import stats

...

  var rs: RunningStat
  for j in 1..5:
    for i in 1 .. 1_000:
      rs.push(gauss())
    doAssert abs(rs.mean-0) < 0.05
    doAssert abs(rs.standardDeviation()-1.0) < 0.1
    let maxVal = 3.0
    doAssert abs(rs.min + maxVal) < 1.0
    doAssert abs(rs.max-maxVal) < 1.0
    rs.clear()
```

tests/stdlib/trandom.nim
```diff
@@ -14,6 +14,8 @@ proc main =
     x = rand(high(occur))
     inc occur[x]
 
+  doAssert max(occur) <= 140 and min(occur) >= 60 # gives some slack
```
timotheecour, line 16: remove `var x = 8234` and use `let x = rand(high(occur))` inside the loop

lib/pure/random.nim
```diff
@@ -255,7 +255,7 @@ proc rand*(r: var Rand; max: range[0.0 .. high(float)]): float {.benign.} =
     result = (cast[float](u) - 1.0) * max
 
 proc rand*[T: Ordinal or SomeFloat](max: T): T {.benign.} =
-  ## Returns a random number in the range `T(0) .. max`.
+  ## Returns a random number in the range `T(0) .. max`. Notes that `max` must be >= T(0).
```
timotheecour: typo: s/Notes/Note/

(also need assert in the implementation otherwise bad errmsg)

tests/stdlib/trandom.nim
ringabout: It is not working at my computer?

```
Error: unhandled exception: trandom.nim(22, 14) `abs(rs.mean - 0) < 0.05`  [AssertionDefect]
Error: execution of an external program failed: 'D:\QQPCmgr\Desktop\Nim\tests\stdlib\trandom.exe '
```

ringabout:
See alternative revert PR: https://github.com/nim-lang/Nim/pull/16291

ringabout:
Now this works(added to tests)
```nim
when true:
  # RangeDefect
  import std/random
  type DiceRoll = range[3..6]
  for i in 0..<100000:
    echo rand(5.DiceRoll)

when true:
  # RangeDefect
  import std/random
  type Foob = enum g0=1,g1,g2
  for i in 0..<100:
    echo i
    echo rand(Foob.high)
```

lib/pure/random.nim
```diff
@@ -277,9 +278,32 @@ proc rand*[T: Ordinal or SomeFloat](max: T): T {.benign.} =
     doAssert rand(100) == 96
     doAssert rand(100) == 66
 
+  result = rand(state, int(max))
+
+proc rand*[T: SomeFloat](max: T): float {.benign.} =
```
timotheecour, line 283: I don't have a good answer to how to fix this, but:
* what happens on a 32bit machine when you call `rand(maximumPositiveValue(float64))` or similar?
* what happens on a 32bit machine when you call `rand(int64(int32.high) + 1)` or similar?
* also on a 64bit machine, things like `echo rand(uint64.high-1)` fail with `RangeDefect`

the problem is that returning `int` or `float` seems flawed; I don't know if it's fixable or not, and whether `rand(max:T)` should simply be deprecated.

Note that `rand(HSlice)` and `rand(typedesc[T])` are much more sound and dont' have those issues and can always be used in place of `rand(max:T)`

timotheecour:
=> https://github.com/nim-lang/Nim/pull/16291#issuecomment-740913908

lib/pure/random.nim
timotheecour: => https://github.com/nim-lang/Nim/pull/16291#issuecomment-740913908

ringabout:
#16291 is merged
improvements are moved to https://github.com/nim-lang/Nim/pull/16297

PR #16300, opened 2020-12-09
Added strscans.scanTuple

beef331:
Added `whenscanf`, which auto declares identifiers so one can just do `whenScanf(input,pattern, x): echo x`. Also implemented all the tests for the normal scanf to test whenscanf aswell.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -56,6 +56,7 @@
 - `writeStackTrace` is available in JS backend now.
 
 - `strscans.scanf` now supports parsing single characters.
+- `strscans.scanTuple` added which uses `strscans.scanf` internally, returning a tuple which can be unpacked for easier usage of `scanf`. 
 
 
 ## Language changes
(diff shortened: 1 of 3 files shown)


Araq:
We need a better name. And declaration injections are bad style, we should consider something like ``let (success, a, b, c) = scanG("$d $d); if success:``. Which has its own problems, of course, you can access `a` even though ``success`` is false. Tuple unpacking of an ``Option[tuple]`` anyone?

beef331:
Well with either implementation, could always `block` it to prevent access from anywhere outside of the macro's domain. As the intent is just for removing the manual declaration of the fields, not to replace scanf in all cases. Although I do realize an issue with using it in any breakable situation. The block idea would result in the following being generated.
```nim
block:
  var a: int
  var b: string
  if scanf(line,pattern, a, b):
     body
```

Araq:
The block helps, yes. There is still the problem of how this works with the "generic symbol lookup prepass" in the long run. ``(let x)`` as expressions looks decent:

```nim

  if scang(line, pattern, (var a), (var b)):
     body

```

We need some syntax for that sooner rather than later.

Araq:
Needs a .since annotation and a changelog entry. Superb work!

PR #16301, opened 2020-12-09
put both funcs and procs under the same section in the documentation

narimiran:
Since there are ongoing efforts to use `func` in the stdlib, having all `proc`s and `func`s under the same section should help with easier discovery of them (no need to look in two places).

### DIFF
--- a/compiler/docgen.nim
+++ b/compiler/docgen.nim
@@ -1045,12 +1045,9 @@ proc generateDoc*(d: PDoc, n, orig: PNode, docFlags: DocFlags = kDefault) =
     let pragmaNode = findPragma(n, wDeprecated)
     d.modDeprecationMsg.add(genDeprecationMsg(d, pragmaNode))
   of nkCommentStmt: d.modDesc.add(genComment(d, n))
-  of nkProcDef:
+  of nkProcDef, nkFuncDef:
     when useEffectSystem: documentRaises(d.cache, n)
     genItemAux(skProc)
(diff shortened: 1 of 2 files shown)


ee7:
Nice. I was going to suggest something like this.

For an example of where this was already bad: see https://nim-lang.org/docs/algorithm.html#13 where we have `sort` and `isSorted` as both `proc` and `func`. In the source code, they're ordered like:

```Nim
func sort*[T](a: var openArray[T], cmp: proc (x, y: T): int {.closure.}, order = SortOrder.Ascending) =
proc sort*[T](a: var openArray[T], order = SortOrder.Ascending) = sort[T](a, system.cmp[T], order) =

func isSorted*[T](a: openArray[T], cmp: proc(x, y: T): int {.closure.}, order = SortOrder.Ascending): bool =
proc isSorted*[T](a: openArray[T], order = SortOrder.Ascending): bool =
```

But in the pre-PR docs, the funcs are at the bottom of the page under a "Funcs" header. 

With this PR, it looks better (see below for a screenshot) - they're all grouped now under a header called "Procs". But given that we separated procs and funcs in the past, would it be better to name that header something like "Procs and Funcs"? 

The current state of this PR means that func-only modules like [math](https://nim-lang.github.io/Nim/math.html#12) and [strutils](https://nim-lang.github.io/Nim/strutils.html#12) have all their funcs under a header called "Procs", instead of "Funcs" as before.

PR #16323, opened 2020-12-11
incremental compilation for the frontend

disruptek:
For purely observational purposes. (read: so we can spam watchers)


planetis-m:
So why did you choose hash, instead of md5/sha1? Is it because of convenience or does it produce better results?

planetis-m:
if you would like to adapt, this works: https://github.com/planetis-m/bingo/blob/master/tests/tserial.nim

disruptek:
> So why did you choose hash, instead of md5/sha1? Is it because of convenience or does it produce better results?

What are you referring to?  `hash` is idiomatic and matches the semantic we want.  We use MD5 for `SigHash`.

compiler/ic/store.nim
```diff
@@ -0,0 +1,149 @@
...
+    let reply = readHeader stream
+    if reply.ok:
+      if hash(config) != reply.config:
+        # NOTE: we just bail if the configs don't hash equivalently
+        return
+      # try to reduce some buffer churn
+      var snap = newStringOfCap(getFileSize $fn)
+      thaw(stream, snap)
+      var m = thaw[Module] uncompress(snap)
+      if hash(m.ast) != reply.hash:
```
planetis-m, line 115: Sorry, I meant here. Checking deserialized ast with hash

AmjadHD:
silly question, What is meant by the frontend here ?

disruptek:
> silly question, What is meant by the frontend here ?

The frontend takes your Nim source and turns it into input for a backend; the backend varies between C, C++, JavaScript, etc. and it generates code of the given language.

compiler/ic/store.nim
disruptek: I think it's more important that the `hash()` for a `Table` key supplies our concept of equivalency.  Can it be done with `MD5`?  Sure.  Should it?  Why?

One reason to leave it this way is that `frosty` has *debug* code to audit de/serialization via an included `Hash` value.

planetis-m: First off, sha1/md5 are widely used for this purpose (together with SHA256, Sha512, etc).

Why not to use `std/hashes`? Because it may break, hash has overloads for pointer types that just cast them to int (this is done ofc for good reasons). There is a conflict of interests, you need a Table[int, ref object] working but also correct de serialization checking. Can't serve both at the same time! It's just not it's use-case! 

Moreover you need to be clear on what exactly are you testing. I suppose that is, the saved file is being retrieved without corruption. By adding the check at later stage, you effectively don't trust your serialization lib and test it's output.

My thought was that since serializing to a string dumps every type into a flat, sequence of bytes, it's clever to use this in order to produce the checksum of the contents. By using `hashes` you effectively add another indirection that needs to be maintained and tested.

planetis-m: I see you use hash in frosty a lot. Maybe you should consider other ways for testing, like making roundtrips, exact equality `==` operators, writing to a string and testing its length, using `setPosition` and reading certain parts of the stream, etc. If you need a performant lib that's also well tested, I might know smth. ;)

disruptek: What we're testing is that the current compiler's idea of what's important about the data in the rodfile matches the prior compiler's idea.  If we can't agree on the value of what we've read, then we know that we need to assume that the contents of the file are stale.  The valuations will be used across the code to measure equivalency between packed objects, so it's not the data that's important -- it's what it *means*.

disruptek: > I see you use hash in frosty a lot. Maybe you should consider other ways for testing, like making roundtrips, exact equality `==` operators, writing to a string and testing its length, using `setPosition` and reading certain parts of the stream, etc. If you need a performant lib that's also well tested, I might know smth. ;)


By all means, feel free to recommend something.  Better yet, submit a PR to `frosty` that adds a benchmark comparison to the repository so we can measure the improvement given identical input.  There are existing tests that you can use to ensure your alternative meets the requirements of PackedAst.

`frosty` hasn't been optimized for static data as `jason` has -- I think my approach there sucks and I'm waiting until concepts work better to try again with `frosty` -- so there is clearly a lot of room for improvement.

planetis-m: Man, hash is NOT equivalency, its used to shortcircuit `==` even in [tables](https://github.com/nim-lang/Nim/blob/5514b299ebc1de4ef0c1c182d4f21a02e3c2fa90/lib/pure/collections/hashcommon.nim#L60) Lets not make another lib war, do whats correct.

disruptek: What's confusing about this?

PR #16336, opened 2020-12-12
Fix broken links in docs

elliotwaite:


juancarlospaco:
Awesome work!.

elliotwaite:
Ah, got it. Fixed.

lib/pure/random.nim
```diff
@@ -220,7 +220,8 @@ proc rand*(r: var Rand; max: Natural): int {.benign.} =
   ## * `rand proc<#rand,int>`_ that returns an integer using the default
   ##   random number generator
   ## * `rand proc<#rand,Rand,range[]>`_ that returns a float
-  ## * `rand proc<#rand,Rand,HSlice[T,T]>`_ that accepts a slice
+  ## * `rand proc<#rand,Rand,HSlice[T:+Ordinal+or+float+or+float32+or+float64,T:+Ordinal+or+float+or+float32+or+float64]>`_
```
ee7: I believe that none of the HSlice links in this module work.

lib/system.nim
```diff
@@ -96,7 +96,7 @@ proc defined*(x: untyped): bool {.magic: "Defined", noSideEffect, compileTime.}
   ## defined.
   ##
   ## `x` is an external symbol introduced through the compiler's
-  ## `-d:x switch <nimc.html#compiler-usage-compile-time-symbols>`_ to enable
+  ## `-d:x switch <nimc.html#compiler-usage-compileminustime-symbols>`_ to enable
```
ee7, line 99: Nice catch.

It's nasty that the correct link has "minus" though. Especially because the symbol in "compile-time" is a hyphen, not a minus sign.

ee7 (commented):
I had a very quick look - I think there's a problem in `random.nim`. But otherwise, looks great. Thanks for working on this.

Ideally, we should have link checking in CI. And adding it to `nim doc` does seem like a sensible way of doing that.

But if we wanted a short-term hack, at least we have the option of running your script before every release...

lib/pure/random.nim
elliotwaite: Oh, you're right. I had used the URLs of the links in the left menu which are actually also broken. I updated the links in the body of the page to be correct now (replacing the plus signs with spaces), and I also fixed the other HSlice links I had missed. And since I'm not sure how to fix those broken left menu links, I added a note about them here: https://github.com/nim-lang/Nim/issues/16337

lib/system.nim
elliotwaite: Yep, kind of funny.

elliotwaite:
Sounds good. Yeah, having link checking in CI would be ideal.

I think I fixed the errors I made on the `random` page, nice catch.

PR #16339, opened 2020-12-13
Add jsxmlserializer

juancarlospaco:
- Added `jsxmlserializer` module for `XMLSerializer` for the JavaScript target.
- https://developer.mozilla.org/en-US/docs/Web/API/XMLSerializer
- Serialize DOM to XML very fast. Tiny diff.

```nim
let cerealizer: XMLSerializer = newXMLSerializer()
echo cerealizer.serializeToString(node = document)
```

temp

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -56,10 +56,13 @@
 - `writeStackTrace` is available in JS backend now.
 
 - `strscans.scanf` now supports parsing single characters.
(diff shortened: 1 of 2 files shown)


metagn:
Can you make a case for why this should be in the standard library? I think anything in "Web APIs" on MDN should be in nimble/fusion.

Araq:
Please add it to fusion instead.

PR #16356, opened 2020-12-15
More colorful error messages

beef331:
This PR uses ANSI colours to make the terminal output of errors quicker to read and spacing out the errors for overload errors. Highlighting the actual issues. Also adds a explict message about immutable variables being the cause of errors.
```nim
#Sample code used for this test.
let stringValue = "Hmm"
stringValue.add(10)
```

image

image

## (EDIT) links
* refs https://github.com/nim-lang/RFCs/issues/87
* refs https://github.com/nim-lang/Nim/issues/2844#issuecomment-107060552

### DIFF
--- a/compiler/msgs.nim
+++ b/compiler/msgs.nim
@@ -8,7 +8,7 @@
 #
 
 import
-  std/[strutils, os, tables, terminal, macros, times],
+  std/[strutils, os, tables, terminal, macros, times, colors, strformat, parseutils],
   std/private/miscdollars,
   options, ropes, lineinfos, pathutils, strutils2
 
@@ -477,12 +477,68 @@ proc sourceLine*(conf: ConfigRef; i: TLineInfo): string =
 
   result = conf.m.fileInfos[i.fileIndex.int32].lines[i.line.int-1]
 
-proc getSurroundingSrc(conf: ConfigRef; info: TLineInfo): string =
+proc colorError(s: string, color: ForegroundColor, conf: ConfigRef): string =
+  if optUseColors in conf.globalOptions:
+    template isQuote(val: untyped): untyped = val == '\''
+    template isNotQuote(val: untyped): untyped = val != '\''
+    let parsable = (s.count('"').mod 2) == 0
+    var 
+      pos = 0
+    while pos < s.len:
+      if s[pos].isQuote:
+        if pos < s.high and s[pos + 1].isQuote:
+          result.add s[pos..(pos + 1)]
+        var inDoubleQuote = false
+        inc pos
+        let start = pos
+        while (pos < s.len and s[pos].isNotQuote) or inDoubleQuote:
+          if s[pos] == '"' and parsable:
+            inDoubleQuote = not inDoubleQuote
+          inc pos
+        #Highlight error
+        if pos < s.len and parsable:
+          if (s[pos].isQuote):
+            result.add fmt"""{color.ansiForegroundColorCode}{s[start..(pos - 1)]}{ansiResetCode}"""
+          else:
+            result.add s[start..(pos - 1)]
+        else: 
+          result.add s[start..s.high]
+          break
+      else:
+        result.add s[pos]
+      inc pos
+  else:
+    result = s
+
+proc getSurroundingSrc(result: var string, conf: ConfigRef; info: TLineInfo, color: ForegroundColor) =
   if conf.hasHint(hintSource) and info != unknownLineInfo:
     const indent = "  "
-    result = "\n" & indent & $sourceLine(conf, info)
-    if info.col >= 0:
-      result.add "\n" & indent & spaces(info.col) & '^'
+    if optUseColors in conf.globalOptions:
+      var msg = $sourceLine(conf, info)
+      let
+        colorEnd =
+          case msg[info.col]:
+          of IdentStartChars:
+            info.col + 1 + msg.skipUntil({'\0'..'\255'} - IdentChars, info.col)
+          of '[': # + 2 due to wanting to include the sym in the coloring
+            info.col + 2 + msg.skipUntil(']', info.col)
+          of '(': # + 2 due to wanting to include the sym in the coloring
+            info.col + 2 + msg.skipUntil(')', info.col)
+          of '.': # + 2 due to wanting to include the sym in the coloring
+            info.col + 2 + msg.skipUntil({'.', ' ', '(', '['}, info.col)
+          else: msg.high
+
+      msg.insert("'", info.col)
+      msg.insert("'", colorEnd)
+      msg.insert("\n  ", 0)
+
+      if info.col >= 0:
+        msg.add "\n" & indent & spaces(info.col) & "'^'"
+      result.add msg.colorError(color, conf)
+    else:
+      result.add "\n" & indent & $sourceLine(conf, info)
+      if info.col >= 0:
+        result.add "\n" & indent & spaces(info.col) & '^'
 
 proc formatMsg*(conf: ConfigRef; info: TLineInfo, msg: TMsgKind, arg: string): string =
   let title = case msg
@@ -539,7 +595,8 @@ proc liMessage*(conf: ConfigRef; info: TLineInfo, msg: TMsgKind, arg: string,
     color = HintColor
     inc(conf.hintCounter)
 
-  let s = if isRaw: arg else: getMessageStr(msg, arg)
+  let strMsg = if isRaw: arg else: getMessageStr(msg, arg)
+  var s = colorError(strMsg, color, conf)
   if not ignoreMsg:
     let loc = if info != unknownLineInfo: conf.toFileLineCol(info) & " " else: ""
     # we could also show `conf.cmdInput` here for `projectIsCmd`
@@ -550,8 +607,9 @@ proc liMessage*(conf: ConfigRef; info: TLineInfo, msg: TMsgKind, arg: string,
       if msg == hintProcessing and conf.hintProcessingDots:
         msgWrite(conf, ".")
       else:
+        getSurroundingSrc(s, conf, info, color)
         styledMsgWriteln(styleBright, loc, resetStyle, color, title, resetStyle, s, KindColor, kindmsg,
-                         resetStyle, conf.getSurroundingSrc(info), conf.unitSep)
+                         resetStyle, conf.unitSep)
         if hintMsgOrigin in conf.mainPackageNotes:
           # xxx needs a bit of refactoring to honor `conf.filenameOption`
           styledMsgWriteln(styleBright, toFileLineCol(info2), resetStyle,
--- a/compiler/semcall.nim
+++ b/compiler/semcall.nim
@@ -11,7 +11,7 @@
 # included from sem.nim
 
 from algorithm import sort
-
+from renderer import quoteExpr
 proc sameMethodDispatcher(a, b: PSym): bool =
   result = false
   if a.kind == skMethod and b.kind == skMethod:
@@ -234,16 +234,16 @@ proc presentFailedCandidates(c: PContext, n: PNode, errors: CandidateErrors):
         let wanted = err.firstMismatch.formal.typ
         doAssert err.firstMismatch.formal != nil
         candidates.add("\n  required type for " & nameParam &  ": ")
-        candidates.addTypeDeclVerboseMaybe(c.config, wanted)
-        candidates.add "\n  but expression '"
+        candidates.add typeToString(wanted)
+        candidates.addDeclaredLocMaybe(c.config, wanted)
+        candidates.add "\n  but expression "
         if err.firstMismatch.kind == kVarNeeded:
-          candidates.add renderNotLValue(nArg)
-          candidates.add "' is immutable, not 'var'"
+          candidates.add "$1 is immutable, not 'var'" % renderNotLValue(nArg).quoteExpr
         else:
-          candidates.add renderTree(nArg)
-          candidates.add "' is of type: "
-          let got = nArg.typ
-          candidates.addTypeDeclVerboseMaybe(c.config, got)
+          candidates.add "$1 is of type: " % renderTree(nArg).quoteExpr
+          var got = nArg.typ
+          candidates.add typeToString(got).quoteExpr
+          candidates.addDeclaredLocMaybe(c.config, got)
           doAssert wanted != nil
           if got != nil:
             if got.kind == tyProc and wanted.kind == tyProc:
@@ -274,8 +274,8 @@ const
   errTypeMismatch = "type mismatch: got <"
   errButExpected = "but expected one of:"
   errUndeclaredField = "undeclared field: '$1'"
-  errUndeclaredRoutine = "attempting to call undeclared routine: '$1'"
-  errBadRoutine = "attempting to call routine: '$1'$2"
+  errUndeclaredRoutine = "attempting to call undeclared routine: $1"
+  errBadRoutine = "attempting to call routine: $1$2"
   errAmbiguousCallXYZ = "ambiguous call; both $1 and $2 match for: $3"
 
 proc notFoundError*(c: PContext, n: PNode, errors: CandidateErrors) =
@@ -287,7 +287,7 @@ proc notFoundError*(c: PContext, n: PNode, errors: CandidateErrors) =
     globalError(c.config, n.info, "type mismatch")
     return
   if errors.len == 0:
-    localError(c.config, n.info, "expression '$1' cannot be called" % n[0].renderTree)
+    localError(c.config, n.info, "expression $1 cannot be called" % n[0].renderTree.quoteExpr)
     return
 
   let (prefer, candidates) = presentFailedCandidates(c, n, errors)
@@ -325,7 +325,7 @@ proc getMsgDiagnostic(c: PContext, flags: TExprFlags, n, f: PNode): string =
     var o: TOverloadIter
     var sym = initOverloadIter(o, c, f)
     while sym != nil:
-      result &= "\n  found $1" % [getSymRepr(c.config, sym)]
+      result &= "\n  found $1" % [getSymRepr(c.config, sym).quoteExpr]
       sym = nextOverloadIter(o, c, f)
 
   let ident = considerQuotedIdent(c, f, n).s
@@ -344,8 +344,8 @@ proc getMsgDiagnostic(c: PContext, flags: TExprFlags, n, f: PNode): string =
     let suffix = if result.len > 0: " " & result else: ""
     result = errUndeclaredField % ident & typeHint & suffix
   else:
-    if result.len == 0: result = errUndeclaredRoutine % ident
-    else: result = errBadRoutine % [ident, result]
+    if result.len == 0: result = errUndeclaredRoutine % ident.quoteExpr
+    else: result = errBadRoutine % [ident.quoteExpr, result.quoteExpr]
 
 proc resolveOverloads(c: PContext, n, orig: PNode,
                       filter: TSymKinds, flags: TExprFlags,
@@ -432,8 +432,8 @@ proc resolveOverloads(c: PContext, n, orig: PNode,
       return
     elif result.state != csMatch:
       if nfExprCall in n.flags:
-        localError(c.config, n.info, "expression '$1' cannot be called" %
-                   renderTree(n, {renderNoComments}))
+        localError(c.config, n.info, "expression $1 cannot be called" %
+                   renderTree(n, {renderNoComments}).quoteExpr)
       else:
         if {nfDotField, nfDotSetter} * n.flags != {}:
           # clean up the inserted ops
@@ -457,8 +457,8 @@ proc resolveOverloads(c: PContext, n, orig: PNode,
       args.add(")")
 
       localError(c.config, n.info, errAmbiguousCallXYZ % [
-        getProcHeader(c.config, result.calleeSym),
-        getProcHeader(c.config, alt.calleeSym),
+        getProcHeader(c.config, result.calleeSym).quoteExpr,
+        getProcHeader(c.config, alt.calleeSym).quoteExpr,
         args])
 
 proc instGenericConvertersArg*(c: PContext, a: PNode, x: TCandidate) =
@@ -622,7 +622,7 @@ proc semOverloadedCall(c: PContext, n, nOrig: PNode,
       notFoundError(c, n, errors)
 
 proc explicitGenericInstError(c: PContext; n: PNode): PNode =
-  localError(c.config, getCallLineInfo(n), errCannotInstantiateX % renderTree(n))
+  localError(c.config, getCallLineInfo(n), errCannotInstantiateX % renderTree(n).quoteExpr)
   result = n
 
 proc explicitGenericSym(c: PContext, n: PNode, s: PSym): PNode =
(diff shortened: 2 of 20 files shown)


juancarlospaco:
This needs a switch, can break Testament tests that check output.

beef331:
Yea I've noticed that

juancarlospaco:
I dont remember but investigate this, it should allow to customize stacktrace messages without breaking the compiler and all tests:
https://nim-lang.github.io/Nim/stackframes.html#setFrameMsg.t%2Cstring%2Cstring
Maybe you can improve it.

beef331:
I'm uncertain how that module would help.

timotheecour:
I support the idea of colorful error messages, but I'm not sure this is the right approach, because it pollutes compiler sources with formatting, in potentially lots of places.

D's approach seems better IMO, see how it was implemented here: https://github.com/dlang/dmd/pull/6777
and corresponding discussion here https://forum.dlang.org/post/of9oao$230j$1@digitalmars.com

in D's case, syntax highlighting is triggered when encountering backtick
```
`
```
in nim's case, we could trigger with single quote 
```
'
```
instead, but the idea is the same.
benefits:
* reuse existing logic for syntax highlight (and honors user customizations etc)
* works out of the box with compiler messages, without cluttering code with formatting logic; all that's needed is ensuring quotes are consistently, which can be done using an auxiliary function

## example

image

corresponding source code in D:
src/dmd/expressionsem.d:2656:24:
```d
exp.error("undefined identifier `%s`, did you mean %s `%s`?", exp.ident.toChars(), s2.kind(), s2.toChars());
```

beef331:
Yea I agree that the D approach is better,  although I think something that lets you tell the style of the value would be better(immutable error being yellow and the immutable var in the error message being yellow). So I'm thinking of `value, style` tuples that use enums styles which means it can be switched off or customized by users rather easily.

beef331:
It now uses `'`s for marking the message colour and `''` for keeping the symbol information, and after being fixed should be easily toggable and not mess up testing, probably still can be done smarter.
image

beef331:
Now on parity with my original PR, though I think I'm proceeding to still require changing a lot of the compiler to have the `'` and `''` in the error messages shoehorning in the feature instead of properly supporting it. At the very least it's easily toggable and can generate the same error as what's expected(minus the extra new line to make the overload error more readable).

compiler/msgs.nim
```diff
@@ -486,6 +486,28 @@ proc formatMsg*(conf: ConfigRef; info: TLineInfo, msg: TMsgKind, arg: string): s
               else: ErrorTitle
   conf.toFileLineCol(info) & " " & title & getMessageStr(msg, arg)
 
+proc colorError(s: string, color: ForegroundColor): string =
+  when not defined(colorfulError):
```
timotheecour: `nimColorErrors` (see contributing.rst for rationale of nim prefix)

compiler/msgs.nim
```diff
@@ -486,6 +486,28 @@ proc formatMsg*(conf: ConfigRef; info: TLineInfo, msg: TMsgKind, arg: string): s
...
+    template isQuote(val: untyped): untyped = val == '\''
```
timotheecour, line 482: and also, should be:
`conf.isDefined("nimColorErrors")` so that you don't need to recompile nim just to toggle this flag

compiler/msgs.nim
beef331: Will do it's just for testing mostly at the moment.

compiler/msgs.nim
```diff
@@ -486,6 +486,31 @@ proc formatMsg*(conf: ConfigRef; info: TLineInfo, msg: TMsgKind, arg: string): s
...
+  template isQuote(val: untyped): untyped = val == '\''
+  template isNotQuote(val: untyped): untyped = val != '\''
+  var pos = 0
+  while pos < s.len:
+    if s[pos].isQuote:
+      inc pos
+      let start = pos
+      let nested = pos < s.high and s[pos].isQuote
+      inc pos
+      while (pos < s.len and s[pos].isNotQuote and not nested) or (nested and pos < s.high and (s[pos].isNotQuote or s[pos + 1].isNotQuote)):
```
timotheecour: would be good to know when this logic breaks down, eg:
```nim
proc fn=
  let a = 0
  a = "will a sneaky quote in user code ' confuse this? "
```
and gracefully handle (eg just not colorize strings that aren't correctly parsed)

compiler/semcall.nim
```diff
@@ -232,20 +232,20 @@ proc presentFailedCandidates(c: PContext, n: PNode, errors: CandidateErrors):
...
-        candidates.add "\n  but expression '"
+        candidates.add "\n  but expression ''"
         if err.firstMismatch.kind == kVarNeeded:
           candidates.add renderNotLValue(nArg)
-          candidates.add "' is immutable, not 'var'"
+          candidates.add "'' 'is immutable, not' ''var''"
         else:
           candidates.add renderTree(nArg)
-          candidates.add "' is of type: "
+          candidates.add "'' 'is of type:' ''"
```
timotheecour: instead of concatenating half quoted expressions, use `renderer.quoteExpr` which I had added some time ago in preparation for something like this (feel free to move it if needed).

It can be made more robust by escaping quotes that it finds in the input, and then your `colorError` can correctly handle such escaped quotes to avoid getting confused by it.

Let me know if this needs more clarification!

compiler/msgs.nim
```diff
@@ -486,6 +486,31 @@ proc formatMsg*(conf: ConfigRef; info: TLineInfo, msg: TMsgKind, arg: string): s
               else: ErrorTitle
   conf.toFileLineCol(info) & " " & title & getMessageStr(msg, arg)
 
+proc colorError(s: string, color: ForegroundColor): string =
```
timotheecour: add a tests/compiler/tmsgs.nim test file with:
```nim
block: # colorError
  # add tests here for `colorError`
```
(make some procs public if needed)

=> will make it easier to debug / improve it once we have tests (and would also avoid having to recompile nim just to test/improve this proc)

compiler/msgs.nim
```diff
@@ -549,7 +555,7 @@ proc liMessage*(conf: ConfigRef; info: TLineInfo, msg: TMsgKind, arg: string,
     color = HintColor
     inc(conf.hintCounter)
 
-  let s = if isRaw: arg.colorError(color) else: getMessageStr(msg, arg).colorError(color)
+  let s = if isRaw: arg.colorError(color, conf.isDefined("nimColorErrors")) else: getMessageStr(msg, arg).colorError(color, conf.isDefined("nimColorErrors"))
```
timotheecour: * DRY
* why not use the existing `optUseColors` instead of a new `nimColorErrors` ?

compiler/renderer.nim
```diff
@@ -1680,6 +1680,10 @@ proc quoteExpr*(a: string): string {.inline.} =
   ## can be used for quoting expressions in error msgs.
   "'" & a & "'"
 
+proc doubleQuoteExpr*(a: string): string {.inline.} =
```
timotheecour: why double quotes instead of single quotes?
that would mean most messages would need to be changed to a less readable syntax:
```nim
before:
expected '$1' got '$2'

after:
expected ''$1'' got ''$2''
```

If you're worried about corner cases where `a` contains a `'`, the fix is simply to make `quoteExpr` escape `'` so that `colorError` handles these correctly.

beef331: Well `''word''` is for keeping the `'` in the error message and `'word'` is for highlighting word, to be more to the point. So I figured using quotes and double single quotes for achieving this would be best, but i suppose if the word highlighting isn't wanted it can be thrown away.

timotheecour: I think it's simpler if error msgs stay simple (single quotes, as currently done), and the renderer can, depending on `-d:nimColorErrors`, either return input as is or transform colorize inputs in single quotes and drop the quotes in the process, which become redundant.

A special case can be made for `''` (ie when expression inside quotes is empty) to render as is, because with colors it would otherwise not show anything.

beef331: Well it's certainly simpler, so will go back to previous state, should I also use `-d:nimNewLineOverloadError` for opt-in spacing out the overload error(Can be seen in the 2 images)

timotheecour: > -d:nimNewLineOverloadError

meh on this... in any case this seems orthogonal to this PR :)
(I'm all for customizations but it should be designed well, eg via user defined templates of how error messages should look like, eg: `{file}:{col}:{line}: {msg}`, see how lldb does this to customize prompt/etc settings in ~/.lldbinit:
```
settings set frame-format frame #${frame.index}: ${frame.pc} start-addr:${line.start-addr} { ${module.file.basename}{ ${function.name-with-args}${function.pc-offset}}}{ at ${ansi.fg.yellow} ${line.file.fullpath}:${line.number}:${line.column} ${ansi.normal}}\n
```
=> more flexible. In any way, that's out of scope for this PR IMO

EDIT: related: see msgs.nim:
```
    hintSuccessX: "${loc} lines; ${sec}s; $mem; $build build; proj: $project; out: $output",
```

compiler/msgs.nim
```diff
@@ -555,7 +557,9 @@ proc liMessage*(conf: ConfigRef; info: TLineInfo, msg: TMsgKind, arg: string,
     color = HintColor
     inc(conf.hintCounter)
 
-  let s = if isRaw: arg.colorError(color, conf.isDefined("nimColorErrors")) else: getMessageStr(msg, arg).colorError(color, conf.isDefined("nimColorErrors"))
+  let
+    strMsg = if isRaw: arg else: getMessageStr(msg, arg)
+    s = colorError(strMsg, color, conf.isDefined("optUseColors"))
```
timotheecour: `optUseColors in conf.globalOptions`

(enbled via `nim --colors:on main`)

compiler/semcall.nim
```diff
@@ -198,6 +198,7 @@ proc presentFailedCandidates(c: PContext, n: PNode, errors: CandidateErrors):
   var candidatesAll: seq[string]
   var candidates = ""
   var skipped = 0
+  var immutParam = "" #Stores the name of an immutable param that needs to be 'var'
```
timotheecour: space after `#` (+ everywhere relevant)

compiler/msgs.nim
```diff
@@ -486,6 +486,39 @@ proc formatMsg*(conf: ConfigRef; info: TLineInfo, msg: TMsgKind, arg: string): s
...
+        inc pos
+        let start = pos
+        while (pos < s.len and s[pos].isNotQuote) or inDoubleQuote:
+          if s[pos] == '"' and parsable:
+            inDoubleQuote = not inDoubleQuote
+          inc pos
+        #Highlight error
+        if pos < s.len and parsable:
+          if (s[pos].isQuote):
+            result.add fmt"""{color.ansiForegroundColorCode}{s[start..(pos - 1)]}{ansiResetCode}"""
```
timotheecour, line 501: use sensible defaults; right now quoted expression is red and looks too similar to Error when i tried it:
```nim
nim --colors -d:optUseColors --eval:asdf
```
(but see other comment: `-d:optUseColors` is a misuse, --colors should be all that's needed)


image


=> D's choice of colors seems reasonable but anything distinctive is fine; we can distinguish:
current:
warning(yellow)
error(red)
hint(green)
`file(line,col)` (inconsistent, sometimes bold, sometimes regular)

new from this PR:
quoted expr(bold? cyan? magenta?)

compiler/msgs.nim
```diff
@@ -9,7 +9,7 @@
 
 import
   options, strutils, os, tables, ropes, terminal, macros,
-  lineinfos, pathutils
+  lineinfos, pathutils, strformat
```
timotheecour: `nim --colors -d:optUseColors --eval:'var a: int = "abc"'`

image

"abc"
shouldn't be in double quotes, since it's in color already just abc is fine

timotheecour: my bad, the double quotes comes from string here, works as intented

compiler/semcall.nim
```diff
@@ -253,6 +255,8 @@ proc presentFailedCandidates(c: PContext, n: PNode, errors: CandidateErrors):
       candidates.add(diag & "\n")
     candidatesAll.add candidates
   candidatesAll.sort # fix #13538
+  if immutParam.len > 0:
+    candidatesAll.insert(("\nAtleast one mismatch is due to " & immutParam & " being immutable\n\n").quoteExpr)
```
timotheecour: good idea; I'd go even further and only show the procs with mismatch due to mutable/immutable when there's at least one of them, and rely on `--showAllMismatches:on` for showing all, but that would definitely belong to future PR

compiler/semcall.nim
```diff
@@ -235,24 +236,27 @@ proc presentFailedCandidates(c: PContext, n: PNode, errors: CandidateErrors):
         candidates.add "\n  but expression '"
         if err.firstMismatch.kind == kVarNeeded:
           candidates.add renderNotLValue(nArg)
+          immutParam = renderNotLValue(nArg)
```
timotheecour: pre-existing but now that you use `quoteExpr` how about avoiding un-matched single quotes as much as possible so code is correct "by construction" (and also deals with nested quotes in a correct way, see my other comment regarding escaping quotes):
```nim
candidates.add "\n  but expression "
if err.firstMismatch.kind == kVarNeeded:
  immutParam = renderNotLValue(nArg)
  candidates.add "$1 is immutable, not 'var'" % renderNotLValue(nArg).quoteExpr
else:
  candidates.add "$1 is of type: " % renderTree(nArg).quoteExpr
  ...
```

compiler/semcall.nim
```diff
@@ -256,7 +256,7 @@ proc presentFailedCandidates(c: PContext, n: PNode, errors: CandidateErrors):
     candidatesAll.add candidates
   candidatesAll.sort # fix #13538
   if immutParam.len > 0:
-    candidatesAll.insert(("\nAtleast one mismatch is due to " & immutParam & " being immutable\n\n").quoteExpr)
+    candidatesAll.insert(("\nAtleast one mismatch is due to " & immutParam.quoteExpr & " being immutable\n\n"))
```
timotheecour: general comment: use `%` instead of `&`
```nim
"\nAtleast one mismatch is due to $# being immutable\n\n" % immutParam.quoteExpr
```

compiler/msgs.nim
beef331: Well I see a `styleString` and a `quoteExpr` overload which takes the same params as `styleString` that means we can drop all `'` and then just call `quoteExpr` with our desired colour for each part, only thing is getting access to user defined colours and the conf as currently the procs that generate errors as far as i can see cannot access the `conf`.

timotheecour: this works:
change to:
```nim
proc colorError(conf: ConfigRef, s: string, color: ForegroundColor): string =
  let doColor = optUseColors in conf.globalOptions
  let colorExpr = # get it from conf
```

`quoteExpr` doesn't need to have access to ConfigRef, all it does is quote expressions (and escaping quotes); btw: see how quoting works in addQuoted (which is for `"` instead of `'`)

beef331: Well you've said `much as possible so code is correct "by construction"`, as far as I can see it's best to add styling when generating the messages, versus using escaped characters to get style after the message is created. Starting to think the best thing is to completely rewrite the message emitting so that we can create user defined messages, which have their own styling embedded, and just paste the data where it's wanted.

timotheecour: > Well you've said

this is what I meant in https://github.com/nim-lang/Nim/pull/16356#discussion_r544537851 :
the following isn't correct by construction:
```nim
msg.add " foo '" # unmatched quote by itself
msg.add n.renderTree
msg.add "' rest of msg"
```
which is bad for 2 reasons:
* P1: msg is composed by half-matched atoms, and gets trickier with complex control flow (see sigmatch)
* P2: incorrect if renderTree contains some quote (would likely fail colorizing with your PR)

this is correct by construction:
```nim
msg.add " foo "
msg.add n.renderTree.quoteExpr
msg.add " rest of msg"
```
or even better:
```nim
msg.add " foo $1 rest of msg" % n.renderTree.quoteExpr
```

> as far as I can see it's best to add styling when generating the messages, versus using escaped characters to get style after the message is created
> best thing is to completely rewrite ...

The advantage of what I'm suggesting is that it requires minimum or no patching to most existing compiler code generating messages.

eg, this will automatically work with your PR (except for corner case where n.renderTree contains a quote):
```nim
msg.add " foo '$1' rest of msg" % n.renderTree
```
ideally this would be written as
```nim
msg.add " foo $1 rest of msg" % n.renderTree.quoteExpr
```
but it's not needed outside of above mentioned edge case.

compiler/semcall.nim
```diff
@@ -265,9 +263,9 @@ proc presentFailedCandidates(c: PContext, n: PNode, errors: CandidateErrors):
 const
   errTypeMismatch = "type mismatch: got <"
   errButExpected = "but expected one of: "
-  errUndeclaredField = "undeclared field: '$1'"
-  errUndeclaredRoutine = "attempting to call undeclared routine: '$1'"
-  errBadRoutine = "attempting to call routine: '$1'$2"
+  errUndeclaredField = "undeclared field: $1"
```
timotheecour: I tried your PR again, it's starting to look good!
- [ ] bad:
```nim
when true:
  var a: string
  a = 1+2
```
XDG_CONFIG_HOME= nim r main # ok
XDG_CONFIG_HOME= nim r --hint:Source main # bad

```
t11525.nim(12, 8) Error: type mismatch: got int literal(3) for 3 but expected string
    a = 1+2
         ^
```
`a = 1+2` isn't colorized


- [ ] bad:
```
echo 'proc fn(a = 0) = discard\nproc fn() = discard\nfn()' | nim r -
stdinfile.nim(3, 3) Error: ambiguous call; both stdinfile.fn(a: int) [proc declared in stdinfile.nim(1, 6)] and stdinfile.fn() [proc declared in stdinfile.nim(2, 6)] match for: ()
```

image

=> `stdinfile.fn(a: int)` should be colorized but not the whole `stdinfile.fn(a: int) [proc declared in stdinfile.nim(1, 6)]` (or at least only `stdinfile.nim(1, 6)` should)


- [ ] bad
```
`nim --eval:'let a = "foo"; a.add 2'`
```

```
cmdfile.nims(1, 17) Error: type mismatch: got <string, int literal(2)>
but expected one of:
proc add(x: var string; y: char)
  required type for x: var string
  but expression a is immutable, not var

```

image


some improvements can be done there too:
<string, int literal(2)>
=>
'string, int literal(2)' so it colorizes

would be very cool to colorize only the mismatching argument in `proc add(x: var string; y: char)` but that can be done in future work


- [ ] some quotes use backtick and should be changed to single quote `'` so they colorize properly:
```nim
when true:
  proc fn[T]()
  fn()
```

```
t11525.nim(59, 5) template/generic instantiation of `fn` from here
t11525.nim(58, 11) Error: cannot instantiate: T
```

- [ ] etc
there are other cases not properly colorized (eg hintCC); of course, can be done in future work

beef331: I didnt even realize that `--hint:source` was an option.

juancarlospaco: Try this with some Regex and PEG and see if it crashes.

beef331:
Just want to update that I am continuing this in a different branch, but with manual highlighting as I and Clyybber feel it's more beneficial, although for more work. Using the quote method is quick, but it does not allow highlighting as much as possible in my view.

timotheecour:
@beef331 now that your other https://github.com/nim-lang/Nim/pull/18384 was merged, any interest in continuing this? 

I still prefer this approach to https://github.com/nim-lang/Nim/pull/16446, because it's simpler, and is also simpler for code that generates error messages which don't need to be aware of colors. It was also the option adopted in D.

I don't think we need advanced coloring options, simply highlighting any text inside delimiters (either single quotes or backticks TBD, easily changeable) plus maybe also highlighting known tokens like `file(line,col)` would be enough IMO.

beef331:
I probably should :sweat_smile: the second method although more helpful is also much more to maintain and add to, so I sorta agree this method is the "better" one just due to the tedium otherwise required. Now I sorta agree we dont need hand coloured highlighting cause if the error messages need that, probably better to just clean them up.

timotheecour:
great. Note also that for this PR, you don't need to cover all cases, IIRC what you had was already close to mergeable, it can always improve in future work; I'd recommend rebasing to fix bitrot conflicts, then i can take another look to see what should be done before merge vs what can be left for future work

compiler/msgs.nim
```diff
@@ -480,9 +480,17 @@ proc sourceLine*(conf: ConfigRef; i: TLineInfo): string =
 proc getSurroundingSrc(conf: ConfigRef; info: TLineInfo): string =
```
timotheecour: how about add an inplace overload:
```nim
proc getSurroundingSrc(result: var string, conf: ConfigRef; info: TLineInfo) = ...
proc getSurroundingSrc(conf: ConfigRef; info: TLineInfo): string = getSurroundingSrc(result, conf, info)
```
and then just used append ops

beef331: I think I get what you mean though don't see much benefit, the bigger concern to me presently is applying some heursitics to selecting only the pertinent information in the source hint. Though that gets rather complex rather quickly and may hide information.
Presently some messages look like this 
image
though helpful for this kind
image

timotheecour: > I think I get what you mean though don't see much benefit

just ignore it for now then (it's for performance reason).

> the bigger concern to me presently is applying some heursitics to selecting only the pertinent information in the source hint.

the problem is we're not tracking an interval in the source; TLineInfo only captures a point instead of interval (unlike tools like clang etc which can thus generate nicer diagnostics). I think we'll have to address this at some point but for this PR, how about simply colorize the next token starting from TLineInfo; i think it's a simple enough heuristic.

using lexer should tell you where the token ends, or maybe using a heuristic can work well (eg stop at whitespace or non-identifier character), although it'd require care for tokens of type stringlit or number literals, which is why re-using lexer would be the simplest/most robust. Let me know if that makes sense.

beef331: Yea it all makes sense, I mainly forgot about the fact the source can be printed out more than once which is why the confusion due to the in place version.

beef331:
Well hopefully all the tests pass now, but dont know with the colouring here due to hastily fixing the crashing, the required type probably shouldnt be quoted :smile: 
image

tests/concepts/t3330.nim
```diff
@@ -5,29 +5,29 @@ t3330.nim(70, 4) Error: type mismatch: got <Bar[system.int]>
 but expected one of:
```
timotheecour, line 5: how about coloring paths consistently in 1 color? eg:
`nim r --colors tests/misc/t9039.nim`
shows 
image

=> can the `../../lib/system/basic_types.nim(2, 3)` be colored like `t9039.nim(30, 22)` ?

before doing anything, is that hard to do with what you have?

tests/concepts/t3330.nim
```diff
@@ -5,29 +5,29 @@ t3330.nim(70, 4) Error: type mismatch: got <Bar[system.int]>
...
 proc test(foo: Foo[int])
```
timotheecour, line 6: can you color expression using a different color than Error / warning / hint ?

eg for 
image

using bold white is an option (but would still be good to distinguish expressions from paths but maybe not critical)

somehow it looks good in D:

image

tests/concepts/t3330.nim
```diff
@@ -5,29 +5,29 @@ t3330.nim(70, 4) Error: type mismatch: got <Bar[system.int]>
...
   first type mismatch at position: 1
-  required type for foo: Foo[int]
```
timotheecour, line 8: IMO we shouldn't remove the quotes when coloring an expression, so that copy pasting a msg on terminal gives the same message with --colors:on vs --colors:off (avoids un-necessary complications)

as can be seen in https://github.com/nim-lang/Nim/pull/16356/files#r670864680, the quotes are currently removed when colored

tests/concepts/t3330.nim
beef331: Could be done in the declared locs using the `colors` as a control to toggle it.

tests/concepts/t3330.nim
beef331: Yea this could be done, though I personally dont like the boldness in the D message as i can barely notice a difference(though it may just be your font), ideally we could have a user defined string with a `$color` as a marker for if they want colour, so then styling could be user definable using colour.

PR #16362, opened 2020-12-15
std/lists: O(1) concatenation of singly- and doubly linked lists.

salvipeter:
There is also new `toSinglyLinkedList` and `toDoublyLinkedList`
functions for conversion from `openArray`s, similarly
to `toHashSet` or `toTable`.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -77,7 +77,8 @@
 
 - Added `lists.copy` for shallow copying singly- and doubly linked lists.
 
-- `add` is now overloaded also for the O(1) concatenation of singly- and doubly linked lists.
+- `add` is now overloaded also for the concatenation of singly- and doubly linked lists;
+  an O(1) variation that consumes its argument, `addMove`, is also supplied.
 
 ## Language changes
 
--- a/changelog.md
+++ b/changelog.md
@@ -75,7 +75,7 @@
 - Added `lists.toSinglyLinkedList` and `lists.toDoublyLinkedList` for conversion from
   `openArray`s.
 
-- `add` is now overloaded also for the O(1) concatenation of singly- and doubly linked lists.
+- `add` is now overloaded also for the O(1) concatenation of `SinglyLinkedList`s.
 
 ## Language changes
 
--- a/changelog.md
+++ b/changelog.md
@@ -78,7 +78,7 @@
 - Added `lists.copy` for shallow copying singly- and doubly linked lists.
 
 - `add` is now overloaded also for the concatenation of singly- and doubly linked lists;
-  an O(1) variation that consumes its argument, `addMove`, is also supplied.
+  an O(1) variation that consumes its argument, `addMoved`, is also supplied.
 
 ## Language changes
 
--- a/changelog.md
+++ b/changelog.md
@@ -82,8 +82,7 @@
 - Added `lists.toSinglyLinkedList` and `lists.toDoublyLinkedList` for conversion from
   `openArray`s.
 
-- `lists.append` is now overloaded also for the O(1) concatenation of singly- and doubly
-  linked lists.
+- `add` is now overloaded also for the O(1) concatenation of singly- and doubly linked lists.
 
 ## Language changes
 
--- a/changelog.md
+++ b/changelog.md
@@ -65,6 +65,9 @@
 
 - `echo` and `debugEcho` will now raise `IOError` if writing to stdout fails.  Previous behavior
   silently ignored errors.  See #16366.  Use `-d:nimLegacyEchoNoRaise` for previous behavior.
+- Added `lists.toSinglyLinkedList` and `lists.toDoublyLinkedList` for conversion from
+  `openArray`s, and `lists.concat` for the O(1) concatenation of singly- and doubly
+  linked lists.
 
 ## Language changes
 
--- a/changelog.md
+++ b/changelog.md
@@ -75,8 +75,7 @@
 - Added `lists.toSinglyLinkedList` and `lists.toDoublyLinkedList` for conversion from
   `openArray`s.
 
-- `lists.append` is now overloaded also for the O(1) concatenation of singly- and doubly
-  linked lists.
+- `add` is now overloaded also for the O(1) concatenation of singly- and doubly linked lists.
 
 ## Language changes
 
--- a/changelog.md
+++ b/changelog.md
@@ -84,7 +84,8 @@
 
 - Added `lists.copy` for shallow copying singly- and doubly linked lists.
 
-- `add` is now overloaded also for the O(1) concatenation of singly- and doubly linked lists.
+- `add` is now overloaded also for the concatenation of singly- and doubly linked lists;
+  an O(1) variation that consumes its argument, `addMove`, is also supplied.
 
 ## Language changes
 
--- a/changelog.md
+++ b/changelog.md
@@ -71,7 +71,7 @@
 - Added `lists.copy` for shallow copying singly- and doubly linked lists.
 
 - `add` is now overloaded also for the concatenation of singly- and doubly linked lists;
-  an O(1) variation that consumes its argument, `addMove`, is also supplied.
+  an O(1) variation that consumes its argument, `addMoved`, is also supplied.
 
 ## Language changes
 
--- a/changelog.md
+++ b/changelog.md
@@ -85,7 +85,7 @@
 - Added `lists.copy` for shallow copying singly- and doubly linked lists.
 
 - `add` is now overloaded also for the concatenation of singly- and doubly linked lists;
-  an O(1) variation that consumes its argument, `addMove`, is also supplied.
+  an O(1) variation that consumes its argument, `addMoved`, is also supplied.
 
 ## Language changes
 
--- a/changelog.md
+++ b/changelog.md
@@ -80,7 +80,9 @@
 - `add` is now overloaded also for the concatenation of singly- and doubly linked lists;
   an O(1) variation that consumes its argument, `addMoved`, is also supplied.
 - Added `lists.toSinglyLinkedList` and `lists.toDoublyLinkedList` for conversion from
-  `openArray`s, and `lists.concat` for the O(1) concatenation of singly- and doubly
+  `openArray`s.
+
+- `lists.append` is now overloaded also for the O(1) concatenation of singly- and doubly
   linked lists.
 
 ## Language changes
--- a/changelog.md
+++ b/changelog.md
@@ -66,7 +66,9 @@
 - `echo` and `debugEcho` will now raise `IOError` if writing to stdout fails.  Previous behavior
   silently ignored errors.  See #16366.  Use `-d:nimLegacyEchoNoRaise` for previous behavior.
 - Added `lists.toSinglyLinkedList` and `lists.toDoublyLinkedList` for conversion from
-  `openArray`s, and `lists.concat` for the O(1) concatenation of singly- and doubly
+  `openArray`s.
+
+- `lists.append` is now overloaded also for the O(1) concatenation of singly- and doubly
   linked lists.
 
 ## Language changes
--- a/changelog.md
+++ b/changelog.md
@@ -73,7 +73,9 @@
 - `add` is now overloaded also for the concatenation of singly- and doubly linked lists;
   an O(1) variation that consumes its argument, `addMoved`, is also supplied.
 - Added `lists.toSinglyLinkedList` and `lists.toDoublyLinkedList` for conversion from
-  `openArray`s, and `lists.concat` for the O(1) concatenation of singly- and doubly
+  `openArray`s.
+
+- `lists.append` is now overloaded also for the O(1) concatenation of singly- and doubly
   linked lists.
 
 ## Language changes
--- a/changelog.md
+++ b/changelog.md
@@ -65,27 +65,11 @@
 
 - `echo` and `debugEcho` will now raise `IOError` if writing to stdout fails.  Previous behavior
   silently ignored errors.  See #16366.  Use `-d:nimLegacyEchoNoRaise` for previous behavior.
-- Added `lists.toSinglyLinkedList` and `lists.toDoublyLinkedList` for conversion from
-  `openArray`s.
 
-- Added `lists.copy` for shallow copying singly- and doubly linked lists.
-
-- `add` is now overloaded also for the concatenation of singly- and doubly linked lists;
-  an O(1) variation that consumes its argument, `addMoved`, is also supplied.
-- Added `lists.toSinglyLinkedList` and `lists.toDoublyLinkedList` for conversion from
-  `openArray`s.
-
-- Added `lists.copy` for shallow copying singly- and doubly linked lists.
-
-- `add` is now overloaded also for the concatenation of singly- and doubly linked lists;
-  an O(1) variation that consumes its argument, `addMoved`, is also supplied.
-- Added `lists.toSinglyLinkedList` and `lists.toDoublyLinkedList` for conversion from
-  `openArray`s.
-
-- Added `lists.copy` for shallow copying singly- and doubly linked lists.
-
-- `add` is now overloaded also for the concatenation of singly- and doubly linked lists;
-  an O(1) variation that consumes its argument, `addMoved`, is also supplied.
+- Added new operations for singly- and doubly linked lists: `lists.toSinglyLinkedList`
+  and `lists.toDoublyLinkedList` convert from `openArray`s; `lists.copy` implements
+  shallow copying; `lists.add` concatenates two lists - an O(1) variation that consumes
+  its argument, `addMoved`, is also supplied.
 
 ## Language changes
 
--- a/changelog.md
+++ b/changelog.md
@@ -68,7 +68,9 @@
 - Added `lists.toSinglyLinkedList` and `lists.toDoublyLinkedList` for conversion from
   `openArray`s.
 
-- `add` is now overloaded also for the O(1) concatenation of `SinglyLinkedList`s.
+- Added `lists.copy` for shallow copying singly- and doubly linked lists.
+
+- `add` is now overloaded also for the O(1) concatenation of singly- and doubly linked lists.
 
 ## Language changes
 
--- a/changelog.md
+++ b/changelog.md
@@ -68,7 +68,7 @@
 - Added `lists.toSinglyLinkedList` and `lists.toDoublyLinkedList` for conversion from
   `openArray`s.
 
-- `add` is now overloaded also for the O(1) concatenation of singly- and doubly linked lists.
+- `add` is now overloaded also for the O(1) concatenation of `SinglyLinkedList`s.
 
 ## Language changes
 
--- a/changelog.md
+++ b/changelog.md
@@ -72,6 +72,9 @@
 
 - `add` is now overloaded also for the concatenation of singly- and doubly linked lists;
   an O(1) variation that consumes its argument, `addMoved`, is also supplied.
+- Added `lists.toSinglyLinkedList` and `lists.toDoublyLinkedList` for conversion from
+  `openArray`s, and `lists.concat` for the O(1) concatenation of singly- and doubly
+  linked lists.
 
 ## Language changes
 
--- a/changelog.md
+++ b/changelog.md
@@ -79,6 +79,9 @@
 
 - `add` is now overloaded also for the concatenation of singly- and doubly linked lists;
   an O(1) variation that consumes its argument, `addMoved`, is also supplied.
+- Added `lists.toSinglyLinkedList` and `lists.toDoublyLinkedList` for conversion from
+  `openArray`s, and `lists.concat` for the O(1) concatenation of singly- and doubly
+  linked lists.
 
 ## Language changes
 
--- a/changelog.md
+++ b/changelog.md
@@ -82,7 +82,9 @@
 - Added `lists.toSinglyLinkedList` and `lists.toDoublyLinkedList` for conversion from
   `openArray`s.
 
-- `add` is now overloaded also for the O(1) concatenation of `SinglyLinkedList`s.
+- Added `lists.copy` for shallow copying singly- and doubly linked lists.
+
+- `add` is now overloaded also for the O(1) concatenation of singly- and doubly linked lists.
 
 ## Language changes
 
--- a/changelog.md
+++ b/changelog.md
@@ -75,7 +75,9 @@
 - Added `lists.toSinglyLinkedList` and `lists.toDoublyLinkedList` for conversion from
   `openArray`s.
 
-- `add` is now overloaded also for the O(1) concatenation of `SinglyLinkedList`s.
+- Added `lists.copy` for shallow copying singly- and doubly linked lists.
+
+- `add` is now overloaded also for the O(1) concatenation of singly- and doubly linked lists.
 
 ## Language changes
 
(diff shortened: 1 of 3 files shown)


Araq:
Thanks! Please add ``.since`` annotations, tests (edit tests/stdlib/tlists.nim) and a changelog entry.

Clyybber:
I think these should be overloads of `append`/`prepend`.

salvipeter:
Added the annotations (I hope they're OK), the tests, and the changelog.
As for the naming... I also first searched for this functionality as `append` (and yes, a `prepend` version can also be done), but on the other hand, `sequtils` has `concat`.

Clyybber:
Yes but `sequtils.concat` is not an inplace operation (and has varargs).

salvipeter:
Okay, I renamed it. But thinking about it, I found that `prepend` is not really a good match, because always the first member changes, and it would be strange to say `l1.prepend(l2)`, modifying `l2` while `l1` stays the same. We should just use `l2.append(l1)`.

Araq:
Instead of "concat" you can use "&", but I don't like using prepend/append for it.

salvipeter:
I personally like the `append` idea, it is very straightforward naming, and consistent with other languages. But I'm open to other suggestions, and an operator (even if only as an alias) is useful. How are these questions decided? By votes?

Araq:
Well ``&`` is Nim's concat operator, nothing controversial about it. We tend to vote for controversial things.

salvipeter:
The problem with `&` that it gives the wrong impression - that it is a pure function returning a new instance. I would certainly not expect `l1 & l2` to change `l1` (but I would expect it to return a value!).

timotheecour:
> Instead of "concat" you can use "&", but I don't like using prepend/append for it.

`&` is outplace, not inplace, so you probably mean `&=`, but it's much less standard than `add` and an operator has no benefit here.

Almost all of stdlib uses `add` for inplace append, so we should use that instead of this PR's `append`.

```
ropes: add(a: var Rope; b: Rope)
ropes: add(a: var Rope; b: string)
system: add[T](x: var seq[T]; y: openArray[T])
system: add(x: var string; y: cstring)
unicode: add(s: var string; c: Rune)
macros: add(father, child: NimNode): NimNode
macros: add(father: NimNode; children: varargs[NimNode]): NimNode
...
```

it's also what's recommended in nep1 explicitly: https://nim-lang.github.io/Nim/nep1.html
`use add instead of append`

(note: `append` is currently only used inside `lists` module and the (AFAIK unused) `std/chains` module, we can `{.deprecated: [append: add].}` in future work)

salvipeter:
@timotheecour has a point there... if the others agree, I'll rename it to `add` then.

Araq:
> & is outplace, not inplace, so you probably mean &=, but it's much less standard than add and an operator has no benefit here.

Indeed. However, why not instead add ``&`` which is outplace. People who are into lists are likely more into FP so "consistency with the rest of Nim" where we have inplace + dup doesn't really convince me.

salvipeter:
> why not instead add & which is outplace

Because the whole point of this PR is to create an O(1) concatenation :)
We can add an outplace `&` operator, but it will be O(n).

salvipeter:
By the way, `dup` does not work on lists, because it is just an assignment under the hood, IIUC, not a `deepCopy`, but we could have
```nim
import lists, sugar
func `&`(L1, L2: SinglyLinkedList): SinglyLinkedList = L1.deepCopy.dup(add(L2.deepCopy))
```
... or `L2` may be without `deepCopy`, which would be what I would expect (as someone coming from Lisp), but it does not match the functionality of `&` elsewhere in Nim.

Araq:
> Because the whole point of this PR is to create an O(1) concatenation :)
> We can add an outplace & operator, but it will be O(n).

Ah, good one.

salvipeter:
I've just realized that this operation is not really meaningful for doubly linked lists, as it will always leave one in an inconsistent state. Consider
```nim
var
  a = [1, 2, 3].toDoublyLinkedList
  b = [4, 5, 6].toDoublyLinkedList
  c = a.add(b)
```
If I set the `prev` pointer of the first node in `b` to the last node of `a`, after the `add` operation `c` will be consistent, but `b` won't be. OTOH if `b` is not changed (as in my implementation), then `c` will be inconsistent.

I can't see a really good solution, so I removed the `DoublyLinkedList` implementation.

tests/stdlib/tlists.nim
```diff
@@ -81,3 +81,39 @@ block tlistsToString:
     l.append('2')
     l.append('3')
     doAssert $l == """['1', '2', '3']"""
+
+block SinglyLinkedListConversion:
```
timotheecour: name the block after the symbols you're testing: `toSinglyLinkedList`

it makes it easier to search for tests for a given symbol
(note that merely using a symbol in some test isn't the same as a dedicated block testing it)

if a block tests multiple symbols, use:
```nim
block: # foo1, foo2
  
```

tests/stdlib/tlists.nim
```diff
@@ -81,3 +81,39 @@ block tlistsToString:
...
+  let l: seq[int] = @[]
```
timotheecour: `let l = seq[int].default`

tests/stdlib/tlists.nim
```diff
@@ -81,3 +81,39 @@ block tlistsToString:
...
+  doAssert $l.toSinglyLinkedList == "[]"
```
timotheecour: use toSeq for tests; in general it's more robust and independent of how something is rendered

`doAssert [1].toSinglyLinkedList.toSeq == [1]`

timotheecour:
> I've just realized that this operation is not really meaningful for doubly linked lists, as it will always leave one in an inconsistent state.

Actually there's a similar problem with `SinglyLinkedList`: `b` is left in inconsistent state after being added twice, and other problems, see RFC https://github.com/nim-lang/RFCs/issues/303

> I can't see a really good solution, so I removed the DoublyLinkedList implementation.

I have one, see RFC

> But thinking about it, I found that prepend is not really a good match, because always the first member changes, and it would be strange to say l1.prepend(l2), modifying l2 while l1 stays the same. We should just use l2.append(l1).

see RFC, which explains that can't work and `prepend` still needs to be implemented

salvipeter:
Revised the tests according to @timotheecour's comments.

lib/pure/collections/lists.nim
```diff
@@ -458,20 +458,32 @@ proc prepend*[T](L: var SinglyLinkedList[T], value: T) {.inline.} =
...
-proc add*[T](L1: var SinglyLinkedList[T], L2: SinglyLinkedList[T]) {.since: (1, 5).} =
-  ## Adds `L2` to the end of `L1`. Efficiency: O(1).
-  ## Note that the two lists share structure after the operation.
+func copy*[T](L: SinglyLinkedList[T]): SinglyLinkedList[T] {.since: (1, 5).} =
+  ## Creates a shallow copy of `L`.
   runnableExamples:
-    var a = [1, 2, 3].toSinglyLinkedList
-    let b = [4, 5].toSinglyLinkedList
+    let a = [1, 2, 3].toSinglyLinkedList
+    assert $a == $a.copy
```
timotheecour: can you expand on this example to illustrate the semantics of shallow copying? eg using a ref object as in https://github.com/nim-lang/RFCs/issues/303#issuecomment-747350870 or something better

lib/pure/collections/lists.nim
```diff
@@ -458,20 +458,32 @@ proc prepend*[T](L: var SinglyLinkedList[T], value: T) {.inline.} =
...
+  result = initSinglyLinkedList[T]()
+  for x in L:
+    result.append(x)
+
+proc add*[T](L1, L2: var SinglyLinkedList[T]) {.since: (1, 5).} =
+  ## Moves `L2` to the end of `L1`. Efficiency: O(1).
+  ## Note that `L2` becomes empty after the operation.
```
timotheecour: needs to explain what happens for `a.add a`

tests/stdlib/tlists.nim
```diff
@@ -95,25 +95,51 @@ block toDoublyLinkedList:
...
+        l1 = [1].toSinglyLinkedList
+        l2 = [2, 3].toSinglyLinkedList
+        l3 = [4, 5, 6].toSinglyLinkedList
+      l0.add(l3.copy)
+      l1.add(l3.copy)
+      l2.add(l3)
+      doAssert l0.toSeq == [4, 5, 6]
+      doAssert l1.toSeq == [1, 4, 5, 6]
+      doAssert l2.toSeq == [2, 3, 4, 5, 6]
+      doAssert l3.toSeq == []
```
timotheecour, line 124: add:
```nim
l2.add(l3) # re-adding l3 that was destroyed is now a noop
doAssert l3.toSeq == []
doAssert l2.toSeq == [2, 3, 4, 5, 6]
```

salvipeter: Sure, but that seems pretty self-explanatory from `l3.toSeq == []` and the `l3.add(l0)` example in the other block.

lib/pure/collections/lists.nim
salvipeter: It becomes empty, with the current implementation.

lib/pure/collections/lists.nim
salvipeter: OK

lib/pure/collections/lists.nim
timotheecour: empty sucks though;
how about raising instead? ideally would be disallowed at CT (like rust) but for now raising might be best we can do

salvipeter: We can do that, but I don't know ... somehow I feel that we concentrate on a special case too much. A lot worse than empty can happen when someone starts shuffling `head`s and `tail`s, even when sticking with the API's `append`. (You can do `a.append(a.head)`, after all, which results in a list of just the first element.)
I'm just wondering how far we should intervene... I'm all for liberty :)

lib/pure/collections/lists.nim
```diff
@@ -461,22 +461,35 @@ proc prepend*[T](L: var SinglyLinkedList[T], value: T) {.inline.} =
...
-    assert $a == $a.copy
+    type Foo = ref object
+      x: int
+    var f = Foo(x: 1)
+    let
+      a = [f].toSinglyLinkedList
+      b = a.copy
+    f.x = 42
+    assert a.head.value.x == 42
+    assert b.head.value.x == 42
```
timotheecour: add this or similar to emphasize the copy semantics. Makes behavior even more crystal clear.
(requiers changing a to var)
```nim
a.add [f].toSinglyLinkedList
doAssert a.toSeq == [f, f]
doAssert b.toSeq == [f] # b isn't modified...

f.x = 42
assert a.head.value.x == 42
assert b.head.value.x == 42 # ... but elements are not deep copied
```

tests/stdlib/tlists.nim
timotheecour: but the test file currently doesn't even exercise the case where you're appending an empty list; test files should test edge cases to prevent regressions from future implementation changes (or seemingly unrelated changes).

so actually I'd add that snippet, as well as a snippet that calls `a.add b` where `a` is empty. (IIRC it wasn't exercised)

tests/stdlib/tlists.nim
```diff
@@ -81,3 +81,88 @@ block tlistsToString:
...
+      var
+        l0 = initSinglyLinkedList[int]()
+        l1 = [1].toSinglyLinkedList
+        l2 = [2, 3].toSinglyLinkedList
+        l3 = [4, 5, 6].toSinglyLinkedList
+      l3.add(l0)
+      l2.add(l1)
+      doAssert l3.toSeq == [4, 5, 6]
+      doAssert l2.toSeq == [2, 3, 1]
+  block: # DoublyLinkedList
```
timotheecour: there's a ton of duplication here. instead use this:
```nim
template testCommon(initList, toList) =
  var
    l0 = initDoublyLinkedList[int]()
    l1 = [1].toList
    l2 = [2, 3].toList
    l3 = [4, 5, 6].toList
    l4 = l3.copy
    l5 = l3.copy
  l0.add(l3)
  l1.add(l4)
  l2.add(l5)

testCommon(initDoublyLinkedList, toDoublyLinkedList)
testCommon(initSinglyLinkedList, toSinglyLinkedList)
```

it guarantees equal coverage, and simplifies maintenance etc

tests specific to doubly or singly can be in a different block, or, if it's a small diff, you can always use:
```nim
when type(l3) is DoublyLinkedList: ...
```

tests/stdlib/tlists.nim
salvipeter: Empty lists on both the left- and right-hand sides are tested (`l0.add(l3)` in the first sub-block, and vice versa in the second sub-block).

lib/pure/collections/lists.nim
salvipeter: If you insist... but (at least for me) the immutable argument in the declaration is enough to be sure that `b` is not modified.

tests/stdlib/tlists.nim
salvipeter: Thanks, that will improve the code a lot, I was also bothered by the code duplication, but templates didn't occur to me.

lib/pure/collections/lists.nim
timotheecour: > the immutable argument in the declaration is enough to be sure that b is not modified

unfortunately that's not the case, see https://github.com/nim-lang/Nim/issues/16312#issuecomment-742892857

tests/stdlib/tlists.nim
```diff
@@ -82,87 +82,60 @@ block tlistsToString:
...
-      l2.add(l1)
+        l0 = initList[int]()
+        l1 = [1].toList
+        l2 = [2, 3].toList
+        l3 = [4, 5, 6].toList
+      l3.add l0
+      l2.add l1
       doAssert l3.toSeq == [4, 5, 6]
       doAssert l2.toSeq == [2, 3, 1]
+  testCommon(initSinglyLinkedList, toSinglyLinkedList)
```
timotheecour: nit: you'd have less boilerplate with a single `testCommon`, and sub-blocks for each common proc

```nim
block: # SinglyLinkedList, DoublyLinkedList
  template testCommon(initList, toList) =
    block copy: ...
    block add: ...
    ...
  testCommon(initSinglyLinkedList, toSinglyLinkedList)
```

tests/stdlib/tlists.nim
timotheecour: ok!

tests/stdlib/tlists.nim
salvipeter: I also considered that, but I assumed that a single top-level block for each tested procedure was the standard :)

lib/pure/collections/lists.nim
salvipeter: I'm not sure I understand ... from that thread I see that an assignment can make a (bad) shallow copy, and mess up things. But surely a `func` such as this `copy` cannot do that?
... or maybe it can. But - to quote you - this is really bad.

lib/pure/collections/lists.nim
```diff
@@ -498,6 +499,20 @@ proc add*[T](L1, L2: var SinglyLinkedList[T]) {.since: (1, 5).} =
   L2.head = nil
   L2.tail = nil
 
+proc add*[T](L1: var SinglyLinkedList[T], L2: SinglyLinkedList[T]) {.since: (1, 5).} =
```
timotheecour: let's not propagate pre-existing bad code :)
this violates style guide (caps)

instead use:
```nim
proc add*[T](a: var SinglyLinkedList[T], b: SinglyLinkedList[T]) {.since: (1, 5).} =
```
(or s/a/list/ but it's implied by the type)

future PR can fix the pre-existing style violations

lib/pure/collections/lists.nim
```diff
@@ -626,6 +642,20 @@ proc add*[T](L1, L2: var DoublyLinkedList[T]) {.since: (1, 5).} =
   L2.head = nil
   L2.tail = nil
 
+proc add*[T](L1: var DoublyLinkedList[T], L2: DoublyLinkedList[T]) {.since: (1, 5).} =
```
timotheecour: factor (as other procs do in lists.nim):
```nim
proc add*[T: SomeLinkedList](a: var T, b: T) {.since: (1, 5).} =
  var c = b.copy
  a.addMove c
```

since its implementation is identical
(runnableExamples can show an example of each type, or not, (up to you); in any case both are tested in tlists.nim)

And a future `prepend` can also do like that

salvipeter:
There was a conflict with changelog.md so I did a rebase ... not entirely sure what happened. Anyway, added the extended example to `lists.copy`. Anything else left to do?

lib/pure/collections/lists.nim
```diff
@@ -178,6 +179,28 @@ proc newSinglyLinkedNode*[T](value: T): <//>(SinglyLinkedNode[T]) =
   new(result)
   result.value = value
 
+func toSinglyLinkedList*[T](elems: openArray[T]): SinglyLinkedList[T] {.since: (1, 6).} =
```
timotheecour: `(1,5,1)`
otherwise you break devel code; 1.6.x hasn't been released yet

(+ elsewhere)

timotheecour:
> There was a conflict with changelog.md so I did a rebase ... not entirely sure what happened.

- [ ] explanation here https://github.com/nim-lang/Nim/pull/14098#issuecomment-619454614 (`/changelog.md merge=union` is the lesser evil, things were worse before)

> Anyway, added the extended example to lists.copy. Anything else left to do?

- [ ] CI red, eg: 2020-12-18T20:55:19.3172340Z tlists.nim(89, 15) Error: undeclared field: 'toSinglyLinkedList' , probably because of https://github.com/nim-lang/Nim/pull/16362/files#r546099031

- [ ] please rebase against devel, right now your PR shows a giant unrelated diff (eg tools/dochack/dochack.nim), so it's impossible to review

lib/pure/collections/lists.nim
salvipeter: OK. So these will be later updated automatically to (1, 6)? (since these will actually only be in that version, at least according to the changelog :))

timotheecour: > So these will be later updated automatically to (1, 6)?

there will be no need to update anything, look at the implementation of `since`. It will "just work", because once 1.6 is released, `1.6 >= (1,5,1)`

changelog.md
```diff
@@ -65,6 +65,27 @@
 
 - `echo` and `debugEcho` will now raise `IOError` if writing to stdout fails.  Previous behavior
   silently ignored errors.  See #16366.  Use `-d:nimLegacyEchoNoRaise` for previous behavior.
+- Added `lists.toSinglyLinkedList` and `lists.toDoublyLinkedList` for conversion from
```
timotheecour: * changelog needs to be fixed (reason for duplication is `/changelog.md merge=union`, but it's there for a good reason)

* group everything for this PR under just 1 entry

* newline between preceding entry (`echo and debugEcho...`) and your new entry

lib/pure/collections/lists.nim
```diff
@@ -178,6 +179,28 @@ proc newSinglyLinkedNode*[T](value: T): <//>(SinglyLinkedNode[T]) =
   new(result)
   result.value = value
 
+func toSinglyLinkedList*[T](elems: openArray[T]): SinglyLinkedList[T] {.since: (1, 5, 1).} =
+  ## Creates a new `SinglyLinkedList` with the members of the
+  ## given collection (seq, array, or string) `elems`.
```
timotheecour: that's not the place to describe what openArray is; it would violate "one definition rule" for docs; note that openArray definition is 1 click away thanks to docgen
=> simply:
```nim
Creates a new `SinglyLinkedList` from members of `elems`
```

(+ elsewhere)

tests/stdlib/tlists.nim
```diff
@@ -81,3 +81,63 @@ block tlistsToString:
...
+
+template testCommon(initList, toList) =
+
+  block: # toSinglyLinkedList, toDoublyLinkedList
+    let l = seq[int].default
+    doAssert l.toList.toSeq == []
+    doAssert [1].toList.toSeq == [1]
+    doAssert [1, 2, 3].toList.toSeq == [1, 2, 3]
+
+  block copy:
```
timotheecour, line 93: this test doesn't actually test correctness of `copy` (2 parts: source container is not modified; but elements are references)

(Remember that runnableExamples is only run for 1 specific backend, unlike this which will be tested on all backends (eventually, once https://github.com/nim-lang/Nim/issues/16384 is fixed))

EDIT: how about this:
```nim
  type Foo = ref object
    x: int
  let f0 = Foo(x: 0)
  let f1 = Foo(x: 1)
  var a = [f0]. toList
  var b = a.copy
  b.append f1
  doAssert a.toSeq == [f0]
  doAssert b.toSeq == [f0, f1]
```

lib/pure/collections/lists.nim
```diff
@@ -523,6 +591,72 @@ proc prepend*[T](L: var DoublyLinkedList[T], value: T) =
...
+    a.addMoved a
+    assert a.toSeq == []
+  if b.head != nil:
+    b.head.prev = a.tail
+  if a.tail != nil:
+    a.tail.next = b.head
+  a.tail = b.tail
+  if a.head == nil:
+    a.head = b.head
+  b.head = nil
```
timotheecour: change to:
```nim
  if a.addr != b.addr:
    b.head = nil
    b.tail = nil
```

and add this test:
```nim
  import lists, sequtils
  import std/enumerate
  import std/sugar
  var a = [0,1].toSinglyLinkedList # actualy, `toList` in your test
  a.addMoved a
  let s = collect:
    for i, ai in enumerate(a):
      if i >= 6: break
      ai
  doAssert s == [0,1,0,1,0,1]
```

and edit doc comment to:
```nim
  ## Note that `b` becomes empty after the operation.
  ## Self-adding results in an empty list.
```
=>
```nim
  ## Note that `b` becomes empty after the operation unless it has same address as `a`.
```

that's IMO a saner behavior

* ditto with SinglyLinkedList

lib/pure/collections/lists.nim
```diff
@@ -435,7 +456,54 @@ proc prepend*[T](L: var SinglyLinkedList[T], value: T) {.inline.} =
...
+    assert a.toSeq == [f, f]
+    assert b.toSeq == [f] # b isn't modified...
+    f.x = 42
+    assert a.head.value.x == 42
+    assert b.head.value.x == 42 # ... but the elements are not deep copied
+
+    let c = [1, 2, 3].toSinglyLinkedList
+    assert $c == $c.copy
+  result = initSinglyLinkedList[T]()
+  for x in a:
```
timotheecour: * `a.items` is needed until https://github.com/nim-lang/Nim/issues/11167 is fixed
(ie same workaround as done in https://github.com/nim-lang/Nim/pull/16060)

* copy also needs this fix

lib/pure/collections/lists.nim
salvipeter: If I change just this, it will create a circular list. I assume that you meant that self-addition should be a no-op, right?

salvipeter:
Done... btw, why can't I write `a.enumerate` instead of `enumerate(a)` ? This seems to be against the general logic of the language.

timotheecour:
> btw, why can't I write a.enumerate instead of enumerate(a) ? This seems to be against the general logic of the language.

=> followup in https://github.com/timotheecour/Nim/issues/473 to avoid hijacking this

timotheecour (approved):
LGTM. thanks for your patience!

salvipeter:
And thank you for your comments, I learned a lot in the process :)

PR #16382, opened 2020-12-17
RST enumlist followup

a-mr:
This adds possibility to use `1` as auto-enumerator in enumerated lists (follow-up to #16295)

```
1. x
1. y
```
generates `1-2` list.

`#` and `1` can not be used in one list as auto-numerator simultaneously — it will split the list to two.

However the exception:
```
1. x
#. y
```
works fine as a pure RST list, 1 does not go as auto-enumerator.

Also i noticed a bug that special case of auto-enumerator in format `#)` did not work, fix's in commit 1.

### DIFF
--- a/lib/packages/docutils/rst.nim
+++ b/lib/packages/docutils/rst.nim
@@ -48,6 +48,10 @@
 ## * markdown code blocks
 ## * markdown links
 ## * markdown headlines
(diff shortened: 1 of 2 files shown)


a-mr:
CC @narimiran

PR #16396, opened 2020-12-18
ARM-based CI for Linux [ci skip]

alaviss:
Only a proof-of-concept atm.

Progress can be tracked here: https://cloud.drone.io/alaviss/Nim/


drone.yml
```diff
@@ -0,0 +1,45 @@
...
+clone:
+  depth: 1
+
+steps:
+- name: runci
+  image: ghcr.io/alaviss/nim-ci:sha-1d368dc@sha256:b03ad81192e7e14dec4471b63a3d98c78ceada2e675d9075151193d972814b26
+  commands:
+  - git clone --depth 1 https://github.com/nim-lang/csources.git
+  - export PATH=$PWD/bin:$PATH
+  - make -C csources -j$(nproc)
```
timotheecour:
```suggestion
  - . ci/funs.sh && nimBuildCsourcesIfNeeded
  - export PATH=$PWD/bin:$PATH
```

(updates to new csources)
ditto in the other places where `git clone --depth 1 https://github.com/nim-lang/csources.git` is used

also, the more the CI's get refactored to share common code the better; ideally most of the logic would be in reusable "APIs" in ci/funs.sh

EDIT: see https://github.com/nim-lang/Nim/pull/18035 which refactors some more to further reduce duplication

drone.yml
```diff
@@ -0,0 +1,45 @@
...
+platform:
+  os: linux
+  arch: arm64
+
+clone:
+  depth: 1
+
+steps:
+- name: runci
+  image: ghcr.io/alaviss/nim-ci:sha-1d368dc@sha256:b03ad81192e7e14dec4471b63a3d98c78ceada2e675d9075151193d972814b26
```
timotheecour: can this be hosted in nim-lang/nimdockers (or some other repo), along with a Dockerfile so that it's reproducible from sources?

alaviss: It's already hosted at alaviss/nim-ci. I can move it to nim-lang, sure.

timotheecour:
you're hitting the dreaded https://github.com/nim-lang/Nim/issues/17456 again; i've restarted osx CI for you

tools/ci_generate.nim
```diff
@@ -102,6 +102,51 @@ nimBuildCsourcesIfNeeded "$@"
 {genBuildExtras("echo_run", "./koch", "bin/nim")}
 """
 
+proc genDroneScript(): string =
+  const Image = "ghcr.io/alaviss/nim-ci:sha-87f5a84"
+  type
+    Arch = enum
+      arm64
+      arm
+
```
timotheecour, line 111:
```suggestion
  result = fmt"# {doNotEdit}"
```
(replace `#` with whatever is recognized as comment marker for drone yml)

see how other pipelines use this so that ppl know where the generated code comes from and are not tempted to edit the generated code directly;

alaviss: If you notice the latest commit, I had to remove it. It doesn't seem to play well with drone yaml.

timotheecour: * well the last commit also adds `nim c koch` so it's unclear it's due to that comment?

* where can i see CI logs? if i click on links inside https://cloud.drone.io/alaviss/Nim/ it shows 404's

* how about using yaml syntax instead of json, since our other pipelines use yaml and it's arguably more readable; eg see https://github.com/hbokh/docker-drone-builder/blob/master/.drone.yml (which btw also contains `#` for comments), or pretty much most drone.yml files one can find on github
* should it be drone.yml or .drone.yml ? (i honestly don't know)

tools/ci_generate.nim
```diff
@@ -140,6 +185,7 @@ packages:
   writeFile("build_all.sh", genPosixScript())
   writeFile("build_all.bat", genWindowsScript(buildAll = true))
   writeFile("ci/build_autogen.bat", genWindowsScript(buildAll = false))
+  writeFile("drone.yml", genDroneScript())
```
timotheecour, line 188: shouldn't that be .drone.yml? (IIRC that's what's commonly used, and the default)

I've enabled drone for nim-lang/nim, see https://cloud.drone.io/nim-lang/Nim/settings
(and assumes .drone.yml, was was the default, although we could configure it but not sure it'd work for forks unless the fork also sets this?)


image

tools/ci_generate.nim
alaviss: > where can i see CI logs? if i click on links inside https://cloud.drone.io/alaviss/Nim/ it shows 404's

Hmm, not sure, it used to work without logging in.

> how about using yaml syntax instead of json, since our other pipelines use yaml and it's arguably more readable; eg see https://github.com/hbokh/docker-drone-builder/blob/master/.drone.yml (which btw also contains # for comments), or pretty much most drone.yml files one can find on github

It's easier to build with json, that's all. And the code is not even that complicated...

> should it be drone.yml or .drone.yml ? (i honestly don't know)

we can name it whatever, you get to set it on repository setup

PR #16402, opened 2020-12-19
Update parsecfg.nim  Add:  1. Users can specify comment symbols.  2, keep the original file format, white space and comments will be retained.

lihf8515:


lib/pure/parsecfg.nim
```diff
@@ -295,252 +371,440 @@ proc getEscapedChar(c: var CfgParser, tok: var Token) =
...
   open(p, stream, filename)
+  p.commentSymbol = commentSymbol
   while true:
     var e = next(p)
     case e.kind
     of cfgEof:
       break
     of cfgSectionStart: # Only look for the first time the Section
+      var tp: SectionItem
+      var t = <//>OrderedTableRef[string, KeyValueItem]()
```
juancarlospaco: I think theres other PR that removes `<//>`, I think is Deprecated or almost deprecated.

lib/pure/parsecfg.nim
```diff
@@ -622,20 +905,53 @@ proc getSectionValue*(dict: Config, section, key: string, defaultVal = ""): stri
...
+      kvi.value = value
+      if not allCharsInSet(value, SymChars): ## Non system character
+        kvi.keyValueRelated.valueStringKind = skString
+      else:
+        kvi.keyValueRelated.valueStringKind = skSymbol
+      kv.add(key, kvi)
+    tp.keyValue = kv
+    dict[section] = tp
+  else: # add section
+    if key.startsWith("--") or key.startsWith("-"):
```
juancarlospaco, line 955: On proc that take `char` and if you use a single char string, use the `char` overload, 
because operations over a `char` tends to be more efficient than over a `string`.

lib/pure/parsecfg.nim
```diff
@@ -562,57 +826,76 @@ proc replace(s: string): string =
...
+        s.add(section)
+      s.add(sectionRelated.sectionRearBlank & sectionRelated.tokenRight &
+            sectionRelated.tokenRearBlank & sectionRelated.comment)
+      stream.writeLine(s)
+    for key, kv in keyValue.pairs():
+      var newKey = ""
+      s = ""
+      s.add(kv.keyValueRelated.keyFrontBlank)
+      newKey = key
+      if kv.keyValueRelated.keyStringKind == skLongString:
```
juancarlospaco, line 865: Some of these `if` should be `case` switches.

lib/pure/parsecfg.nim
```diff
@@ -295,252 +371,440 @@ proc getEscapedChar(c: var CfgParser, tok: var Token) =
...
+      dict[curSection] = tp
     of cfgError:
-      break
+      close(p)
+      raise newException(Exception, e.msg)
   close(p)
   result = dict
 
-proc loadConfig*(filename: string): <//>Config =
+proc loadConfig*(filename: string, commentSymbol = "#;"): <//>Config =
```
juancarlospaco: `commentSymbols = {'#', ';'}`

lihf8515: All of these optimizations are necessary, but for now My focus is on the inspection not passing.

juancarlospaco:
If you are rewriting most of it, give it a Destructor IMHO.

lihf8515:
I don't quite understand. How to destruct?

lib/pure/parsecfg.nim
juancarlospaco: Was not about optimization, is more correctness of the code too, 
when something is a char then must be a `char`, Nim has a concrete type for that.

e.g. `commentSymbol = "##"` and :bomb: :boom:  
But `commentSymbol = {'#', '#'}` can not be.

lihf8515: The main thing is that I now add the function of empty lines and comment lines to 'parsecfg', but 'nimble' checks that it doesn't read 'key' as desired when he calls' parsecfg 'to read' key ', so the checks are always wrong. It seems we need to modify 'nimble'.

Araq:
Sorry for this terribly long feedback cycle (1 year!), but this change is large and adds much code that we then have to support for good. Please publish your work first as a Nimble package.

PR #16431, opened 2020-12-21
add sequtils to prelude

shadowninja55:
i would argue that sequtils is used just as often as the other imports in prelude, and it'd be nice for it to be included.

### DIFF
--- a/lib/prelude.nim
+++ b/lib/prelude.nim
@@ -19,5 +19,5 @@
 ##   import os, strutils, times, parseutils, hashes, tables, sets
 ##   when not defined(js): import parseopt
 
-import os, strutils, times, parseutils, hashes, tables, sets
+import os, strutils, times, parseutils, hashes, tables, sets, sequtils
 when not defined(js): import parseopt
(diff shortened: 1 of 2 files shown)


juancarlospaco:
Yes, please, some people asked for this, `sequtils` on `prelude` is very useful. :+1:

lib/prelude.nim
```diff
@@ -19,5 +19,5 @@
 ##   import os, strutils, times, parseutils, hashes, tables, sets
```
ringabout: Please update docs too

shadowninja55: got it

ringabout (approved):
Looks good to me

timotheecour:
LGTM, no objection to this PR.

Although I still think `import std/stdimports` instead of `include prelude` is a better approach as discussed in https://github.com/nim-lang/Nim/issues/16238

Note: this PR shouldn't be affected by the sequtils.toSeq issue of https://github.com/nim-lang/RFCs/issues/512 (see https://github.com/nim-lang/RFCs/issues/512 for details)

timotheecour (approved):

Issue #1211, opened 2014-05-20
Docutils: ReST: Add support for admonitions


moigagoo:
Currently, none of the admonition directives (note, important, hint, admonition, etc.) are supported.

Can I help?

Araq:
No feedback, feature not important.

PR #16438, opened 2020-12-22
RST: implement admonitions

a-mr:
Here is how they look (in dark theme): 
admonitions

Latex rendering is more simplistic:
image

Admonitions are grouped to 3 different types: info (note in the picture), warnings, errors.

Also documentation in rst.nim is expanded.

CC @narimiran

### DIFF
--- a/lib/packages/docutils/rst.nim
+++ b/lib/packages/docutils/rst.nim
@@ -7,11 +7,18 @@
 #    distribution, for details about the copyright.
 #
 
-## This module implements a `reStructuredText`:idx: parser. A large
-## subset is implemented. Some features of the `markdown`:idx: wiki syntax are
-## also supported.
+## This module implements a `reStructuredText`:idx: (RST) parser. A large
+## subset is implemented. Some features of the `markdown`:idx: syntax are
+## also supported. Nim can output the result to HTML (command ``rst2html``)
+## or Latex (command ``rst2tex``).
 ##
-## Supported RST features:
+## If you are new to RST please consider reading the following:
+##
+## 1) a short `quick introduction`_
+## 2) an `RST reference`_: a comprehensive cheatsheet for RST
+## 3) a more formal 50-page `RST specification`_.
+##
+## Supported standard RST features:
 ##
 ## * body elements
 ##   + sections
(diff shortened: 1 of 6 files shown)


timotheecour:
maybe add this to PR message: fixes https://github.com/nim-lang/Nim/issues/1211
(even though that issues was already closed)

lib/packages/docutils/rstast.nim
```diff
@@ -316,3 +320,13 @@ proc renderRstToJson*(node: PRstNode): string =
   ##     "sons":optional node array
   ##   }
   renderRstToJsonNode(node).pretty
+
+proc dumpRstTree*(node: PRstNode, indent=0) =
```
timotheecour: procs that echo are almost never the right thing to do, even for debugging. Instead, return a string and let test code use echo if they want to

lib/packages/docutils/rst.nim
```diff
@@ -1891,28 +1930,36 @@ proc dirRaw(p: var RstParser): PRstNode =
...
+  "code-block": dirCodeBlock,
+  "container": dirContainer,
+  "contents": dirContents,
+  "figure": dirFigure,
+  "image": dirImage,
+  "include": dirInclude,
+  "index": dirIndex,
+  "raw": dirRaw,
+  "title": dirTitle
+}.toTable
```
planetis-m: Not to sound picky but for a Table this small Im not really sure if its worth it, linear search is pbl better

FedericoCeratto:
Thanks for this!
I find the color bars to be a bit overwhelming compared to other elements like headers. May I suggest a smaller visual element like: https://bootsnipp.com/snippets/qAdNQ ?

a-mr:
@FedericoCeratto ,
now it looks like this.
```
## .. Note:: By default nim has ``roSupportMarkdown`` turned **on**.
##
## .. warning:: Using Nim-specific features can cause other RST implementations
##   to fail on your document.
##
## Limitations:
##
## .. error:: Unknown error
```

image

image

lib/packages/docutils/rstast.nim
a-mr: thanks, changed to returning a string.

lib/packages/docutils/rst.nim
a-mr: I'm not sure. Let @narimiran decide.

I moved to the table-based implementation only because it seemed more compact and clear than previous one.

Also note that though there are only 20 directives currently, but there are more of them in the RST spec.

timotheecour:
is this feasible? 
image

from https://bluescarni.github.io/mppp/

lib/packages/docutils/rst.nim
planetis-m: ~I only count ten~ but I will make a bench and return.

planetis-m: So yes its faster/uses less memory https://gist.github.com/planetis-m/fa12f6c4ff528babf9f15f3d189038de Its also as clear as the dispatch table approach but doesn't use function pointers.

planetis-m: ...and you dont need all those workarounds because you need a uniform function signature

a-mr: yes, but there is no need in enums then. A large case statement on strings works even faster https://gist.github.com/a-mr/fe3596a928df860503d7f946026d50a3

a-mr: thank you @planetis-m ,
moved to a simple case statement.

a-mr:
> is this feasible?

Yeah, but it's a bit too bulky and obtrusive.

Also I like FedericoCeratto's design because it corresponds closely to its RST text. So it may be easier to memorize the syntax.

Using icons may be a good idea for Latex backend, which is more probable to be used in black-and-white printing.

PR #16450, opened 2020-12-23
math: add div overload for float

ringabout:
ref RFC https://github.com/timotheecour/Nim/issues/396

### DIFF
--- a/lib/pure/math.nim
+++ b/lib/pure/math.nim
@@ -845,6 +845,18 @@ func round*[T: float32|float64](x: T, places: int): T =
     var mult = pow(10.0, places.T)
     result = round(x*mult)/mult
 
+func `div`*[T: float32|float64](x, y: T): T =
+  ## Computes the truncated division operation for float values.
+  runnableExamples:
+    assert 0.5 div 0.1 == 4.0
+    assert 0.4 div 0.1 == 4.0
+    assert 0.1 div 0.5 == 0.0
+    assert almostEqual(12.5 div 3.7, 3.0)
+    assert almostEqual(-19.6 div 5.2, -3.0)
+    assert almostEqual(19.6 div -5.2, -3.0)
+    assert almostEqual(-19.6 div -5.2, 3.0)
+  result = (x - x mod y) / y
+
 func floorDiv*[T: SomeInteger](x, y: T): T =
   ## Floor division is conceptually defined as ``floor(x / y)``.
   ##
@@ -864,18 +876,6 @@ func floorDiv*[T: SomeInteger](x, y: T): T =
   result = x div y
   let r = x mod y
   if (r > 0 and y < 0) or (r < 0 and y > 0): result.dec 1
-
-func `div`*[T: float32|float64](x, y: T): T =
-  ## Computes the division operation for float values.
-  runnableExamples:
-    assert 0.5 div 0.1 == 4.0
-    assert 0.4 div 0.1 == 4.0
-    assert 0.1 div 0.5 == 0.0
-    assert almostEqual(12.5 div 3.7, 3.0)
-    assert almostEqual(-19.6 div 5.2, -3.0)
-    assert almostEqual(19.6 div -5.2, -3.0)
-    assert almostEqual(-19.6 div -5.2, 3.0)
-  result = (x - x mod y) / y
 
 func floorMod*[T: SomeNumber](x, y: T): T =
   ## Floor modulus is conceptually defined as ``x - (floorDiv(x, y) * y)``.
--- a/lib/pure/math.nim
+++ b/lib/pure/math.nim
@@ -865,6 +865,18 @@ func floorDiv*[T: SomeInteger](x, y: T): T =
   let r = x mod y
   if (r > 0 and y < 0) or (r < 0 and y > 0): result.dec 1
 
+func `div`*[T: float32|float64](x, y: T): T =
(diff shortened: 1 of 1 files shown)


lib/pure/math.nim
```diff
@@ -865,6 +865,18 @@ func floorDiv*[T: SomeInteger](x, y: T): T =
   let r = x mod y
   if (r > 0 and y < 0) or (r < 0 and y > 0): result.dec 1
 
+func `div`*[T: float32|float64](x, y: T): T =
+  ## Computes the division operation for float values.
```
Araq: Which "division"? ``floor(x / y)``?

ringabout: trunc(default) division just like `mod(x, y: float)`.

>   func `mod`*(x, y: float32): float32 {.importc: "fmodf", header: "<math.h>".}
>  func `mod`*(x, y: float64): float64 {.importc: "fmod", header: "<math.h>".}
    ## Computes the modulo operation for float values (the remainder of ``x`` divided by ``y``).

ringabout: Ah, I put it at the wrong place.

ringabout: The implementation for Python floor(float) div
https://github.com/python/cpython/blob/52a327c1cbb86c7f2f5c460645889b23615261bf/Objects/floatobject.c#L636

I verified floor div for integer between Python and Nim, they are the same.
such as `floorDiv(5.6, 2.5)`
And `floorDiv` is based on `div`.

lib/pure/math.nim
```diff
@@ -845,6 +845,18 @@ func round*[T: float32|float64](x: T, places: int): T =
     var mult = pow(10.0, places.T)
     result = round(x*mult)/mult
 
+func `div`*[T: float32|float64](x, y: T): T =
```
timotheecour, line 848: `T: SomeFloat`

and in future PR, we should also use `SomeFloat` instead of `float32|float64` in this module

ringabout: Does it need since?
pre-existing `floorDiv` needs this one.

timotheecour: This is a recurring problem for added overloads and cases where we add optional parameters. 

The solution is https://github.com/timotheecour/Nim/issues/479 (see `exportSince` + linked PR which allows picking a particular overload symbol) but until then IMO the simplest here is to omit {.since.}. In the unlikely case this cause a regression (or one that's unacceptable) we can either implement https://github.com/timotheecour/Nim/issues/479 or add a one-off `-d:nimLegacyDisableDivFloat`

lib/pure/math.nim
```diff
@@ -845,6 +845,18 @@ func round*[T: float32|float64](x: T, places: int): T =
...
+    assert 0.5 div 0.1 == 4.0
+    assert 0.4 div 0.1 == 4.0
+    assert 0.1 div 0.5 == 0.0
+    assert almostEqual(12.5 div 3.7, 3.0)
+    assert almostEqual(-19.6 div 5.2, -3.0)
+    assert almostEqual(19.6 div -5.2, -3.0)
+    assert almostEqual(-19.6 div -5.2, 3.0)
+  result = (x - x mod y) / y
+
 func floorDiv*[T: SomeInteger](x, y: T): T =
```
timotheecour, line 860: now this can be `SomeFloat` too
(and needs 1 line that reflects that in the runnableExamples block)

timotheecour:
in following case at least, div1 provides a better answer than this PR's div2, returning 99999999.0 instead of 99999999.00000001 
```nim
import std/math

func div1[T: float32|float64](x, y: T): T = trunc(x / y)
func div2[T: float32|float64](x, y: T): T = (x - x mod y) / y

proc fn(x, y: float) =
  let a = div1(x, y)
  let b = div2(x, y)
  echo (x, y, a, b, x / y, x mod y)
  doAssert a == b, $(x, y, a, b, a - b)

proc main() =
  fn(1001.0, 1.0) # ok
  fn(10000000000000001.0, 100000000.21)
    #  `a == b` (1e+16, 100000000.21, 99999999.0, 99999999.00000001, -1.490116119384766e-08) [AssertionDefect]
main()
```

ringabout:
```nim
import std/math

func div1[T: float32|float64](x, y: T): T = trunc(x / y)
func div2[T: float32|float64](x, y: T): T = round((x - x mod y) / y)

proc fn(x, y: float) =
  let a = div1(x, y)
  let b = div2(x, y)
  doAssert a == b, $(x, y, a, b, a - b)

proc main() =
  fn(0.5, 0.1)
```
It seems that `trunc` is not correct.

nim r --eval:"import math; echo 0.5 mod 0.1" 
0.09999999999999998
And according to pre-existing mod for float, reminder is 0.09999999999999998 not zero.
So 0.4 is consistent and correct.

lib/pure/math.nim
ringabout: will do

timotheecour:
* this https://stackoverflow.com/a/39966752/1426932 sheds a lot of light on the weird value of `fmod(1.0, 0.1)` not being consistent with `trunc(1.0 / 0.1)`

* see also description of math.fmod(x, y) vs % in python in https://python.readthedocs.io/en/latest/library/math.html

Araq:
Too risky to add. Few people need it and the danger that it'll become another patchfest of subtle edge cases is huge.

PR #16454, opened 2020-12-23
Add missing HTTP codes

juancarlospaco:
- Add missing HTTP Codes to `httpcore`.


Araq:
Yes, I know these have no ``.since`` annotation. I think in practice these don't need it.

timotheecour (approved):

Issue #556, opened 2013-07-31
Template `` and do


DAddYE:
I've a template like:

``` nimrod
template `->` *(fun: proc()) =
   # ...
```

Calling it with:

``` nimrod
->:
  # ...
```

or

``` nimrod
-> do:
  # ...
```

Result in a compiler error.

PR #16459, opened 2020-12-24
add system random to stdlib: std/sysrand

ringabout:
Cryptographically secure pseudorandom number generator

https://en.wikipedia.org/wiki/Cryptographically_secure_pseudorandom_number_generator

# Many popular and mature programming languages put `system random` on **their standard library**

- Python
https://www.python.org/dev/peps/pep-0506
https://www.python.org/dev/peps/pep-0524
https://docs.python.org/3/library/os.html#os.urandom
https://docs.python.org/3/library/secrets.html

- Rust
https://github.com/rust-lang/rust/blob/master/library/std/src/sys/unix/rand.rs

- Go
https://golang.org/pkg/crypto/rand
https://sourcegraph.com/github.com/golang/go/-/blob/src/internal/syscall/unix/getrandom_linux.go

- JS(WebAPI)
https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues

- Nodejs
https://nodejs.org/api/crypto.html#crypto_crypto_randomfillsync_buffer_offset_size

- Crystal
https://github.com/crystal-lang/crystal/blob/master/src/random/secure.cr

- Zig
https://github.com/ziglang/zig/blob/master/lib/std/crypto/tlcsprng.zig

- Java/Scala/Kotlin
https://docs.oracle.com/javase/7/docs/api/java/security/SecureRandom.html

- C#
https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.rngcryptoserviceprovider?view=net-5.0

- PHP 7
random_bytes — Generates cryptographically secure pseudo-random bytes
https://www.php.net/manual/en/function.random-bytes.php

- Ruby
https://ruby-doc.org/stdlib-2.5.1/libdoc/securerandom/rdoc/SecureRandom.html

- Dart
https://api.dart.dev/stable/2.10.5/dart-math/Random/Random.secure.html

- Erlang
https://erlang.org/doc/man/crypto.html#strong_rand_bytes-1

-------------------------------------------------------------------------------------------------------------------

libsodium
https://github.com/jedisct1/libsodium/blob/master/src/libsodium/randombytes/sysrandom/randombytes_sysrandom.c

blog:
https://paragonie.com/blog/2016/05/how-generate-secure-random-numbers-in-various-programming-languages

implementation:
image

## future work
- [ ] refs https://github.com/nim-lang/Nim/pull/16459#issuecomment-767523249
> It need not hold up this PR being merged, but if this is merge-worthy then I think it is probably better to use this instead of a high resolution clock in lib/pure/random.nim:randomize.

`oids` can also benefits from this PR(use a secure seed).

If stdlib can accept `UUID`, the module will benefit from this PR too.

It can also be used to generate salts, keys or initialization vectors.


tests/magics/t10307.nim
```diff
@@ -4,7 +4,7 @@ discard """
 res: yes
 yes
 running someProc(false)
-res: 
+res:
```
ringabout:
```suggestion
res: 
```

lib/std/secrets.nim
```diff
@@ -0,0 +1,98 @@
...
+  proc urandom*(size: int): string =
+    result = newString(size)
+    if urandom(result.cstring, size) < 0:
+      raiseOsError(osLastError())
+
+else:
+  proc urandom*(p: pointer, size: int): int =
+    result = getDevUrandom(p, size)
+
+  proc urandom*(size: int): string =
```
juancarlospaco:
```suggestion
  proc urandom*(size: Positive): string =
```

Same for all others proc that take the `int` argument. Or `Natural`.

lib/std/secrets.nim
```diff
@@ -0,0 +1,98 @@
...
+  doAssert urandom(0).len == 0
+  doAssert urandom(10).len == 10
+  doAssert urandom(20).len == 20
+  doAssert urandom(120).len == 120
+
+
+when defined(posix):
+  import posix
+
+  template processReadBytes(readBytes: int, p: pointer) =
```
timotheecour: in almost all procs here, instead of a pointer param, why not use a `openArray`; either
generic `openArray[T]` or `openArray[char]` or `openArray[byte]`; not sure yet.

(UncheckedArray also comes to mind, but I don't think it's needed here, you know the size of the openArray)

(Note, you can convert a string to openArray[char], so no problem here)

lib/std/secrets.nim
```diff
@@ -0,0 +1,98 @@
...
+    if readBytes == 0:
+      break
+    elif readBytes > 0:
+      inc(result, readBytes)
+      cast[ptr pointer](p)[] = cast[pointer](cast[ByteAddress](p) + readBytes)
+    else:
```
timotheecour: elif

lib/std/secrets.nim
```diff
@@ -0,0 +1,98 @@
...
+    if fd > 0:
+      var stat: Stat
+      if fstat(fd, stat) != -1 and S_ISCHR(stat.st_mode):
+        while result < size:
+          let readBytes = posix.read(fd, p, size - result)
+          processReadBytes(readBytes, p)
+
+      discard posix.close(fd)
+
+when defined(windows):
```
timotheecour: elif defined windows?

timotheecour: nevermind; doesn't work bc of `elif defined(linux):`

lib/std/secrets.nim
```diff
@@ -0,0 +1,98 @@
+import os
```
timotheecour: only because of osLastError ? maybe we can refactor to avoid this large dependency?
(eg: import/re-export from the usual `std/private/...`)

lib/std/secrets.nim
```diff
@@ -0,0 +1,98 @@
...
+      cast[ptr pointer](p)[] = cast[pointer](cast[ByteAddress](p) + readBytes)
+    else:
+      if osLastError().int in {EINTR, EAGAIN}:
+        discard
+      else:
+        result = -1
+        break
+
+  proc getDevUrandom(p: pointer, size: int): int =
+    let fd = posix.open("/dev/urandom", O_RDONLY)
```
Araq: If available, use an OS API instead. See https://insanecoding.blogspot.com/2014/05/a-good-idea-with-bad-usage-devurandom.html for more information.

ringabout: Yeah, It's convenient for me to test specific functions in Windows and Linux VM. But I can't test a function specific to openbsd/freebsd and OSX. So I haven't worked on them because I don't want to use CI to debug.

timotheecour: @xflywind there's https://www.macincloud.com/ for OSX (I've used it occasionally even though i have osx, when i needed to test other OSX versions); i can help also if needed for osx

for openbsd/freebsd, both should also have urandom, it should work

xflywind:
<p class="module-desc"><table border="1" class="docutils"><tr><th>Targets</th><th>Implementation</th></tr>
--
  | <tr><td>Windows</td><td><a class="reference external" href="https://docs.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptgenrandom">BCryptGenRandom</a></td></tr>
  | <tr><td>Linux</td><td><a class="reference external" href="https://man7.org/linux/man-pages/man2/getrandom.2.html">getrandom</a> system call when available, otherwise <a class="reference external" href="https://en.wikipedia.org/wiki//dev/random">/dev/urandom</a> will be used</td></tr>
  | <tr><td>Macosx</td><td><a class="reference external" href="https://developer.apple.com/documentation/security/1399291-secrandomcopybytes?language=objc">SecRandomCopyBytes</a> system call when available, otherwise <a class="reference external" href="https://en.wikipedia.org/wiki//dev/random">/dev/urandom</a> will be used</td></tr>
  <tr><td>OpenBSD</td><td><a class="reference external" href="https://man.openbsd.org/getentropy.2">getentropy</a> system call when available, otherwise <a class="reference external" href="https://en.wikipedia.org/wiki//dev/random">/dev/urandom</a> will be used</td></tr>
  <tr><td>JS(Web Browser)</td><td><a class="reference external" href="https://www.w3.org/TR/WebCryptoAPI/#Crypto-method-getRandomValues">getRandomValues</a></td></tr>
  | <tr><td>Other platforms</td><td><a class="reference external" href="https://en.wikipedia.org/wiki//dev/random">/dev/urandom</a></td></tr>
  | </table></p>

lib/std/secrets.nim
```diff
@@ -0,0 +1,186 @@
...
+
+  proc getRandomValues(arr: Uint8Array) {.importjs: "window.crypto.getRandomValues(#)".}
+
+  const
+    maxBrowserCryptoBufferSize = 256
+
+  proc urandom*(p: var openArray[byte]): int =
+    let size = p.len
+    if size > 0:
+      let chunks = (size - 1) div maxBrowserCryptoBufferSize
```
ringabout: I also find `chunks` apis are useful:
https://doc.rust-lang.org/std/slice/struct.Chunks.html
https://doc.rust-lang.org/std/slice/struct.ChunksMut.html

lib/std/secrets.nim
planetis-m: there are nimble packages that provide these procs. See nim-crypto

planetis-m: https://github.com/cheatfate/nimcrypto/blob/master/nimcrypto/sysrand.nim

lib/std/secrets.nim
```diff
@@ -0,0 +1,215 @@
...
+## .. _getRandomValues: https://www.w3.org/TR/WebCryptoAPI/#Crypto-method-getRandomValues
+## 
+
+import std/os
+
+runnableExamples:
+  doAssert urandom(0).len == 0
+  doAssert urandom(10).len == 10
+  doAssert urandom(20).len == 20
+  doAssert urandom(120).len == 120
```
timotheecour:
```suggestion
runnableExamples:
  doAssert urandom(0).len == 0
  doAssert urandom(113).len == 113
  doAssert urandom(1234) != urandom(1234) # unlikely to fail in practice
```

lib/std/secrets.nim
```diff
@@ -4,17 +4,19 @@
 ## | :---         | ----:       |
 ## | Windows| `BCryptGenRandom`_ |
 ## | Linux| `getrandom`_ system call when available, otherwise `/dev/urandom`_ will be used|
-## | MacOS| `SecRandomCopyBytes`_ system call when available, otherwise `/dev/urandom`_ will be used|
-## | OpenBSD| `getentropy`_ system call when available, otherwise `/dev/urandom`_ will be used|
+## | MacOS| `getentropy`_ system call when available, otherwise `/dev/urandom`_ will be used|
```
timotheecour:
```suggestion
## | MacOSX| `getentropy`_ system call when available, otherwise `/dev/urandom`_ will be used|
```

lib/std/secrets.nim
```diff
@@ -153,11 +156,10 @@ elif defined(openbsd):
     let size = p.len
     if size > 0:
       result = randomBytes(addr p[0], size)
-      if result < 0:
-        result = getDevUrandom(p, size)
 
 elif defined(freebsd):
-  proc getrandom(p: pointer, size: csize_t, flags: cuint): int {.importc: "getrandom", header: "<unistd.h>".}
+  type ssize_t = int
```
timotheecour: `type ssize_t {.importc, header: "<sys/types.h>".}`

lib/std/secrets.nim
timotheecour: this would be easy with std/views (from https://github.com/nim-lang/Nim/pull/14869)

lib/std/secrets.nim
```diff
@@ -0,0 +1,226 @@
...
+elif defined(windows):
+  proc urandom*(size: Natural): seq[byte] =
+    result = newSeq[byte](size)
+    if urandom(result) != STATUS_SUCCESS:
+      raiseOsError(osLastError())
+else:
+  proc urandom*(size: Natural): seq[byte] =
+    result = newSeq[byte](size)
+    if urandom(result) < 0:
+      raiseOsError(osLastError())
```
timotheecour:
```suggestion
proc urandom*(size: Natural): seq[byte] =
  ## some doc comment
  result = newSeq[byte](size)
  let ret = urandom(result)
  when defined(js): discard ret
  elif defined(windows):
    if ret != STATUS_SUCCESS:
      raiseOsError(osLastError())
  else:
    if ret < 0:
      raiseOsError(osLastError())
```suggestion

File: lib/std/secrets.nim [hunk 13]
```diff
@@ -0,0 +1,226 @@
...
+        inc(base, maxBrowserCryptoBufferSize)
+
+      let left = size - chunks * maxBrowserCryptoBufferSize
+      var src = newUint8Array(left)
+      getRandomValues(src)
+      for i in 0 ..< left:
+        p[base + i] = src[i]
+
+elif defined(windows):
+  type
```
timotheecour: optional, could be done in future work: `std/private/wintypes.nim` to avoid each module re-defining those? (later can be refactored to avoid duplication with winlean, io.nim and others)

lib/std/secrets.nim
```diff
@@ -0,0 +1,226 @@
...
+
+  proc urandom*(p: var openArray[byte]): int =
+    let size = p.len
+    if size > 0:
+      result = randomBytes(addr p[0], size)
+
+elif defined(linux):
+  let SYS_getrandom {.importc: "SYS_getrandom", header: "<syscall.h>".}: clong
+
+  proc syscall(n: clong, buf: pointer, bufLen: cint, flags: cuint): int {.importc: "syscall", header: "syscall.h".}
```
timotheecour:
```suggestion
  proc syscall(n: clong, buf: pointer, bufLen: cint, flags: cuint): clong {.importc: "syscall", header: "<sys/syscall.h>".}
```
(source: https://man7.org/linux/man-pages/man2/syscall.2.html)

lib/std/secrets.nim
```diff
@@ -0,0 +1,226 @@
...
+    bCryptGenRandom(nil, cast[PUCHAR](pbBuffer), ULONG(cbBuffer),
+                            BCRYPT_USE_SYSTEM_PREFERRED_RNG)
+
+  proc urandom*(p: var openArray[byte]): int =
+    let size = p.len
+    if size > 0:
+      result = randomBytes(addr p[0], size)
+
+elif defined(linux):
+  let SYS_getrandom {.importc: "SYS_getrandom", header: "<syscall.h>".}: clong
```
timotheecour:
```suggestion
  let SYS_getrandom {.importc: "SYS_getrandom", header: "<sys/syscall.h>".}: clong
```

?

lib/std/secrets.nim
```diff
@@ -0,0 +1,226 @@
...
+
+  proc syscall(n: clong, buf: pointer, bufLen: cint, flags: cuint): int {.importc: "syscall", header: "syscall.h".}
+
```
timotheecour: why not https://man7.org/linux/man-pages/man2/getrandom.2.html ? 

> getrandom() was introduced in version 3.17 of the Linux kernel.
> Support was added to glibc in version 2.25.

ringabout: use `sys/syscall.h` is more safer. Because glibc 2.25 is a bit new

ringabout: Better solution is to check whether `getrandom` is available, I will do it later.

ringabout: See what Python does:
https://github.com/python/cpython/blob/15bd9efd01e44087664e78bf766865a6d2e06626/Python/bootstrap_hash.c#L13

lib/std/secrets.nim
```diff
@@ -4,7 +4,7 @@
 ## | :---         | ----:       |
 ## | Windows| `BCryptGenRandom`_ |
 ## | Linux| `getrandom`_ system call when available, otherwise `/dev/urandom`_ will be used|
-## | MacOS| `getentropy`_ system call when available, otherwise `/dev/urandom`_ will be used|
+## | MacOSX| `getentropy`_ system call when available, otherwise `/dev/urandom`_ will be used|
```
timotheecour: > `otherwise `/dev/urandom`_ will be used|`  ?

where in the code does it do this fallback for osx? (and is that fallback needed there?)

lib/std/secrets.nim
```diff
@@ -0,0 +1,220 @@
+## Cryptographically secure pseudorandom number generator.
```
timotheecour: * cross reference needed from std/random
* changelog entry
* entry in doc/lib.rst (https://nim-lang.github.io/Nim/lib.html#pure-libraries-math-libraries)

lib/std/secrets.nim
ringabout: It haven't been implemented. I need to figure out a good solution(Rust is more mature than Nim in low level primitives).

lib/std/secrets.nim
```diff
@@ -0,0 +1,220 @@
...
+  proc urandom*(p: var openArray[byte]): int =
+    let size = p.len
+    if size > 0:
+      result = randomBytes(addr p[0], size)
+else:
+  proc urandom*(p: var openArray[byte]): int =
+    let size = p.len
+    if size > 0:
+      result = getDevUrandom(p, size)
+
```
timotheecour: change all `proc urandom*(p: var openArray[byte]): int =` procs to `urandomImpl` and then introduce:
```nim
proc urandom*(p: var openArray[byte]): int {.inline.} =
  ## doc comment... in particular explaning return value
  result = urandomImpl(p)
```
so that it can have a doc comment (`when defined nimdoc` is possible too but this is simpler IMO)

lib/std/secrets.nim
ringabout: Does Nim provide some convenient methods to check whether a C flags is defined or not>
https://android.googlesource.com/platform/external/expat/+/refs/heads/master/expat_config.h#31

lib/std/sysrand.nim
```diff
@@ -0,0 +1,258 @@
...
+## See also
+## ========
+## * `random module <random.html>`_
+## 
+
+import std/os
+
+
+
+when defined(posix):
```
Araq, line 53: This has to come **after** ``defined(linux)`` etc!

lib/std/secrets.nim
timotheecour: proposal 5 from https://github.com/nim-lang/RFCs/issues/205 would allow doing this in a robust, clean way (it's not even that hard to implement)
until then we need hacks similar to `NIM_EmulateOverflowChecks`, or move the `defined` logic to the C side (eg in nimbase.h which is ugly but works)

lib/std/sysrand.nim
```diff
@@ -146,29 +165,40 @@ elif defined(linux):
...
-    while result < size:
-      let readBytes = getrandom(p, csize_t(size - result), 0)
-      processReadBytes(readBytes, p, result)
+  proc getrandom(p: pointer, size: csize_t, flags: cuint): cssize_t {.importc: "getrandom", header: "<sys/random.h>".}
 
   proc urandom*(p: var openArray[byte]): int =
     let size = p.len
     if size > 0:
-      result = randomBytes(addr p[0], size)
+      let
```
timotheecour: optional: can you abstract this pattern into a template? it's used twice (and in future work it could be exposed as an API `byChunks`)

lib/std/sysrand.nim
ringabout: I don't understand.
The planned implementation is that first try `getrandom` in linux then fall back to read `/dev/urandom`.

timotheecour: ah ok, the code is like this:
```nim
when defined(posix): ...

when defined(js):
elif defined(windows):
elif defined(linux):
elif ...
```
which seems fine

c-blake:
It need not hold up this PR being merged, but if this is merge-worthy then I think it is probably better to use this instead of a high resolution clock in `lib/pure/random.nim:randomize`.

(EDIT: added as a TODO for future work in PR description)

ringabout:
TODO:
- [x] cross reference needed from std/random
- [x] changelog entry
- [x] entry in doc/lib.rst (https://nim-lang.github.io/Nim/lib.html#pure-libraries-math-libraries)
- [x] docs

- [ ] `std/sysrand` is very low level interface, we could provide a high level interface in the following PR Ref https://docs.python.org/3/library/secrets.html

lib/std/sysrand.nim
```diff
@@ -0,0 +1,287 @@
...
+
+        result = posix.read(fd, addr dest[base], left)
+      discard posix.close(fd)
+
+proc urandom*(dest: var openArray[byte]): int =
+  ## Fills `dest` with random bytes suitable for cryptographic use.
+  ## 
+  ## If `dest` is empty, `urandom` immediately returns success,
+  ## without calling underlying operating system api.
+  urandomImpl(dest)
```
timotheecour: instead of handling `dest.len == 0` in each specialization use this:
```suggestion
  if dest.len == 0: result = 0
  else: urandomImpl(dest)
```
and each specialization can implicitly assume dest.len > 0

lib/std/sysrand.nim
```diff
@@ -0,0 +1,287 @@
...
+    result = randomBytes(addr dest[0], size)
+
+elif defined(linux):
+  let SYS_getrandom {.importc: "SYS_getrandom", header: "<sys/syscall.h>".}: clong
+  const syscallHeader = """#include <unistd.h>
+#include <sys/syscall.h>"""
+
+  proc syscall(
+    n: clong, buf: pointer, bufLen: cint, flags: cuint
+  ): int {.importc: "syscall", header: syscallHeader.}
```
timotheecour:
```suggestion
  ): clong {.importc: "syscall", header: syscallHeader.}
```
according to https://man7.org/linux/man-pages/man2/syscall.2.html

(same comment as https://github.com/nim-lang/Nim/pull/16459#pullrequestreview-575088663)

lib/std/sysrand.nim
```diff
@@ -0,0 +1,287 @@
...
+    result = getentropy(p, cint(size)).int
+
+  template urandomImpl(dest: var openArray[byte]) =
+    batchImpl(result, dest, getRandomImpl)
+
+elif defined(freebsd):
+  type cssize_t {.importc: "ssize_t", header: "<sys/types.h>".} = int
+
+  proc getrandom(p: pointer, size: csize_t, flags: cuint): cssize_t {.importc: "getrandom", header: "<sys/random.h>".}
+    # Upon successful completion, the number of bytes which were	actually read
```
timotheecour:
```suggestion
    # Upon successful completion, the number of bytes which were actually read
```
(there was a weird tab char)

lib/std/sysrand.nim
```diff
@@ -0,0 +1,287 @@
...
+    # is returned. For requests larger than 256	bytes, this can	be fewer bytes
```
timotheecour: what character is that? check your editor to make sure it doesn't return weird characters.
image

lib/std/sysrand.nim
```diff
@@ -0,0 +1,287 @@
...
+  when not defined(linux):
+    const batchSize = 256
+
+    template batchImpl(result: var int, dest: var openArray[byte], getRandomImpl) =
+      let size = dest.len
+      if size == 0:
+        return
+
+      let
+        chunks = (size - 1) div batchSize
```
timotheecour:
```suggestion
        chunks = size div batchSize
```

ditto in other similar instances; 
and change:
`result = getRandomImpl(addr dest[base], left)`
to
```nim
if left > 0:
  result = getRandomImpl(addr dest[base], left)
```

lib/std/sysrand.nim
```diff
@@ -0,0 +1,287 @@
...
+      let size = dest.len
+      if size == 0:
+        return
+
+      let chunks = (size - 1) div batchSize
+      for i in 0 ..< chunks:
+        for j in 0 ..< batchSize:
+          var src = newUint8Array(batchSize)
+          getRandomValues(src)
+          dest[result + j] = src[j]
```
timotheecour: I don't understand, this seems buggy.

lib/std/sysrand.nim
```diff
@@ -0,0 +1,287 @@
...
+        for i in 0 ..< chunks:
+          let readBytes = posix.read(fd, addr dest[base], batchSize)
+          if readBytes < 0:
+            return readBytes
+          inc(base, batchSize)
+
+        result = posix.read(fd, addr dest[base], left)
+      discard posix.close(fd)
+
+proc urandom*(dest: var openArray[byte]): int =
```
timotheecour:
```suggestion
proc urandom*(dest: var openArray[byte]): bool =
```

and change the rest of the code accordingly to return true on success, false on error.
The result is otherwise un-portable and meaningless, for several reasons.

(`proc urandom*(size: Natural): seq[byte] {.inline.} =` also needs to be adapted accordingly)

lib/std/sysrand.nim
```diff
@@ -0,0 +1,287 @@
...
+      if fstat(fd, stat) != -1 and S_ISCHR(stat.st_mode):
+        let
+          chunks = (size - 1) div batchSize
+          left = size - chunks * batchSize
+
+        var base = 0
+        for i in 0 ..< chunks:
+          let readBytes = posix.read(fd, addr dest[base], batchSize)
+          if readBytes < 0:
+            return readBytes
```
timotheecour, line 270: that's a bug, it leaks fd.

in this case IMO `defer` is justified, and avoids making the code uglier than needed:
```nim
let fd = posix.open("/dev/urandom", O_RDONLY)
defer: discard posix.close(fd)
```
alternatively, use try/finally

lib/std/sysrand.nim
```diff
@@ -0,0 +1,287 @@
...
+          inc(base, batchSize)
+
+        result = posix.read(fd, addr dest[base], left)
+      discard posix.close(fd)
```
timotheecour: let's re-evaluate this after all other comments are addressed, but I wonder if it's possible to write this in a correct way without raising exceptions. One problem is that `errno` after an un-successful posix.read for example (errno read by `raiseOsError(osLastError()))` may be over-written by subsequent OS calls such as posix.close, which would then mask the original errno.

lib/std/sysrand.nim
```diff
@@ -0,0 +1,287 @@
...
+
+  template urandomImpl(dest: var openArray[byte]) =
+    batchImpl(result, dest, getRandomImpl)
+else:
+  template urandomImpl(dest: var openArray[byte]) =
+    let size = dest.len
+    if size == 0:
+      return
+
+    let fd = posix.open("/dev/urandom", O_RDONLY)
```
timotheecour, line 257:
```suggestion
    let fd = posix.open("/dev/urandom", O_RDONLY)
    # see: https://www.2uo.de/myths-about-urandom/ which justifies using urandom instead of random
```

lib/std/sysrand.nim
```diff
@@ -0,0 +1,287 @@
...
+
+  template urandomImpl(dest: var openArray[byte]) =
+    let size = dest.len
+    if size == 0:
+      return
+
+    result = randomBytes(addr dest[0], size)
+
+elif defined(linux):
+  let SYS_getrandom {.importc: "SYS_getrandom", header: "<sys/syscall.h>".}: clong
```
timotheecour, line 161: on linux, you should use `getrandom` when system has this;
refs:

* https://github.com/jedisct1/libsodium/blob/e4206f13370d0deda3bcfac2085a24911fdf4007/src/libsodium/randombytes/internal/randombytes_internal_random.c#L26 (cf link you mentioned)

* [  TITLE: conditional C declarations · Issue #556 · timotheecour/Nim](https://github.com/timotheecour/Nim/issues/556) (shows how to do this on the nim side)

ringabout: A bit ugly, I won't use it in this PR. `syscall` is elegant enough to satisfy most of our needs.

lib/std/sysrand.nim
ringabout: I like the first way.

lib/std/sysrand.nim
xflywind: `sysrand` is very low level, so I think returning `int` is enough.

lib/std/sysrand.nim
ringabout: It should be same as the `batchImpl`

lib/std/sysrand.nim
```diff
@@ -91,13 +91,16 @@ when defined(js):
     const batchSize = 256
 
     proc getRandomValues(p: Uint8Array) {.importjs: "window.crypto.getRandomValues(#)".}
+      # The requested length of `p` must not be more than 65536.
```
timotheecour, line 100:
```suggestion
      doAssert p.len <= 65536
```
better check than assume

ringabout: I mean `getRandomValues` has this restriction. But I use a batch impl internal, so it doesn't hurt.

timotheecour: makes sense. should we have `const batchSize = 256` or `const batchSize = 65536` for js?

ringabout: I'm not sure.

timotheecour:
closed/reopened to get https://github.com/nim-lang/Nim/pull/16842 in

ringabout:
Almost ready to review, please take a look.

lib/std/sysrand.nim
```diff
@@ -0,0 +1,304 @@
+#
+#
+#            Nim's Runtime Library
+#        (c) Copyright 2021 Nim contributors
+#
+#    See the file "copying.txt", included in this
+#    distribution, for details about the copyright.
+#
+
+## `std/sysrand` generates random numbers from a secure source provided by the operating system.
```
timotheecour, line 10: the acronym CSPRNG should also be mentioned somewhere in this file for searchability

lib/std/sysrand.nim
```diff
@@ -0,0 +1,304 @@
...
+
+  template urandomImpl(result: var int, dest: var openArray[byte]) =
+    let size = dest.len
+    if size == 0:
+      return
+
+    while result < size:
+      let readBytes = syscall(SYS_getrandom, addr dest[result], cint(size - result), 0).int
+      if readBytes == 0:
+        break
```
timotheecour: seems like a bug; you're breaking but result < size; as a result, the urandom call will succeed even though the buffer wasn't filled

lib/std/sysrand.nim
```diff
@@ -0,0 +1,304 @@
...
+  type
+    SecRandom {.importc: "struct __SecRandom".} = object
+
+    SecRandomRef = ptr SecRandom
+      ## An abstract Core Foundation-type object containing information about a random number generator.
+
+  proc secRandomCopyBytes(
+    rnd: SecRandomRef, count: csize_t, bytes: pointer
+    ): cint {.importc: "SecRandomCopyBytes", header: "<Security/SecRandom.h>".}
+    ## Generates an array of cryptographically secure random bytes.
```
timotheecour:
```suggestion
    ## https://developer.apple.com/documentation/security/1399291-secrandomcopybytes
```

a link is worth a thousand words

(ditto elsewhere but can be fixed later)

lib/std/sysrand.nim
```diff
@@ -0,0 +1,304 @@
...
+
+  template urandomImpl(result: var int, dest: var openArray[byte]) =
+    let size = dest.len
+    if size == 0:
+      return
+
+    # `kSecRandomDefault` is a synonym for NULL.
```
timotheecour:
```suggestion
```
kSecRandomDefault isn't referenced here and users who click on above link i suggested get all the context for nil

lib/std/sysrand.nim
```diff
@@ -0,0 +1,304 @@
...
+  template urandomImpl(result: var int, dest: var openArray[byte]) =
+    let size = dest.len
+    if size == 0:
+      return
+
+    # see: https://www.2uo.de/myths-about-urandom/ which justifies using urandom instead of random
+    let fd = posix.open("/dev/urandom", O_RDONLY)
+    defer: discard posix.close(fd)
+
+    if fd > 0:
```
timotheecour, line 260: what if fd <= 0? shouldn't we raise?

lib/std/sysrand.nim
```diff
@@ -0,0 +1,304 @@
...
+          chunks = (size - 1) div batchSize
+          left = size - chunks * batchSize
+
+        for i in 0 ..< chunks:
+          let readBytes = posix.read(fd, addr dest[result], batchSize)
+          if readBytes < 0:
+            return readBytes
+          inc(result, batchSize)
+
+        result = posix.read(fd, addr dest[result], left)
```
timotheecour, line 273: seems buggy, in those cases:
* if you have a short read (result < left)
* or if result == -1 and errno is restartable (eg EINTR)

lib/std/sysrand.nim
```diff
@@ -0,0 +1,304 @@
...
+    if size == 0:
+      return
+
+    # see: https://www.2uo.de/myths-about-urandom/ which justifies using urandom instead of random
+    let fd = posix.open("/dev/urandom", O_RDONLY)
+    defer: discard posix.close(fd)
+
+    if fd > 0:
+      var stat: Stat
+      if fstat(fd, stat) != -1 and S_ISCHR(stat.st_mode):
```
timotheecour, line 262: bug: if this branch is not taken, result would be 0, indicating success

timotheecour (approved):
LGTM; @araq how about we merge this now, and we can do followup PR's to fix the remaining bugs; since it's a new module it's fine IMO to do this so we can make faster progress (maybe with a note: experimental module, unstable API)

lib/std/sysrand.nim
ringabout: `getrandom` in Linux is blocking on default. It won't succeed unless enough bytes are returned. If no random bytes are available, it blocks; if the entropy pool has not yet been initialized, it blocks.

>  By default, when reading from the random source,
  getrandom() blocks if no random bytes are available, and
  when reading from the urandom source, it blocks if the
  entropy pool has not yet been initialized.

https://man7.org/linux/man-pages/man2/getrandom.2.html

Clyybber:
> LGTM; @Araq how about we merge this now, and we can do followup PR's to fix the remaining bugs; since it's a new module it's fine IMO to do this so we can make faster progress (maybe with a note: experimental module, unstable API)

We cannot merge something with potential security issues/bugs like this, especially when it should cryptographically secure. Let's fix the bugs instead of rushing.

lib/std/sysrand.nim
ringabout: only small percentages of OS fall back to read `/dev/urandom`.
Reading from `/dev/urandom` is unsafe, that's why `getrandom` apis exists.

timotheecour:
> We cannot merge something with potential security issues/bugs like this, especially when it should cryptographically secure. Let's fix the bugs instead of rushing.

What I'm arguing for here is to allow iterating in a more effective way on new modules (or even new APIs in existing modules), including this one. This could be as simple as:
```nim
# in std/sysrand.nim
when not defined(nimExperimentalSysrand): static: doAssert "experimental module, unstable API"
# rest of code ...
```

so that:
* this module remains hidden by default, but people can easily try it in devel with explicit flag `-d:nimExperimentalSysrand`; more eyeballs => higher likelyhood to find bugs/design issues
* more than 1 person can easily contribute to it
* no need to constantly rebase long running PR's and resolve conflicts since they'd be merged faster
* we can more freely iterate without fear of breaking changes for regular devel users

Unlike the proposal discussed in https://github.com/nim-lang/RFCs/issues/310, there'd be no need for `std2` namespace, and no need for introducing an `experimental` namespace (which has the issues of causing problems when module is moved), since the only difference once module is stabilized is that the flag `-d:nimExperimentalFoo` would not be needed anymore;

This is essentially a lightweight form of `--experimental:foo` that can be used for new (non-trivial) modules/API's.
(refs https://github.com/timotheecour/Nim/issues/575)

lib/std/sysrand.nim
timotheecour: true (but the list include netbsd as of your PR), but at least it's better to be conservative here:
```nim
result = posix.read(fd, addr dest[result], left)
if result >=0 and result < left: result = -1 # short reads
# xxx handle EINTR
```

lib/std/sysrand.nim
timotheecour: but then `doAssert false` is better than `break`

(doAssert instad of assert because it's security related and the cost of the check is dwarfed by surrouding cost)

lib/std/sysrand.nim
ringabout: yeah, we didn't have netbsd CI, it that matters, we could use `getrandom` for netbsd too
https://man.netbsd.org/getrandom.2

timotheecour: totally fine to defer as future work with a TODO/xxx in the code, but up to you

ringabout: For a Rust project, they often supports many platforms,
image

timotheecour: link: https://docs.rs/getrandom/0.2.2/getrandom/

good reference; maybe should be added in top section as a comment

lib/std/sysrand.nim
ringabout: make sense

lib/std/sysrand.nim
```diff
@@ -0,0 +1,308 @@
...
+  else:
+    let ret = urandomInternalImpl(dest)
+    when defined(windows):
+      if ret != STATUS_SUCCESS:
+        result = false
+    else:
+      if ret < 0:
+        result = false
+
+proc urandom*(size: Natural): seq[byte] {.inline.} =
```
juancarlospaco, line 302: Should be `Positive` ?, whats `size = 0` useful for?.

What if it works in-place?, 
remove the `size`, take a `var seq[byte]`, rename to `fromUrandom` or similar.

dom96 (approved):
Beautiful, a perfect example of something that should be in the stdlib.

I'm mostly writing this because I feel like I very often reject PRs for new modules to the stdlib, it's nice to accept a case that I think does deserve to be in the stdlib :)

The API is also solid and docs look good. Implementation we can always easily improve as long as API doesn't need to change so IMO we can merge this now tbh.

dom96:
> We cannot merge something with potential security issues/bugs like this, especially when it should cryptographically secure. Let's fix the bugs instead of rushing.

Yes, we can. We just need to make it clear that the module is new, all this needs is a warning in the docs, something like:

> **Warning:** This module was added in Nim 1.6. If you are using it for crypto purposes then keep in mind that so far this has not been audited by any security professionals, as such may not be secure.

In fact, we should keep this warning until a reputable company does audit it and then link to the audit in the docs.

lib/std/sysrand.nim
timotheecour: > Should be Positive ?, whats size = 0 useful for?.

that'd introduce a weird special case for no benefit. 

> What if it works in-place?,

this PR already has that overload:
```nim
proc urandom*(dest: var openArray[byte]): bool =
```

but I'd argue we only need `proc urandom*(dest: var openArray[byte]): bool =`, because of `sugar.dup`

timotheecour:
> Yes, we can. We just need to make it clear that the module is new, all this needs is a warning in the docs, something like:

indeed. This is exactly what https://github.com/timotheecour/Nim/issues/575 is about: avoiding analysis paralysis where a PR slowly collects bitrot over time, and needs constant rebasing. More eyes = more bugfixes. 
We can hide this behind `when defined(nimExperimentalSysrandom)` in a next PR to allow fixing API in meantime as needed, until it stabilizes.

> reputable company does audit it and then link to the audit in the docs

it can't hurt but I'm not sure if it buys you much in practice beyond a marketing argument; this module depends on some parts of stdlib (system.nim) and compiler for correctness; that one-time audit wouldn't   cover those, and even if it did it wouldn't guard against regressions.

euantorano:
This is completely broken on NetBSD at the moment (and likely Dragonfly, though I haven’t tested there yet). 

Note that there are a few existing Nimble packages that aim to achieve the same thing, like:

- https://github.com/euantorano/sysrandom.nim/blob/master/src/sysrandom.nim
- https://github.com/cheatfate/nimcrypto/blob/master/nimcrypto/sysrand.nim

I would almost always base this kind of code on the fantastic libsodium, who’s approach can be found here: https://github.com/jedisct1/libsodium/blob/master/src/libsodium/randombytes/sysrandom/randombytes_sysrandom.c

I’m very happy to see this module get added to the stdlib though :)

timotheecour:
@euantorano PR welcome; also regarding running netbsd in CI

for netbsd, see also this comment https://github.com/nim-lang/Nim/pull/16459#discussion_r570899805 which has some potential solutions

ringabout:
see https://github.com/nim-lang/Nim/commit/18c24eb4d0694beb971581d59a4570826c5fa01f#r47295760

euantorano:
> @euantorano PR welcome; also regarding running netbsd in CI
> 
> for netbsd, see also this comment [#16459 (comment)](https://github.com/nim-lang/Nim/pull/16459#discussion_r570899805) which has some potential solutions

I have some nightly tests running here already: http://jenkins.euantorano.co.uk/job/Nim/124/console - it just runs on a schedule rather than on every PR/commit which is good enough for now until we get some of the failures fixed. 

> see [18c24eb#r47295760](https://github.com/nim-lang/Nim/commit/18c24eb4d0694beb971581d59a4570826c5fa01f#r47295760)

Yep, looks like it. It was broken when I commented, likely due to the nature of it running nightly.

PR #16460, opened 2020-12-24
Add assertions for jsconsole

juancarlospaco:
- Add assertions for `jsconsole`.
- Generates a `console.assert(expresion, message);`. You can keep using normal assertions too.
- Changelog, since, doc with links, `runnableExamples`, etc.

### DIFF
--- a/lib/js/jsconsole.nim
+++ b/lib/js/jsconsole.nim
@@ -10,7 +10,7 @@
 ## Wrapper for the `console` object for the `JavaScript backend
 ## <backends.html#backends-the-javascript-target>`_.
 
-import std/private/since, std/private/miscdollars
+import std/private/since, std/private/miscdollars  # toLocation
 
 when not defined(js) and not defined(Nimdoc):
   {.error: "This module only works on the JavaScript platform".}
--- a/lib/js/jsconsole.nim
+++ b/lib/js/jsconsole.nim
@@ -10,10 +10,12 @@
 ## Wrapper for the `console` object for the `JavaScript backend
 ## <backends.html#backends-the-javascript-target>`_.
 
+import std/private/since
+
 when not defined(js) and not defined(Nimdoc):
   {.error: "This module only works on the JavaScript platform".}
 
-type Console* = ref object of RootObj
+type Console* = ref object of JsRoot
 
 proc log*(console: Console) {.importcpp, varargs.}
   ## https://developer.mozilla.org/docs/Web/API/Console/log
@@ -66,5 +68,16 @@ proc timeLog*(console: Console, label = "".cstring) {.importcpp.}
 
 proc table*(console: Console) {.importcpp, varargs.}
   ## https://developer.mozilla.org/docs/Web/API/Console/table
+
+since (1, 5):
+  # proc nor func can not be used here, because may eval assertion at compile-time,
+  # writing directly false or true to the JavaScript instead of the expression.
+  # importjs nor importcpp can not be used with template.
+
+  template doAssert*(console: Console; assertion) =
+    ## https://developer.mozilla.org/en-US/docs/Web/API/Console/assert
+    runnableExamples: console.doAssert(42 == 42)
+    {.emit: "console.assert(" & astToStr(assertion) & ");".}
+
 
 var console* {.importc, nodecl.}: Console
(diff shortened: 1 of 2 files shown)


lib/js/jsconsole.nim
```diff
@@ -67,4 +69,15 @@ proc timeLog*(console: Console, label = "".cstring) {.importcpp.}
 proc table*(console: Console) {.importcpp, varargs.}
   ## https://developer.mozilla.org/docs/Web/API/Console/table
 
+since (1, 5):
```
timotheecour, line 72: this doesn't work, you're hitting A1 https://github.com/nim-lang/Nim/issues/13943
```nim
import jsconsole
proc main()=
  var a = 2
  console.doAssert 1 == a
  a = 3
  console.doAssert 1 == a
main()
```

ReferenceError: a is not defined
(look at the codegen)

this works:
```nim
{.emit: ["console.assert(", assertion, ");"].}
```

lib/js/jsconsole.nim
```diff
@@ -67,4 +69,15 @@ proc timeLog*(console: Console, label = "".cstring) {.importcpp.}
...
+  # proc nor func can not be used here, because may eval assertion at compile-time,
+  # writing directly false or true to the JavaScript instead of the expression.
+  # importjs nor importcpp can not be used with template.
+
+  template doAssert*(console: Console; assertion) =
```
timotheecour: rename it (jassert or similar), and make it clear in doc comment and runnableExamples that assert failure keeps going and just prints to console; for nodejs, specify whether it prints to stdout or stderr.

lib/js/jsconsole.nim
```diff
@@ -67,4 +69,15 @@ proc timeLog*(console: Console, label = "".cstring) {.importcpp.}
...
+    ## https://developer.mozilla.org/en-US/docs/Web/API/Console/assert
+    runnableExamples: console.doAssert(42 == 42)
+    {.emit: "console.assert(" & astToStr(assertion) & ");".}
```
timotheecour: the error message sucks, it gives zero context nor line number.
instead start with this:
```nim
    var a = ""
    a.addQuoted(astToStr(assertion))
    {.emit: ["console.assert(", assertion, ",",  a.cstring, ");"].}
```
and modify it to also file/line/col (same as in assertions.nim using `toLocation`)

lib/js/jsconsole.nim
juancarlospaco: But this hardcodes `true` or `false` at compile-time instead of emitting the actual expression. :thinking:

lib/js/jsconsole.nim
juancarlospaco: But is not meant to be better or equal than normal assertions, 
is just for when you need faster performance AND assertions, otherwise use normal assertions.

lib/js/jsconsole.nim
timotheecour: why is that a problem?

lib/js/jsconsole.nim
timotheecour: suppose you use this in a program, it'll give you 0 context as to where an jsAssert failure was triggered; and since it keeps going on failure, you'll just get:
```
assert failed
assert failed
assert failed
```

also, how would the suggestion result in slower performance?

juancarlospaco: Done, kinda similar message to Nim assertions.

lib/js/jsconsole.nim
```diff
@@ -67,4 +69,37 @@ proc timeLog*(console: Console, label = "".cstring) {.importcpp.}
...
+      console.jsAssert('`' == '\n' and '\t' == '\0') # Message correctly formatted
+      assert 42 == 42  # Normal assertions keep working
+
+    const
+      loc = instantiationInfo(fullPaths = compileOption("excessiveStackTrace"))
+      ploc = cstring($loc)
+    var msg = ""
+    msg.addQuoted(astToStr(assertion))
+    {.line: loc.}:
+      {.emit: ["console.assert(", assertion, ", ", ploc, ", ", msg.cstring, ");"].}
```
timotheecour: the ploc + msg can be computed at CT, eg:
```nim
const msg = getMsg(astToStr(assertion), $loc).cstring
{.line: loc.}:
  {.emit: ["console.assert(", assertion, ", ", msg, ");"].}
```
or similar

PR #16463, opened 2020-12-25
Add httpcore.toString

juancarlospaco:
- Continuing the work done on https://github.com/nim-lang/Nim/pull/16454
- Add `httpcore.toString` same as `$` but only accepts valid input `HttpCode` as standarized by IANA for correctness and safety.
- Changelog, docs with links, `runnableExamples` with `doAssert` and `doAssertRaises`, since, etc.
- A safer `$` basically.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -56,7 +56,7 @@
 - `writeStackTrace` is available in JS backend now.
 
 - `strscans.scanf` now supports parsing single characters.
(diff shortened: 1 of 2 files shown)


Araq:
Sorry, rejected because:

- Unclear how many people need the feature.
- Poorly named.
- A string representation's business is **not** constraint enforcements.

PR #16464, opened 2020-12-25
Don't use `unittest.suite` and `unittest.test` in testament

ringabout:
Ref https://github.com/nim-lang/Nim/pull/16448#issuecomment-750455031

### DIFF
--- a/doc/contributing.rst
+++ b/doc/contributing.rst
@@ -47,7 +47,7 @@ which only gets executed when the tester is building the file.
 Each test should be in a separate ``block:`` statement, such that
 each has its own scope. Use boolean conditions and ``doAssert`` for the
 testing by itself, don't rely on echo statements or similar; in particular, avoid
-things like `echo "done"`.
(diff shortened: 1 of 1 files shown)


doc/contributing.rst
```diff
@@ -47,7 +47,7 @@ which only gets executed when the tester is building the file.
 Each test should be in a separate ``block:`` statement, such that
 each has its own scope. Use boolean conditions and ``doAssert`` for the
 testing by itself, don't rely on echo statements or similar; in particular, avoid
-things like `echo "done"`.
+things like `echo "done"`. Don't use `unittest.suite` and `unittest.test`.
```
juancarlospaco, line 50:
```suggestion
things like `echo "done"`. Do not use `unittest` module on Testament.
```

ringabout: Yeah, you could USE unittest.check in testament.

timotheecour (approved):

PR #16473, opened 2020-12-26
fix nim js cmp fails at CT

### LINKED ISSUE #9: nim js cmp fails at CT: cannot generate VM code for asm

ringabout:
fix https://github.com/timotheecour/Nim/issues/9

```nim
import std/strtabs

proc fun()=
  let ret = newStringTable(modeCaseSensitive)
  ret["foo"] = "bar"

  echo ret

  let b = ret["foo"]
  echo b

proc main()=
  static: fun()
  fun()

main()
```
- nim js -r test.nim
system/jssys.nim(319, 3) Error: cannot generate VM code for asm " if ("a

### DIFF
--- a/lib/system.nim
+++ b/lib/system.nim
@@ -2394,14 +2394,8 @@ when notJSnotNims:
     """.}
 
 when defined(js):
-  when not defined(nimscript):
-    include "system/jssys"
-    include "system/reprjs"
-  else:
-    proc cmp(x, y: string): int =
-      if x == y: return 0
-      if x < y: return -1
-      return 1
+  include "system/jssys"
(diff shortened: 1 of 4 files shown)


timotheecour (commented):
LGTM

PR #16478, opened 2020-12-27
minor improvement for nim secret

ringabout:
Before
```
D:\QQPCmgr\Desktop\Nim>nim secret
Hint: used config file 'C:\Users\blue\.choosenim\toolchains\nim-#devel\config\nim.cfg' [Conf]
Hint: used config file 'C:\Users\blue\.choosenim\toolchains\nim-#devel\config\config.nims' [Conf]
.....>>> 12 + 4
16
>>>
```

After

```
D:\QQPCmgr\Desktop\Nim>nim secret
Hint: used config file 'C:\Users\blue\.choosenim\toolchains\nim-#devel\config\nim.cfg' [Conf]
Hint: used config file 'C:\Users\blue\.choosenim\toolchains\nim-#devel\config\config.nims' [Conf]
.....
>>> 12 + 4
16
>>>
```

### DIFF
--- a/compiler/llstream.nim
+++ b/compiler/llstream.nim
@@ -110,7 +110,7 @@ proc llReadFromStdin(s: PLLStream, buf: pointer, bufLen: int): int =
   s.rd = 0
   var line = newStringOfCap(120)
   var triples = 0
-  while readLineFromStdin(if s.s.len == 0: ">>> " else: "... ", line):
+  while readLineFromStdin(if s.s.len == 0: "\n>>> " else: "... ", line):
     s.s.add(line)
     s.s.add("\n")
     inc triples, countTriples(line)


timotheecour:
@xflywind 
I think we should revert this; this breaks `linenoise` integration in nim secret, making it painful to use s a REPL.

before this PR: we can use the usual bash/zsh-like navigation to edit cmdline
after this PR: each keyboard interaction (eg left arrow to move cursor when editing cmdline) creates a newline

The (minor) annoyance you're observing is fixable though.
## links
* https://github.com/nim-lang/Nim/pull/14418 which introduced dots
* https://github.com/nim-lang/Nim/issues/16334 which is an issue caused by dots that was recently fixed, see https://github.com/nim-lang/Nim/pull/16335

ringabout:
Let me think.
Could you give me some screenshots?

I will fix this differently soon.

timotheecour:
image


after clicking on left arrow a few times:

image


(might be easier to fix in a linux vm to see this behavior rather than windows ;-) )

ringabout:
Oh, I see.

timotheecour:
in future work, we should probably add some tests for `nim secret` (at least a few basic ones, to prevent regressions); this should be doable despite stdin interaction using osproc (iirc inim has some like that but this can be done rather simply)

(EDIT: basic tests in https://github.com/nim-lang/Nim/pull/16495)

ringabout:
Use `nim secret` in Linux, when pressing up button
```nim
(base) root@iZuf60lvzs716yak68vciwZ:~/Nim# nim secret
Hint: used config file '/root/.choosenim/toolchains/nim-#devel/config/nim.cfg' [Conf]
Hint: used config file '/root/.choosenim/toolchains/nim-#devel/config/config.nims' [Conf]
.....>>> 12 + 4

16
>>> ^[[A^[[A^[[A
```

ringabout:
How to compile nimsecret quick? It's painful to debug it.
I use `koch temp --d:useLinenoise secret` now.

ringabout:
It is reverted now.
https://github.com/nim-lang/Nim/pull/16483

krux02:
> Use `nim secret` in Linux, when pressing up button

use `rlwrap` like this: ``rlwrap nim secret``, and up button will provide you access to history. Also editing works it even has a vi mode if you care about that. You need to install rlwrap though.

ringabout:
thanks

PR #16480, opened 2020-12-27
Add mimetypes.mimesLongest

juancarlospaco:
- Add `mimetypes.mimesLongest` that return the length of the longest "ext" and "mime" from `mimes` database,
  this is useful for optimizations with `newStringOfCap` and `newString`.
- If you use it, all your `mimetypes` stuff can reduce allocs. Nothing changes if you do not use it.
- Changelog, since, doc with links, `runnableExamples` with `doAssert`, `strictFuncs` OK, etc.

```nim
const capForString: tuple[ext: int, mime: int] = mimesLongest()  # Use for newStringOfCap()
```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -56,7 +56,7 @@
 - `writeStackTrace` is available in JS backend now.
 
 - `strscans.scanf` now supports parsing single characters.
-- `strscans.scanTuple` added which uses `strscans.scanf` internally, returning a tuple which can be unpacked for easier usage of `scanf`. 
+- `strscans.scanTuple` added which uses `strscans.scanf` internally, returning a tuple which can be unpacked for easier usage of `scanf`.
 
 - Added `setutils.toSet` that can take any iterable and convert it to a built-in set,
   if the iterable yields a built-in settable type.
@@ -70,11 +70,13 @@
   and `lists.toDoublyLinkedList` convert from `openArray`s; `lists.copy` implements
   shallow copying; `lists.add` concatenates two lists - an O(1) variation that consumes
   its argument, `addMoved`, is also supplied.
-  
+
 - Added `sequtils` import to `prelude`.
 
 - Added `euclDiv` and `euclMod` to `math`.
 - Added `httpcore.is1xx` and missing HTTP codes.
+
+- Added `mimetypes.mimesLongest` that returns the lenght of the longest "ext" and "mime" from `mimes`.
 
 
 ## Language changes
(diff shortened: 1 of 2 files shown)


lib/pure/mimetypes.nim
```diff
@@ -1917,6 +1917,24 @@ func register*(mimedb: var MimeDB, ext: string, mimetype: string) =
   {.noSideEffect.}:
     mimedb.mimes[ext.toLowerAscii()] = mimetype.toLowerAscii()
 
+func mimesLongest*(): tuple[ext: int, mime: int] {.since: (1, 5).} =
+  ## Return the length of the longest "ext" and "mime" from `mimes`,
```
ringabout:
```suggestion
  ## Returns the length of the longest "ext" and "mime" from `mimes`,
```

Araq:
Ok, this is nice, but not computing this at compile-time is a shame. Compute it via a ``static`` block or similar.

juancarlospaco:
@Araq I used 2 `const`, to save a tuple unpacking at run-time, simpler UX, and not publish a func that returns static result.

temp

Araq:
Some bikshedding of the name: ``mimesMaxLen``, ``mimesExtMaxLen`` ?

changelog.md
```diff
@@ -70,12 +70,14 @@
...
   its argument, `addMoved`, is also supplied.
-  
+
 - Added `sequtils` import to `prelude`.
 
 - Added `euclDiv` and `euclMod` to `math`.
 - Added `httpcore.is1xx` and missing HTTP codes.
 
+- Added `mimetypes.mimesLongest` that returns the length of the longest "ext" and "mime" from `mimes`.
+
```
Araq, line 81: You need to update the changelog.

lib/pure/mimetypes.nim
```diff
@@ -1917,6 +1917,33 @@ func register*(mimedb: var MimeDB, ext: string, mimetype: string) =
   {.noSideEffect.}:
     mimedb.mimes[ext.toLowerAscii()] = mimetype.toLowerAscii()
 
+
+since (1, 5):
+  func mimesLongest(): array[2, int] {.compiletime.} =
+    runnableExamples:
+      static:
+        doAssert mimesLongest() >= (ext: 24, mime: 73)
```
timotheecour, line 1925: @juancarlospaco this is bad for several reasons:
* runnableExamples inside a private proc is never run/compiled
* the code is in fact invalid (>= not defined on array)
* even if it were valid (say it returned a tuple), it's bad because  `(23, 73) >= (22, 100)`


and as reported elsewhere it causes this regression: https://github.com/nim-lang/Nim/issues/16703 (EDIT: which itself is caused by https://github.com/nim-lang/Nim/issues/16790)

PR #16527, opened 2020-12-31
refactor the tests of options

ringabout:
I planed to work on https://github.com/timotheecour/Nim/issues/6#issuecomment-633500785, however https://github.com/nim-lang/Nim/issues/15730 blocks me.

So I could only refactor the tests of options.
I also find a bug:

```nim
import options

# change to templates, it works
# closure anonymous functions doesn't work in VM with JS
# Error: cannot evaluate at compile time: procRan
proc main() =
  var procRan = 0
  some(123).map(proc (v: int) = procRan = v)

static: main()
```

### DIFF
--- a/tests/stdlib/toptions.nim
+++ b/tests/stdlib/toptions.nim
@@ -1,19 +1,8 @@
 discard """
-  output: '''{"foo":{"test":"123"}}'''
+  targets: "c js"
 """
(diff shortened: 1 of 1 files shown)


juancarlospaco:
Looks good :+1:

PR #16529, opened 2020-12-31
Algorithm improvements

konsumlamm:
This PR aims to improve the documentation for the algorithm module. It replaces double with single backticks and changes the descriptions to imperative form (as recommended in the contributing guideline). It also fixes some grammar issues and typos.

Some noteworthy changes:
* There was a bug in the `binarySearch` proc where the result of a comparison was compared to 1 instead of 0 (line 199). In practice this probably didn't cause any issues, because most comparison functions return -1, 0 or 1, but that isn't guaranteed.
* Changes the examples for `binarySearch` to use sorted arrays, since the algorithm assumes that the input is sorted. The previous examples did work, but that was only a special case and it invokes the impression that it works fine if the input is unsorted.
* The documentation for `lowerBound` and `upperBound` was swapped.

Two things I noticed:
* [`sort`](https://nim-lang.github.io/Nim/algorithm.html#sort%2CopenArray%5BT%5D%2Cproc%28T%2CT%29) and [`isSorted`](https://nim-lang.github.io/Nim/algorithm.html#isSorted%2CopenArray%5BT%5D%2Cproc%28T%2CT%29) are `func`s when given a custom `cmp`, but the other versions are `proc`s, eventhough they just delegate to the first version.
* The `last` parameter of [`reversed`](https://nim-lang.github.io/Nim/algorithm.html#reversed%2CopenArray%5BT%5D%2CNatural%2Cint) has type `int`, while the `first` parameter and the `last` parameter of [`reverse`](https://nim-lang.github.io/Nim/algorithm.html#reverse%2CopenArray%5BT%5D%2CNatural%2CNatural) have type `Natural`. I think this is so that `reversed` works on empty `openArray`s, but that could be changed by calling `reversed(a, 0, max(0, a.high))`.

I assume these are unintended, but I didn't change it, because I'm not sure and it's probably backwards incompatible.

### DIFF
--- a/lib/pure/algorithm.nim
+++ b/lib/pure/algorithm.nim
@@ -47,7 +47,7 @@ type
     Descending, Ascending
 
 proc `*`*(x: int, order: SortOrder): int {.inline.} =
-  ## Flip `x` if `order == Descending`.
+  ## Flips the sign of `x` if `order == Descending`.
   ## If `order == Ascending` then `x` is returned.
   ##
   ## `x` is supposed to be the result of a comparator, i.e.
@@ -69,9 +69,9 @@ template fillImpl[T](a: var openArray[T], first, last: int, value: T) =
     inc(x)
 
 proc fill*[T](a: var openArray[T], first, last: Natural, value: T) =
-  ## Fill the slice `a[first..last]` with `value`.
+  ## Assigns `value` to all elements of the slice `a[first..last]`.
   ##
-  ## If an invalid range is passed, it raises IndexDefect.
+  ## If an invalid range is passed, it raises `IndexDefect`.
   runnableExamples:
     var a: array[6, int]
     a.fill(1, 3, 9)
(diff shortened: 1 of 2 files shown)


lib/pure/algorithm.nim
```diff
@@ -49,18 +47,18 @@ type
     Descending, Ascending
 
 proc `*`*(x: int, order: SortOrder): int {.inline.} =
-  ## Flips ``x`` if ``order == Descending``.
-  ## If ``order == Ascending`` then ``x`` is returned.
+  ## Flip `x` if `order == Descending`.
```
ringabout: `Flips` is preferred.

konsumlamm: The [contributing guideline](https://nim-lang.github.io/Nim/contributing.html#documentation) says:
```
The preferred documentation style is to begin with a capital letter and use the imperative (command) form.
```

Araq:
Old version was better, we updated the contributing docs.

konsumlamm:
@Araq This PR does more than just changing the documentation to imperative form, please read my summary above. I added a new commit that reverts the changes to imperative form and has some additional improvements.

PR #16535, opened 2020-12-31
Add short description on memory management

FedericoCeratto:
From time to time people comment that GC makes Nim unfit for games or other uses, often under the assumption that Nim uses an intrusive "stop-the-world" model.
Add a short introduction and a note to help clarify memory management.


Araq:
Nim also always offered manual memory management...

Issue #303, opened 2013-01-09
Provide all symbols (incl. stdlib and babel).


reactormonk:
Provide all symbols that are known to the stdlib and babel. Intended for documentation lookup.

PR #16536, opened 2020-12-31
std/lists: Various changes to `lists` (RFC #303)

salvipeter:


salvipeter:
As discussed [here](https://github.com/nim-lang/RFCs/issues/303#issuecomment-748515887).

Clyybber (commented):
`append` should not be deprecated IMO, as it looks more natural in code that also uses `prepend`.

Araq:
Yeah, please don't deprecate ``append``, I don't appreciate ecosystem-damaging micro-optimizations.

lib/pure/collections/lists.nim
```diff
@@ -225,7 +225,7 @@ iterator items*[T](L: SomeLinkedList[T]): T =
   ## .. code-block::
```
timotheecour: * replace with:
```nim
  ## ...
  ## * `nodes iterator <#nodes.i,SomeLinkedList[T]>`_
  runnableExamples:
    from std/sugar import collect
    let a = collect(initSinglyLinkedList):
      for i in 1..3: 10*i
    doAssert toSeq(items(a)) == toSeq(a)
    doAssert toSeq(a) == @[10, 20, 30]
```

* ditto with remaining code-blocks

lib/pure/collections/lists.nim
```diff
@@ -373,29 +373,79 @@ proc contains*[T](L: SomeLinkedCollection[T], value: T): bool {.inline.} =
...
+    assert a.toSeq == [1, 2, 3, 4, 5]
+    assert b.toSeq == [1, 2, 3]
+    a.prepend a
+    assert a.toSeq == [1, 2, 3, 4, 5, 1, 2, 3, 4, 5]
+  var tmp = b.copy
+  tmp.addMoved a
+  a = tmp
+
+proc prependMoved*[T: SomeLinkedList](a, b: var T) {.since: (1, 5, 1).} =
+  ## Moves `b` before the head of `a`. Efficiency: O(1).
```
timotheecour, line 385:
```suggestion
  ## Moves `b` before the head of `a`. Complexity: O(1).
```

lib/pure/collections/lists.nim
```diff
@@ -373,29 +373,79 @@ proc contains*[T](L: SomeLinkedCollection[T], value: T): bool {.inline.} =
...
+  ## Note that `b` becomes empty after the operation unless it has the same address as `a`.
+  ## Self-prepending results in a cycle.
+  ##
+  ## See also:
+  ## * `prepend proc <#prepend,T,T>`_
+  ##   for prepending a copy of a list
+  runnableExamples:
+    import sequtils, std/enumerate, std/sugar
```
timotheecour:
```suggestion
    import std/[sequtils, enumerate, sugar]
```

lib/pure/collections/lists.nim
timotheecour: EDIT: looks like sets and tables also use Efficiency, IMO the appropriate term is Complexity, but this should rather be changed in some future PR

lib/pure/collections/lists.nim
```diff
@@ -669,13 +719,33 @@ proc add*[T: SomeLinkedList](a: var T, b: T) {.since: (1, 5, 1).} =
...
+  ## Removes a node `n` from `L`.
+  ## Efficiency: O(n), unless `n` is the head element.
+  runnableExamples:
+    var
+      a = initSinglyLinkedList[int]()
+      n = newSinglyLinkedNode[int](5)
+    a.add(n)
+    assert 5 in a
+    a.remove(n)
+    assert 5 notin a
```
timotheecour:
```suggestion
  runnableExamples:
    from sequtils import toSeq
    var a = [0,1,2].toSinglyLinkedList
    let n = a.head.next
    doAssert n.value == 1
    a.remove n
    doAssert toSeq(a) == [0,2]
```

tests/stdlib/tlists.nim
```diff
@@ -146,5 +146,59 @@ template testCommon(initList, toList) =
...
+  block: # prepend, prependMoved
+    block:
+      var
+        l0 = initList[int]()
+        l1 = [1].toList
+        l2 = [2, 3].toList
+        l3 = [4, 5, 6].toList
+      l0.prepend l3
+      l1.prepend l3
+      l2.prependMoved l3
```
timotheecour, line 167: to check that prepend didn't affect l3
```suggestion
      l3.toSeq == [4, 5, 6]
      l2.prependMoved l3
```

lib/pure/collections/lists.nim
```diff
@@ -669,13 +719,33 @@ proc add*[T: SomeLinkedList](a: var T, b: T) {.since: (1, 5, 1).} =
...
+
+  if n == L.head:
```
timotheecour, line 722: this gives a SIGSEGV
```nim
when true:
  import std/lists
  from sequtils import toSeq
  var a = [3].toSinglyLinkedList
  let n = newSinglyLinkedNode[int](5)
  a.remove n
```

instead, either:
* raise `KeyError` when `n` is not found
* or make this a noop when `n` not found

And also:
* specify it in doc comment
* add corresponding example with `doAssertRaises`
* likewise for `DoublyLinkedList`

note that for `DoublyLinkedList` it's a noop:
```nim
when true:
  import std/lists
  from sequtils import toSeq
  var a = [3].toDoublyLinkedList
  let n = newDoublyLinkedNode[int](5)
  a.remove n
```

so the most logical is a noop here too

lib/pure/collections/lists.nim
salvipeter: Also those in the `basic usage` part?

lib/pure/collections/lists.nim
salvipeter: Nice!

salvipeter:
While I'm at it, there are a couple of other functions that would be helpful, such as:
- `length`
- `isEmpty`
- `[]` (indexing operator, giving the n-th element)
- `first` & `rest` (maybe `pop` and `popFirst` ?)

And some unification:
- no uppercase variables (`L`) - maybe `a`?
- in `runnableExamples`, always use `from ... import` (instead of just `import`)?
- in `runnableExamples`, always use `assert` (instead of `doAssert`)?
- in `runnableExamples`, always use `toSeq` (instead of `$... == "..."`)?
- there is considerable variation in style, e.g., `add(a, b)` vs. `a.add(b)` vs. `a.add b` ... what is the policy?

What do you think? Also, is the reference to the list-list `add` in the reST comment of `append` (l. 379) correct? Can I specify it better?

lib/pure/collections/lists.nim
timotheecour: yes please! `.. code-block::` is not even tested (refs https://github.com/nim-lang/RFCs/issues/309#issuecomment-752821157)

EDIT: but this can be done in future PR

timotheecour:
> While I'm at it, there are a couple of other functions that would be helpful, such as:

yes, but all these in another PR so that this PR can be merged sooner rather than later :)
in fact, 2 PRs: one for style-like changes, and one for symbol additions.

notes:
* `length` => should be `len`
* `isEmpty` => not sure; `a.len == 0` would be less efficient indeed, but I'm wondering whether a completely generic `isDefault` would subsume this (see https://github.com/nim-lang/Nim/pull/13526)
* `no uppercase variables (L) - maybe a?` yes
* `[]` yes (and also `[]=`)
* `first & rest`; no; IMO it's better (more general) to split on a Node n: `split(a: list, b: Node): tuple[lhs, rhs: list]`
* `(maybe pop and popFirst ?)` I need to think; it's a bit redundant with `remove`; let's re-discuss this

* `in runnableExamples, always use assert (instead of doAssert)?` no, because there's no consensus yet on assert vs doAssert inside runnableExamples (only inside tests, where doAssert should be used)
* `in runnableExamples, always use toSeq (instead of $... == "...")?` yes, except of course for the  runnableExamples for `$`
* `in runnableExamples, always use from ... import (instead of just import)?` 
  EDIT: how about: only if there's a single import, so that examples are kept small; eg:
  `from sequtils import toSeq` : ok
  `import sequtils, sugar`: ok
  if/when https://github.com/nim-lang/RFCs/issues/76 is implemented, we'd be able to have:
  `from sequtils, sugar import collect, toSeq`
   

> there is considerable variation in style, e.g., add(a, b) vs. a.add(b) vs. a.add b ... what is the policy?

historical baggage and lack of specification / enforcement of https://nim-lang.github.io/Nim/nep1.html, let's discuss this separately

> Also, is the reference to the list-list add in the reST comment of append (l. 379) correct? Can I specify it better?

we currently can't do better than what you wrote (`lists.html#add,T,T`) because docgen ignores the `: SomeLinkedList` in the declaration `proc add[T: SomeLinkedList](a: var T; b: T)` (so in particular would be incorrect with another overload `proc add[T: Foo](a: var T; b: T)`), so what you wrote is correct; however, the https://github.com/nim-lang/RFCs/issues/125 I proposed will simplify all this and also make doc links more robust.

lib/pure/collections/lists.nim
```diff
@@ -382,9 +368,16 @@ proc contains*[T](L: SomeLinkedCollection[T], value: T): bool {.inline.} =
 
   result = find(L, value) != nil
 
-proc append*[T](L: var SomeLinkedCollection[T], n: SomeLinkedNode[T] | T)
-    {.deprecated: "Deprecated since v1.6; use `add`".} =
-  add(L, n)
+proc append*[T](a: var SomeLinkedCollection[T], b: SomeLinkedNode[T] | T) =
+  ## Alias for `a.add(b)`.
```
timotheecour: I aknowledge this comment `Yeah, please don't deprecate append, I don't appreciate ecosystem-damaging micro-optimizations.` but can we at least write:
```nim
## Deprecated alias for `a.add(b)`.
```

this won't cause any warning but at least will nudge users to use `add` instead of `append`, for reasons already discussed before in https://github.com/nim-lang/RFCs/issues/303#issuecomment-747787419; there should be 1 preferred way to do things.

> it looks more natural in code that also uses prepend.

that's not a strong argument; if we add `prepend` to some API that already contained `add`, that'd mean we'd have to also add `append` just for language symmetry; IMO the arguments in https://github.com/nim-lang/RFCs/issues/303#issuecomment-747787419 make a stronger case.

timotheecour:
@salvipeter please remember to mark as resolved comments you've resolved :)

timotheecour:
- [ ] @salvipeter for `remove` how about instead of a noop (or before your recent commit, a SIGSEGV), do as follows:
```nim
proc remove*[T](L: var SinglyLinkedList[T], n: SinglyLinkedNode[T]) =
  ## ...
=>
proc remove*[T](L: var SinglyLinkedList[T], n: SinglyLinkedNode[T]): bool =
  ## ...
  ## Returns true if `n` was found in `L`
```
which is more flexible, eg client code can discard the result if they don't care, of `doAssert` if they care that element was in list prior to remove. 


Note that the pre-existing `remove(DoublyLinkedList, n)` is O(1) instead of O(n) and cannot do such a check, so returning void in that case was fine.

lib/pure/collections/lists.nim
```diff
@@ -669,13 +712,36 @@ proc add*[T: SomeLinkedList](a: var T, b: T) {.since: (1, 5, 1).} =
   var tmp = b.copy
   a.addMoved tmp
 
+proc remove*[T](L: var SinglyLinkedList[T], n: SinglyLinkedNode[T]) =
+  ## Removes a node `n` from `L`.
+  ## Efficiency: O(n), unless `n` is the head element.
```
timotheecour: ` unless `n` is the head element.` doesn't really make sense in the context of discussing complexity, and also is too restrictive because runtime is also low if element appears early in list; instead do this:
```suggestion
  ## Efficiency: O(n); the list is traversed until `n` is found.
```

timotheecour:
- [ ] `proc append*[T](a: var SomeLinkedCollection[T], b: SomeLinkedNode[T] | T) =` isn't correct as it'd accept this:
```nim
var a = [10,11,12,13].toDoublyLinkedList
a.append newSinglyLinkedNode[int](5)
```
(and then would give an error in the wrong place, inside instantiation of `append` instead of giving a sigmatch error before instantiating `append`)

instead do this:
```nim
proc append*[T](a: var (SinglyLinkedList[T] | SinglyLinkedRing[T]), b: SinglyLinkedNode[T] | T) =
  ## ...
  a.add b
proc append*[T](a: var (DoublyLinkedList[T] | DoublyLinkedRing[T]), b: DoublyLinkedNode[T] | T) =
  ## ...
  a.add b
```

I don't think concepts or generics are powerful enough to allow writing this using a single proc (maybe @araq knows how?), but this would be easy with `{.enableif.}` (refs https://github.com/nim-lang/Nim/pull/12048):
```nim
proc append*[T](a: var SomeLinkedCollection[T]), b: SomeLinkedNode[T] | T) {.enableif: type(a.head) is type(b).} =
  a.add b
```
for eg, this fails:
```nim
type Bar[T] = object
  foo: int
proc fn[T](a: Bar[T], b: typeof(a.foo)) = discard # Error: undeclared field: 'foo'
```

tests/stdlib/tlists.nim
```diff
@@ -146,5 +146,62 @@ template testCommon(initList, toList) =
...
+      var c = [0, 1].toList
+      c.prependMoved c
+      let s = collect:
+        for i, ci in enumerate(c):
+          if i == 6: break
+          ci
+      doAssert s == [0, 1, 0, 1, 0, 1]
+
+  block remove:
+    var l = [0, 1, 2, 3].toList
```
timotheecour, line 193: * also add this test case where you remove an element from a cycle:
```nim
when true:
  import lists
  template fn(init) =
    var a = [10,11,12].toDoublyLinkedList
    a.addMoved a
    doAssert take(a, 6) == @[10, 11, 12, 10, 11, 12]
    remove(a, a.head.next)
    doAssert take(a, 6) == @[10, 12, 10, 12, 10, 12]
  fn(toDoublyLinkedList)
  fn(toSinglyLinkedList)
```

* and specify in doc comment for both `remove` overloads that removing an element from a cycle maintains a cycle (it's not obvious, it could've also cut the cycle; but note that for for `DoublyLinkedList`, since we're limited to O(1) for remove, that's the only possible behavior, so might as well preserve this behavior for `SinglyLinkedList`)
* and also specify this behavior in a runnableExamples (can combine into 1 to avoid duplication)

* I'm using this definition for take, which I want to add to sequtils https://github.com/timotheecour/Nim/issues/504; until we add `take`, you can use this in tslists.nim for now (and use this instead of enumerate everywhere in this file), but for runnableExamples, you can use the usual enumerate

salvipeter:
Yes, this was bothering me, too - it would be nice if it could be written as a single proc, but for the time being I changed it as you recommended. Also added `SinglyLinkedList` / `DoublyLinkedList` as a possible type for `b`, which was missing.
To be consistent, we also need an `appendMoved` which is an alias for `addMoved` ... but it is a bit strange to introduce a new symbol and call it a "deprecated alias" right away.

salvipeter:
As for `remove` returning a `bool`, I am not so sure. I think it has little use, but it would clutter up "normal" code with `discard`s.

tests/stdlib/tlists.nim
salvipeter: At this point I am starting to doubt the usefulness of the ...Ring types. Maybe they should be removed, and replaced with just this:
```nim
proc toRing[T](a: var SomeLinkedList) = a.addMoved a
```

salvipeter: How can I combine the runnableExamples of two different `remove` procs into one?

timotheecour: just refer to it, eg:
```nim
see examples for <link...>
```

timotheecour: > At this point I am starting to doubt the usefulness of the ...Ring types. Maybe they should be removed, and replaced with just this:

semantics are different, eg `toSeq(someRing)` makes sense, whereas `toSeq(someLinkedListWithLoop)` will go in infinite loop and then likely go out of memory

timotheecour:
> To be consistent, we also need an appendMoved which is an alias for addMoved

not sure we need; the point of re-adding append after the rename append=>add was to avoid breaking code; but no code ever used appendMoved so no need to introduce this alias

timotheecour:
> it would clutter up "normal" code with discards.

there's also `{.discardable.}`

timotheecour:
ping @salvipeter

lib/pure/collections/lists.nim
```diff
@@ -669,13 +722,47 @@ proc add*[T: SomeLinkedList](a: var T, b: T) {.since: (1, 5, 1).} =
...
+  else:
+    var prev = L.head
+    while prev.next != n and prev.next != nil:
+      prev = prev.next
+    if prev.next != nil:
+      prev.next = n.next
+
 proc remove*[T](L: var DoublyLinkedList[T], n: DoublyLinkedNode[T]) =
   ## Removes a node `n` from `L`. Efficiency: O(1).
+  ## When the list is cyclic, the cycle is preserved after removal.
```
timotheecour, line 739: need to specify what happens when `L` is not in `n`; note that we can't check whether `L` is in `n` as it'd break the `O(1)` cost, but at least this should be explained

salvipeter:
> ping @salvipeter

I'm still alive, just work & family have priority :) I'll do a revision soon though.

lib/pure/collections/lists.nim
```diff
@@ -752,12 +753,18 @@ proc remove*[T](L: var SinglyLinkedList[T], n: SinglyLinkedNode[T]) =
...
+      return false
+    prev.next = n.next
+  true
 
 proc remove*[T](L: var DoublyLinkedList[T], n: DoublyLinkedNode[T]) =
   ## Removes a node `n` from `L`. Efficiency: O(1).
-  ## When the list is cyclic, the cycle is preserved after removal.
+  ## This function assumes, for the sake of efficiency, that `n` is contained in `L`,
+  ## otherwise the effects are undefined.
+  ## When the list is cyclic, the cycle is preserved after removal,
```
timotheecour: can you add a test for that?

tests/stdlib/tlists.nim
```diff
@@ -217,3 +217,13 @@ template testCommon(initList, toList) =
...
+
+block remove: # return value check
+  var l = [0, 1, 2, 3].toSinglyLinkedList
+  let n = l.head.next.next
+  doAssert l.remove(n) == true
+  doAssert l.toSeq == [0, 1, 3]
+  doAssert l.remove(n) == false
+  doAssert l.toSeq == [0, 1, 3]
+  doAssert l.remove(l.head) == true
+  doAssert l.toSeq == [1, 3]
```
timotheecour, line 229: also test removing last element:
```nim
  doAssert l.remove(l.head.next) == true
  doAssert l.toSeq == [1]
  doAssert l.remove(l.head) == true
  doAssert l.toSeq == []
```

lib/pure/collections/lists.nim
salvipeter: How would you test undefined effects? :)
If you mean the cycle preservation, I included a link as you recommended...

timotheecour: yes i meant cycle preservation

lib/pure/collections/lists.nim
```diff
@@ -22,50 +22,43 @@
...
-##
-##   assert a.next == b
-##   assert a.prev == c
-##   assert c.next == a
-##   assert c.next.next == b
-##   assert c.prev == nil
-##   assert b.next == nil
-##
-##
+runnableExamples:
```
ringabout, line 25: please check the doc generated. runnableExamples doesn't work well with  headers.

salvipeter:
JavaScript compilation seems to break because the `swap a, b` call at `lists.nim:408` has no effect, I don't know why.

tests/stdlib/tlists.nim
```diff
@@ -2,15 +2,15 @@ discard """
   targets: "c js"
 """
```
Clyybber, line 4:
```suggestion
  targets: "c"
"""
#XXX: js target currently disabled for this test because `swap a, b` call at lists.nim:408 has no effect
```

timotheecour: @salvipeter this is bad for 2 reasons:
* it's a test coverage regression (previously js was being tested, now it's not tested at all and won't catch future regressions); when you encounter such cases, instead of disabling the entire test, disable just what's needed, as follows:
```nim
block: # prepend, prependMoved
  when not defined(js): # xxx swap broken in js, refs: bug #16771
    ...
```

* furthermore, the issue is actually easy to work around, simply write:
```nim
# in lists.prependMoved:
  when defined(js): # xxx swap broken in js, refs: bug #16771
    (b, a) = (a, b)
  else: swap a, b
```

Clyybber: @timotheecour It was me who commited the suggestion, not salvipeter.
If it's easy to workaround, sure let's do that.

tests/stdlib/tlists.nim
```diff
@@ -1,16 +1,16 @@
 discard """
-  targets: "c js"
+  targets: "c"
 """
-
-import lists, sequtils, std/enumerate, std/sugar
+#XXX: js target currently disabled for this test because `swap a, b` call at lists.nim:408 has no effect
```
Clyybber, line 4:
```suggestion

```

tests/stdlib/tlists.nim
```diff
@@ -1,16 +1,15 @@
 discard """
   targets: "c js"
 """
-
-import lists, sequtils, std/enumerate, std/sugar
+import lists, sequtils
```
Clyybber, line 4:
```suggestion

import lists, sequtils
```

tests/stdlib/tlists.nim
Clyybber: Done

lib/pure/collections/lists.nim
```diff
@@ -980,3 +952,31 @@ proc remove*[T](L: var DoublyLinkedRing[T], n: DoublyLinkedNode[T]) =
...
+  ## * `add proc <#add,T,T>`_
+  a.add b
+
+proc appendMoved*[T: SomeLinkedList](a, b: var T) {.since: (1, 5, 1).} =
+  ## Deprecated alias for `a.addMoved(b)`.
+  ##
+  ## See also:
+  ## * `addMoved proc <#addMoved,SinglyLinkedList[T],SinglyLinkedList[T]>`_
+  ## * `addMoved proc <#addMoved,DoublyLinkedList[T],DoublyLinkedList[T]>`_
+  a.addMoved b
```
timotheecour: remove this, refs https://github.com/nim-lang/Nim/pull/16536#issuecomment-753599392
```suggestion
```

lib/pure/collections/lists.nim
```diff
@@ -880,3 +954,31 @@ proc remove*[T](L: var DoublyLinkedRing[T], n: DoublyLinkedNode[T]) =
       L.head = nil
     else:
       L.head = L.head.prev
+
+proc append*[T](a: var (SinglyLinkedList[T] | SinglyLinkedRing[T]),
+                b: SinglyLinkedList[T] | SinglyLinkedNode[T] | T) =
+  ## Deprecated alias for `a.add(b)`.
```
Clyybber, line 960:
```suggestion
  ## Alias for `a.add(b)`.
```

lib/pure/collections/lists.nim
```diff
@@ -880,3 +954,31 @@ proc remove*[T](L: var DoublyLinkedRing[T], n: DoublyLinkedNode[T]) =
...
+  ##
+  ## See also:
+  ## * `add proc <#add,DoublyLinkedList[T],DoublyLinkedNode[T]>`_
+  ## * `add proc <#add,DoublyLinkedList[T],T>`_
+  ## * `add proc <#add,T,T>`_
+  a.add b
+
+proc appendMoved*[T: SomeLinkedList](a, b: var T) {.since: (1, 5, 1).} =
+  ## Deprecated alias for `a.addMoved(b)`.
```
Clyybber, line 979:
```suggestion
  ## Alias for `a.addMoved(b)`.
```

changelog.md
```diff
@@ -109,6 +109,9 @@ with other backends. see #9125. Use `-d:nimLegacyJsRound` for previous behavior.
 - Added `math.signbit`.
 
 - Removed the optional `longestMatch` parameter of the `critbits._WithPrefix` iterators (it never worked reliably)
+- In `lists`: renamed `append` to `add` and retained `append` (and added `appendMoved`) as an alias;
```
```suggestion
- In `lists`: renamed `append` to `add` and retained `append` as an alias;
```

refs https://github.com/nim-lang/Nim/pull/16536#issuecomment-753599392

changelog.md
```diff
@@ -109,6 +109,9 @@ with other backends. see #9125. Use `-d:nimLegacyJsRound` for previous behavior.
...
+  added `prepend` and `prependMoved` analogously to `add` and `andMoved`;
```
```suggestion
  added `prepend` and `prependMoved` analogously to `add` and `addMoved`;
```

lib/pure/collections/lists.nim
```diff
@@ -880,3 +954,31 @@ proc remove*[T](L: var DoublyLinkedRing[T], n: DoublyLinkedNode[T]) =
...
+  ## * `add proc <#add,T,T>`_
+  a.add b
+
+proc appendMoved*[T: SomeLinkedList](a, b: var T) {.since: (1, 5, 1).} =
+  ## Alias for `a.addMoved(b)`.
+  ##
+  ## See also:
+  ## * `addMoved proc <#addMoved,SinglyLinkedList[T],SinglyLinkedList[T]>`_
+  ## * `addMoved proc <#addMoved,DoublyLinkedList[T],DoublyLinkedList[T]>`_
+  a.addMoved b
```
timotheecour, line 984:
```suggestion
```

refs https://github.com/nim-lang/Nim/pull/16536#issuecomment-753394613

salvipeter: I think it would be a mistake not to add `appendMoved`, especially now that we are not deprecating the `append` functions at all. If there is `prependMoved`, symmetry asks for `appendMoved`. But that's just my 10 cents.

timotheecour: how about removing it in this PR and so that we can move forward with this PR and adding it in another PR so it can be discussed separately without block this?

apart from this point, I think it's ready to go

salvipeter: I'm not that emotionally attached to the idea, just wanted to give you a chance to reconsider :)

timotheecour (approved):
LGTM, excellent work!

lib/pure/collections/lists.nim
Clyybber: I agree with salvipeter, and it should be decided in this PR IMO.

lib/pure/collections/lists.nim
```diff
@@ -880,3 +954,23 @@ proc remove*[T](L: var DoublyLinkedRing[T], n: DoublyLinkedNode[T]) =
...
+
+proc append*[T](a: var (DoublyLinkedList[T] | DoublyLinkedRing[T]),
+                b: DoublyLinkedList[T] | DoublyLinkedNode[T] | T) =
+  ## Alias for `a.add(b)`.
+  ##
+  ## See also:
+  ## * `add proc <#add,DoublyLinkedList[T],DoublyLinkedNode[T]>`_
+  ## * `add proc <#add,DoublyLinkedList[T],T>`_
+  ## * `add proc <#add,T,T>`_
+  a.add b
```
Clyybber, line 976:
```suggestion
  a.add b

proc appendMoved*[T: SomeLinkedList](a, b: var T) {.since: (1, 5, 1).} =
  ## Alias for `a.addMoved(b)`.
  ##
  ## See also:
  ## * `addMoved proc <#addMoved,SinglyLinkedList[T],SinglyLinkedList[T]>`_
  ## * `addMoved proc <#addMoved,DoublyLinkedList[T],DoublyLinkedList[T]>`_
  a.addMoved b
```

lib/pure/collections/lists.nim
Clyybber: @timotheecour I'm readding it to get on with this and not subject salvipeter to this review ping pong :D
Removing `appendMoved` can also be done in a further PR (until the release); but as of now there is no argument to not add it.

EDIT: Let's wait for @Araq to decide.

salvipeter:
ping ?

timotheecour:
@salvipeter thanks for pinging, don't hesitate to in future, it's easy to forget PR's ; now that this was merged, PR welcome to address remaining points that were discussed here (ideally, breaking down in multiple smaller PR's whenever possible will make reviewing/merging faster)

ringabout:
The regression devel docs

image

lib/pure/collections/lists.nim
konsumlamm: I know you already fixed this, but the issue was not the `runnableExamples` itself, but a missing blank line, this would have worked:
```nim
## Lists
## -----
##
runnableExamples:
  ...
```

ringabout: not really, that still had other issues.  If using rst syntax, the second sub headers won't be rendered correctly. You can compare with the right ome.

timotheecour:
fixed in https://github.com/nim-lang/Nim/pull/16981; thanks @xflywind

lib/pure/collections/lists.nim
timotheecour: indeed, tracked in https://github.com/nim-lang/Nim/issues/16990

PR #16540, opened 2021-01-01
Added repr and dollar for set[char] using range notation where appropriate

zevv:
So {'a'..'z'} will also be repr'ed and `$`-ed as {'a'..'z'}, instead of {'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z'}


zevv:
So `{'a'..'z'}` will also be repr'ed as `{'a'..'z'}`, instead of `{'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z'}`

lib/system/repr.nim
```diff
@@ -117,9 +117,15 @@ proc reprSetAux(result: var string, p: pointer, typ: PNimType) =
         inc(elemCounter)
   add result, "}"
 
+proc reprSetChar(result: var string, p: pointer) =
+  result = $cast[ptr set[char]](p)[]
```
zevv, line 121: Is this allowed, and if so, why is this not used by other reprs as well?

metagn:
Excuse me if I'm missing something, but why only chars?

zevv:
Because that scratches my personal itch. I guess this could safely be extended to SomeNumber as well, but it might be confusing for enums or other types.

Araq:
Ok, but a range like '%'..'$' is not understandable. You need even more specialization so that only 0-9, a-z, A-Z or the '\numberHere' things are affected. And this code shouldn't be active for ``-d:nimLeanLib`` or similar, code bloat on embedded systems is a real problem.

zevv:
> Ok, but a range like '%'..'$' is not understandable. You need even more specialization so that only 0-9, a-z, A-Z or the '\numberHere' things are affected. And this code shouldn't be active for `-d:nimLeanLib` or similar, code bloat on embedded systems is a real problem.

Well, 'repr' on a set is not something one would typically do on an embedded system, if you don't want lib bloat, simple don't call the function. Simple as that.

PR #16552, opened 2021-01-02
fix `is "closure"`

ringabout:
I want to use this feature too. It is very useful.

The feature/bug is introduced in this https://github.com/nim-lang/Nim/commit/dd9ad9e49730cec954e9113f0136053c5020aafd

At that time, Nim doesn't have closure iterator.

The docs: https://github.com/nim-lang/Nim/blob/32d0ac01dcd5cebfdd1da5182312d308da955837/doc/manual.txt


ringabout:
```nim
import std/hashes


var x = 12
iterator hello(): int {.closure.} =
  yield x

echo hash(hello)
```
Before this PR:
D:\QQPCmgr\Desktop\Nim\pair.nim(30, 10) template/generic instantiation of `hash` from here
C:\Users\blue\.choosenim\toolchains\nim-#devel\lib\pure\hashes.nim(183, 26) Error: type mismatch: got 'iterator (): int{.closure, gcsafe, locks: 0.}' for 'x' but expected 'pointer'

After this PR:
compiles and print the hash value

PR #16570, opened 2021-01-03
stdlib/linux: Add unshare

rominf:
Added unshare function to linux module with a simple test with comments.

Depends on #16573.

### DIFF
--- a/lib/posix/linux.nim
+++ b/lib/posix/linux.nim
@@ -37,3 +37,5 @@ proc clone*(fn: pointer; child_stack: pointer; flags: cint;
             ctid: ptr Pid): cint {.importc, header: "<sched.h>".}
 
 proc pipe2*(a: array[0..1, cint], flags: cint): cint {.importc, header: "<unistd.h>".}
+
+proc unshare*(flags: cint): cint {.importc, header: "<sched.h>", sideEffect.}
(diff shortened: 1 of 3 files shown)


lib/posix/linux.nim
```diff
@@ -37,3 +37,59 @@ proc clone*(fn: pointer; child_stack: pointer; flags: cint;
...
+    from posix import fork, getgid, getuid, waitpid
+    from posix_utils import uname
+    import oids
+    import os
+    import osproc
+    import strformat
+    import strutils
+
+    proc isSupportedByKernel: bool =
+      let kernelConfigFilePath = "/boot/config-" & uname().release
```
alaviss: This test depends on the kernel configuration being installed in `/boot/`. Many distributions omit this, which makes it not useful as an example.

Please put tests within the `tests` folder. Please see our [contributing guidelines](https://github.com/nim-lang/Nim/blob/devel/doc/contributing.rst#stdlib) for more information.

alaviss (changes requested):

lib/posix/linux.nim
rominf: OK, I've put this into tests.

rominf:
I guess it's out of scope, closing.

PR #16588, opened 2021-01-04
Improve documentation for complex

konsumlamm:
* Add missing doc comments
* Fix grammar issues
* Replace double with single backticks
* Minor formatting

I wanted to add an example to the module description, but I'm not sure how to do it, since we rather shouldn't use `==` or `echo`. Maybe some examples with a custom "approximately equal" operator? On the other hand, it's probably obvious enough how to use the procs...

### DIFF
--- a/lib/pure/complex.nim
+++ b/lib/pure/complex.nim
@@ -12,6 +12,27 @@
 ##
 ## Complex numbers are currently generic over 64-bit or 32-bit floats.
 
+runnableExamples:
+  from math import almostEqual, sqrt
+
+  func almostEqual(a, b: Complex): bool =
+    almostEqual(a.re, b.re) and almostEqual(a.im, b.im)
+
+  let
+    z1 = complex(1.0, 2.0)
+    z2 = complex(3.0, -4.0)
+
+  assert almostEqual(z1 + z2, complex(4.0, -2.0))
+  assert almostEqual(z1 - z2, complex(-2.0, 6.0))
+  assert almostEqual(z1 * z2, complex(11.0, 2.0))
+  assert almostEqual(z1 / z2, complex(-0.2, 0.4))
+
+  assert almostEqual(abs(z1), sqrt(5.0))
+  assert almostEqual(conjugate(z1), complex(1.0, -2.0))
+
+  let (r, phi) = z1.polar
+  assert almostEqual(rect(r, phi), z1)
+
 {.push checks: off, line_dir: off, stack_trace: off, debugger: off.}
 # the user does not want to trace a part of the standard library!
 
@@ -27,20 +48,17 @@ type
     ## Alias for a complex number using 32-bit floats.
 
 func complex*[T: SomeFloat](re: T; im: T = 0.0): Complex[T] =
-  ## Returns a `Complex[T]` with real part `re` and imaginary part `im`
-  ## (default: 0).
+  ## Returns a `Complex[T]` with real part `re` and imaginary part `im`.
   result.re = re
   result.im = im
 
 func complex32*(re: float32; im: float32 = 0.0): Complex32 =
-  ## Returns a `Complex32` with real part `re` and imaginary part `im`
-  ## (default: 0).
+  ## Returns a `Complex32` with real part `re` and imaginary part `im`.
   result.re = re
   result.im = im
 
 func complex64*(re: float64; im: float64 = 0.0): Complex64 =
-  ## Returns a `Complex64` with real part `re` and imaginary part `im`
-  ## (default: 0).
+  ## Returns a `Complex64` with real part `re` and imaginary part `im`.
   result.re = re
   result.im = im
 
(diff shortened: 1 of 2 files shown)


lib/pure/complex.nim
```diff
@@ -17,96 +19,124 @@ import math
...
   Complex64* = Complex[float64]
-    ## Alias for a pair of 64-bit floats.
+    ## Alias for a complex number using 64-bit floats.
   Complex32* = Complex[float32]
-    ## Alias for a pair of 32-bit floats.
+    ## Alias for a complex number using 32-bit floats.
 
 func complex*[T: SomeFloat](re: T; im: T = 0.0): Complex[T] =
+  ## Returns a `Complex[T]` with real part `re` and imaginary part `im`
+  ## (default: 0).
```
timotheecour: `## (default: 0).` is redundant; docs should only complement what's not implied in the declaration otherwise it's just noise

image

lib/pure/complex.nim
```diff
@@ -17,96 +19,124 @@ import math
...
 
 func conjugate*[T](z: Complex[T]): Complex[T] =
-  ## Conjugates of complex number ``z``.
+  ## Returns the complex conjugate of `z` (`complex(z.re, -z.im)`).
   result.re = z.re
   result.im = -z.im
 
 func inv*[T](z: Complex[T]): Complex[T] =
-  ## Multiplicatives inverse of complex number ``z``.
+  ## Returns the multiplicative inverse of `z` (`1/z`).
```
timotheecour, line 91: this is where a doc comment (or runnableExamples) would be the most valuable; specifying the behavior for edge cases eg when `abs2(z) == 0`

lib/pure/complex.nim
```diff
@@ -17,96 +19,124 @@ import math
...
-func `-` *[T](z: Complex[T]): Complex[T] =
+func `-`*[T](z: Complex[T]): Complex[T] =
   ## Unary minus for complex numbers.
   result.re = -z.re
   result.im = -z.im
 
-func `-` *[T](x: T; y: Complex[T]): Complex[T] =
+func `-`*[T](x: T; y: Complex[T]): Complex[T] =
   ## Subtracts a complex number from a real number.
   x + (-y)
```
timotheecour: pre-existing but could we make this more efficient, avoiding a temporary? (could also be done in another PR)

lib/pure/complex.nim
```diff
@@ -180,28 +195,34 @@ func sqrt*[T](z: Complex[T]): Complex[T] =
       result.re = z.im / (result.im + result.im)
 
 func exp*[T](z: Complex[T]): Complex[T] =
-  ## ``e`` raised to the power ``z``.
+  ## Computes the exponential function (`e^z`).
   var
```
timotheecour: pre-existing but: let

lib/pure/complex.nim
```diff
@@ -180,28 +195,34 @@ func sqrt*[T](z: Complex[T]): Complex[T] =
...
     rho = exp(z.re)
     theta = z.im
   result.re = rho * cos(theta)
   result.im = rho * sin(theta)
 
 func ln*[T](z: Complex[T]): Complex[T] =
-  ## Returns the natural log of ``z``.
+  ## Returns the (principal value of the) natural logarithm of `z`.
```
timotheecour: how about link to https://en.wikipedia.org/wiki/Principal_value

lib/pure/complex.nim
```diff
@@ -223,126 +244,137 @@ func pow*[T](x, y: Complex[T]): Complex[T] =
...
-  ## | ``result.re = r * cos(phi)``
-  ## | ``result.im = r * sin(phi)``
+  ## **See also:**
+  ## * `polar func<#polar,Complex[T]>`_ for the inverse operation
   complex(r * cos(phi), r * sin(phi))
 
 
 func `$`*(z: Complex): string =
-  ## Returns ``z``'s string representation as ``"(re, im)"``.
+  ## Returns `z`'s string representation as `"(re, im)"`.
```
timotheecour, line 400: a runnableExamples would be better than this description

timotheecour (approved):
LGTM, all comments can be addressed later;

> I'm not sure how to do it, since we rather shouldn't use == or echo

a single top-level runnableExamples would be nice; it could defined `proc almostEqual(a, b: Complex): bool = almostEqual(a.re, b.re) and almostEqual(a.im, b.im)` if need be

lib/pure/complex.nim
konsumlamm: I decided to leave that to a later PR.

lib/pure/complex.nim
```diff
@@ -7,8 +7,31 @@
...
 
-## This module implements complex numbers.
-## Complex numbers are currently implemented as generic on a 64-bit or 32-bit float.
+## This module implements complex numbers
+## and basic mathematical operations on them.
+##
+## Complex numbers are currently generic over 64-bit or 32-bit floats.
+
+runnableExamples:
+  from math import almostEqual, sqrt
```
ringabout:
```suggestion
  from std/math import almostEqual, sqrt
```

konsumlamm: Should I also change the imports in the module itself and the tests or would that cause problems?

ringabout: I don't think that would cause problems(you could give a try :). `import std/math` is prefered to `import math`. I don't care about the old parts. But I think the newer additions should use `std` namespace.

PR #16592, opened 2021-01-05
add math.signbit

ringabout:
C99 function

Ref https://github.com/timotheecour/Nim/issues/505

### DIFF
--- a/lib/pure/math.nim
+++ b/lib/pure/math.nim
@@ -56,7 +56,7 @@ import std/private/since
 {.push debugger: off.} # the user does not want to trace a part
                        # of the standard library!
 
-import bitops, fenv
+import std/[bitops, fenv]
 
 when defined(c) or defined(cpp):
   proc c_isnan(x: float): bool {.importc: "isnan", header: "<math.h>".}
@@ -64,6 +64,8 @@ when defined(c) or defined(cpp):
 
   proc c_copysign(x, y: cfloat): cfloat {.importc: "copysignf", header: "<math.h>".}
   proc c_copysign(x, y: cdouble): cdouble {.importc: "copysign", header: "<math.h>".}
+
+  proc c_signbit(x: SomeFloat): cint {.importc: "signbit", header: "<math.h>".}
 
 func binom*(n, k: int): int =
   ## Computes the `binomial coefficient <https://en.wikipedia.org/wiki/Binomial_coefficient>`_.
@@ -155,6 +157,29 @@ func isNaN*(x: SomeFloat): bool {.inline, since: (1,5,1).} =
   else:
     when defined(js): fn()
     else: result = c_isnan(x)
+
+when defined(js):
+  proc toBitsImpl(x: float): array[2, uint32] =
+    asm """
+    const buffer = new ArrayBuffer(8);
+    const floatBuffer = new Float64Array(buffer);
+    const uintBuffer = new Uint32Array(buffer);
+    floatBuffer[0] = `x`;
+    `result` = uintBuffer
+    """
+
+proc signbit*(x: SomeFloat): bool {.inline, since: (1, 5, 1).} =
+  ## Returns true if `x` is negative, false otherwise.
+  runnableExamples:
+    doAssert not signbit(0.0)
+    doAssert signbit(-0.0)
+    doAssert signbit(-0.1)
+    doAssert not signbit(0.1)
+  when defined(js):
+    let uintBuffer = toBitsImpl(x)
+    result = (uintBuffer[1] shr 31) != 0
+  else:
+    result = c_signbit(x) != 0
 
 func copySign*[T: SomeFloat](x, y: T): T {.inline, since: (1, 5, 1).} =
   ## Returns a value with the magnitude of `x` and the sign of `y`;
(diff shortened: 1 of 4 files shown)


lib/pure/math.nim
```diff
@@ -65,6 +65,8 @@ when defined(c) or defined(cpp):
   proc c_copysign(x, y: cfloat): cfloat {.importc: "copysignf", header: "<math.h>".}
   proc c_copysign(x, y: cdouble): cdouble {.importc: "copysign", header: "<math.h>".}
 
+  proc c_signbit(x: SomeFloat): cint {.importc: "signbit", header: "<math.h>".}
```
timotheecour, line 68: Is signbit always implemented as a macro in c? Otherwise we may need non generics here

tests/stdlib/tmath.nim
```diff
@@ -304,6 +304,12 @@ block:
 
 template main =
   # xxx wrap all under `main` so it also gets tested in vm.
+  block: # signbit
```
timotheecour, line 307: Needs inf, nan tests

lib/pure/math.nim
```diff
@@ -156,6 +158,29 @@ func isNaN*(x: SomeFloat): bool {.inline, since: (1,5,1).} =
...
+  runnableExamples:
+    doAssert not signbit(0.0)
+    doAssert signbit(-0.0)
+    doAssert signbit(-0.1)
+    doAssert not signbit(0.1)
+  when defined(js):
+    let uintBuffer = toBitsImpl(x)
+    result = (uintBuffer[1] shr 31) != 0
+  else:
+    result = c_signbit(x) != 0
```
timotheecour, line 182: Would cast[bool] work to avoid branching?

Araq: I'm pretty sure that there is no branch involved for the code as it is written for any backend that we care about.

timotheecour: the codegen and asm is different, `c_signbit(x) != 0` seems to generate an extra instruction:

image

but the asm seems the same in -d:danger, so i guess it's fine

lib/pure/math.nim
ringabout: C99 standard
```
7.12.3.6 The signbit macro
Synopsis
1 #include <math.h>
int signbit(real-floating x);
Description
2 The signbit macro determines whether the sign of its argument value is negative.198)
Returns
3 The signbit macro returns a nonzero value if and only if the sign of its argument value
is negative.
```

timotheecour: and C++11 also
C11 standard (ISO/IEC 9899:2011):
https://en.cppreference.com/w/c/numeric/math/signbit

ok

tests/stdlib/tmath.nim
```diff
@@ -310,6 +310,11 @@ template main =
     doAssert signbit(-0.1)
     doAssert not signbit(0.1)
 
+    doAssert not signbit(Inf)
+    doAssert signbit(-Inf)
+    doAssert not signbit(NaN)
+    doAssert not signbit(-NaN)
```
timotheecour: that's a because of a bug in `-` that we'll eventually have to fix, refs https://github.com/timotheecour/Nim/issues/499

not a blocker though, we'll just change this assert (and similar ones in this file)

timotheecour: and indeed, CI fails:
```
2021-01-06T02:17:10.5225810Z ../../lib/system/fatal.nim(53, 5) Error: unhandled exception: tmath.nim(316, 5) `not signbit(-NaN)`  [AssertionDefect]
```
=> just remove this line, it's wrong (NaN's have a sign bit too)

xflywind: It works in C/JS backend?

tests/stdlib/tmath.nim
```diff
@@ -304,6 +304,17 @@ block:
...
+  block: # signbit
+    doAssert not signbit(0.0)
+    doAssert signbit(-0.0)
+    doAssert signbit(-0.1)
+    doAssert not signbit(0.1)
+
+    doAssert not signbit(Inf)
+    doAssert signbit(-Inf)
+    doAssert not signbit(NaN)
+    doAssert not signbit(-NaN)
```
ringabout:
```suggestion
```

timotheecour (approved):

Araq:
We target C89 though. Also, why do we need it? One can always cast to uint64 and check the sign bit.

timotheecour:
> We target C89 though

plenty of code in stdlib use functions that require >= C99, see https://en.wikibooks.org/wiki/C_Programming/math.h, so there's no reason to single this one out; future work can improve general support for C89 (via `when not defined(nimc99)` maybe) but this shouldn't be at the expense of performance for the vast majority of use cases that can assume >= C99.

> One can always cast to uint64 and check the sign bit.

that wouldn't be portable, it'd depend on the FP type, endian-ness, architecture; note that FP endian-ness is not simple, eg see https://en.wikipedia.org/wiki/Endianness

> Although the ubiquitous x86 processors of today use little-endian storage for all types of data (integer, floating point), there are a number of hardware architectures where floating-point numbers are represented in big-endian form while integers are represented in little-endian form
> here are ARM processors that have half little-endian, half big-endian floating-point representation for double-precision numbers: both 32-bit words are stored in little-endian like integer registers, but the most significant one first

etc...

ringabout:
It is useful for making copysign work robust in JS backend.
https://github.com/nim-lang/Nim/pull/16609

It is also useful for making nan consistent in CT and RT

https://github.com/timotheecour/Nim/issues/499

PR #16594, opened 2021-01-05
fix #16516 method dispatch for sink args

### LINKED ISSUE #16516: Sink parameters break method resolution
If methods use sink parameters then they appear to not be able to dynamically resolve the types of their arguments. See the example where in `main`, variable `b` is declared as type `A` (the root of type `B`), and is treated like an `A` in `method foo`.

Note: The `lst` parameter may not be necessary to get a minimal example. However; if you remove the keywords `sink` from the arguments, but then compile with `--gc:arc --sinkInference:on`, then adding to `lst` triggers sink inference and should yield you the exact same problem.

### Example
```nim
type
  A = ref object of RootObj
    x: int

  B = ref object of A

method foo(v: sink A, lst: var seq[A]) {.base,locks:0.} =
  echo "type A"
  lst.add v

method foo(v: sink B, lst: var seq[A]) =
  echo "type B"
  lst.add v

proc main() =
  let
    a = A(x: 5)
    b: A = B(x: 5)

  var lst: seq[A]

  foo(a, lst)
  foo(b, lst)

main()
```

### Current Output

```
type A
type A
```

### Expected Output
```
type A
type B
```

### Additional Information

Compiled with 
```
nim c -r sinkmethod.nim
```

Issue is present in 1.4.2, checked with latest devel release (version dump below) and verified it still exists.

This has impact beyond simply not being able to use sink parameters in methods. A Nim source file that has methods similar to the example above cannot participate in sink inference using the new ARC garbage collector at all. The sink inference mechanism will rightly decide that `v` should be a sink parameter, but then the methods will no longer be considered related. The compiler does give an appropriate warning in this case.

```
$ nim --version
Compiled at 2020-12-22

git hash: 297c8e403d110dd872e070563328f4e0c734cd01
```

cooldome:
fix #16516 method dispatch for sink args

### DIFF
--- a/compiler/cgmeth.nim
+++ b/compiler/cgmeth.nim
@@ -65,8 +65,8 @@ proc sameMethodBucket(a, b: PSym; multiMethods: bool): MethodResult =
     var aa = a.typ[i]
     var bb = b.typ[i]
     while true:
-      aa = skipTypes(aa, {tyGenericInst, tyAlias})
-      bb = skipTypes(bb, {tyGenericInst, tyAlias})
+      aa = skipTypes(aa, {tyGenericInst, tyAlias, tySink})
+      bb = skipTypes(bb, {tyGenericInst, tyAlias, tySink})
       if aa.kind == bb.kind and aa.kind in {tyVar, tyPtr, tyRef, tyLent}:
         aa = aa.lastSon
         bb = bb.lastSon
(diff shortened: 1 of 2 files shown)


Araq:
I think you need to watch out that the methods agree on the "sink" parameters.

cooldome:
Agree, fixed

PR #16599, opened 2021-01-05
Fixes the asynchttpserver example some more

dom96:
I dislike this example a lot (busy looping for FDs to be closed is a very
poor waste of resources) but at least with these changes it's a little bit
better.

### DIFF
--- a/lib/pure/asynchttpserver.nim
+++ b/lib/pure/asynchttpserver.nim
@@ -18,29 +18,28 @@ runnableExamples:
   # This example will create an HTTP server on port 8080. The server will
   # respond to all requests with a `200 OK` response code and "Hello World"
   # as the response body. Run locally with:
-  # `nim doc --doccmd:-d:nimAsynchttpserverEnableTest --lib:lib lib/pure/asynchttpserver.nim`
+  # `nim doc --doccmd:-d:nimAsyncHttpServerEnableTest --lib:lib lib/pure/asynchttpserver.nim`
   import asyncdispatch
-  if defined(nimAsynchttpserverEnableTest):
+  if defined(nimAsyncHttpServerEnableTest):
     proc main {.async.} =
       const port = 8080
       var server = newAsyncHttpServer()
       proc cb(req: Request) {.async.} =
         echo (req.reqMethod, req.url, req.headers)
-        let headers = {"Date": "Tue, 29 Apr 2014 23:40:08 GMT",
-            "Content-type": "text/plain; charset=utf-8"}
+        let headers = {"Content-type": "text/plain; charset=utf-8"}
         await req.respond(Http200, "Hello World", headers.newHttpHeaders())
 
       echo "test this with: curl localhost:" & $port & "/"
-      server.listen Port(port)
+      server.listen(Port(port))
       while true:
         if server.shouldAcceptRequest():
           await server.acceptRequest(cb)
         else:
           # too many concurrent connections, `maxFDs` exceeded
-          poll()
+          # wait 500ms for FDs to be closed
+          await sleepAsync(500)
 
-    asyncCheck main()
-    runForever()
+    waitFor main()
 
 import asyncnet, asyncdispatch, parseutils, uri, strutils
 import httpcore


lib/pure/asynchttpserver.nim
```diff
@@ -18,29 +18,28 @@ runnableExamples:
...
-      server.listen Port(port)
+      server.listen(Port(port))
       while true:
         if server.shouldAcceptRequest():
           await server.acceptRequest(cb)
         else:
           # too many concurrent connections, `maxFDs` exceeded
-          poll()
+          # wait 500ms for FDs to be closed
+          await sleepAsync(500)
```
timotheecour, line 40: is there a way to wait until a FD is closed to avoid a forced wait?

timotheecour:
> busy looping for FDs to be closed

why is that busy looping? shouldn't poll only return when an event happened?

in any case, see https://github.com/nim-lang/Nim/issues/16603 which shows that there's a bigger problem, `shouldAcceptRequest` doesn't seem to work

dom96:
> why is that busy looping? shouldn't poll only return when an event happened?

Sure, but just because an event happened doesn't mean an FD was freed.

> in any case, see #16603 which shows that there's a bigger problem, shouldAcceptRequest doesn't seem to work

I'll let @Araq take a look. He should be the best to resolve this :)

lib/pure/asynchttpserver.nim
dom96: That is the ideal indeed. There is no way to do this out of the box, you could potentially get a good approximation if you add some code into each `close` proc that would then complete a `Future` to let code awaiting that future know that the FDs have been freed. But that assumes all your FDs are created within Nim, if you call into C you're screwed.

More and more, I'm having a hard time keeping an open mind to this "improvement". I'm curious whether other languages implement something similar. I've never seen somebody work around FD limits like this so this really smells like a bad idea to me still, especially now that it's needing so many fixes.

dom96:
In fact, thinking about this some more. I'm thinking this should be resolved at the `accept` call-site. The code should detect that the FD limit is reached and just loop again.

You even get an [error code for this](https://man7.org/linux/man-pages/man2/accept.2.html)...

image

Araq:
Making ``accept`` more robust is fine with me but my protection would still be useful on top of that.

dom96:
> Making accept more robust is fine with me but my protection would still be useful on top of that.

How so? This `shouldAcceptRequest` mechanism could and should be replaced completely by this unless I'm missing something. Also it's not about making `accept` more robust, it's about doing better error handling in asynchttpserver specifically.

Araq:
What if you know the request opens 4 files etc, then you can prevent ``accept`` from accepting it anyway. Just because accept found a single FD to do its work doesn't mean the request handler can accept the request.

dom96:
I would rather have the request handler fail too. It would even compose well with other error handling. You can easily handle these errors in the request handler too.

In addition, imagine you have other threads, now you'll need a lock around your call to `shouldAcceptRequest` because other threads might have opened more FDs.

lib/pure/asynchttpserver.nim
aboisvert: I know this PR has already been merged but I figured I'd share my opinion (two cents) regardless.

IMO, the real pitfall here is trying to cope with too few FDs *silently*.  If a system runs of out FDs (or any resource for that matter), it should somehow tell the administrator that it's operating beyond its capabilities and is no longer able to fulfill requests.  (i.e. a simple log statement would help)

It's great that we're making httpserver more resilient, but in almost all occurrences this is still an undesirable situation and so should probably be communicated.   

EDIT:  Ooops I didn't realize this change was for the `runnableExamples` section, but still my comment applies to the [`serve()`](https://github.com/nim-lang/Nim/blob/4fd444152a28839c97cff942ac28199c6b32634f/lib/pure/asynchttpserver.nim#L341) proc declared in `asynchttpserver.nim`, which still uses the [`poll()`](https://github.com/nim-lang/Nim/blob/4fd444152a28839c97cff942ac28199c6b32634f/lib/pure/asynchttpserver.nim#L363) method.   It seems these two should be consistent and include some logging for the out-of-FD situation.  If it helps, LMK if you'd like me to submit a PR.

dom96: If you're interested in making a PR then I would encourage you to implement what I described below (https://github.com/nim-lang/Nim/pull/16599#issuecomment-755772989). That is unless you disagree with it. Personally I don't think we should expose this mechanism in the way that we do at all. Asynchttpserver should just call `accept` and handle the errors it receives appropriately, and yes it should log when that happens.

timotheecour: potential solution: https://github.com/timotheecour/Nim/pull/750

PR #16611, opened 2021-01-06
refactor cmpNimIdentifier

ringabout:
I think some modules uses `cmpIgnoreStyle` instead of `cmpNimIdentifier` which seems wrong.

Though I don't change them in this PR.

### DIFF
--- a/lib/core/macros.nim
+++ b/lib/core/macros.nim
@@ -1390,7 +1390,6 @@ when defined(nimVmEqIdent):
 
 else:
   from std/private/strimpl import cmpNimIdentifier
-  # this procedure is optimized for native code, it should not be compiled to nimVM bytecode.
 
   proc eqIdent*(a, b: string): bool = cmpNimIdentifier(a, b) == 0
     ## Check if two idents are equal.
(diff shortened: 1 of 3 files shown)


lib/core/macros.nim
```diff
@@ -1389,12 +1389,10 @@ when defined(nimVmEqIdent):
     ## these nodes will be unwrapped.
 
 else:
-  from std/private/strimpl import cmpIgnoreStyleImpl
+  from std/private/strimpl import cmpNimIdentifier
   # this procedure is optimized for native code, it should not be compiled to nimVM bytecode.
```
Clyybber:
```suggestion
```

lib/std/private/strimpl.nim
```diff
@@ -70,3 +70,7 @@ template endsWithImpl*[T: string | cstring](s, suffix: T) =
     if s[i+j] != suffix[i]: return false
     inc(i)
   if i >= suffixLen: return true
+
+
+func cmpNimIdentifier*[T: string | cstring](a, b: T): int {.noSideEffect.} =
```
Clyybber:
```suggestion
func cmpNimIdentifier*[T: string | cstring](a, b: T): int =
```

timotheecour (approved):

PR #16614, opened 2021-01-06
RST: implement internal targets

a-mr:
Currently rst.nim supports only
1. automatic generating anchors for sections
2. automatic generating anchors for procs and other Nim code in HTML

This PR implements ability to **manually** mark any RST body elements and text as anchors. 

Both ["outline" internal hyperlink targets](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#hyperlink-targets) and [inline internal targets](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#inline-internal-targets) are supported according to the spec.

Examples of "outline" targets:
``` RST
.. _`concept definition`:

Paragraph.

.. _`my code example`:

.. code:: nim
   const pi = 3.14
```

Those targets can mark any top-level body elements, not only paragraphs and code, but also images, block quotes, line blocks, etc.

To anchor something inside a top level element, use inline targets, e.g. to make an anchor inside a paragraph:
``` RST
A _`concept definition` is a thing, that...
```

They can be referenced with the standard RST link syntax:

``` RST
Ref. `concept definition`_. See `my code example`_.
```

CC @narimiran .

### DIFF
--- a/lib/packages/docutils/rst.nim
+++ b/lib/packages/docutils/rst.nim
@@ -677,8 +677,8 @@ proc isInlineMarkupStart(p: RstParser, markup: string, twoTokens=false): bool =
   if not twoTokens:
     result = currentTok(p).symbol == markup
   else:
-    result = currentTok(p).symbol == markup[0..0] and
-               nextTok(p).symbol == markup[1..1]
+    result = currentTok(p).symbol == $markup[0] and
+               nextTok(p).symbol == $markup[1]
   if not result: return
   # Rule 6:
   result = p.idx == 0 or prevTok(p).kind in {tkIndent, tkWhite} or
--- a/lib/packages/docutils/rst.nim
+++ b/lib/packages/docutils/rst.nim
@@ -7,18 +7,28 @@
 #    distribution, for details about the copyright.
 #
 
+## ==================================
+## rst: Nim-flavored reStructuredText
+## ==================================
+##
 ## This module implements a `reStructuredText`:idx: (RST) parser.
 ## A large subset is implemented with some limitations_ and
 ## `Nim-specific features`_.
-## A few features_ of the `markdown`:idx: syntax are also supported.
-## Nim can output the result to HTML (commands ``nim doc`` for \*.nim files and
-## ``nim rst2html`` for \*.rst files) or Latex (command ``nim rst2tex``).
+## A few `extra features`_ of the `Markdown`:idx: syntax are
+## also supported.
+##
+## Nim can output the result to HTML (commands ``nim doc`` for
+## ``*.nim`` files and ``nim rst2html`` for ``*.rst`` files) or
+## Latex (command ``nim rst2tex`` for ``*.rst``).
 ##
 ## If you are new to RST please consider reading the following:
 ##
 ## 1) a short `quick introduction`_
 ## 2) an `RST reference`_: a comprehensive cheatsheet for RST
 ## 3) a more formal 50-page `RST specification`_.
+##
+## Features
+## --------
 ##
 ## Supported standard RST features:
 ##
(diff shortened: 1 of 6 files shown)


lib/packages/docutils/rst.nim
```diff
@@ -626,10 +671,14 @@ proc isInlineMarkupEnd(p: RstParser, markup: string): bool =
...
 
-proc isInlineMarkupStart(p: RstParser, markup: string): bool =
+proc isInlineMarkupStart(p: RstParser, markup: string, twoTokens=false): bool =
   # rst rules: https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#inline-markup-recognition-rules
   var d: char
-  result = currentTok(p).symbol == markup
+  if not twoTokens:
+    result = currentTok(p).symbol == markup
+  else:
+    result = currentTok(p).symbol == markup[0..0] and
```
narimiran: Why these strange-looking ranges here?

a-mr: it was a way to convert char->string. Changed to more conventional style.

lib/packages/docutils/rst.nim
```diff
@@ -626,10 +671,14 @@ proc isInlineMarkupEnd(p: RstParser, markup: string): bool =
     if markup != "``" and prevTok(p).symbol == "\\":
       result = false
 
-proc isInlineMarkupStart(p: RstParser, markup: string): bool =
+proc isInlineMarkupStart(p: RstParser, markup: string, twoTokens=false): bool =
```
narimiran: Can't we write the condition of this function as:

```nim
if markup.len == 1:
  # (...)
```

and then we don't need to pass the extra parameter `twoTokens` to it?

lib/packages/docutils/rst.nim
```diff
@@ -577,6 +591,37 @@ proc findRef(p: var RstParser, key: string): PRstNode =
...
+      break
+    var toLeave = false
+    for anchor in subst.aliases:
+      if anchor == refn:  # this anchor will be named as mainAnchor
+        result = subst.mainAnchor
+        toLeave = true
+    if toLeave:
+      break
+
+proc newRstNodeA(kind: RstNodeKind, p: var RstParser): PRstNode =
```
narimiran: In all other procs where it is one of their arguments, `p: var RstParser` is usually the first argument. Is there a reason why here should be in the last place?

lib/packages/docutils/rst.nim
```diff
@@ -577,6 +591,37 @@ proc findRef(p: var RstParser, key: string): PRstNode =
     if key == p.s.refs[i].key:
       return p.s.refs[i].value
 
+proc updateAnchors(p: var RstParser, refn: string, reset: bool) =
```
narimiran: Minor: this updates just the *last* anchor, maybe the name should reflect that?

lib/packages/docutils/rst.nim
```diff
@@ -43,13 +45,17 @@
...
+##     internal links (inline and outline)
 ##   + \`interpreted text\` with roles ``:literal:``, ``:strong:``,
 ##     ``emphasis``, ``:sub:``/``:subscript:``, ``:sup:``/``:supscript:``
 ##     (see `RST roles list`_ for description).
+##   + inline internal targets
 ##
-## Additional features:
+## Additional features of rst.nim:
+##
+## .. _`Nim-specific features`:
```
narimiran, line 66: This should be above the line `Additional features of rst.nim:`, otherwise when you click on the link, you don't see the "headline"

lib/packages/docutils/rst.nim
```diff
@@ -75,6 +83,8 @@
 ##
 ## Limitations:
 ##
+## .. _limitations:
```
narimiran: Same as above. (Change the order)

lib/packages/docutils/rst.nim
```diff
@@ -59,6 +65,8 @@
 ## Optional additional features, turned on by ``options: RstParseOption`` in
 ## `rstParse proc <#rstParse,string,string,int,int,bool,RstParseOptions,FindFileHandler,MsgHandler>`_:
 ##
+## .. _features:
```
narimiran: Same as above. (Change the order)

narimiran (commented):
Some minor things to fix, otherwise it looks good!

lib/packages/docutils/rst.nim
a-mr: not really, because other passed markup arguments include 1,2,3 characters, e.g. *, **, ***. But because this _ \` is an only case of 2 tokens, it's reasonable to get rid of `twoTokens` by hard-coding condition `markup == ..`

lib/packages/docutils/rst.nim
a-mr: agree with your comment, changed everywhere.

lib/packages/docutils/rst.nim
a-mr: changed to `addAnchor`, hopefully it's better.

lib/packages/docutils/rst.nim
a-mr: changed it here and in other places.

PR #16618, opened 2021-01-06
DELETE requests should always have a content-length header

jyapayne:
Not having DELETE in this list is causing hanging when trying to close webdriver sessions in [halonium](https://github.com/halonium/halonium/issues/10) and likely any other implementation of the webdriver protocol. Both at least chromedriver and geckodriver are affected by this issue.

### DIFF
--- a/lib/pure/httpclient.nim
+++ b/lib/pure/httpclient.nim
@@ -980,7 +980,7 @@ proc requestAux(client: HttpClient | AsyncHttpClient, url, httpMethod: string,
   var data: seq[string]
   if multipart != nil and multipart.content.len > 0:
     data = await client.format(multipart)
-  elif httpMethod in ["POST", "PATCH", "PUT"] or body.len != 0:
+  elif httpMethod in ["POST", "PATCH", "PUT", "DELETE"] or body.len != 0:
     client.headers["Content-Length"] = $body.len
 
   when client is AsyncHttpClient:
--- a/lib/pure/httpclient.nim
+++ b/lib/pure/httpclient.nim
@@ -980,8 +980,11 @@ proc requestAux(client: HttpClient | AsyncHttpClient, url, httpMethod: string,
   var data: seq[string]
   if multipart != nil and multipart.content.len > 0:
     data = await client.format(multipart)
-  elif httpMethod in ["POST", "PATCH", "PUT", "DELETE"] or body.len != 0:
-    client.headers["Content-Length"] = $body.len
+  else:
+    if body.len != 0:
+      client.headers["Content-Length"] = $body.len
+    elif httpMethod notin ["GET", "HEAD"] and not client.headers.hasKey("Content-Length"):
+      client.headers["Content-Length"] = "0"
 
   when client is AsyncHttpClient:
     if not client.parseBodyFut.isNil:


dom96 (commented):
I did some research on this and the RFCs aren't very clear about whether the `Content-Length` is necessary for the "DELETE" method. I did find the following though.

RFC7230 states the following:

> A user agent SHOULD send a Content-Length in a request message when
   no Transfer-Encoding is sent and the request method defines a meaning
   for an enclosed payload body.  For example, a Content-Length header
   field is normally sent in a POST request even when the value is 0
   (indicating an empty payload body).  A user agent SHOULD NOT send a
   Content-Length header field when the request message does not contain
   a payload body and the method semantics do not anticipate such a
   body.

Then RFC7231 states the following:

> A payload within a DELETE request message has no defined semantics;
   sending a payload body on a DELETE request might cause some existing
   implementations to reject the request.

Which implies to me that the "Content-Length" is not expected for "DELETE" requests.

That said, it does appear that other HTTP libraries do send it, at the same time though the Stack Overflow question asks how to remove it: https://stackoverflow.com/questions/41351657/how-to-do-a-http-delete-request-with-python-requests-module-without-content-len.

Since we do not currently have a way to remove this header (do correct me if I'm wrong here) and it does appear to have caused problems for some, but do have a way to add it without changing the stdlib code I would lean towards doing so and thus not merging this. But maybe I am missing something in the spec, if you can find something more concrete that would be helpful.

jyapayne:
@dom96 thanks for the research and the links. Just for reference, the Python requests library does [this](https://github.com/psf/requests/blob/master/requests/models.py#L532),

```python
   def prepare_content_length(self, body):
        """Prepare Content-Length header based on request method and body"""
        if body is not None:
            length = super_len(body)
            if length:
                # If length exists, set it. Otherwise, we fallback
                # to Transfer-Encoding: chunked.
                self.headers['Content-Length'] = builtin_str(length)
        elif self.method not in ('GET', 'HEAD') and self.headers.get('Content-Length') is None:
            # Set Content-Length to 0 for methods that can have a body
            # but don't provide one. (i.e. not GET or HEAD)
            self.headers['Content-Length'] = '0'
```

which leads me to believe it is required by a lot of endpoints (even if it's not in the spec). Setting the `Content-Length` to `0` is the only way that the Delete request is accepted by the webdriver. And there are some repos with [issues](https://github.com/node-fetch/node-fetch/issues/101) as well. 

Looking at [this rfc](https://tools.ietf.org/html/rfc2616#page-52), there is a section that states:


>    For compatibility with HTTP/1.0 applications, HTTP/1.1 requests
>    containing a message-body MUST include a valid Content-Length header
>    field unless the server is known to be HTTP/1.1 compliant. If a
>    request contains a message-body and a Content-Length is not given,
>    the server SHOULD respond with 400 (bad request) if it cannot
>    determine the length of the message, or with 411 (length required) if
>    it wishes to insist on receiving a valid Content-Length.

So, I'm assuming that for HTTP/1.0, there will be issues if there is no `Content-Length` header.

EDIT: Looks like I misread the above quote. I thought it said "messages not containing a message-body". However, there has to be a reason the requests module is setting the content-length to 0.

Maybe it would be best to change the code to be more like the Python requests version?

jyapayne:
Hmm, interpreting the spec a bit more:

>   A payload within a DELETE request message has no defined semantics;
>   sending a payload body on a DELETE request might cause some existing
>   implementations to reject the request.

BUT, that means that a payload **is** supported, it just doesn't have any default defined behavior for the server. It is the duty of the server to respond with whatever the application design permits.

And then, since a body is _supported_, this line from the [HTTP/1.0 spec states that](https://www.w3.org/Protocols/HTTP/1.0/draft-ietf-http-spec.html#Entity-Body):

> For response messages, whether or not an entity body is included with a message is dependent on both the request method and the response code. All responses to the HEAD request method must not include a body, even though the presence of entity header fields may lead one to believe they do. All 1xx (informational), 204 (no content), and 304 (not modified) responses must not include a body. **All other responses must include an entity body or a Content-Length header field defined with a value of zero (0).**

So I would say the Python implementation is correct and we should instead implement that.

jyapayne:
@dom96 let me know what you think of the new solution. The doc build seems to have failed for unrelated reasons. I see other PRs failed on the same error.

dom96 (approved):
Sure, makes sense.

Will merge, but could you test one thing? Verify that if an empty "Content-Length" header is added to the client headers that it does not get sent? (I have a feeling there might be a bug where it sends "Content-Length: " which would be bad).

PR #16624, opened 2021-01-07
minor fix in tests/stdlib/tbitops.nim.cfg

rockcavera:


tests/stdlib/tbitops.nim.cfg
```diff
@@ -1 +1 @@
--d:noUndefinedBitOps
+-d:noUndefinedBitOpts
```
Araq, line 1: What? This should be noUndefinedBitOps somewhere else.

rockcavera: I thought it was wrong here, considering that in the `bitops` documentation and in the codes everyone is like` noUndefinedBitOpts`.

rockcavera: So the only correct place is in this .cfg.

Because even `noIntrinsicsBitOpts` is Opts.

rockcavera: Now, which one would be correct?

- no option of bit intrinsics?
- no intrinsic bitops (bit operations)?

Araq: My mistake.

PR #16634, opened 2021-01-07
Make small text changes in the docs

elliotwaite:
Let me know if any of these suggestions are incorrect or not preferred and I can update this PR.


doc/contributing.rst
```diff
@@ -589,4 +589,4 @@ use `lib/std/collections/foo.nim`, not `lib/pure/collections/foo.nim`.
 `std/sums.nim` instead of `std/sum.nim`. In particular, this reduces chances of conflicts
 between module name and the symbols it defines. Furthermore, is should use `snake_case`
```
awr1: replace "is" with "module names"

awr1 (changes requested):

doc/contributing.rst
elliotwaite: Ah, nice catch. Not sure how I missed that. Fixed.

awr1 (approved):

PR #16636, opened 2021-01-08
Add support for Transfer-Encoding: chunked

vabresto:
I've spent the past few days trying to use Nim with OpenFaas (my first Nim project, just a proof-of-concept, see: https://gitlab.com/vabresto/nim-sudoku-faas/-/tree/master/) and had some issues with the Nim HTTP server because the OpenFaas of-watchdog proxy doesn't pass the `Content-Length` header, but rather uses chunked encoding (see: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding).

I've added support for this for my own uses and wanted to contribute it back to Nim.

I'd like to add tests but I'm not sure what the best way to do so is. Additionally, I can't actually compile and run the devel branch locally (OSX 15.6), so I made these changes based on the Nim 1.4 code I have locally.

Any feedback would be very welcome. Thanks!

### DIFF
--- a/lib/pure/asynchttpserver.nim
+++ b/lib/pure/asynchttpserver.nim
@@ -144,15 +144,14 @@ proc sendStatus(client: AsyncSocket, status: string): Future[void] =
 
 func hasChunkedEncoding(request: Request): bool = 
   ## Searches for a chunked transfer encoding
-  var isChunked = false
-  for encoding in request.headers.iter("Transfer-Encoding"):
-    if "chunked" == encoding.strip:
-      isChunked = true
-      break
+  const transferEncoding = "Transfer-Encoding"
 
-  return (request.reqMethod == HttpPost and
-          request.headers.hasKey("Transfer-Encoding") and
-          isChunked)
+  if request.headers.hasKey(transferEncoding):
+    for encoding in seq[string](request.headers[transferEncoding]):
+      if "chunked" == encoding.strip:
+        # Returns true if it is both an HttpPost and has chunked encoding
+        return request.reqMethod == HttpPost
+  return false
 
 proc processRequest(
   server: AsyncHttpServer,
--- a/lib/pure/asynchttpserver.nim
+++ b/lib/pure/asynchttpserver.nim
@@ -301,8 +301,9 @@ proc processRequest(
           break
 
         # Read bytesToRead and add to body
-        await client.recvLineInto(lineFut, maxLength = bytesToRead)
-        request.body = request.body & lineFut.mget
+        # Note we add +2 because the line must be terminated by \r\n
+        let chunk = await client.recv(bytesToRead + 2)
+        request.body = request.body & chunk
 
       inc sizeOrData
   elif request.reqMethod == HttpPost:
--- a/lib/pure/asynchttpserver.nim
+++ b/lib/pure/asynchttpserver.nim
@@ -142,6 +142,18 @@ proc parseProtocol(protocol: string): tuple[orig: string, major, minor: int] =
 proc sendStatus(client: AsyncSocket, status: string): Future[void] =
   client.send("HTTP/1.1 " & status & "\c\L\c\L")
 
+func hasChunkedEncoding(request: Request): bool = 
+  ## Searches for a chunked transfer encoding
+  var isChunked = false
+  for encoding in request.headers.iter("Transfer-Encoding"):
+    if "chunked" == encoding.strip:
+      isChunked = true
+      break
+
+  return (request.reqMethod == HttpPost and
+          request.headers.hasKey("Transfer-Encoding") and
+          isChunked)
+
 proc processRequest(
   server: AsyncHttpServer,
   req: FutureVar[Request],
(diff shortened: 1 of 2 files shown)


lib/pure/httpcore.nim
```diff
@@ -183,6 +183,13 @@ proc `[]=`*(headers: HttpHeaders, key, value: string) =
   ## Replaces any existing values.
   headers.table[headers.toCaseInsensitive(key)] = @[value]
 
+iterator iter*(headers: HttpHeaders, key: string): string =
+  ## Yields each value associated with the requested key
+  ## If no such key exists, this does nothing
+  if headers.table.contains(headers.toCaseInsensitive(key)):
+    for value in headers.table[headers.toCaseInsensitive(key)]:
+        yield value
```
ringabout:
```suggestion
      yield value
```

lib/pure/httpcore.nim
```diff
@@ -183,6 +183,13 @@ proc `[]=`*(headers: HttpHeaders, key, value: string) =
   ## Replaces any existing values.
   headers.table[headers.toCaseInsensitive(key)] = @[value]
 
+iterator iter*(headers: HttpHeaders, key: string): string =
+  ## Yields each value associated with the requested key
+  ## If no such key exists, this does nothing
```
ringabout:
```suggestion
  ## If no such key exists, this does nothing.
```

vabresto: Fixed

lib/pure/httpcore.nim
vabresto: Fixed (good catch!)

lib/pure/asynchttpserver.nim
```diff
@@ -261,6 +273,38 @@ proc processRequest(
...
+          await request.respond(Http411, ("Invalid chunked transfer encoding - " &
+                                          "chunk data size must be hex encoded"))
+          return true
+      else:
+        if bytesToRead == 0:
+          # Done reading chunked data
+          break
+
+        # Read bytesToRead and add to body
+        await client.recvLineInto(lineFut, maxLength = bytesToRead)
```
dom96: Pretty sure this is not what you want. You're not reading ``bytesToRead``, you're reading a line. You need to use `recv` here.

lib/pure/httpcore.nim
```diff
@@ -183,6 +183,13 @@ proc `[]=`*(headers: HttpHeaders, key, value: string) =
   ## Replaces any existing values.
   headers.table[headers.toCaseInsensitive(key)] = @[value]
 
+iterator iter*(headers: HttpHeaders, key: string): string =
+  ## Yields each value associated with the requested key
+  ## If no such key exists, this does nothing.
+  if headers.table.contains(headers.toCaseInsensitive(key)):
+    for value in headers.table[headers.toCaseInsensitive(key)]:
+      yield value
+
```
dom96: We would need to come up with a good naming scheme for this kind of iterators. I would prefer to just remove it and for you to inline this code above.

dom96 (changes requested):
Thank you for this!

I left some comments, please also include info about how you tested this locally (i.e. a test plan).

vabresto:
I've been testing locally with my repo above, though the shortest possible test case I've used is:
```
# main.nim
import asynchttpserver, asyncdispatch

  await req.respond(Http200, "Got " & req.body)

waitFor server.serve(Port(7000), cb)
```
and then running
`curl --data "hello=world" -H transfer-encoding:chunked http://127.0.0.1:7000/`
which should output
`Got hello=world` instead of the current `Content-Length required.`

If there's a way to include this kind of curl testing in the nim tests, I'd be happy to do that too.

lib/pure/httpcore.nim
vabresto: I agree that the naming is poor, but I think this style of accessor is very valuable because it offers exception-free access, similar to range-based for loops on lists. Maybe simply calling it `range` instead of `iter`? 

I don't mind inlining it for now, but it's actually a little awkward because `toCaseInsensitive` isn't exported, so I'd have to make it public now. For now, I haven't made this change, but I will do so if we're ok with exporting that property (or renaming the iterator).

lib/pure/asynchttpserver.nim
vabresto: Yes, you're right. I tried using `recv` initially but ran into issues because of the trailing `\r\n` characters, but this is the more correct way of doing it, and I've fixed it now.

lib/pure/httpcore.nim
dom96: You don't need the `toCaseInsensitive` call though. `[]` and `contains` already handle this. So just inline this without those calls.

vabresto: Fixed.

lib/pure/asynchttpserver.nim
```diff
@@ -261,6 +272,39 @@ proc processRequest(
...
+                                          "chunk data size must be hex encoded"))
+          return true
+      else:
+        if bytesToRead == 0:
+          # Done reading chunked data
+          break
+
+        # Read bytesToRead and add to body
+        # Note we add +2 because the line must be terminated by \r\n
+        let chunk = await client.recv(bytesToRead + 2)
```
dom96, line 304: This looks suspicious to me, I have a feeling that chunked encoding doesn't delimit data with \r\n. You should test with larger files to make sure they work (if you're just testing with "hello world" then you're unlikely to run into this).

dom96 (approved):
> If there's a way to include this kind of curl testing in the nim tests, I'd be happy to do that too.

What might be the best way to test it is to capture the data that curl sends, then just create a simple Nim script that does the following:

* Creates an asynchttpserver instance
* sends the data that curl would send using a raw socket (you can use newAsyncSocket() + connect + send for this)

As I mentioned in the comment I think you need to test this more thoroughly manually as well. I will merge this anyway though, we can always make fixes to the broken implementation without any problems. The main thing for me is that the API either doesn't change or is changed in a way that makes sense :)

dom96:
Also thank you for going the extra mile to resolve my comments, I saw that you had some trouble with the converter in IRC.

The reason it's there is for backwards compatibility: so that you don't need to change `headers["Blah"]` to `headers["Blah"][0]`. Indeed it shouldn't be removed.

lib/pure/asynchttpserver.nim
vabresto: @dom96 The two sources I've found with details on this topic say that the chunks are delimited with `\r\n`, see:
* (MDN): https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding
* (Wikipedia): https://en.wikipedia.org/wiki/Chunked_transfer_encoding#Example

However, you were right to be suspicious! The code needs to discard these newlines, which I did not correctly do. I've fixed this (and added tests!) in this follow-up PR: https://github.com/nim-lang/Nim/pull/16678

I'd appreciate it if you could take a look and review that too! Thanks!

PR #16643, opened 2021-01-08
add example of hashing an object by all of its fields with `fields`

ghost:


lib/pure/hashes.nim
```diff
@@ -38,6 +38,15 @@
...
 ##    result = !$h
 ##
+## You can hash an object by all of its fields with the `fields` iterator:
+## 
+## .. code-block:: Nim
+##  proc hash(x: MyObject): Hash =
+##    var h: Hash = 0
+##    for val in obj.fields:
+##      h = h !& hash(val)
+##    result = !$h
```
ringabout:
```suggestion
##  proc hash(x: object): Hash =
##    var h: Hash = 0
##    for val in x.fields:
##      h = h !& hash(val)
##    result = !$h
```

lib/pure/hashes.nim
```diff
@@ -54,6 +54,22 @@ runnableExamples:
...
+      h = h !& hash(val)
+    result = !$h
+
+  type
+    Obj = object
+      x: int
+      y: string
+
+  var x = Obj(x: 520, y: "Nim")
+  discard hash(x)
```
```suggestion
  doAssert hash(Obj(x: 520, y: "Nim")) != hash(Obj(x: 520, y: "Nim2"))
```

timotheecour (approved):

lib/pure/hashes.nim
```diff
@@ -54,6 +54,21 @@ runnableExamples:
 ## **Note:** If the type has a `==` operator, the following must hold:
 ## If two values compare equal, their hashes must also be equal.
 ##
+## You can hash an object by all of its fields with the `fields` iterator:
+runnableExamples:
+  proc hash(x: object): Hash =
+    var h: Hash = 0
+    for val in x.fields:
+      h = h !& hash(val)
+    result = !$h
```
```suggestion
  proc hash(x: object): Hash =
    for f in fields(x):
      result = result !& hash(f)
    result = !$result
```

timotheecour:
LGTM
maybe in future PR we should simply add this overload, since we already have it defined for `tuple` (and since users can always override `hash` for a given object)

dom96 (commented):
AFAIK there is a reason that this overload isn't specified, and that is to force you to explicitly decide which fields should be part of the hash. Because of this the example added here could be considered an anti-pattern, dunno maybe @Araq can comment

timotheecour:
> AFAIK there is a reason that this overload isn't specified, and that is to force you to explicitly decide which fields should be part of the hash

But with the same logic you'd disable `$` (or `==`, assignment, or toJson/fromJson/`json.%*` etc) for object and would require users to explicitly define `$` (etc) for each of their object types, just in case some particular object should be treated specially. This doesn't make sense IMO.

`hash` shouldn't be any different.

Instead what I suggest is to define `hash` by default for object, and users can always customize `hash` for their own type.

Araq:
You're both right... I consider "everything has a $ operator" a design mistake  so for ``ref object`` I draw the line, refs shouldn't have an `$` operator. And equality is complex, don't simply guess the implementation.

Having said that, @timotheecour has a good point, given that we have `==` for object we might as well add a `hash` operation for it.

timotheecour:
=> PR https://github.com/nim-lang/Nim/pull/17175

PR #4970, opened 2016-10-31
Fix problems with persistent HTTP connections

nigredo-tori:
Fixes #4969 

1. More robust request parsing in accordance with https://tools.ietf.org/html/rfc7230#section-3.5 .
2. Add `overrideHeaders` parameter to `request` procs for headers that shouldn't persist between requests.

PR #16667, opened 2021-01-10
A new request should always have a new content-length

jyapayne:
In [my last PR](https://github.com/nim-lang/Nim/pull/16618) I made a mistake by assuming that the client.headers were cleared on every request, like the Python version. So, due to the fact that Nim keeps the client headers, we need to clear the Content-Length header on each request (which makes sense because you almost never want to use the same Content-Length twice, but you may want to reuse other headers)

### DIFF
--- a/lib/pure/httpclient.nim
+++ b/lib/pure/httpclient.nim
@@ -1041,7 +1041,7 @@ proc request*(client: HttpClient | AsyncHttpClient, url: Uri | string,
   ## You need to make sure that the ``url`` doesn't contain any newline
   ## characters. Failing to do so will raise ``AssertionDefect``.
   ##
-  ## ``headers`` are HTTP headers that override the ``client.headers`` for
+  ## `headers` are HTTP headers that override the `client.headers` for
   ## this specific request only and will not be persisted.
   ##
   ## **Deprecated since v1.5**: use HttpMethod enum instead; string parameter httpMethod is deprecated
--- a/lib/pure/httpclient.nim
+++ b/lib/pure/httpclient.nim
@@ -973,19 +973,6 @@ proc requestAux(client: HttpClient | AsyncHttpClient, url: Uri,
   if url.scheme == "":
     raise newException(ValueError, "No uri scheme supplied.")
 
-  var data: seq[string]
-  if multipart != nil and multipart.content.len > 0:
-    data = await client.format(multipart)
-  else:
-    # A new request should always have a new content length
-    if client.headers.hasKey("Content-Length"):
-      client.headers.del("Content-Length")
-  
-    if body.len != 0:
-      client.headers["Content-Length"] = $body.len
-    elif httpMethod notin {HttpGet, HttpHead}:
-      client.headers["Content-Length"] = "0"
-
   when client is AsyncHttpClient:
     if not client.parseBodyFut.isNil:
       # let the current operation finish before making another request
@@ -995,6 +982,18 @@ proc requestAux(client: HttpClient | AsyncHttpClient, url: Uri,
   await newConnection(client, url)
 
   let newHeaders = client.headers.override(headers)
+
+  var data: seq[string]
+  if multipart != nil and multipart.content.len > 0:
+    data = await client.format(multipart)
+  else:
+    # Only change headers if they have not been specified already
+    if not newHeaders.hasKey("Content-Length"):
+      if body.len != 0:
+        newHeaders["Content-Length"] = $body.len
+      elif httpMethod notin {HttpGet, HttpHead}:
+        newHeaders["Content-Length"] = "0"
+
   if not newHeaders.hasKey("user-agent") and client.userAgent.len > 0:
     newHeaders["User-Agent"] = client.userAgent
 
@@ -1041,6 +1040,9 @@ proc request*(client: HttpClient | AsyncHttpClient, url: Uri | string,
   ##
   ## You need to make sure that the ``url`` doesn't contain any newline
   ## characters. Failing to do so will raise ``AssertionDefect``.
+  ##
+  ## ``headers`` are HTTP headers that override the ``client.headers`` for
+  ## this specific request only and will not be persisted.
   ##
   ## **Deprecated since v1.5**: use HttpMethod enum instead; string parameter httpMethod is deprecated
   when url is string:
--- a/lib/pure/httpclient.nim
+++ b/lib/pure/httpclient.nim
@@ -977,9 +977,13 @@ proc requestAux(client: HttpClient | AsyncHttpClient, url: Uri,
   if multipart != nil and multipart.content.len > 0:
     data = await client.format(multipart)
   else:
+    # A new request should always have a new content length
+    if client.headers.hasKey("Content-Length"):
+      client.headers.del("Content-Length")
+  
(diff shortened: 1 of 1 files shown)


dom96 (commented):
This is a departure from the Python version though. How can the user override the "Content-Length" header after this change? The Python version allows this and contains that `client.headers.hasKey` check IIRC

jyapayne:
> How can the user override the "Content-Length" header after this change?

Isn't the way to do that using the `headers` param in the `request` proc? 

You're right that it's a departure, however Python's `requests` prepares the request a bit differently. `requests` has a `Request` object, which is created every time you want to send a new request. In Nim, the connection and headers are not cleared after a new request, so there could be some headers that are left over from the previous request.

Right now, the issue I'm seeing is that I send a POST with a body, and then try to send a DELETE request without a body. This leaves the `Content-Length` header in from the POST request and the DELETE request fails. Maybe it would be better to simply clear the `client.headers` table before every request (or after)?

dom96:
> Isn't the way to do that using the headers param in the request proc?

But the code you have here will override this, no?

> Maybe it would be better to simply clear the client.headers table before every request (or after)?

Looking at the code it seems to me that the intention is for `client.headers` to be the headers specified by the user. As such they shouldn't be changed by the code, maybe that is the change that is needed here, there is a reason that the code below creates a `newHeaders` variable.

jyapayne:
> As such they shouldn't be changed by the code, maybe that is the change that is needed here, there is a reason that the code below creates a newHeaders variable.

Looking at the code for this line:

```nim
let newHeaders = client.headers.override(headers)
```

reveals the override proc:

```nim
proc override(fallback, override: HttpHeaders): HttpHeaders =
  # Right-biased map union for `HttpHeaders`
  if override.isNil:
    return fallback

  result = newHttpHeaders()
  # Copy by value
  result.table[] = fallback.table[]
  for k, vs in override.table:
    result[k] = vs
```

If I'm reading this code correctly, isn't then the `headers` param that provides the overrides and is provided by the user? My understanding was that the `client.headers` was used for storing headers that needed to be computed by the lib and then the `headers` param is used by the user to override any precomputed values.

> But the code you have here will override this, no?

If I'm right about the above (please correct me if I've misunderstood), then no, the user is still able to override the value with the `headers` param.

yglukhov:
Storing request specific headers in client-wide headers looks wrong indeed.

dom96:
> If I'm right about the above (please correct me if I've misunderstood), then no, the user is still able to override the value with the headers param.

Yeah, it appears you're right. So the additional `headers` argument was added to `request` (and `request` only, the `getContent`, `get`, `post`, etc. functions don't have it) in https://github.com/nim-lang/Nim/pull/4970. It looks like these were added so that `get` et al. can pass additional headers without changing `client.headers`, similar to your issue here.

In any case, `client.headers` shouldn't be changed. I think you need to modify the code so that it is left alone and only modify `newHeaders`.

Might be worthwhile to also update the docs for `headers` to mention that this is for additional headers that shouldn't be persisted across requests.

jyapayne:
> Yeah, it appears you're right. So the additional headers argument was added to request (and request only, the getContent, get, post, etc. functions don't have it) in #4970. It looks like these were added so that get et al. can pass additional headers without changing client.headers, similar to your issue here.
> 
> In any case, client.headers shouldn't be changed. I think you need to modify the code so that it is left alone and only modify newHeaders.
> 
> Might be worthwhile to also update the docs for headers to mention that this is for additional headers that shouldn't be persisted across requests.
> 

@dom96 Alright I can do that. Also, do you think that `get`, `post`, etc should be modified as well in order to override headers?

And one more question (perhaps for a different PR), this proc modifies the Content-Length header as well, but it looks more suited to be pulled out of the proc. What are your thoughts?

```nim
proc format(client: HttpClient | AsyncHttpClient,
            multipart: MultipartData): Future[seq[string]] {.multisync.} =
  let bound = getBoundary(multipart)
  client.headers["Content-Type"] = "multipart/form-data; boundary=" & bound
  await client.readFileSizes(multipart)
  var length: int64
  for entry in multipart.content:
    result.add(format(entry, bound) & httpNewLine)
    if entry.isFile:
      length += entry.fileSize + httpNewLine.len
  result.add "--" & bound & "--"
  for s in result: length += s.len
  client.headers["Content-Length"] = $length
```

dom96:
> Also, do you think that get, post, etc should be modified as well in order to override headers?

Not quite sure what you're proposing. But I think we should avoid unnecessary changes in this PR at least. Feel free to make further changes in a separate PR if you think it makes the API better.


> And one more question (perhaps for a different PR), this proc modifies the Content-Length header as well, but it looks more suited to be pulled out of the proc. What are your thoughts?

Yeah, this should be fixed in a separate PR.

timotheecour:
you're hitting https://github.com/timotheecour/Nim/issues/410 again; I've just restart CI just for `Nim SSL CI / ubuntu-18.04 (amd64) (pull_request)`

we really need to merge https://github.com/nim-lang/Nim/pull/16221 so that those tests always run and the corresponding bug gets fixed instead of letting it run once in a blue moon and letting regressions potentially accumulate in between

jyapayne:
@dom96 

> Not quite sure what you're proposing. 

Basically, just adding a `headers` param to each of `get`, `post`, etc. Will do it in a separate PR.

@timotheecour thanks for the info. I was wondering why a test failed. I agree the test should be fixed. Seems like a tough one.

lib/pure/httpclient.nim
```diff
@@ -1042,6 +1041,9 @@ proc request*(client: HttpClient | AsyncHttpClient, url: Uri | string,
   ## You need to make sure that the ``url`` doesn't contain any newline
   ## characters. Failing to do so will raise ``AssertionDefect``.
   ##
+  ## ``headers`` are HTTP headers that override the ``client.headers`` for
```
timotheecour: https://nim-lang.github.io/Nim/contributing.html
> In nim sources, prefer single backticks to double backticks since it's simpler and nim doc supports it (even in rst files with nim rst2html). [...]

jyapayne: Ah, alright. I was just following the other docs on this proc. Should I change all of them?

timotheecour: just for the code you're touching unless it's a cleanup-only PR

dom96:
> Basically, just adding a headers param to each of get, post, etc. Will do it in a separate PR.

Hrm, not sure we want that to be honest.

jyapayne:
Ah alright, not a problem then. Just thought it was weird not having that param on all of them.

The Pr is ready for review again. Not sure why CI is failing after I changed the doc comment.

ringabout:
You could follow instructions to restart CI:
https://nim-lang.github.io/Nim/contributing.html#the-git-stuff-debugging-ci-failures-flaky-tests-etc

jyapayne:
@xflywind I don't remember seeing that option in the Checks tab. Does it require any sort of permissions?

ringabout:
For example github action, I think author should be enough,

image

jyapayne:
@dom96 ping for review again when you have time

dom96 (approved):
Nice, LGTM

PR #16696, opened 2021-01-12
Check for errors after sendPendingSslData

yglukhov:
Not that I reproduced any issues related to this, but it just looks wrong. It is the only yielded future in the file that is not checked for errors. Also git blame shows that this line was intact since the very beginning. Other similar lines gained their error checking over time, so I assume this one just got missed.

### DIFF
--- a/lib/pure/asyncnet.nim
+++ b/lib/pure/asyncnet.nim
@@ -260,10 +260,9 @@ when defineSsl:
       # Call the desired operation.
       opResult = op
       # Bit hackish here.
-      # TODO: Introduce an async template transformation pragma?
 
       # Send any remaining pending SSL data.
-      yield sendPendingSslData(socket, flags)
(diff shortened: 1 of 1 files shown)


lib/pure/asyncnet.nim
```diff
@@ -263,7 +263,10 @@ when defineSsl:
       # TODO: Introduce an async template transformation pragma?
 
       # Send any remaining pending SSL data.
-      yield sendPendingSslData(socket, flags)
+      block:
+        let fut = sendPendingSslData(socket, flags)
+        yield fut
+        fut.read() # Check for errors
```
dom96: Nowadays you can just replace this with `await` afaik

yglukhov: I missed that somehow. Fixed.

dom96 (approved):

PR #16716, opened 2021-01-14
fix bug: unbuffered socket recv 1 byte per call

fun04wr0ng:
unbuffered socket recv bytes with timeout args will cause cpu 100%, because of recving strategy ( 1 byte per call)

### DIFF
--- a/lib/pure/net.nim
+++ b/lib/pure/net.nim
@@ -1400,7 +1400,7 @@ proc recv*(socket: Socket, data: pointer, size: int, timeout: int): int {.
     let avail = waitFor(socket, waited, timeout, size-read, "recv")
     var d = cast[cstring](data)
     assert avail <= size-read
-    result = recv(socket, addr(d[read]), avail)
+    result = recv(socket, addr(d[read]), size-read)
     if result == 0: break
     if result < 0:
       return result


lib/pure/net.nim
```diff
@@ -1400,7 +1400,7 @@ proc recv*(socket: Socket, data: pointer, size: int, timeout: int): int {.
     let avail = waitFor(socket, waited, timeout, size-read, "recv")
     var d = cast[cstring](data)
     assert avail <= size-read
-    result = recv(socket, addr(d[read]), avail)
+    result = recv(socket, addr(d[read]), size-read)
```
Araq, line 1403:
```suggestion
  if avail > 0:
    result = recv(socket, addr(d[read]), avail)
```

What about this variant?

dom96 (commented):
What's the case where this will cause 100% CPU usage? AFAIK `waitFor` should block.

fun04wr0ng:
> What's the case where this will cause 100% CPU usage? AFAIK `waitFor` should block.

When socket is unbuffered, waitFor just wait 1 byte per call, it wait and recv 1 byte N times if I need recv N bytes.

Actually it shall waitFor 1 byte and recv as many as possible per call. @dom96

fun04wr0ng:
In reality, I `recv(socket, size=1024*1024, timeout=10000),` it will call `waitFor & recv(1)` 1024*1024 times. 
That's the problem

PR #16723, opened 2021-01-14
conservative approach to fix #15184

### LINKED ISSUE #15184: Blank first Nim documentation tag ('##') causes text to be quoted in HTML output

a-mr:
Another approach w.r.t. PR #16699

Removes appearance of block quote when first line in a doc comment is empty:
```
##
## x
```

Now blank lines at the beginning of a doc comment are ignored in setting of baseline indentation level inside the comment.

Also the similar situation was wrong for multi-line doc comments:
``` nim
proc foo1* =
  ##[

  Long documentation comment1
  here.
  ]##
```

### DIFF
--- a/compiler/lexer.nim
+++ b/compiler/lexer.nim
@@ -955,8 +955,10 @@ proc skipMultiLineComment(L: var Lexer; tok: var Token; start: int;
   # detect the amount of indentation:
   if isDoc:
     toStrip = getColNumber(L, pos)
-    while L.buf[pos] == ' ': inc pos
-    if L.buf[pos] in {CR, LF}:
+    while L.buf[pos] == ' ':
+      inc pos
+      inc toStrip
+    while L.buf[pos] in {CR, LF}:  # skip blank lines
       pos = handleCRLF(L, pos)
       toStrip = 0
       while L.buf[pos] == ' ':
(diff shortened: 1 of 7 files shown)


Araq:
Thanks, I love it. :-)

Issue #10625, opened 2019-02-10
`setjmp` on linux mangles `ebp` leading to early collection


arnetheduck:
d50336837dfa1a3b81c1b5338b5a9b5d41cf6221

when the GC calls `setjmp`, the assumption is that it will put register values on the stack. some registers however, notably `RBP`, are mangled so as to protect against stack overwrite attacks. This leads to trouble, because mark-and-sweep will miss pointers and prematurely reclaim memory.

`RBP` is used as a general-purpose register whenever frame pointers are omitted: https://stackoverflow.com/questions/41912684/what-is-the-purpose-of-the-rbp-register-in-x86-64-assembler

Generally, `-fomit-frame-pointer` is enabled at all optimization levels, in gcc, meaning that even though the test suite might be passing, release compiles will be more likely to crash randomly.

https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/sysdeps/x86_64/setjmp.S#L33 shows how `RBP` is mangled.

arnetheduck:
One workaround for the `rbp` issue is to force frame pointers - this removes `RBP` from the set of general-purpose registers available for optimization..

Araq:
I'd prefer to use inline assembler for the "store registers on the stack" operation. We can base the implementations on setjmp's implementation. Hey, we only support about 18 different CPU archs, cannot be hard, right?

arnetheduck:
well, yeah, that would work as well, but strikes me as the less attractive option, at least to begin with - one could start with no-omit-fp in the standard options and then selectively use `asm` if it turns out to be beneficial for real.. 

I kind of doubt the impact is _that_ great on `x86_64` for example - it has what.. 16 registers already, so one more or less isn't that much of a difference (and probably dwarfed by other codegen bloat, like casts and stack reads/writes due to the many temporaries generated in general). 

In addition, on `x86` and `arm`, gcc has the `-momit-leaf-frame-pointer` option which should be safe and lessen the impace.

jlokier:
This GC bug occurs on MacOS as well.  Not sure which version at started at, but disassembling MacOS 10.15's `_setjmp` shows the same mangling as Glibc:

```
libsystem_platform.dylib`_setjmp:
->  0x7fff72f4cc9c <+0>:  movq   %rbx, (%rdi)
    0x7fff72f4cc9f <+3>:  movq   %rbp, %rax
[...]

Issue #4851, opened 2016-10-03
[GC] Illegal storage access when collecting cycle


arnetheduck:
The following causes a crash:

```
$ nim --version
Nim Compiler Version 0.15.0 (2016-10-03) [Linux: amd64]
Copyright (c) 2006-2016 by Andreas Rumpf

git hash: 6e1313515df4ee5808c77af5aeacfa77fce42474
$ nim c test ; ./test
Traceback (most recent call last)
test.nim(19)           test
system.nim(347)          prepareDealloc
SIGSEGV: Illegal storage access. (Attempt to read from nil?)

$ nim c -d:useSysAssert test ; ./test
[SYSASSERT] CollectZCT: isAllocatedPtr
```

test.nim:

```
type
  PNode = ref TNode
  TNode = object
    le: PNode

proc finalizeNode(n: PNode) =
  var s = @[0]

proc returnTree() =
  var cycle: PNode
  new(cycle, finalizeNode)
  cycle.le = cycle

for i in 1..100:
  returnTree()

GC_fullCollect()
GC_fullCollect()
```

yglukhov:
+1. Sometimes observed similar issue when used `echo` in a finalizer. Looks like anything that allocates inside finalizer may potentially crash the GC.

genotrance:
Still crashes with latest devel.

ghost:
Same behaviour on latest devel:
```
~/P/stuff ❯❯❯ nimq c seq.nim && ./seq
Traceback (most recent call last)
/home/dian/Projects/stuff/seq.nim(18) seq
/home/dian/Things/Nim/lib/system/gc_common.nim(417) prepareDealloc
SIGSEGV: Illegal storage access. (Attempt to read from nil?)

~/P/stuff ❯❯❯ nimq c -d:useSysAssert seq.nim && ./seq
[SYSASSERT] CollectZCT: isAllocatedPtr
```

Although seems to work just fine with ``--gc:orc`` :)
```
~/P/stuff ❯❯❯ nimq c --gc:orc -d:useMalloc seq.nim
~/P/stuff ❯❯❯ valgrind --leak-check=full ./seq
==23485== Memcheck, a memory error detector
==23485== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==23485== Using Valgrind-3.16.0.GIT and LibVEX; rerun with -h for copyright info
==23485== Command: ./seq
==23485== 
==23485== 
[...]

PR #16761, opened 2021-01-19
useMalloc for standard GC

### LINKED ISSUE #15394: -d:useMalloc for the default GC
At Status, we have been fighting against resource leaks for months (sockets, file descriptors, futures, timers and most importantly memory)
- https://github.com/status-im/nim-libp2p/pull/174
- https://github.com/status-im/nim-libp2p/pull/134
- https://github.com/status-im/nim-libp2p/pull/318
- https://github.com/status-im/nim-libp2p/pull/371
- https://github.com/status-im/nim-eth/commit/6ec942a195d3e3fcdff5e5fa39a458f06f4028aa
- https://github.com/status-im/nim-eth/pull/39
- https://github.com/status-im/nim-chronos/pull/102
- https://github.com/status-im/nim-chronos/pull/130
- https://github.com/status-im/nim-chronos/pull/65
- https://github.com/status-im/nim-chronos/issues/53
- https://github.com/status-im/nim-beacon-chain/pull/1673
- https://github.com/status-im/nim-beacon-chain/pull/1011
- https://github.com/status-im/nim-beacon-chain/issues/850
- https://github.com/status-im/nim-beacon-chain/issues/1629
- https://github.com/status-im/nim-beacon-chain/issues/1123
- https://github.com/status-im/nim-beacon-chain/issues/1121
- https://github.com/status-im/nim-beacon-chain/pull/1474#issuecomment-672043600
- https://github.com/status-im/nim-beacon-chain/issues/1010
- https://github.com/status-im/nim-beacon-chain/issues/447
- https://github.com/status-im/nim-beacon-chain/issues/784
- https://github.com/status-im/nim-beacon-chain/issues/779
- https://github.com/status-im/nim-libp2p/pull/128

We built several tools to help us track the issues
- https://github.com/status-im/nim-libp2p/pull/136
- https://github.com/status-im/nim-libp2p/issues/207
- https://github.com/status-im/nim-chronos/pull/33
- https://github.com/status-im/nim-chronos/issues/31
- https://github.com/status-im/nim-chronos/pull/85
- https://github.com/status-im/nim-chronos/issues/32
- https://github.com/status-im/nim-beacon-chain/issues/869
- https://github.com/status-im/nim-chronos/pull/106

However we are still leaking. As a workaround we are currently advising people to restart every 6 hours or so but for production we need to remove all possible source of leaks so that user can run that application for months without restart.

The computational part of our application is relatively easy to debug for leaks but the async/IO/networking part as been leaking resources quite often via closure iterators/futures due to unattended cancellation/expiration.

This is incredibly hard to debug.

We would like to have ``-d:useMalloc`` available for the default GC, backported to the 1.2.x branch so that we can use conventional C tools like Valgrind to detect those memory leaks and also run memory leak detection in dedicated CI.

Furthermore, we are currently investigating an issue with misreported memory accounting by Nim GC, which makes it even harder to debug with Nim standard tools (memory fragmentation?).

arnetheduck:
Closes #15394

unoptimized, malloc-based allocator for the GC - uses less memory, is a
bit slower for small chunks, less code


arnetheduck:
Using this branch halves the memory footprint of nimbus, as measured by RSS in linux (400 mb instead of 800).

It's not intended for production use really, more as a debugging tool - it's branched off the beginning of the 1.2 release cycle roughly, so can easily be applied to any version after that.

Araq:
Does it support the alignment requirements that the conservative stack scan needs?

arnetheduck:
> Does it support the alignment requirements that the conservative stack scan needs?

well, it assumes 8-byte alignment on anything coming out of `malloc` and redefines `pageIndex` and boundary checking accordingly - seems to work no worse than the "normal" GC allocator, judging by GC tests and nimbus (which is a long-running memory-allocation-heavy process). 

Nimbus is the reason this branch was written - we're having GC-related crashes and it's unclear what the culprit is: unfixed bugs in the cycle collector like https://github.com/nim-lang/Nim/issues/4851 and https://github.com/nim-lang/Nim/issues/10625 or the pointer tricks we sometimes subject the code to, that perhaps break the scanner.

mratsim:
Closes https://github.com/nim-lang/Nim/issues/15394

arnetheduck:
the test fails here are likely because this was branched off where 1.2.0 was released which might be missing some commits to update test scripts

arnetheduck:
Oh, and with something like `tcmalloc` you regain most of the lost performance - there are a few obvious improvements one could make which probably would bring the performance up to par - in particular, this branch has a large per-allocation overhead that mostly could go away, but it was simply more easy to write it this way so that one wouldn't have to modify the rest of the gc

Araq:
Ping @arnetheduck, can you please rebase?

tests/collections/ttables.nim
```diff
@@ -28,7 +28,13 @@ block tableadds:
   main()
   echo "done tableadds"
 
+block:
+  block:
+    var t1 = OrderedTable[int, string]()
+    var t2 = OrderedTable[int, string]()
 
+    doAssert t1 == t2
+
```
xflywind, line 37:
```suggestion
```

lib/pure/collections/tables.nim
```diff
@@ -1653,8 +1653,10 @@ proc `==`*[A, B](s, t: OrderedTable[A, B]): bool =
 
   if s.counter != t.counter:
     return false
+
```
xflywind, line 1656:
```suggestion
```

lib/pure/collections/tables.nim
```diff
@@ -1653,8 +1653,10 @@ proc `==`*[A, B](s, t: OrderedTable[A, B]): bool =
...
   if s.counter == 0 and t.counter == 0:
     return true
+
```
xflywind, line 1659:
```suggestion
```

tests/collections/ttables.nim
```diff
@@ -28,7 +28,6 @@ block tableadds:
   main()
   echo "done tableadds"
```
xflywind, line 30:
```suggestion


```

tests/collections/ttables.nim
```diff
@@ -443,4 +442,3 @@ block emptyOrdered:
   var t1: OrderedTable[int, string]
   var t2: OrderedTable[int, string]
   doAssert t1 == t2
```
xflywind, line 444:
```suggestion
  doAssert t1 == t2

```

ringabout:
Sorry for messing up commits messages(maybe you can use `force push` or somethings to clean the commits messages)
I only fetch the latest commits, didn't change the codes.

lib/system/gcmalloc.nim
```diff
@@ -0,0 +1,540 @@
...
+#            Nim's Runtime Library
+#        (c) Copyright 2012 Andreas Rumpf
+#        (c) Copyright 2021 Jacek Sieka
+#
+#    See the file "copying.txt", included in this
+#    distribution, for details about the copyright.
+#
+
+# Malloc-based allocator for the GC - similar to what's found in alloc.nim
+# without all the fancy small chunk allocation stuff - enable with -d:useMalloc
```
ringabout, line 12: maybe rename `useMalloc`? so ARC or ORC can benefit from this allocator.

arnetheduck: you mean the module? I'm not against though I don't have enough insight into orc/arc really - I imagine it could be simplified a bit for those use cases though if they are no longer stack-scanning, no?

arnetheduck:
> Sorry for messing up commits messages

I'm a bit unsure what you've done here? did you rebase against devel?

ringabout:
yeah

arnetheduck:
hm, the 32-bit issue might be legit, need to recheck some arithmetic

arnetheduck:
Closing this for now - it's more unstable than I remember it, ie with modern nim it fails to bootstrap nim itself when enabled where it used to, so something has changed and I'm not going to be investigating what it was.

Issue #16743, opened 2021-01-18
SIGSEGV httpclient basic example


juancarlospaco:
# Examples

```nim
import httpclient
let client = newHttpClient()  # OK
let rspons = client.head("http://httpbin.org/head")  # OK
echo rspons.body  # BUG
```

# Current Output

```nim
nim-#devel/lib/pure/httpclient.nim(266) body
nim-#devel/lib/pure/streams.nim(282) readAll
nim-#devel/lib/pure/streams.nim(228) readData

SIGSEGV: Illegal storage access. (Attempt to read from nil?)

$
```

# Expected Output

```nim
""
```

# Additional Information

- If you do `repr(response)` you see `body = ""`, so body is empty string.
- I know `HEAD` is not supposed to have a body, but still not a reason to `SIGSEGV`.
- Repro on all versions.

# System Information

<details>

```json

{
  "compiled": "2021-01-18T00:58:04",
  "NimVersion": "1.5.1",
  "hostCPU": "amd64",
  "hostOS": "linux",
  "cpuEndian": "littleEndian",
  "getTempDir": "/tmp/",
  "now": "2021-01-17T22:01:07-03:00",
  "getCurrentEncoding": "UTF-8",
  "getFreeMem": 397312,
  "getTotalMem": 528384,
  "getOccupiedMem": 97856,
  "countProcessors": 4,
  "arch": "x86_64",
  "FileSystemCaseSensitive": true,
  "nimcacheDir": "/home/juan/.cache/nim/nimbug_r",
  "ccompilerPath": "",
  "currentCompilerExe": "/home/juan/.choosenim/toolchains/nim-#devel/bin/nim",
  "nimpretty": "0.2",
  "nimble": "nimble v0.12.0 compiled at 2021-01-17 01:03:49\ngit hash: couldn't determine git hash",
  "nimgrep": "1.6.0",
  "nimsuggest": "Nim Compiler Version 1.5.1 [Linux: amd64]\nCompiled at 2021-01-17\nCopyright (c) 2006-2021 by Andreas Rumpf\n\ngit hash: 44ceefa9fefca58ba32e9638f754d7f4ca8a2b82\nactive boot switches: -d:release -d:danger --gc:markAndSweep",
  "choosenim": "choosenim v0.7.4 (2020-10-20 17:30:52) [linux/amd64]",
  "gcc": "gcc (GCC) 10.2.0",
  "clang": "clang version 11.0.1",
  "git": "2.30.0",
[...]

PR #16766, opened 2021-01-20
Fix SIGSEGV in httpclient response body

daehee:
Prevent SIGSEGV on response body when getBody is false; ref: https://github.com/nim-lang/Nim/issues/16743

### DIFF
--- a/lib/pure/httpclient.nim
+++ b/lib/pure/httpclient.nim
@@ -833,7 +833,7 @@ proc parseResponse(client: HttpClient | AsyncHttpClient,
     httpError("Connection was closed before full request has been made")
 
   when client is HttpClient:
-      result.bodyStream = newStringStream()
+    result.bodyStream = newStringStream()
   else:
       result.bodyStream = newFutureStream[string]("parseResponse")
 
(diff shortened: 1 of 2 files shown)


lib/pure/httpclient.nim
```diff
@@ -832,14 +832,16 @@ proc parseResponse(client: HttpClient | AsyncHttpClient,
   if not fullyRead:
     httpError("Connection was closed before full request has been made")
 
+  when client is HttpClient:
+      result.bodyStream = newStringStream()
```
Araq:
```suggestion
    result.bodyStream = newStringStream()
```

lib/pure/httpclient.nim
```diff
@@ -832,14 +832,16 @@ proc parseResponse(client: HttpClient | AsyncHttpClient,
...
+  else:
+      result.bodyStream = newFutureStream[string]("parseResponse")
```
Araq:
```suggestion
    result.bodyStream = newFutureStream[string]("parseResponse")
```

Araq:
Ping @dom96

ringabout (approved):

Issue #4123, opened 2016-04-28
Async await TODO list


dom96:
- [ ] Ability to cancel futures. Probably implemented as a `cancel` procedure.
- [ ] Epoll/kqueue/IOCP system timers instead of emulated ones.
- [ ] `callSoon`

v3ss0n:
EPoll/kqueu/IOCP should be first , then cancelling futures.

dom96:
Don't worry about the order, it's not ordered by importance.

endragor:
OS-based timer system will not work for everyone. It implies dealing with OS file descriptor limit (or other imposed limit), which does not allow to create a system that schedules a lot of timers. So if you decide to implement that, please make it optional.

Araq:
@dom96 Update this list or close it.

dom96:
Both of the items are still TODO. I don't see a point in putting them into a separate issue or closing this one.

PR #16769, opened 2021-01-20
Added possibility to cancel the futures using complete() and fail()

Q-Master:
I've made a fix to asyncfutures and asyncmacro to be able to complete or fail any pending future at a cancellation points, like I've asked in [https://forum.nim-lang.org/t/7287](https://forum.nim-lang.org/t/7287)

### DIFF
--- a/lib/pure/asyncfutures.nim
+++ b/lib/pure/asyncfutures.nim
@@ -23,6 +23,9 @@ type
     callbacks: CallbackList
 
     finished: bool
+    cancelled: bool
+    firstCancelled: bool
+    recursiveFuture*: FutureBase        ## Stored future for recursive cancellation
     error*: ref Exception              ## Stored exception
     errorStackTrace*: string
     when not defined(release):
@@ -37,6 +40,8 @@ type
 
   FutureError* = object of Defect
     cause*: FutureBase
+  
+  FutureCancelledError* = object of FutureError
 
 when not defined(release):
   var currentID = 0
@@ -108,6 +113,9 @@ proc callSoon*(cbproc: proc ()) =
 template setupFutureBase(fromProc: string) =
   new(result)
   result.finished = false
+  result.cancelled = false
+  result.firstCancelled = false
+  result.recursiveFuture = nil
   when not defined(release):
     result.stackTrace = getStackTraceEntries()
     result.id = currentID
@@ -132,16 +140,19 @@ proc newFutureVar*[T](fromProc = "unspecified"): owned(FutureVar[T]) =
   result = typeof(result)(fo)
   when isFutureLoggingEnabled: logFutureStart(Future[T](result))
 
-proc clean*[T](future: FutureVar[T]) =
+proc clean*[T](futureVar: FutureVar[T]) =
   ## Resets the `finished` status of `future`.
-  Future[T](future).finished = false
-  Future[T](future).error = nil
+  template future: untyped = Future[T](futureVar)
+  future.finished = false
+  future.cancelled = false
+  future.firstCancelled = false
+  future.error = nil
 
 proc checkFinished[T](future: Future[T]) =
   ## Checks whether `future` is finished. If it is then raises a
   ## `FutureError`.
   when not defined(release):
-    if future.finished:
+    if future.finished and not future.cancelled:
       var msg = ""
       msg.add("An attempt was made to complete a Future more than once. ")
       msg.add("Details:")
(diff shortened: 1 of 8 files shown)


juancarlospaco:
Needs changelog entry, documentation should be updated to explain it, needs at least 1 test.

Q-Master:
Added changelogs and tests.

Q-Master:
Fixed some doctstrings. Marked cancellation points.

dom96:
Ooh, thanks for the PR.

I'm afraid I won't have time to review this until next week. In the meantime, would be nice to get the thoughts of those that worked on chronos to see what they've learned from their implementation of cancellation and what they think of this one.

Q-Master:
Their cancellation is different from this and it seems too complicated and not that functional for me. They're just cancelling the futures and that's all. This implementation is much like python one which makes it possible either to complete future with value(which will be transmitted to all the awaiters of it) or fail it with exception which will also transmit to all who is awaiting for that future.

lib/pure/asyncfutures.nim
```diff
@@ -23,6 +23,8 @@ type
     callbacks: CallbackList
 
     finished: bool
+    cancelled: bool
+    recursiveFuture*: FutureBase        ## Stored future for recursive cancellation
```
dom96: I finally got to this and it's midnight so I'm likely missing something. But my initial impressions from reading this are that this looks awfully like emulating exception propagation logic, couldn't we simply create a `FutureCancelled` exception and have it propagate up the futures?

The reason I ask this is because if we can then it seems like it would be an even simpler and cleaner addition to what we already have.

Q-Master: We need to store recursive futures because the iterator loop is not storing them anywhere, just adding callbacks to them and that's all. The idea from chronos to store all the futures in hashtable seems too complicated and not so needed to me. The cancelled flag is needed only for those recursive futures to stop immediately without any error.

The idea of replacing the cancelled flag to failure with FutureCancelled might lead to situation when the recurrent future will raise the cancelled exception while it is not realy cancelled. The difference with chronos is that my implementation is not cancelling the future - it is completing or failing it, so we just break the iterator loop and call all the callbacks which are connected to that future with setting the value or error for that future, but to completely stop we need to stop all the recurrent futures silently (IMO) without raising the cancelled error because this could lead to an exception mess and stopping of the application.

dom96: > The idea of replacing the cancelled flag to failure with FutureCancelled might lead to situation when the recurrent future will raise the cancelled exception while it is not realy cancelled.

How could it? We pretty much control when the Future exception gets re-raised.

Q-Master: > How could it? We pretty much control when the Future exception gets re-raised.

The cancellation of all subsequent futures will not be silent, they all will get this FutureCancelled exception, so they all will reraise it. Not sure that it is a good behaviour, especially when this exception will return to a parent one which is not cancelled at all, but just completed or failed. BTW, in this case the failure of a parent future might be replaced with a failure of a subsequent exception, which is not good too.

If you insist on replacing this flag to a FutureCancelled exception then we should separate the parent future and all subsequent ones by some flag also to not allow such a behaviour which I mentioned before.

ringabout:
It fixes #4123.

dom96:
> It fixes #4123.

Not quite, I think it's the first step towards fixing it. But I'm still not certain it's the right step.

Q-Master:
>Not quite, I think it's the first step towards fixing it. But I'm still not certain it's the right step.

It's really only the first step. I nowadays am already using various async realizations for other languages for years, so I am rather sure that it is a little right step.
The best thing is that this patch doesn't brake anything working before, so there's no need to rewrite previously working code.

yglukhov:
Just a recap of what I said on the call. My main concern is that cancelling coroutines by just terminating their code execution makes it unsafe to use any kind of statefullness. In simpler words, a cancellation exception should be raised upon reading the innermost future in the corresponding coroutine so that it can react appropriately to restore its state, free its resources, etc.

Q-Master:
I've thought that it would be easy to implement, but it is not. I'm trying to figure out how to do so that raising the exception might be caught by try/except, because it is somehow propagates up and evades the try block. I think that I'm doing something wrong.

Q-Master:
> Just a recap of what I said on the call. My main concern is that cancelling coroutines by just terminating their code execution makes it unsafe to use any kind of statefullness. In simpler words, a cancellation exception should be raised upon reading the innermost future in the corresponding coroutine so that it can react appropriately to restore its state, free its resources, etc.

I think I've fixed this problem. Every time the future is cancelled the FutureCancelledError is raised, so the future might catch it and do some work reacting.

Q-Master:
seems too outdated

PR #16787, opened 2021-01-21
Finer analysis for array access

Clyybber:
This also allows optimizing assignments such as `a = a[i]` where `i` is a runtime value into sinks.

### DIFF
--- a/compiler/dfa.nim
+++ b/compiler/dfa.nim
@@ -574,32 +574,87 @@ proc skipConvDfa*(n: PNode): PNode =
       result = result[1]
     else: break
 
-proc aliases*(obj, field: PNode): bool =
+type AliasKind* = enum
+  yes, no, maybe
+
+from trees import exprStructuralEquivalent
+
+proc aliases*(obj, field: PNode): AliasKind =
+  # obj -> field:
+  # x -> x: true
+  # x -> x.f: true
+  # x.f -> x: false
+  # x.f -> x.f: true
+  # x.f -> x.v: false
+  # x -> x[0]: true
+  # x[0] -> x: false
+  # x[0] -> x[0]: true
+  # x[0] -> x[1]: false
+  # x -> x[i]: true
+  # x[i] -> x: false
+  # x[i] -> x[i]: MAYBE Further analysis could make this return true when i is a runtime-constant
+  # x[i] -> x[j]: MAYBE also returns MAYBE if only one of i or j is a compiletime-constant
   var n = field
   var obj = obj
+  var objImportantNodeStack: seq[PNode]
   while true:
     case obj.kind
-    of {nkObjDownConv, nkObjUpConv, nkAddr, nkHiddenAddr, nkDerefExpr, nkHiddenDeref}:
+    of PathKinds0 - {nkDotExpr, nkCheckedFieldExpr, nkBracketExpr}:
       obj = obj[0]
     of PathKinds1:
       obj = obj[1]
-    else: break
+    of nkDotExpr, nkCheckedFieldExpr, nkBracketExpr:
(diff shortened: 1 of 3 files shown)


Araq:
Why is that correct? `a` can alias `a[i]` for `ref` types. More importantly, we should focus on real-world optimizations, don't make the compiler as smart as you possibly can because it's fun, it's futile. In the end all it accomplishes is that it's harder to understand for everybody.

Clyybber:
`ref` aliases are handled with by `ref`s `=sink` hook. This optimization is a very real-world optimization, especially for the compiler where we often have `a = a.sons[i]` (which aliases itself uses for recursing :D).
This PR simply extends the optimization from #16775 by removing the bail-out check in isAnalysableFieldAccess and then refines `aliases` to properly analyse array access with an unknown index for which it returns the `maybe` state.

The most complex change to `aliases` is that we now also skip `PathKinds0 - {nkDotExpr, nkCheckedFieldExpr, nkBracketExpr}` and `PathKinds1` nodes that are nested in between the important nodes `nkDotExpr, nkCheckedFieldExpr, nkBracketExpr`. Thats a bugfix/refactoring that should be done even without enabling this optimization.

Araq:
>  This optimization is a very real-world optimization, especially for the compiler where we often have `a = a.sons[i]` (which aliases itself uses for recursing :D).

This is a cursor and should be optimized as such, and not transformed into a sink.

Clyybber:
Sure, I'm just outlining why this is always correct, even without cursor inference/when `a` cant be turned into a cursor.

Clyybber:
The refactoring of aliases also gets rid of the quadratic complexity (sameTree being called in each iteration), and paves the way for using information like invariants for further analysis, because now the nodes that lead to locations `insloc` and `loc` are compared pairwise.

compiler/dfa.nim
```diff
@@ -574,49 +574,87 @@ proc skipConvDfa*(n: PNode): PNode =
...
+        result.add n
+        n = n[0]
+      of nkSym:
+        result.add n; break
+      else: return no
+
+  collectImportantNodes(objImportantNodes, obj)
+  collectImportantNodes(fieldImportantNodes, field)
+
+  if fieldImportantNodes.len < objImportantNodes.len: return no
```
Araq, line 615: Why is this correct? Why is this not ``if fieldImportantNodes.len != objImportantNodes.len``?

compiler/dfa.nim
```diff
@@ -574,49 +574,87 @@ proc skipConvDfa*(n: PNode): PNode =
...
+
+  result = yes
+  for i in 1..objImportantNodes.len:
+    let currFieldPath = fieldImportantNodes[^i]
+    let currObjPath = objImportantNodes[^i]
```
Araq: Use `countdown` instead?

Araq:
Ok, if you want it, add comments to the code sketching a proof of correctness.

compiler/dfa.nim
Clyybber: Because aliases(obj: `x`, field:`x.f`) should return true,
but aliases(obj: `x.f`, field: `x`) should return false.
For the first case `fieldImportantNodes.len` is 2 and `objImportantNodes.len` is 1
but for the second case `objImportantNodes.len` is 2 and `fieldImportantNodes.len` is 1, which means that field cannot be part of/aliased by obj.

compiler/dfa.nim
Clyybber: `fieldImportantNodes` might be larger than `objImportantNodes` and we want to start at the last element of both.

Clyybber:
ping @Araq

PR #16793, opened 2021-01-22
show msys2 as foreignDep for windows

capocasa:
The `distros` module provides example install commands for installation packages declared with `foreignDep`.

Of course, on windows it is unclear which one to use, because there are no package managers provided by the OS and many 3rd party ones.

The current one chosen, `Chocolatey`, is unlikely to be useful for installing build dependencies- for example, the Nim version provided is 1.0.0. Also, `Chocolatey` is no longer the current install command used by Chocolatey- the command is `choco`.

Instead, I would like to suggest using `pacman` from `msys2`- msys2 provides current versions for nim, mingw64, autoconf, libtool, bash, perl and pretty much any tool that might be needed to build code, as well as many C libraries and other unixy packages, rendering foreignDep for Windows actually useful.

I just used it to build 3 different nimterop packages with very little fuss.

### DIFF
--- a/lib/pure/distros.nim
+++ b/lib/pure/distros.nim
@@ -222,7 +222,7 @@ proc foreignDepInstallCmd*(foreignPackageName: string): (string, bool) =
   ## and whether it requires root/admin rights.
   let p = foreignPackageName
   when defined(windows):
-    result = ("Chocolatey install " & p, false)
+    result = ("pacman -S " & p, false)
   elif defined(bsd):
     result = ("ports install " & p, true)
   elif defined(linux):


Araq:
I'm not a fan of emulating Unix on Windows, at least Chocolatey was designed for Windows... Maybe we can simply remove any suggestion for a package manager for Windows though.

capocasa:
You don't have to use the msys2 unix environment, you can just call the installed programs from the windows shell, including package installation. Then it's just another package manager- just one with up-to-date and mature packages for every unix build dependency out there, common and esoteric.

What that allows you to do is wrap C libraries using their own build system. Doing that without msys2 is terrible- you have to Frankenstein together your toolchain from various windows installers, if you're lucky enough to find the right ones. I was extracting msys2 packages to get tools I couldn't find when it occurred to me I might just use the package manager.

Not using the native build system for a wrap is terrible as well- getting all the flags right in Nim requires intimate knowledge of the library *and* the tooling. That's brain cycles that could be used to make a nice Nim-API on top of the wrapper- even worse, it likely prevents a *lot* of wraps from happening in the first place.

msys2-as-package-manager is just a logical extension of providing a ZIP with mingw- just, you get everything you need to build, not just 50%. But even if you don't buy that one- in my mind, the trade-off between more purity and dramatically lowering the barrier to a robust Nim package ecosystem is a no-brainer.

Araq:
Ok, please write a changelog entry and I can approve this change.

capocasa:
Great!!!

I'll try to place a detailed description on the msys2 wiki and link to it with a short summary for the notes.

capocasa:
msys2 replied! My approach for Window-shell use of msys2 (place msys2 paths in windows path, .bat-wrap scripts) is actually new to the msys2 world and not currently in use, discouraged citing unspecified problems. This came as a surprise to me because it worked so well. The msys2 folks wouldn't mind to link to my instructions but don't want to host them, the better to underscore it's inofficial.

I see three options to move forward with Nim-msys2 integration, providing the mentioned benefits:

1. Proceed boldly: Recommend placing msys2 paths into Windows path, and accept Nim might take some mild heat for condoning inofficial msys2 use if problems to arise.
2. Proceed cautiously: Provide an activation script that places msys2 paths temporarily into the Windows path. This would require the user to perform an additional step before nimble-installing packages with foreign deps, but would limit exposure to possible breakage.
3. Be clever: Create a script that copies installed msys2 executables into the Nim bin path, to be run after installing foreign deps.

I think I am inclined to recommend solution 3, because it is similar to the way Nim already provides zipped .exe files of some unix utilities. This approach would also lend itself to be extended into an installer down the line, which is how the Ruby ecosystem provides Windows dependencies for its c-library-wrapping "Gems". The next step would be to talk to the creator of the current zipped .exe unix tools that comes with Nim, to avoid duplicate work.

Araq:
> But even if you don't buy that one- in my mind, the trade-off between more purity and dramatically lowering the barrier to a robust Nim package ecosystem is a no-brainer.

Now compare that with your own:

> msys2 replied! My approach for Window-shell use of msys2 (place msys2 paths in windows path, .bat-wrap scripts) is actually new to the msys2 world and not currently in use, discouraged citing unspecified problems.

This doesn't sound like a way for a "robust Nim package ecosystem" at all anymore.

> I see three options to move forward with Nim-msys2 integration, providing the mentioned benefits: ...

Sorry, but I don't like either option. Instead we should improve Nimble and add a good cmake integration.

capocasa:
I understand- sorry for initially overselling! The solution worked so well in practice it was a surprise that msys2 doesn't support it officially.

Araq:
No worries.

capocasa:
How are the current binary dependency packages that are distributed with Nim put together? It would probably be reasonably easy to extend those a bit, which is probably what you mean.

Araq:
They are currently hand-made using Visual C++ and the various native setups these libraries provide. But I think cmake emerged as a defacto standard nowadays.

capocasa:
>They are currently hand-made using Visual C++

Sounds nice.

>cmake emerged as a defacto standard nowadays

There is certainly some convergence on cmake, but it also polarizes. Meson also has traction, and so do a few others.

To realize nimterop's potential to yield a lot of wrapped functionality with nice Nim interfaces, I don't see any way around being prepared to cover as many build systems as the C library makers come up with- hence, the package management idea.

When you said improving nimble, did you mean add a capability to install and remove system specific binaries?

I suppose I could make do for now by making scoop install scripts for cmake.zip, meson.zip and similar and providing them as foreign deps.

PR #16816, opened 2021-01-25
nimsuggest deprecated symbols less likely to be suggested

saem:
* penalizes the quality score of deprecated symbols
* uses quality more pervasively in order to reflect deprecation impact
* impacts both sug and con

### DIFF
--- a/compiler/suggest.nim
+++ b/compiler/suggest.nim
@@ -258,17 +258,22 @@ proc fieldVisible*(c: PContext, f: PSym): bool {.inline.} =
       result = true
       break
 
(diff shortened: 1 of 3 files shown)


saem:
based off of #16814, so that needs to be merged in first, leaving this as draft for now.

saem:
For some silly reason in the i386 build the test is crashing and I have no idea why. 😦 

It'll take me a bit to get a i386 build env, pretty sure Fedora dropped support for that as part of the standard distro so I need to work out some container business.

saem:
OK, it's back to working. \o/

Issue #12595, opened 2019-11-04
Gensym'd symbol in template can't be found in curly braces


jyapayne:
### Example
```nim
template checkRaises(): untyped =
  let
    codeStr = "adfadf" # Warning here about how "codeStr`gensymXXXXXX" is not used
    vals = {codeStr: "No Exc"} # Error here about no symbol `codeStr` defined
  echo vals

checkRaises()
```

### Current Output
```
Hint: used config file '/Users/jyakimowich-payne/Nim/config/nim.cfg' [Conf]
Hint: used config file '/Users/jyakimowich-payne/JoeyProjects/tani/nim.cfg' [Conf]
Hint: used config file '/Users/jyakimowich-payne/Nim/config/config.nims' [Conf]
Hint: system [Processing]
Hint: widestrs [Processing]
Hint: io [Processing]
Hint: x [Processing]
/Users/jyakimowich-payne/JoeyProjects/tani/x.nim(7, 12) template/generic instantiation of `checkRaises` from here
/Users/jyakimowich-payne/JoeyProjects/tani/x.nim(4, 13) Error: undeclared identifier: 'codeStr'
```

### Expected Output
```
[("adfadf", "No Exc")]
```

### Possible Solution

* No idea. Looks like the compiler doesn't understand when a symbol is in a set of curly braces

### Additional Information

* You can work around this by using the more verbose syntax:

```nim
  let
    codeStr = "adfadf"
    vals = [(codeStr, "No Exc")]
  echo vals
```

Nim is on latest devel:

```
$ nim -v
Nim Compiler Version 1.1.1 [MacOSX: amd64]
Compiled at 2019-11-04
```

konsumlamm:
Still present in 1.5.1. This breaks `sugar.collect` with tables when used in templates, in most cases (sadly, the workaround of using `[(k, v)]` doesn't work for that use case).

Issue #10609, opened 2019-02-09
Weird template bug


dawkot:
```nim
template foo =
  proc bar1(x: bool) =
    var x {.inject.} = x
  
  proc bar2(x: bool) =
    var x {.inject.} = x

foo()
```
Fails with:
> (6, 24) Error: undeclared identifier: 'x'

bluenote10:
What is the purpose of using `{.inject.}` on proc scope? Don't they have their own scope anyway?

It looks like using inject once changes the semantics of the `x` symbol:

```nimrod
import macros

template foo =
  proc bar1(x: bool) =
    echo x
    var x {.inject.} = x
    echo x

  proc bar2(x: bool) =
    echo x
    var x {.inject.} = x
    echo x

static:
  echo getAst(foo()).repr

foo()
```

Output:

```
proc bar1(x163031: bool) =
  echo x163031
  var x {.inject.} = x163031
  echo x

proc bar2(x163033: bool) =
  echo x
  var x {.inject.} = x
  echo x
```

Without the injects:

```
proc bar1(x163033: bool) =
  echo x163033
  var x163035 = x163033
  echo x163035

proc bar2(x163037: bool) =
  echo x163037
  var x163039 = x163037
  echo x163039
```

krux02:
This is certainly a bug here:

```
proc bar2(x163033: bool) = <------ these two x should be the same
  echo x                <----/
  var x {.inject.} = x
  echo x
```

metagn:
The version with `static: echo getAst(foo()).repr` now segfaults

juancarlospaco:
!nim c

```nim
import macros
template foo =
  proc bar1(x: bool) =
    echo x
    var x {.inject.} = x
    echo x
  proc bar2(x: bool) =
    echo x
    var x {.inject.} = x
    echo x
static:
  echo getAst(foo()).repr
foo()
```

PR #16820, opened 2021-01-25
Sugar tests

konsumlamm:
Followup to #16802.

* remove unnecessary example in `sugar` (the example also works without `capture` and I decided to not replace it with another example, since situations where you need `capture` in `mapIt` are rare)
* add more tests (for `=>`, `->`, `dump` and `capture`)
* put tests in `block`s

### DIFF
--- a/lib/pure/sugar.nim
+++ b/lib/pure/sugar.nim
@@ -54,6 +54,7 @@ proc createProcType(p, b: NimNode): NimNode {.compileTime.} =
 
 macro `=>`*(p, b: untyped): untyped =
   ## Syntax sugar for anonymous procedures. It also supports pragmas.
+  # TODO: xxx pending #13491: uncomment in runnableExamples
   runnableExamples:
     proc passTwoAndTwo(f: (int, int) -> int): int = f(2, 2)
 
@@ -67,6 +68,9 @@ macro `=>`*(p, b: untyped): untyped =
 
     myBot.call = (name: string) {.noSideEffect.} => "Hello " & name & ", I'm a bot."
     doAssert myBot.call("John") == "Hello John, I'm a bot."
+
+    # let f = () => (discard) # simplest proc that returns void
+    # f()
 
   var
     params = @[ident"auto"]
@@ -149,7 +153,7 @@ macro `->`*(p, b: untyped): untyped =
 
     proc passOne(f: (int {.noSideEffect.} -> int)): int = f(1)
     # is the same as:
-    # proc passOne(f: proc (x, y: int): int {.noSideEffect.}): int = f(1)
+    # proc passOne(f: proc (x: int): int {.noSideEffect.}): int = f(1)
 
     doAssert passOne(x {.noSideEffect.} => x + 1) == 2
 
--- a/lib/pure/sugar.nim
+++ b/lib/pure/sugar.nim
@@ -187,7 +187,7 @@ macro capture*(locals: varargs[typed], body: untyped): untyped {.since: (1, 1).}
   ## Useful when creating a closure in a loop to capture some local loop variables
   ## by their current iteration values.
   runnableExamples:
-    import std/[strformat, sequtils]
+    import std/strformat
 
     var myClosure: () -> string
     for i in 5..7:
--- a/lib/pure/sugar.nim
+++ b/lib/pure/sugar.nim
@@ -197,12 +197,6 @@ macro capture*(locals: varargs[typed], body: untyped): untyped {.since: (1, 1).}
             myClosure = () => fmt"{i} * {j} = 42"
     doAssert myClosure() == "6 * 7 = 42"
 
-    let m = @[(s: string) => "to " & s,
-              (s: string) => "not to " & s]
-    let l = m.mapIt(capture(it, (s: string) => it(s)))
-    let r = l.mapIt(it("be"))
-    doAssert fmt"{r[0]}, or {r[1]}" == "to be, or not to be"
-
   var params = @[newIdentNode("auto")]
   let locals = if locals.len == 1 and locals[0].kind == nnkBracket: locals[0]
                else: locals
--- a/lib/pure/sugar.nim
+++ b/lib/pure/sugar.nim
@@ -53,8 +53,7 @@ proc createProcType(p, b: NimNode): NimNode {.compileTime.} =
   result.add prag
 
 macro `=>`*(p, b: untyped): untyped =
-  ## Syntax sugar for anonymous procedures.
-  ## It also supports pragmas.
+  ## Syntax sugar for anonymous procedures. It also supports pragmas.
   runnableExamples:
     proc passTwoAndTwo(f: (int, int) -> int): int = f(2, 2)
 
@@ -62,7 +61,7 @@ macro `=>`*(p, b: untyped): untyped =
 
     type
       Bot = object
-        call: proc (name: string): string {.noSideEffect.}
+        call: (string {.noSideEffect.} -> string)
 
     var myBot = Bot()
 
@@ -140,14 +139,19 @@ macro `=>`*(p, b: untyped): untyped =
                    procType = kind)
 
 macro `->`*(p, b: untyped): untyped =
-  ## Syntax sugar for procedure types.
+  ## Syntax sugar for procedure types. It also supports pragmas.
   runnableExamples:
     proc passTwoAndTwo(f: (int, int) -> int): int = f(2, 2)
-
     # is the same as:
     # proc passTwoAndTwo(f: proc (x, y: int): int): int = f(2, 2)
 
     doAssert passTwoAndTwo((x, y) => x + y) == 4
+
+    proc passOne(f: (int {.noSideEffect.} -> int)): int = f(1)
+    # is the same as:
+    # proc passOne(f: proc (x, y: int): int {.noSideEffect.}): int = f(1)
+
+    doAssert passOne(x {.noSideEffect.} => x + 1) == 2
 
   result = createProcType(p, b)
 
--- a/tests/stdlib/tsugar.nim
+++ b/tests/stdlib/tsugar.nim
@@ -3,186 +3,213 @@ discard """
 x + y = 30
 '''
 """
+import std/[sugar, algorithm, random, sets, tables, strutils]
 
-import std/sugar
+template main() =
+  block: # `=>`
+    block:
+      let f1 = () => 42
+      doAssert f1() == 42
 
-# `=>`
+      let f2 = (x: int) => x + 1
+      doAssert f2(42) == 43
 
-block:
-  let f = () => 42
-  doAssert f() == 42
+      let f3 = (x, y: int) => x + y
+      doAssert f3(1, 2) == 3
 
-  let g = (x: int) => x + 1
-  doAssert g(42) == 43
+      var x = 0
+      let f4 = () => (x = 12)
+      f4()
+      doAssert x == 12
 
-  let h = (x, y: int) => x + y
-  doAssert h(1, 2) == 3
+    block:
+      proc call1(f: () -> int): int = f()
+      doAssert call1(() => 12) == 12
 
-block:
-  proc call1(f: () -> int): int = f()
-  doAssert call1(() => 12) == 12
+      proc call2(f: int -> int): int = f(42)
+      doAssert call2(x => x) == 42
+      doAssert call2((x) => x) == 42
+      doAssert call2((x: int) => x) == 42
 
-  proc call2(f: int -> int): int = f(42)
-  doAssert call2(x => x) == 42
-  doAssert call2((x) => x) == 42
-  doAssert call2((x: int) => x) == 42
+      proc call3(f: (int, int) -> int): int = f(1, 2)
+      doAssert call3((x, y) => x + y) == 3
+      doAssert call3((x, y: int) => x + y) == 3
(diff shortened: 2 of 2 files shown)


tests/stdlib/tsugar.nim
```diff
@@ -1,4 +1,72 @@
...
+  output: '''
+x + y = 30
+'''
+"""
+
+import std/sugar
+
+# `=>`
+
+block:
```
timotheecour: group lexically, eg:
```nim
block: # `=>`
  block: # (in case you need sub-blocks)
    ...
  block:
     ...
block: # `->`
  ...
```
makes it easier to follow and search for tests for a given proc

tests/stdlib/tsugar.nim
```diff
@@ -1,4 +1,72 @@
...
+  let f = () => 42
+  doAssert f() == 42
+
+  let g = (x: int) => x + 1
+  doAssert g(42) == 43
+
+  let h = (x, y: int) => x + y
+  doAssert h(1, 2) == 3
+
```
timotheecour: nit: i usually prefer f1, f2 etc insetead of f, g when there's a lot of such procs to test
```suggestion
  let f2 = () => ()
  doAssert f2() == () # empty tuple, not void

  const f3 = () => ()
  doAssert f3() == ()

  # const f4 = () => void # xxx we should have a way to generate a void-returning proc 
```

tests/stdlib/tsugar.nim
```diff
@@ -26,85 +94,95 @@ block dup_with_field:
...
-foo()
\ No newline at end of file
+
+block:
+  doAssert collect(for (i, d) in pairs(data): (i, d)) == @[(0, "bird"), (1, "word")]
+  doAssert collect(for d in data.items: (try: parseInt(d) except: 0)) == @[0, 0]
+  doAssert collect(for (i, d) in pairs(data): {i: d}) == {1: "word",
+      0: "bird"}.toTable
+  doAssert collect(for d in data.items: {d}) == data.toHashSet
+
```
timotheecour: future PR welcome to do the following: https://github.com/timotheecour/Nim/issues/534#issuecomment-769565033
(or in this PR if you can)

tests/stdlib/tsugar.nim
konsumlamm: Do you mean for all procs? E.g. for `collect`, there are quite a lot of tests.

timotheecour: at least the ones you're adding/editing would be nice

tests/stdlib/tsugar.nim
konsumlamm: That actually made me find a bug (#12595) with using `{k: v}` in templates.

tests/stdlib/tsugar.nim
konsumlamm: What's the point of testing `() => ()`?

Also, not sure what you mean by the last comment, returning `void` seems to work.

tests/stdlib/tsugar.nim
```diff
@@ -3,186 +3,213 @@ discard """
...
+      proc call3(f: (int, int) -> int): int = f(1, 2)
+      doAssert call3((x, y) => x + y) == 3
+      doAssert call3((x, y: int) => x + y) == 3
+      doAssert call3((x: int, y: int) => x + y) == 3
+
+      var a = 0
+      proc call4(f: int -> void) = f(42)
+      call4((x: int) => (a = x))
+      doAssert a == 42
```
timotheecour, line 45: add this:
```nim
      let f5 = () => (discard) # simplest proc that returns void
      f5()
```
and also add this to runnableExamples in sugar.=> as it's not obvious how to write it

tests/stdlib/tsugar.nim
```diff
@@ -3,186 +3,213 @@ discard """
...
+      doAssert call2((x: int) => x) == 42
+
+      proc call3(f: (int, int) -> int): int = f(1, 2)
+      doAssert call3((x, y) => x + y) == 3
+      doAssert call3((x, y: int) => x + y) == 3
+      doAssert call3((x: int, y: int) => x + y) == 3
+
+      var a = 0
+      proc call4(f: int -> void) = f(42)
+      call4((x: int) => (a = x))
```
timotheecour, line 44: can you also add a proc litteral that takes a pragma? (pick one, maybe `{.inline.}`?) ditto with runnableExamples to show its syntax, however obvious it may be

ditto for `->`

tests/stdlib/tsugar.nim
timotheecour: indeed, following up in https://github.com/nim-lang/Nim/pull/16820#discussion_r567195878

tests/stdlib/tsugar.nim
```diff
@@ -1,110 +1,215 @@
...
+      x = 10
+      y = 20
+    dump(x + y) # x + y = 30
+
+  let
+    x = 10
+    y = 20
+  dump(x + y) # x + y = 30
+
+# TODO: remove this and `when false` above once #12595 is fixed
```
konsumlamm: When #16874 gets merged, this can be put in the `main` template

tests/stdlib/tsugar.nim
```diff
@@ -1,110 +1,215 @@
...
+    block: # bug #14332
+      template foo =
+        discard collect(newSeq, for i in 1..3: i)
+      foo()
+
+static: main()
+main()
+
+block: # dump
+  # symbols in templates are gensym'd
```
timotheecour: you can use `let x {.inject.} = 10` to prevent that

tests/stdlib/tsugar.nim
```diff
@@ -1,110 +1,215 @@
...
+    doAssert collect(initTable(2), for (i, d) in data.pairs: {i: d}) ==
+      {0: "bird", 1: "word"}.toTable
+    doAssert collect(for (i, d) in pairs(data): {i: d}) ==
+      {1: "word", 0: "bird"}.toTable
+
+  let data = @["bird", "word"]
+  doAssert collect(initTable(2), for (i, d) in data.pairs: {i: d}) ==
+    {0: "bird", 1: "word"}.toTable
+  doAssert collect(for (i, d) in pairs(data): {i: d}) ==
+    {1: "word", 0: "bird"}.toTable
```
timotheecour: you can avoid code duplication as follows:

```nim
proc mainProc() =
  # symbols in templates are gensym'd
  block: # dump
    let
      x = 10
      y = 20
    dump(x + y) # x + y = 30

  # TODO: move to `main` and `when false` above pending bug #12595
  block: # collect pairs
    let data = @["bird", "word"]
    doAssert collect(initTable(2), for (i, d) in data.pairs: {i: d}) ==
      {0: "bird", 1: "word"}.toTable
    doAssert collect(for (i, d) in pairs(data): {i: d}) ==
      {1: "word", 0: "bird"}.toTable

static: mainProc()
mainProc()
```

tests/stdlib/tsugar.nim
konsumlamm: Good idea, though that made me rediscover bug #10609 (apparently using `{.inject.}` once applies to the whole template).

timotheecour: note that https://github.com/nim-lang/Nim/pull/16820#pullrequestreview-579800457 obsoletes this suggestion

tests/stdlib/tsugar.nim
```diff
@@ -1,110 +1,215 @@
...
+      let x = collect:
+        for d in data.items:
+          when d is int: "word"
+          else: d
+      doAssert x == @["bird", "word"]
+
+    block:
+      doAssert collect(for (i, d) in pairs(data): (i, d)) == @[(0, "bird"), (1, "word")]
+      doAssert collect(for d in data.items: (try: parseInt(d) except: 0)) == @[0, 0]
+      when false: # bug #12595
```
timotheecour: make sure to update this now that your fix to https://github.com/nim-lang/Nim/issues/12595 got merged :)

tests/stdlib/tsugar.nim
konsumlamm: So for some reason, `runnableExamples` transforms
```nim
let f = () => (discard)
```
to
```nim
let f = () =>
  discard
```
which doesn't compile... (but that doesn't happen when using for example `echo "test"` instead of `discard`)

timotheecour: i see, you're hitting https://github.com/nim-lang/Nim/issues/13491

lib/pure/sugar.nim
```diff
@@ -140,15 +139,20 @@ macro `=>`*(p, b: untyped): untyped =
...
     proc passTwoAndTwo(f: (int, int) -> int): int = f(2, 2)
-
     # is the same as:
     # proc passTwoAndTwo(f: proc (x, y: int): int): int = f(2, 2)
 
     doAssert passTwoAndTwo((x, y) => x + y) == 4
 
+    proc passOne(f: (int {.noSideEffect.} -> int)): int = f(1)
+    # is the same as:
+    # proc passOne(f: proc (x, y: int): int {.noSideEffect.}): int = f(1)
```
timotheecour:
```suggestion
    # proc passOne(f: proc (x: int): int {.noSideEffect.}): int = f(1)
```

tests/stdlib/tsugar.nim
konsumlamm: How should i proceed on this one?

timotheecour: * add it in runnableExamples but commented out
* add a comment in code: # xxx pending #13491: uncomment runnableExamples
* add a test in tsugar

timotheecour (approved):
LGTM; excellent.

future work can:
* maybe re-add to tsugar an improved version of the `mapIt` example that was removed (one where it would actually be needed); 
* add tests for `(x: int) -> int` (there are currently to tests for that, only  `int -> int`) and in particular ones showing you can pass a proc with differently named params to it

timotheecour:
still LGTM but in a followup PR (or feel free to do it here), I'll improve as follows:

* test in js too with:
```
discard """
  targets: "c js"
  output: '''
x + y = 30
'''
"""
```

* this requires only 1 disabled test for js: (please file a bug so we can link to it in this test `# pending bug #xyz`)
```nim
  block: # capture
    var closure1: () -> int
    when not defined js: # xxx gives: Error: cannot evaluate at compile time: i`gensym0
      for i in 0 .. 10:
        if i == 5:
          capture i:
            closure1 = () => i
      doAssert closure1() == 5

      var closure2: () -> (int, int)
      for i in 0 .. 10:
        for j in 0 .. 10:
          if i == 5 and j == 3:
            capture i, j:
              closure2 = () => (i, j)
      doAssert closure2() == (5, 3)
```

* and also put all in the main template, by using `inject` in just those places:

```nim

  block: # dump
    # symbols in templates are gensym'd
    let
      x {.inject.} = 10
      y {.inject.} = 20
    dump(x + y) # x + y = 30

  block: # dumpToString
    template square(x): untyped = x * x
    let x {.inject.} = 10
    doAssert dumpToString(square(x)) == "square(x): x * x = 100"
    let s = dumpToString(doAssert 1+1 == 2)
    doAssert "failedAssertImpl" in s
    let s2 = dumpToString:
    doAssert "except AssertionDefect" in s2

static: main()
main()
```
(I'm not seeing the bug you mentioned here: https://github.com/nim-lang/Nim/pull/16820#discussion_r567321609)

EDIT => see PR https://github.com/nim-lang/Nim/pull/16919

konsumlamm:
> test in js too

I'll leave that to another PR (I currently don't even have JS installed on my PC). What about other targets (C++, ObjC)? Also, shouldn't the default be "all targets"?

> (I'm not seeing the bug you mentioned here: [#16820 (comment)](https://github.com/nim-lang/Nim/pull/16820#discussion_r567321609))

When using `{.inject.}`, the identifier becomes injected for the whole template (i.e. the injection is not scoped), excluding declarations. So for example the following (from one of the tests for `collect`) will break:
```nim
let x = collect: # here the x is gensym'd
  for d in data.items:
    when d is int: "word"
    else: d
doAssert x == @["bird", "word"] # here the x is injected, due to the mentioned bug
```
This doesn't occur in your example, since every `x` is gensym'd there.

timotheecour:
> I'll leave that to another PR

sounds good, this is LGTM already, just needs another LGTM from 1 other person...

> (I currently don't even have JS installed on my PC).

time for https://nodejs.org/en/download/? this will help you in other PRs

> What about other targets (C++, ObjC)?

sad but there's currently only 1 test(!!) for objc: `tests/osproc/passenv.nim:4:3:  targets: "c cpp objc"`; this can be improved in future work; although swift is more common nowadays

> Also, shouldn't the default be "all targets"?

`cpp` is usually on a need by need basis, eg for really important tests or things that have failed in C++ before or things likely to fail in C++ or c++ specific stuff; the only problem with testing all targets is CI time to run those. 

- [ ] proposal: `targets: "default js"`
That said, CI runs `cpp`  in CI on 1 osx instance (see `NIM_COMPILE_TO_CPP`), but it's not bullet proof (eg when `targets: "c js"` is specified, it won't run cpp; PR welcome to change this to `targets: "default js"`)

(EDIT => https://github.com/nim-lang/RFCs/issues/334)

> So for example the following (from one of the tests for collect) will break:

not sure what you mean since it worked for me, but we can discuss this in next PR

ringabout (approved):

Issue #11021, opened 2019-04-13
Autogenerate Wep Api javascript bindings based on Web-IDL


planetis-m:
### Summary
> Web IDL is an interface description language (IDL) format for describing application programming interfaces (APIs) that are intended to be implemented in web browsers.

From wikipedia [article](https://en.wikipedia.org/wiki/Web_IDL)
### Additional Information
This [repo](https://github.com/awkward-ninja/nim-whatwg) has a script, that does exactly that.

Araq:
Should be provided as a Nimble package.

PR #16865, opened 2021-01-29
Add StorageEvent and correct Storage object

planetis-m:
There is no `LocalStorage` type, it's `Storage`. Also adds window.sessionStorage


planetis-m:
Ideally someone could look into #11021 and have that script autogenerate the dom module.

PR #16934, opened 2021-02-04
Allow strings for -- in config.nims

Clyybber:


juancarlospaco:
Changelog because is tiny but user-facing ?.

Clyybber:
> Changelog because is tiny but user-facing ?.

It's more of a bugfix, you'd expect `--path:"somePath"` to just work.

PR #16935, opened 2021-02-04
remove a unused temporary variable in int128.nim

ringabout:


compiler/int128.nim
```diff
@@ -176,7 +176,6 @@ proc toHex*(arg: Int128): string =
   result.addToHex(arg)
 
 proc inc*(a: var Int128, y: uint32 = 1) =
-  let input = a
```
ringabout, line 179: here, the unused temporary variable

PR #16941, opened 2021-02-05
[Random] Add 'rand' alias for sets/openArrays 'sample'

Wh1teDuke:
These procs used to exist, but after their removal, using random with generics/templates/macros became much harder. I added them for consistency
```nim
template myRandom(t): auto = rand(t) - 1 # This now works, no need for `when t is openArray: sample(t) - 1`
```

### DIFF
--- a/lib/pure/random.nim
+++ b/lib/pure/random.nim
@@ -462,6 +462,22 @@ proc sample*[T](a: openArray[T]): T =
     doAssert sample(marbles) == "red"
   result = a[rand(a.low..a.high)]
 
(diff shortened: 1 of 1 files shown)


Araq:
Nah, it's now called ``sample``.

Wh1teDuke:
> Nah, it's now called `sample`.

Yes, and I'm arguing in favor of the addition of the 'rand' alias to make it more consistent with the rest of the module.

Araq:
I'm happier with the "inconsistency" as choosing a value from a set of possible values is fundamentally different than ``rand`` which always return a number.

PR #16960, opened 2021-02-07
RST: implement footnotes and citations

a-mr:
All 5 types are implemented (see RST spec [Footnotes](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#footnotes) and [Footnote References](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#footnote-references)):
* footnotes
   * manually numbered
   * auto-numbered
   * auto-numbered with label
   * auto-symbol footnotes
* citations

Example:
```rst
Nim can output the result to HTML [#html]_ or Latex [#latex]_.

.. [#html] commands ``nim doc`` for ``*.nim`` files and
   ``nim rst2html`` for ``*.rst`` files

.. [#latex] command ``nim rst2tex`` for ``*.rst``.

* directives: ``code-block`` [cmp:Sphinx]_, ``title``,
  ``index`` [cmp:Sphinx]_

.. [cmp:Sphinx] similar but different from the directives of
   Python `Sphinx directives`_ extensions

.. [2] Manually numbered.

.. [*] Auto-symbol1
.. [*] Auto-symbol2
.. [#] Auto-numbered

Ref [*]_ [*]_ [#]_.
```
The result for HTML:
image

cc @narimiran

### DIFF
--- a/lib/packages/docutils/rst.nim
+++ b/lib/packages/docutils/rst.nim
@@ -694,11 +694,7 @@ proc orderFootnotes(p: var RstParser) =
 
   # Report an error if found any mismatch in number of automatic footnotes
   proc listFootnotes(lines: seq[int]): string =
-    result.add $lines.len
-    if lines.len > 0:
-      result.add " (lines " & $lines[0]
-      for line in lines[1..^1]: result.add ", " & $line
-      result.add ")"
+    result.add $lines.len & " (lines " & join(lines, ", ") & ")"
   if p.s.lineFootnoteNum.len != p.s.lineFootnoteNumRef.len:
     rstMessage(p, meFootnoteMismatch,
       "$1 != $2" % [listFootnotes(p.s.lineFootnoteNum),
--- a/lib/packages/docutils/rst.nim
+++ b/lib/packages/docutils/rst.nim
@@ -722,7 +722,8 @@ proc orderFootnotes(p: var RstParser) =
   else:
     # fill gaps between manually numbered footnotes in ascending order
     manuallyN.sort()  # sort by number - its first field
-    var lst = manuallyN.toSinglyLinkedList()
+    var lst = initSinglyLinkedList[FootnoteSubst]()
+    for elem in manuallyN: lst.append(elem)
     var firstAuto = 0
     if lst.head == nil or lst.head.value.number != 1:
       # no manual footnote [1], start numeration from 1 for auto-numbered
@@ -731,7 +732,7 @@ proc orderFootnotes(p: var RstParser) =
     var curNode = lst.head
     var nextNode: SinglyLinkedNode[FootnoteSubst]
     # go simultaneously through `autoN` and `lst` looking for gaps
-    for (kind, _, autoNumIdx, _, label) in autoN[firstAuto .. ^1]:
+    for (kind, x, autoNumIdx, y, label) in autoN[firstAuto .. ^1]:
       while (nextNode = curNode.next; nextNode != nil):
         if nextNode.value.number - curNode.value.number > 1:
           # gap found, insert new node `n` between curNode and nextNode:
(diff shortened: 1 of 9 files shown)


lib/packages/docutils/rst.nim
```diff
@@ -2038,9 +2265,69 @@ proc selectDir(p: var RstParser, d: string): PRstNode =
   else:
     rstMessage(p, meInvalidDirective, d)
 
+proc parseFootnoteContent(p: var RstParser, father: var PRstNode) =
```
narimiran: `Hint: 'parseFootnoteContent' is declared but not used`

lib/packages/docutils/rst.nim
```diff
@@ -2111,14 +2396,82 @@ proc resolveSubs(p: var RstParser, n: PRstNode): PRstNode =
       if s != "":
         result = newRstNode(rnInternalRef)
         n.kind = rnInner
-        result.add(n)
-        result.add(newRstNode(rnLeaf, s))
+        result.add(n)  # visible text of reference
+        result.add(newRstNode(rnLeaf, s))  # link itself
+  of rnFootnote:
+    let (fnType, i) = getFootnoteType(n.sons[0])
```
narimiran: `Hint: 'i' is declared but not used`

narimiran (commented):
IMO, footnotes after a paragraph, in the middle of a section, seem kinda displaced, and one might ask "why not just write that stuff 'inline', instead of one line below?".

What about placing footnotes before the next section (or before the end of the document if there's no next section)? Something like this:

imgur

and

imgur

a-mr:
> What about placing footnotes before the next section (or before the end of the document if there's no next section)?

Please note that footnotes are placed where you place them in the RST text. So you can just place them at the end of section manually.

I think your proposal is not difficult to implement, but I did current implementation intentionally. And python rst2html also works this way.

(Here I'm actually pushing my personal preference.)

The point is that reading paragraph you get the main idea of it, and reading footnotes you get clarifications. Footnotes are like procedures in programming and links to them are like proc calls. As much you may choose to read footnotes or just skip them. But if you need some clarification it's essential to get it fast — without need to scroll down or even move your eyes too far. That's why it can be very convenient when footnotes are placed right below the paragraph.

Of course, as you said, one can use inline text (e.g. in parentheses) but that disrupts the natural flow of reading.

What I don't like the most about footnotes in the end of a section — is that in case of long sections you don't see footnotes and need to scroll down in HTML or jump links. In Latex we could use normal footnotes in the bottom of page, but I like the current style more, some other people also prefer it, that's why there exist Latex packages like [parnotes](http://www.ctan.org/pkg/parnotes).

**I'd like to know opinions of other people on that topic.**

narimiran:
> I'd like to know opinions of other people on that topic.

+1

lib/packages/docutils/rst.nim
```diff
@@ -966,7 +1114,60 @@ proc parseMarkdownLink(p: var RstParser; father: PRstNode): bool =
...
+    try:
+      result = (fnManualNumber, parseInt(label.sons[0].text))
+    except:
+      result = (fnCitation, -1)
+  else:
+    result = (fnCitation, -1)
+
+proc validRefnamePunct(x: string): bool =
+  ## https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#reference-names
+  x in ["-", "_", ".", ":", "+"]
```
Araq:
```suggestion
  x.len == 1 and x[0] in {'-', '_', '.', ':', '+'}
```

lib/packages/docutils/rstast.nim
```diff
@@ -77,6 +79,8 @@ type
                               ## the document or the section; and rnEnumList
                               ## and rnAdmonition; and rnLineBlockItem
     level*: int               ## valid for headlines/overlines only
+    order*: int               ## footnote order (for auto-symbol footnotes and
```
Araq, line 82: Can't you use `level` too here? level is currently only used for headlines and overlines.

a-mr: I can of course.

But what's the purpose? Isn't it too blatant to reuse other names for different unrelated goals?

I expected to move in the opposite direction in the future and split also `text` field into 2 fields `text` and `info` to avoid confusion.

Araq: Memory consumption is always an issue. If you don't like to re-use fields, make the object contain a ``case``. :-) Can be done in a followup PR, of course.

PR #16966, opened 2021-02-08
Remove declPragmas from lambdas [backport:1.0]

metagn:
Refs #16967

It makes little sense for any of `declPragmas` to be in `lambdaPragmas`. `align` should also only be for object fields and variables, as in [manual](https://nim-lang.org/docs/manual.html#implementation-specific-pragmas-align-pragma)


metagn:
This does not fix #16967, will open separate PR for that, for now just adding simple test so `exportc` doesn't compile

metagn:
CI failure seems unrelated? But is a weird failure nonetheless

PR #16980, opened 2021-02-09
tests and docs for call operator

### LINKED ISSUE #16183: Document `callOperator`
Dot operators are currently documented in the experimental manual, however `{.experimental: "callOperator".}` is not. Perhaps because it's too broken (#9831 looks like it might have some flaws), but it's an experimental feature anyway.

metagn:
Closes #16183

Separated from #16924, no breaking changes


timotheecour (approved):
LGTM (see https://github.com/nim-lang/Nim/pull/16924 for review discussions that led to this)

yglukhov (approved):

PR #17001, opened 2021-02-10
[JS] Ref #15952 make toOpenArray works better

ringabout:
ref #15952 toOpenArray works better

### DIFF
--- a/compiler/jsgen.nim
+++ b/compiler/jsgen.nim
@@ -1291,8 +1291,6 @@ template isIndirect(x: PSym): bool =
     v.kind notin {skProc, skFunc, skConverter, skMethod, skIterator,
                   skConst, skTemp, skLet})
 
-import astalgo
-
 proc genAddr(p: PProc, n: PNode, r: var TCompRes) =
   case n[0].kind
   of nkSym:
--- a/compiler/jsgen.nim
+++ b/compiler/jsgen.nim
@@ -14,27 +14,28 @@ The JS code generator contains only 2 tricks:
 
 Trick 1
 -------
-Some locations (for example 'var int') require "fat pointers" (``etyBaseIndex``)
+Some locations (for example 'var int') require "fat pointers" (`etyBaseIndex`)
 which are pairs (array, index). The derefence operation is then 'array[index]'.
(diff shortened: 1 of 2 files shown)


timotheecour:
@Araq I added a label `backport`, just remove that new label if you disagree; I think that's easier than having to specify it in commit msg since it allows decision to be made/unmade after a PR is merged; 
alternatively we can add 3 labels: backport_10, backport_20, backport_40 instead of backport

Araq:
That's for @narimiran to decide as he is doing the backports.

timotheecour:
I've removed the backport label from this PR in light of https://github.com/nim-lang/Nim/pull/17006 (but the label itself is still useful IMO, to tag whether or not to backport PR's)

Araq:
The label is useful but more useful would be "backport-1.2" etc.

timotheecour:
> The label is useful but more useful would be "backport-1.2" etc.

done, added those; only 1 need to be specified each time a backport is needed:
image

It looks like this PR was backported to 1.4.4, but we may need to revisit this for 1.4.6 because of https://github.com/nim-lang/Nim/pull/17006

> note that toOpenArray with var param doesn't seem to actually work (ie source data isn't mutated), so it's not clear that #17001 actually improved things (it turned an error into a silent non-mutation)

narimiran:
Please continue to use commit messages containing `[backport*]` tag for stuff that needs to be backported.

These labels might be easier for y'all to tag, but harder for me to do backports. I'm removing them so we don't have stuff labeled, but without tag.

timotheecour:
> Please continue to use commit messages containing [backport*] tag for stuff that needs to be backported.

the problem with commit messages is that it requires the decision to be made before the PR is merged; a label gives you more flexibility and hindsight since you can tag/untag after a PR is merged; in this case for example, it was discovered after the PR was merged that it causes an issue and probably shouldn't have been backported.

Araq:
Sure but @narimiran is doing the work so we have to adapt. However, maybe there could be an official backport queue instead that lists commits to backport?

metagn:
What's the followup needed label for?

PR #17015, opened 2021-02-11
Improve documentation for random

konsumlamm:
* use `runnableExamples`
* other minor changes

I noticed that the documentation for `initRand(int64)` is generated twice (see https://github.com/nim-lang/Nim/pull/16953#discussion_r574811739), if anybody knows a workaround for this, please let me know.

### DIFF
--- a/lib/pure/random.nim
+++ b/lib/pure/random.nim
@@ -25,28 +25,28 @@ runnableExamples:
   # examples are run.
   randomize()
 
-  # Pick a number between 0 and 100.
+  # Pick a number in 0..100.
   let num = rand(100)
-  echo num
+  doAssert num in 0..100
 
   # Roll a six-sided die.
   let roll = rand(1..6)
-  echo roll
+  doAssert roll in 1..6
 
   # Pick a marble from a bag.
   let marbles = ["red", "blue", "green", "yellow", "purple"]
   let pick = sample(marbles)
-  echo pick
+  doAssert pick in marbles
 
   # Shuffle some cards.
   var cards = ["Ace", "King", "Queen", "Jack", "Ten"]
   shuffle(cards)
-  echo cards
+  doAssert cards.len == 5
 
-## These examples all use the default random number generator. The
-## `Rand type <#Rand>`_ represents the state of a random number generator.
+## These examples all use the default RNG. The
+## `Rand type <#Rand>`_ represents the state of an RNG.
 ## For convenience, this module contains a default Rand state that corresponds
-## to the default random number generator. Most procs in this module which do
+## to the default RNG. Most procs in this module which do
 ## not take in a Rand parameter, including those called in the above examples,
 ## use the default generator. Those procs are **not** thread-safe.
 ##
@@ -93,7 +93,7 @@ type
                  ## Create a new Rand state using the `initRand proc <#initRand,int64>`_.
                  ##
                  ## The module contains a default Rand state for convenience.
-                 ## It corresponds to the default random number generator's state.
+                 ## It corresponds to the default RNG's state.
                  ## The default Rand state always starts with the same values, but the
                  ## `randomize proc <#randomize>`_ can be used to seed the default generator
                  ## with a value based on the current time.
@@ -149,8 +149,8 @@ proc next*(r: var Rand): uint64 =
 proc skipRandomNumbers*(s: var Rand) =
   ## The jump function for the generator.
   ##
-  ## This proc is equivalent to 2^64 calls to `next <#next,Rand>`_, and it can
-  ## be used to generate 2^64 non-overlapping subsequences for parallel
+  ## This proc is equivalent to `2^64` calls to `next <#next,Rand>`_, and it can
+  ## be used to generate `2^64` non-overlapping subsequences for parallel
   ## computations.
   ##
   ## When multiple threads are generating random numbers, each thread must
@@ -165,7 +165,7 @@ proc skipRandomNumbers*(s: var Rand) =
   ## Rand state to a thread, call this proc before passing it to the next one.
   ## By using the Rand state this way, the subsequences of random numbers
   ## generated in each thread will never overlap as long as no thread generates
-  ## more than 2^64 random numbers.
+  ## more than `2^64` random numbers.
   ##
   ## **See also:**
   ## * `next proc<#next,Rand>`_
@@ -178,7 +178,7 @@ proc skipRandomNumbers*(s: var Rand) =
     proc randomSum(rand: Rand): int =
       var r = rand
       for i in 1..numbers:
-        result += rand(1..10)
+        result += rand(0..10)
 
     var r = initRand(2019)
     var vals: array[spawns, FlowVar[int]]
@@ -187,7 +187,7 @@ proc skipRandomNumbers*(s: var Rand) =
       r.skipRandomNumbers()
 
     for val in vals:
-      echo ^val
+      doAssert abs(^val - numbers * 5) / numbers < 0.1
 
   when defined(js):
     const helper = [0xbeac0467u32, 0xd86b048bu32]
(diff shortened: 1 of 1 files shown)


timotheecour:
> I noticed that the documentation for initRand(int64) is generated twice (see #16953 (comment)), if anybody knows a workaround for this, please let me know.

we should simply fix the compiler instead of using a workaround. Shouldn't be too hard IMO

EDIT: but please file a bug, ideally with a minimized example

lib/pure/random.nim
```diff
@@ -7,9 +7,9 @@
 #    distribution, for details about the copyright.
 #
 
-## Nim's standard random number generator.
+## Nim's standard random number generator (RNG).
```
timotheecour, line 10: now that you've defined RNG in a very visible place on top, how about using RNG everywhere in place of `random number generator` ?

konsumlamm: I didn't want to exaggerate using it (in fact I only used it in one place), but sure, I can do that.

lib/pure/random.nim
```diff
@@ -19,69 +19,63 @@
...
-##
+runnableExamples:
+  # Call randomize() once to initialize the default random number generator.
+  # If this is not called, the same results will occur every time these
+  # examples are run.
+  randomize()
+
+  # Pick a number between 0 and 100.
+  let num = rand(100)
+  echo num
```
timotheecour:
```suggestion
  doAssert num <= 100 and num >= 0
```
 ditto in other instances

lib/pure/random.nim
```diff
@@ -19,69 +19,63 @@
...
-##   shuffle(cards)
-##   echo cards
-##
+runnableExamples:
+  # Call randomize() once to initialize the default random number generator.
+  # If this is not called, the same results will occur every time these
+  # examples are run.
+  randomize()
+
+  # Pick a number between 0 and 100.
```
timotheecour:
```suggestion
  # Pick a number in 0..100.
```
(english is not precise enough, 0..100 is better)

lib/pure/random.nim
```diff
@@ -154,12 +149,12 @@ proc next*(r: var Rand): uint64 =
 proc skipRandomNumbers*(s: var Rand) =
   ## The jump function for the generator.
   ##
-  ## This proc is equivalent to 2^64 calls to `next<#next,Rand>`_, and it can
+  ## This proc is equivalent to 2^64 calls to `next <#next,Rand>`_, and it can
```
timotheecour:
```suggestion
  ## This proc is equivalent to `2^64` calls to `next <#next,Rand>`_, and it can
```

konsumlamm:
> > I noticed that the documentation for initRand(int64) is generated twice (see #16953 (comment)), if anybody knows a workaround for this, please let me know.
> 
> we should simply fix the compiler instead of using a workaround. Shouldn't be too hard IMO
> 
> EDIT: but please file a bug, ideally with a minimized example

Done, see  #17016.

lib/pure/random.nim
```diff
@@ -172,32 +167,28 @@ proc skipRandomNumbers*(s: var Rand) =
...
+        result += rand(1..10)
+
+    var r = initRand(2019)
+    var vals: array[spawns, FlowVar[int]]
+    for val in vals.mitems:
+      val = spawn randomSum(r)
+      r.skipRandomNumbers()
+
+    for val in vals:
+      echo ^val
```
timotheecour:
```suggestion
      assert abs(^val - numbers * 5) / numbers < 0.1
```

konsumlamm: This doesn't pass, I'd just check `^val in numbers..(numbers * 10)`.

lib/pure/random.nim
dom96: How will someone that doesn't have a background in- well, hard to say what really, math? comp sci?- know what `..` means?

We should use English here, if it's not accurate then we can certainly make it so at the expense of verbosity.

timotheecour: `..` vs `..<` is litterally used everywhere in nim, it's what you learn within the 1st hour.
A user who encounters `std/random` will have encountered `..` long before.
`between 0 and 100` immediately raises the question of whether 100 is inclusive or exclusive, whereas `0..100` is un-ambiguous.

lib/pure/random.nim
timotheecour: sorry i had also changed `result += rand(1..10)` to `result += rand(0..10)` to make it pass; it's an intuitive test, showing no statistical bias. Better than the weaker `assert abs(^val - numbers * 5) / numbers < 0.1`

konsumlamm: So you mean I should use `result += rand(0..10)` or rather my version?

timotheecour: yes `result += rand(0..10)`; just makes the expected sum more intuitive

lib/pure/random.nim
```diff
@@ -19,70 +19,63 @@
...
 ## See also
 ## ========
-## * `std/sysrand module<sysrand.html>`_ for cryptographically secure pseudorandom number generator
-## * `math module<math.html>`_ for basic math routines
-## * `mersenne module<mersenne.html>`_ for the Mersenne Twister random number
-##   generator
-## * `stats module<stats.html>`_ for statistical analysis
-## * `list of cryptographic and hashing modules
-##   <lib.html#pure-libraries-hashing>`_
+## * `std/sysrand module <sysrand.html>`_ for a cryptographically secure pseudorandom number generator
```
timotheecour, line 68:
```suggestion
## * `std/sysrand module <sysrand.html>`_ for a cryptographically secure pseudorandom number generator (CSPRNG)
```

lib/pure/random.nim
```diff
@@ -257,23 +246,23 @@ proc rand*(max: int): int {.benign.} =
...
+  ## * `rand proc<#rand,float>`_ that returns a float using the default RNG
   ## * `rand proc<#rand,Rand,Natural>`_ that returns an integer
   ## * `rand proc<#rand,Rand,HSlice[T: Ordinal or float or float32 or float64,T: Ordinal or float or float32 or float64]>`_
   ##   that accepts a slice
   ## * `rand proc<#rand,typedesc[T]>`_ that accepts an integer or range type
   runnableExamples:
     var r = initRand(234)
-    let f = r.rand(1.0)
-    ## f = 8.717181376738381e-07
+    let f = r.rand(1.0) # 8.717181376738381e-07
```
timotheecour, line 264: pre-existing but that's a bug (doesn't affect this PR though); will explain later

lib/pure/random.nim
```diff
@@ -171,34 +165,30 @@ proc skipRandomNumbers*(s: var Rand) =
...
+  runnableExamples("--threads:on"):
+    import std/[random, threadpool]
+
+    const spawns = 4
+    const numbers = 100000
+
+    proc randomSum(rand: Rand): int =
+      var r = rand
+      for i in 1..numbers:
+        result += rand(0..10)
```
timotheecour: pre-existing but I don't understand this code.

timotheecour (approved):
LGTM; remaining points can be addressed in future PRs

lib/pure/random.nim
timotheecour: it's: https://github.com/timotheecour/Nim/issues/435

lib/pure/random.nim
Jjp137: Since I wrote this, yeah I just needed something that wouldn't take a whole page (the rest of the code is already long enough for an example) so I wrote something really simple and boring, which is "let's just add a bunch of random numbers together". It might be a good idea to add a comment that goes like "this is a placeholder for demonstration purposes" near that proc.

If you're talking about why `rand` needs to be copied, then I think it's so that each thread's sequence of random numbers doesn't overlap with another thread's sequence.

Honestly, as someone who barely delves into multi-threading, documenting `skipRandomNumbers` was difficult for me so I had to rely on this comment which was really helpful: https://github.com/nim-lang/Nim/pull/10546#issuecomment-461541312

Hope that helps.

timotheecour: but i don't understand the logic.
`var r = rand` => `r` isn't used after this assignment
`result += rand(1..10)` => this uses randState, not `r`

so I'm assuming you meant:
```nim
result += rand(r, 0..10)
```

Jjp137: lol yeah oops

PR #17020, opened 2021-02-12
[superseded] Adding support for user-defined number suffixes

JohnAD:
As a precursor for adding a decimal (base10) library to Nim's stdlib, this PR adds support for user-defined suffixes to numeric literals.

https://github.com/nim-lang/RFCs/issues/308 - Add decimal to standard library

This PR also solves:

https://github.com/nim-lang/RFCs/issues/216 - User defined integer literals
https://github.com/nim-lang/RFCs/issues/228 - user defined quote literals, eg: -12.3E+4022'dec128 ; -128'i8 is transformed as `i8("-128")`

To distinguish between a typical string literal and a string literal that is the result of a numeric suffix, a new token was added to the lexer: `tkStrNumLit`. This token is only generated if, when processing a numeric literal, the suffix does not match one of the pre-defined suffixes used internally.

The rest of the work is done in the parser. When the `tkStrNumLit` token is detected, a "dot" is presumed to follow and the following symbol is considered the indentifier called in a "dotExpr" syntax. So, effectively:

```nim
var a = 12345.678'something
```
becomes the equivalent of:

```nim
var a = "12345.678".something
```

With this, a library, including the up-and-coming decimal library, can parse numbers in source code that cannot be handled by even `int64` or `float64`. For example:

```nim
var a = 5192296858534827628530496329220095'm
```

(The decimal library plans on using `m` as the suffix to match the convention seen in other languages such as C#.)

While I've written tests and don't know of any bugs, this is my first attempt at modifying the core lexer/parser of a compiler, I've marked the PR as Work-In-Progress [WIP]. _Any and all code review is very much appreciated._

## links
* EDIT(timotheecour) see bug list in https://github.com/nim-lang/Nim/pull/17020#issuecomment-803193947


JohnAD:
While I didn't add this check to the lexer, I could see a benefit to it:

I could easily forbid suffixes that start with the letters `a` to `f` and `A` to `F`. Doing this would remove the confusion that can be seen with both hex numbers and numbers with the `e` exponent (such as `12.3e22`.)

Let me know if adding this restriction to the lexer (and adding it to the documentation) is desirable.

tests/lexer/tstrnumlits.nim
```diff
@@ -0,0 +1,91 @@
...
+  likeFloat      = 12345.67890wrap
+  quotedFloat    = 12345.67890wrap # 12345.67890'wrap
+  smushed        = 1wrap*1wrap
+  borderCase     = 1wrap + -1wrap
+  usingTemplate  = 1twrap
+  inProc         = extraContext()
+  hex            = 0x5a3awrap  # Note: starting a suffix with the letters `a` to `f` is NOT a good idea
+  reallyBig      = -38383839292839283928392839283928392839283.928493849385935898243e-50000wrap
+
+stdout.write(likeInteger.string)
```
timotheecour: `doAssert likeInteger.string == ...`

tests/lexer/tstrnumlits.nim
```diff
@@ -0,0 +1,91 @@
...
+  result = (left.string & "plus" & right.string).S
+
+proc `-`(number: S): S =
+  result = ("neg" & number.string).S
+
+proc `$`(number: S): string =
+  result = number.string
+
+const
+  likeInteger    = 1wrap
```
timotheecour: I don't think we should allow this; instead there should be a quote: `1'wrap`

tests/lexer/tstrnumlits.nim
```diff
@@ -0,0 +1,91 @@
...
+  result = wrap(number)
+
+proc ia(number: string): S =   # proc starts with 'i' just like 'i64'
+  result = wrap(number)
+
+proc ua(number: string): S =   # proc starts with 'u' just like 'u8'
+  result = wrap(number)
+
+
+assert $1234.56wrap == $1234.56fa  # using a suffix that starts with `a` to `f` is allowed but not wise.
```
timotheecour: use doAssert in tests, not assert (so that they remain meaningful with --assertions:off or -d:danger)

tests/lexer/tstrnumlits.nim
```diff
@@ -0,0 +1,91 @@
+discard """
+  output: "[[1]]neg[[1]][[1]][[12345.67890]][[12345.67890]][[1]]times[[1]][[1]]plusneg[[1]][[1]][[22.40]][[0x5a3a]]neg[[38383839292839283928392839283928392839283.928493849385935898243e-50000]]"
+"""
+# Test tkStrNumLit
+
```
timotheecour: add some tests showing what the AST looks like:
```nim
import std/macros
macro deb(a: untyped): untyped =
  result = newLit(a.treeRepr)
doAssert deb(-123'foo) == ...
```

tests/lexer/tstrnumlits.nim
```diff
@@ -0,0 +1,91 @@
...
+  hex            = 0x5a3awrap  # Note: starting a suffix with the letters `a` to `f` is NOT a good idea
+  reallyBig      = -38383839292839283928392839283928392839283.928493849385935898243e-50000wrap
+
+stdout.write(likeInteger.string)
+stdout.write(negative.string)
+stdout.write(quoted.string)
+stdout.write(likeFloat.string)
+stdout.write(quotedFloat.string)
+stdout.write(smushed.string)
+stdout.write(borderCase.string)
```
timotheecour: avoid defining symbols used just once and instead use them inline, which makes the test easier to read, eg:
```nim
doAssert string(1wrap + -1wrap) == ...
```

doc/manual.rst
```diff
@@ -565,6 +565,11 @@ The type suffixes are:
   ``'f64``           float64
 =================    =========================
 
+If the suffix is not predefined, then the suffix is assumed to be a call
+to a proc, template, macro or other callable identifier that is passed the
+string containing the literal. Essentially,  ``1234.56'sample`` is
+the equivalent of ``sample("1234.56")``.
```
timotheecour: use `-128'i8custom` as example, it's more interesting as it illustrate something the current parser has a problem with (https://github.com/timotheecour/Nim/issues/125 -128'i8 gives: Error: number out of range) and that user defined literals could fix

doc/manual.rst
```diff
@@ -545,7 +545,7 @@ is optional if it is not ambiguous (only hexadecimal floating-point literals
 with a type suffix can be ambiguous).
 
 
-The type suffixes are:
+The pre-defined type suffixes are:
```
timotheecour, line 549: open question: could we add `proc f64(a: string): float64` etc to system.nim to make those non-builtin?
what would this break? beside fixing issues like https://github.com/timotheecour/Nim/issues/125

doc/manual.rst
Araq: It's not a parser problem, it's.a spec problem: Unary minus is always separate and never part of a literal. This should be fixed within this PR indeed.

tests/lexer/tstrnumlits.nim
ghost: @timotheecour I think it was made so it's consistent with normal Nim numbers - "1i32" is valid

timotheecour: good point, but it causes ambiguities, eg: `echo 12e32` : is it `"12".e32` or the FP `12.0*10^32` ?

i don't see the point of supporting the non-quoted version

another example: `0x32`, which could be `"0".x32` or the hex `0x32`.
etc there are other examples.

Araq:
Two thoughts:

1. Unary minus is an operator and never part of a literal. This sucks for cases like ``-128'i8``. This is an existing problem with the current spec and can easily be fixed within this PR.
2. The design encourages the existing practice of having short top-level proc names like ``re``. I am not sure if this existing practice is so solid that we shouldn't muse about it. For example, in C++ procs that are called as a literal suffix have to defined as ``operator "" X`` (with an empty string literal). However, that doesn't address the scoping aspect.

Araq:
> I could easily forbid suffixes that start with the letters a to f and A to F. Doing this would remove the confusion that can be seen with both hex numbers and numbers with the e exponent (such as 12.3e22.)

> Let me know if adding this restriction to the lexer (and adding it to the documentation) is desirable.

I'm not sure how it currently works, but floating point values containing 'e' and hexadecimal numbers need to use the ``'suffix`` (with the single quote) form. It would be really weird if my user defined literal could not start with A to F.

timotheecour:
> The design encourages the existing practice of having short top-level proc names like re. I am not sure if this existing practice is so solid that we shouldn't muse about it.

how about requiring a pragma, which mostly addresses this concern, and makes it more searchable, eg:
```nim
proc big(a: string) {.litteral.} = # create a JsBigInt from a
echo -123'big # ok
proc bad(a: string) = # create a JsBigInt from a
echo -123'bad # error, bad must have {.litteral.}
```

(big can still be used in other context; but the pragma also allows it to be used in a litteral context)

Araq:
A pragma doesn't address the scoping problem either. It's not that I want to annotate the suffix declaration, I am thinking about a scope extension. For example:

```nim

from bignums import nil

echo 355'bignum # works because it's turned into

echo bignums.bignum("355")

```

If it weren't so silly I would propose that ``123'suffix`` is turned into ``suffix.suffix("123")``. (Or into ``suffixs.suffix("123")`` as the plural ``s`` for a module name is very common.)

timotheecour:
that's the exact same problem with MCS though.

but here, you can use:

## option 1: no scope extension
(but IIUC your problem is "small identifiers" polluting scope?)
```nim
from bignums import bignum
echo 355'bignum
```

## option 2: special litteral rule
```nim
from bignums import 'bignum # makes it only usable as a litteral
echo 355'bignum # ok
echo bignum("355") # `bignum` not declared, was imported via 'bignum
```
but isn't option 1 good enough?

Araq:
> but IIUC your problem is "small identifiers" polluting scope?

Yeah and you're right, we should simply do nothing about it. There seems to be no good solution and the problem is imaginary.

Araq:
Oh there is an option: Suffix procs need to be declared as 

```nim

proc `'bignum` (...)

```

And then the rewrite rule is:

`123'suffix` is turned into `` `'suffix`("123") ``

And then we can import suffixes just like:

``from bignums import `'bignum` ``

timotheecour:
> Oh there is an option: Suffix procs need to be declared as

seems good; in particular it helps with turning f32 (etc) into a user defined litteral (defined in system.nim via ```proc `'f32'*(a: string): float32)``` without polluting scope with `f32`, since we'd have: 
```nim
doAssert not declared(f32)
doAssert declared(`'f32`)
```

this also encourages 1 way to write things:
123'suffix instead of suffix("123"), since the declaration would contain `'`

a user could of course define both, but that'd be opt in:
```nim
proc `'bignum`(...)
proc bignum(...)
```

* can we also deprecate `123f32` syntax and require `123'f32`?
if not then at least not allow the quote-less form for user defined litterals?

1 way to do things seems better (more searchable, etc); see also my argument about ambiguities this would cause otherwise (12e32 vs 12'e32 etc). This would help turning the builtin f32 (etc) into user defined literals, with obvious advantages.

JohnAD:
I like the two big ideas that have emerged from this discussion:

* For suffix procs, require a `'suffix` style declaration. Also require the apostrophe for usage. This makes things cleaner/clearer.
* Modify the lexer so that the minus symbol is part of the numeric literals when the context calls for it.

I'll start work on both; starting with the lexer itself.

Slightly off-topic: after this change is released, I/we will need to encourage linters and code markup tools (such as prettifiers) to properly handle the unpaired "single quotes". I use the sublime text editor and I'll need to fix/hack that sooner rather than later.)

Araq:
> if not then at least not allow the quote-less form for user defined literals?

I doubt we can deprecate the ``0f32`` syntax. But the new form can require the single quote, we can always allow to leave it out later.

JohnAD:
Started work on these changes. Just wrote the new test cases.

But, since we are talking "big picture", would there be value in also allowing more than [0-9a-fA-F] characters in a number?

Perhaps with restrictions such as:

1. It still starts with a numeric digit [0-9].
2. It has an attached `'suffix` after the character sequence.
3. The character sequence does not contain whitespace.
4. The character sequence does not certain punctuation characters such as full-quotes.

This would allow for useful things like this:

```nim
import korean

var a = 0k열여섯'hangul
```
or

```nim
import numericsystems

var a = 0s03jKi3h'base60
```

or

```nim
import roman

assert 256 == 0CCLVI'roman
```

or crazy stuff like:

```nim
import moneystuff
var a = 1,234,332.20'USD
```

I, myself, don't need this functionality personally, so I'm not pushing hard for it. But, I can see it being useful for other libraries.

I can also see danger in allowing any punctuation other than dots and underscores. So perhaps even if we allow more than the just digit/hex characters, punctuation should not be allowed.

timotheecour:
> This would allow for useful things like this:

user can always write: `var a = "열여섯".hangul`, this is probably not common enough to allow it via user defined literal.

> or crazy stuff like:

Commas would be more common and useful to support but I don't see how to avoid the ambiguity this would lead to: `echo (1,234,567'big)` => is it `(1, 234, big"567")` or `(big"1234567")` ? you'd need some arbitrary rule to disambiguate. And we can use "_" for that.

In any case, this could be handled in future work.

tests/lexer/tstrnumlits.nim
hlaaftana: Can't we specify that proper `e32` and `x32` etc always take priority? Couldn't we also require quotes specifically for hex or binary or engineering notation? So `1wrap` is fine, but `1e32wrap` must be `1e32'wrap`.

Araq:
> I, myself, don't need this functionality personally, so I'm not pushing hard for it. But, I can see it being useful for other libraries.

We can always add it later.

Araq:
@JohnAD what's the current state of this PR? I really want it... :-)

JohnAD:
Started work again on this today. I have already added the expanded/corrected unit testing. Should have a new commit for the lexer/parser by end of tomorrow.

JohnAD:
Open question:

The minus symbol (`-`) has two purposes:
  1. acting as sign for a number aka "negative"  `x = 3+ -4`
  2. acting as an operation indicating an operation  `x = 3-4`

And, if we were doing things in the higher levels of the language, I suspect I could easily distinguish between them. But, I'm trying to get the "negative" usage into the numeric literal where it really makes sense. To do that, the logic needs to be in the lexer, which is literally a character-by-character scan. 

In the end, being conservative, I suspect any numeric library will still need to handle the unary negative.

I just added the ability for the lexer to "remember" the previous tokType in the main loop. This solved multiple problems for me. 

So, if the previous tokType is in the

`{tkComma, tkColon, tkParLe, tkBracketLe, tkCurlyLe}`

set **or** is preceded by whitespace, then a `-` followed by a numeric digit will be parsed into the numeric literal. Otherwise the `-` is parsed as an operation.

Sound reasonable? Suggestions?

(Off-topic, but IMO, the conversion to float/int (and the handling of different bases) for numeric literals should not be in the lexer. That is beyond the scope of a lexer. But that is me being nitpicky.  And moving it would be a thought for another day.)

JohnAD:
Still to do:
- [x] Add AST tests to unit testing
- [x] Test invalid negation, such as against a UInt or out-of-range, in unit testing
- [x] Update manual.rst

compiler/lexer.nim
```diff
@@ -296,7 +301,7 @@ proc getNumber(L: var Lexer, result: var Token) =
       if L.buf[pos] == '_':
         if L.buf[pos+1] notin chars:
           lexMessage(L, errGenerated,
-            "only single underscores may occur in a token and token may not " &
+            "only single underscores may occur in a number and number may not " &
```
konsumlamm:
```suggestion
            "only single underscores may occur in a number and a number may not " &
```

timotheecour:
> Sound reasonable? Suggestions?

yes, this is the best approach IMO.

> {tkComma, tkColon, tkParLe, tkBracketLe, tkCurlyLe}

what about `;` ? eg: `nim --eval:"echo 3;-1.echo"`

> (Off-topic, but IMO, the conversion to float/int (and the handling of different bases) for numeric literals should not be in the lexer. That is beyond the scope of a lexer. But that is me being nitpicky. And moving it would be a thought for another day.)

well it's not really off-topic and relates to https://github.com/nim-lang/RFCs/issues/228
> parser becomes lazy, makes no attempt at parsing the string into a numerical type (parsing is deferred till actually needed insde semphase, if at all)

but it can indeed be fixed in followup work

JohnAD:
I'll be adding quite a few more test cases in unit testing regarding "negation" of a numeric literal and thinking through more cases.

```Nim
var x = 9-5 # works
x = 9 - 5   # works
x = 9- 5    # works
x = 9 -5    # doesn't work: Error: in expression '9 -5': identifier expected, but found '9'
```
I might also add a counter-balancing check of a preceding literal token (any kind) or `tkSymbol`.

@timotheecour : yes, I'll also add `tkSemiColon` to the preceding token check.

timotheecour:
> x = 9 -5    # doesn't work: Error: in expression '9 -5': identifier expected, but found '9'
> I might also add a counter-balancing check of a preceding literal token (any kind) or tkSymbol.


## in devel:

```nim
when true:
  import macros
  macro deb2(a): string = newLit treeRepr(a)
  echo deb2(9-5)
  echo deb2(9 - 5)
  echo deb2(9 -5)
```

prints:
```
Infix
  Ident "-"
  IntLit 9
  IntLit 5
Infix
  Ident "-"
  IntLit 9
  IntLit 5
Command
  IntLit 9
  Prefix
    Ident "-"
    IntLit 5
```

and `let a = 9 -5` gives: `Error: in expression '9 -5': identifier expected, but found '9'`

## after your PR as of f145e215767e90f8ab6d4428369efa88568773f4
this prints:
```
Infix
  Ident "-"
  IntLit 9
  IntLit 5
Infix
  Ident "-"
  IntLit 9
  IntLit 5
Command
  IntLit 9
  IntLit -5
```

and `let a = 9 -5` still gives: `Error: in expression '9 -5': identifier expected, but found '9'` (as it should)

IMO the way you did it in the PR is correct as far as how `9 -5` is parsed as a `Command`; no need to special case tkSymbol as you were suggesting above.

## note:
before PR, this gives `Error: number out of range: '9223372036854775808'`:
```nim
  echo deb2(9 -9223372036854775808)
```
after PR, it prints:
```nim
Command
  IntLit 9
  IntLit -9223372036854775808
```

which is correct, great!

Araq:
> So, if the previous tokType is in the

> {tkComma, tkColon, tkParLe, tkBracketLe, tkCurlyLe}

> set or is preceded by whitespace, then a - followed by a numeric digit will be parsed into the numeric literal. Otherwise the - is > parsed as an operation.

> Sound reasonable? Suggestions?

<strike>Too complex, instead a '-' that is followed by '0'..'9' immediately (without whitespace in between) is part of the number.</strike>

timotheecour:
> Too complex, instead a '-' that is followed by '0'..'9' immediately (without whitespace in between) is part of the number.

that's a breaking change though for `1-3`, whereas the way suggested in https://github.com/nim-lang/Nim/pull/17020#issuecomment-787522529 was not a breaking change, and still intuitive enough / not too weird. I would suggest either:
* going with https://github.com/nim-lang/Nim/pull/17020#issuecomment-787522529
* or going with https://github.com/nim-lang/Nim/pull/17020#issuecomment-787522529 but making `1-3` (without space) issue a deprecation

Araq:
> that's a breaking change though for 1-3,

Bummer, ignore my proposal then.

compiler/lexer.nim
```diff
@@ -74,11 +75,15 @@ type
...
+  const weakTokens = {tkComma, tkSemiColon, tkColon,
+                      tkParRi, tkParDotRi, tkBracketRi, tkBracketDotRi,
+                      tkCurlyRi}
+
 const
-  weakTokens = {tkComma, tkSemiColon, tkColon,
-                tkParRi, tkParDotRi, tkBracketRi, tkBracketDotRi,
-                tkCurlyRi} # \
-    # tokens that should not be considered for previousToken
+  negationPrefixes = {tkComma, tkColon, tkParLe, tkBracketLe, tkSemiColon,
```
timotheecour, line 87: add doc comment or at least a comment

compiler/lexer.nim
```diff
@@ -74,11 +75,15 @@ type
 
   TokTypes* = set[TokType]
 
+when defined(nimsuggest):
```
timotheecour, line 78: why `when defined(nimsuggest):` ? (previous code didn't have it)

compiler/lexer.nim
```diff
@@ -394,74 +408,94 @@ proc getNumber(L: var Lexer, result: var Token) =
...
-  if L.buf[postPos] in {'\'', 'f', 'F', 'd', 'D', 'i', 'I', 'u', 'U'}:
-    if L.buf[postPos] == '\'':
-      inc(postPos)
-
+  var hasSuffix = false
+  var internalSuffix = false
+  if L.buf[postPos] == '\'':
+    hasSuffix = true
+    inc(postPos)
+  # 2A: handle the internal literal versions
```
timotheecour:
```suggestion
  # 2A: handle the builtin literal versions
```
(assuming that's what you mean?)

tests/lexer/tstrnumlits.nim
```diff
@@ -0,0 +1,157 @@
+# Test tkStrNumLit
+
+import std/macros
+import strutils
+
+macro assertAST(expected: string, struct: untyped): untyped =
```
timotheecour: DRY: factor into a help, say `tests/lexer/mlexerutils.nim` instead of copy pasting

tests/lexer/tminushandling.nim
```diff
@@ -0,0 +1,90 @@
+# Test tkStrNumLit
+
+import std/macros
+import strutils
+
+macro assertAST(expected: string, struct: untyped): untyped =
```
timotheecour: how about instead the following, simpler:
import std/unittest
```nim
macro listRepr2(a: untyped): untyped = newLit(a.lispRepr)

# now you can write 1-liners:
check listRepr2(1234'abc) == """(DotExpr (StrLit "1234") (Ident "\'abc"))"""
```

and on error, it will show the informative:
````
 Check failed: listRepr2("1234".'abc) == """(DotExpr (StrLit "134") (Ident "\'abc"))"""
listRepr2("1234".'abc) was (DotExpr (StrLit "1234") (Ident "\'abc"))
```

lispRepr is simpler for small AST's like the ones you have (but you can keep treeRepr for cases where you have a complex AST)

timotheecour:
One plausible alternative syntax for user defined litterals would be `!` instead of `'`:
```
-123'big
-123!big
```
## pros of `!`:
* (same) no ambiguity because `!` operator is (or should be) unary (IIRC)
* no weird rendering like this:
image
yes, existing nim syntax highlighters can be patched, but it still feels odd to have non-matched quotes.
* no conflation with builtin literals, which avoids some above mentioned issues such as keeping `12f` as valid but disallowing `12g`
* allows one to fully re-implement in library code the builtin litteral suffixes i8 etc, since it'd be a different syntax (`12!i8` etc)
* less quoting issues, eg: ```nim --eval:'echo "\n", 123!big' ```

## last but not least:
```nim
# this would give parser error for nim < 1.5.1:
since (1,5,1):
  echo 12!big

# this would work fine for nim < 1.5.1:
since (1,5,1):
  echo 12'big
```
(otherwise, you wouldn't be able to use builtin litterals in a module without jeopardizing 1.4 compatibility, unless the lexer patch is at least partially backported)

## ambiguity is bad:
```
echo 0x7d # 125
echo 0x7'd # 3.458459520888726e-323
```

Thoughts?

compiler/lexer.nim
JohnAD: It suppresses a warning message from the compiler. `weakTokens` is only used when `nimsuggest` is defined.

Admittedly, it is off-topic from the commit, but it removes one more warning to see.

timotheecour: ok!

tests/lexer/tstrnumlits.nim
```diff
@@ -0,0 +1,157 @@
...
+  result = ("[[" & number & "]]").S
+
+proc `'u9`(number: string): S =   # proc starts with 'u' just like 'u8'
+  result = ("[[" & number & "]]").S
+
+doAssert $1234.56'wrap == $1234.56'f9
+doAssert $1234.56'wrap == $1234.56'd9
+doAssert $1234.56'wrap == $1234.56'i9
+doAssert $1234.56'wrap == $1234.56'u9
+
```
timotheecour: this fails:
```nim
when true:
  import std/unittest
  proc `'wrap`(number: string): string = number
  check 12.34e+21331'wrap == "12.34e+21331" # ok
  check 12.34E+21331'wrap == "12.34E+21331" # error: E => e
```
but I'm not sure whether it should pass or not.

At least it should be documented that the lettters get lower-cased

timotheecour:
* shouldn't `DotExpr` be a new node kind `NumberLit`? it seems to me shoe-horning this feature into `DotExpr` (and distinguishing based on whether 1st child is a `StrLit`) will only cause more issues.

* also, repr is buggy because it doesnt' return the original stringified AST:

```nim
when true:
  import macros
  proc `'wrap`(number: string): string = number
  macro fn(a) =
    echo()
    echo a.treeRepr
    echo a.repr
  fn 12'wrap
  fn 12'i8
```
prints:
```
DotExpr
  StrLit "12"
  Ident "\'wrap"
"12".'wrap

Int8Lit 12
12'i8
```

we should instead generate:
```
NumberLit
  StrLit "12"
  Ident "\'wrap"
12'wrap
```

and for `i8`, we'd ideally also generate 
```
NumberLit
  StrLit "12"
  Ident "\'i8"
12'i8
```
after introducing `'i8` to system

Clyybber:
>   * shouldn't `DotExpr` be a new node kind `NumberLit`? it seems to me shoe-horning this feature into `DotExpr` (and distinguishing based on whether 1st child is a `StrLit`) will only cause more issues.
> 
>   * also, repr is buggy because it doesnt' return the original stringified AST:
> 

I don't see a reason to introduce a new node kind. The renderer can be patched to recognize these dot exprs and print them as you'd expect, similar to how we do it for `{}=` `{}` `[]` `[]=`.

timotheecour:
> I don't see a reason to introduce a new node kind. The renderer can be patched to recognize these dot exprs and print them as you'd expect, similar to how we do it for {}= {} [] []=.

this is a false analogy, in `{}= {} [] []=` the symbol is fixed and there is no ambiguity, but with number literals the suffix is user defined.

how can you we tell a dot apart from a number litteral after you conflate both into a DotExpr?

eg:
```nim
when true:
  import macros
  proc `'big`(a:string): string = a
  macro mfn1(a: untyped) =
    echo a.lispRepr
  mfn1 "3".len
  mfn1 3'len
```
(DotExpr (StrLit "3") (Ident "len"))
(DotExpr (StrLit "3") (Ident "\'len"))

the only way to tell them apart here is that Ident starts with `'`
But then it's not clear how this would work inside templates or macros eg:
```nim
when true:
  import t11972b # defined `'big'
  import macros
  proc `'big`(a:string): string = a # redefines `'big'
  macro bar(): untyped =
    result = quote do:
      3'big
    echo result.treeRepr
  discard bar()
```

```
DotExpr
  StrLit "3"
  OpenSymChoice
    Sym "\'big"
    Sym "\'big"
```

so now the rule becomes more complicated, you'd have to inspect the children of the RHS to check if it's a `OpenSymChoice`, it's not clear this can be reliable. And I haven't mentioned dot operators yet, which adds another level of ambiguities.

Furthermore, such conflation will break existing macros that handle `DotExpr` in more subtle ways than breakage caused by not handling some new `NumberLit`.

Clyybber:
> so now the rule becomes more complicated, you'd have to inspect the children of the RHS to check if it's a OpenSymChoice, it's not clear this can be reliable. And I haven't mentioned dot operators yet, which adds another level of ambiguities.

It's not clear how this can be unreliable, it's quite simple IMO.
Dot operators can check for the single-quote too or we could use some other member of nkCallKinds instead.

> Furthermore, such conflation will break existing macros that handle DotExpr in more subtle ways than breakage caused by not handling some new NumberLit.

How so? It's just a normal dot call, possibly constrained to `{.compileTime.}`.

JohnAD:
> 
> (DotExpr (StrLit "3") (Ident "len"))
> (DotExpr (StrLit "3") (Ident "'len"))

within the parser it is a very clear narrow path because of the new tkStrNumLit only occurs with a `'` suffix and that token is specially handled. However, this is not making it back upstream to be seen as unique by `treeLisp` as tkStrNumLit is being lumped in with generic string literals.

So, I don't think anything is breaking in the compiler sense, but not distinguishing suffix expressions/numeric strings in the macro expressions could get confusing for developers.

So, I will examine the code more closely and either: 1. have a new type of node.kind for a `StrNumLit`. or 2. have a simplified new `SuffixExpr` that behaves exactly like `DotExpr`. 3. both.

So, either:

```(DotExpr (StrNumLit "3") (Ident "'len"))```
or
```(SuffixExpr (StrLit "3") (Ident "'len"))```
or
```(SuffixExpr (StrNumLit "3") (Ident "'len"))```

I'll put up a commit that seems cleanest. Personally, I like 1 best as it really is a dot expression; but with suffix marker. In fact, the following does work right now also, since a DotExpr is really just Nim's UCFS format:

```nim
import math
import strutils

proc `'foo`(a: string, exp: int): int =
  result = parseInt(a) * (4^exp)

assert 55'foo(2) == 880
```

JohnAD:
Or, more relevant to the upcoming `decimal` library:

```nim
assert $(123.0'm) == "123.0"
assert $(123.0'm(scale=2)) == "123.00"
```

JohnAD:
Hmmm. I'm going to stash my work on adding nkStrNumLit and nnkStrNumLit for the time being. That is quite the rabbit hole.

I'm going to concentrate on the rest first.

I'm also not convinced adding a NimNodeKind is all that useful. Is it bad that the way to generate:

```nim
var x = 123.45'foo
```

in a macro is:

```nim
macro makeFoo(): untyped =
  result = nnkStmtList.newTree(
    nnkVarSection.newTree(
      nnkIdentDefs.newTree(
        newIdentNode("x"),
        newEmptyNode(),
        nnkDotExpr.newTree(
          newLit("123.45"),
          newIdentNode("'foo")
        )
      )
    )
  )

proc `'foo`(num: string): string = num

makeFoo()

echo x
```

??

In fact, I could modify the docs to describe the suffix as a dot expression of the number as a string.

(Even so, adding a new NimNodeKind is a deep exploration of the compiler's setup. In that sense, it would be fun to add.)

tests/lexer/tstrnumlits.nim
JohnAD: done.

tests/lexer/tstrnumlits.nim
JohnAD: done

tests/lexer/tstrnumlits.nim
JohnAD: done

tests/lexer/tstrnumlits.nim
JohnAD: done.

doc/manual.rst
JohnAD: Made it easier to read by putting an example block of code that makes a crude `nibble`, a 4-bit uint that doesn't natively exist.

compiler/lexer.nim
JohnAD: good catch. changed.

compiler/lexer.nim
JohnAD: done.

compiler/lexer.nim
JohnAD: yes. done!

tests/lexer/tstrnumlits.nim
JohnAD: done.

tests/lexer/tminushandling.nim
JohnAD: done.

tests/lexer/tstrnumlits.nim
JohnAD: I'll look at this next.

tests/lexer/mlexerutils.nim
```diff
@@ -0,0 +1,14 @@
...
+
+macro assertAST*(expected: string, struct: untyped): untyped =
+  var ast = newLit(struct.treeRepr)
+  result = quote do:
+    if `ast` != `expected`:
+      echo "Got:"
+      echo `ast`.indent(2)
+      echo "Expected:"
+      echo `expected`.indent(2)
+      raise newException(ValueError, "Failed to lex properly")
```
timotheecour:
```nim
macro assertAST*(expected: string, struct: untyped): untyped =
  var ast = newLit(struct.treeRepr)
  result = quote do:
    if `ast` != `expected`:
      doAssert false, "\nGot:\n" & `ast`.indent(2) & "\nExpected:\n" & `expected`.indent(2)
```

* echo + raise is bad as catcher won't get the full msg.
* splitting echo is bad because it increases chance a msg gets split and intermingled with other processes' stdout when multiple processes are outputting to stdout as is case with `testament pcat`

timotheecour:
> proc `'foo`(a: string, exp: int): int =
> assert 55'foo(2) == 880

clever, I hadn't thought about using this feature like that!

> So, I will examine the code more closely and either: 1. have a new type of node.kind for a StrNumLit. or 2. have a simplified new SuffixExpr that behaves exactly like DotExpr. 3. both.

I don't see the point in `StrNumLit`, this should just be a `StrLit` and behaves like it; the parent (in case of using `NumberLit lhs rhs`) or the sibling (in case of using `DotExpr lhs ident_starting_with_quote`) would be sufficient to disambiguate. The only question is `NumberLit` (aka `SuffixExpr`) vs reusing `DotExpr` for the parent node.

tests/lexer/tstrnumlits.nim
JohnAD: It is no longer forced to lowercase 'e' in the general case. The individual deserializers are handling it just fine.

JohnAD:
With this last commit, I should be done generally. The only remaining issue is whether it is wise to use with the `dotExpr` as the parent; or should we create a new node such as `suffixExpr` / `numberLit` that behaves just like `dotExpr`.

IMO, of adding a new node:

PROS:
* improved clarity when writing macros
* improved clarity when debugging the compiler

CONS:
* adds "one more" node to debug, maintain, handle
* adds another node to learn when writing macros (arguable)

We could, for now, leave it like this and add the node in a later PR if the community finds it very beneficial to add.

Personally, I don't find re-using `dotExpr` confusing; but I'm the author so that opinion *really* doesn't count :grin:  .

Araq:
This PR seems to break `tests/array/tarray.nim`.

JohnAD:
I will take a look at `tarray.nim` tonight.

Also, this PR is against a older version of Nim. To make it more up-to-date. should I simply rebase it or is there a cleaner method via git?

timotheecour:
> should I simply rebase it or is there a cleaner method via git?

I prefer rebasing (running this locally) `git pull --rebase origin devel` and fixing conflicts, then force-pushing so that commits in your PR aren't inter-mingled with other commits, but you can also use `git merge` if you prefer; the PR will be merged with squash and merge so it won't make a difference to the git history in the devel branch (only in the commits shown in this PR for people reviewing it)

JohnAD:
All is looking good!

timotheecour:
it's starting to look very good. Went through everything, most review comments should be easy to address. The bugs below don't need to be addressed in this PR but bug 2 and bug 9 would be nice to fix in this PR if possible.

-------------

ok i think we can live with `DotExpr` instead of new `NumberLit` for the following reason:
`-12'wrap` is transformed into this untyped AST:
```nim
DotExpr
  StrLit "-12"
  Ident "\'wrap"
```
and this typed AST:
```nim
Call
  Sym "\'wrap"
  StrLit "-12"
```

we're basically saying that `-12'wrap` and ``` "-12".`'wrap` ``` are un-distinguishable.
This is a departure from other cases where untyped AST maintains the syntactical differences (eg `a.b` is represented as `DotExpr`, not `Call`); but this is probably ok.

the following problems remain, but aren't blockers for this PR:

- [x] bug 1: macros.repr
`macros.repr` should be patched to handle this:
```nim
when true:
  import macros
  func `'wrap`(a: string): string = "{" & a & "}"
  macro deb1(a): untyped =
    echo a.repr
    echo a.lispRepr
  macro deb2(a: typed): untyped =
    echo a.repr
    echo a.lispRepr
  deb1(-12'wrap)
  deb2(-12'wrap)
```
prints:
```
"-12".'wrap # bug: this should be: -12'wrap
(DotExpr (StrLit "-12") (Ident "\'wrap")) # ok
'wrap("-12") # ok: typed can change repr, as it does for fn(1) vs 1.fn
(Call (Sym "\'wrap") (StrLit "-12")) # ok
```
(EDIT: see https://github.com/nim-lang/Nim/pull/17020#issuecomment-803459003 for a fix)
EDIT: fixed in devel, see test in https://github.com/nim-lang/Nim/pull/17500

- [x] bug 2: suffix doesn't work if it's a template param
```nim
when true:
  func `'wrap`(a: string): string = "{" & a & "}"
  template toSuf(suf): untyped = 
    -12'suf
  echo toSuf(`'wrap`)
```
gives:
`Error: undeclared field: ''suf'`
but this should work

compiler/semtempl.nim `semTemplBody` should be patched to handle this

=> see test in https://github.com/nim-lang/Nim/pull/17500; using instead this works:
```nim
template toSuf(`'suf`): untyped = 
```

- [ ] bug 3
all the existing nim syntax highlighters are broken and need to handle this, eg this renders badly in sublime text, github (which depends on github linguist, which itself IIRC depends on sublime text nimlime nim package)

- [ ] bug 4
unless we backport this and upgrade csources, we won't be able to benefit from this in stdlib modules used by bootstrap (dependencies of koch, compiler, etc), and using it in other stdlib modules would make `--useversion:1.2|...` unusuable because `when false`, `runnableExamples`, `{.since: (1,5,1).}` etc cannot protect from lexer/parser errors.

However this can be largely mitigated by https://github.com/nim-lang/RFCs/issues/348 (which itself needs to be backported including to csources/csources2)

- [ ] bug 5
we should turn the builtin `123'i64` + friends into non-builtin reusing this feature so that compiler logic simplifies

- [ ] bug 6 (minor): nim doc renders badly
```nim
when true:
  runnableExamples:
    func `'wrap`(a: string): string = "{" & a & "}"
    let a = -128'i8
    echo -12'wrap
```
the syntax highlight of ``` `'wrap` ``` is bad
image

EDIT: relevant code to fix is in lib/packages/docutils/highlite.nim eg:
```nim
    of '\'':
      inc(pos)
      g.kind = gtCharLit
```

- [x] bug 7
this fails; the refactoring mentioned in https://github.com/nim-lang/Nim/pull/17020/files#r598031603 would help to fix this in a DRY way:
```nim
when true:
  func `'f33a`(a: string): string = "{" & a & "}"
  echo -12'f33a

  func `'f32a`(a: string): string = "{" & a & "}"
  echo -12'f32a # Error: invalid token: no whitespace between number and identifier
```

- [x] bug 8
we can't define a litteral proc in quote do:
Error: undeclared identifier: ''wrap'
```nim
when true:
  import std/macros
  macro bar(): untyped =
    result = quote do:
      func `'wrap`(a: string): string = "{" & a & "}"
  bar()
```
EDIT: https://github.com/nim-lang/Nim/pull/17426 will fix this by using `genAst` instead of `quote`

- [ ] bug 9
we can't use a litteral proc in quote do: 
this is possibly a consequence of bug 2
```nim
when true:
  import std/macros
  macro bar(): untyped =
    func wrap1(a: string): string = "{" & a & "}"
    func `'wrap2`(a: string): string = "{" & a & "}"
    result = quote do:
      let a1 = wrap1"-128" # ok
      let a2 = -128'wrap2 # Error: undeclared field: ''wrap2' # bug
  bar()
```

- [x] bug 10
```nim
when defined case36:
  func `'wrap`(a: string): string = "{" & a & "}"
  template toSuf(): untyped =
    -12'wrap
  echo toSuf()
```
negationPrefixes wrong here?
```
Error: type mismatch: got <string>
but expected one of:
func `-`[T](x, y: set[T]): set[T] [func declared in /Users/timothee/git_clone/nim/Nim_prs/lib/system/setops.nim(62, 6)]
```

=> fixed in devel, test in https://github.com/nim-lang/Nim/pull/17500

- [x] bug 11 (not really a bug but a TODO):
add a test for this which now works instead of giving Error: number out of range: '9223372036854775808':
```nim
let a = -9223372036854775808
```
(next to: `    doAssert -128'i8 == (-128).int8`)

(a bit different from other tests since there's no suffix)
=> https://github.com/nim-lang/Nim/pull/17500

compiler/lexer.nim
```diff
@@ -386,82 +402,102 @@ proc getNumber(L: var Lexer, result: var Token) =
...
+  if L.buf[postPos] == '\'':
+    hasSuffix = true
+    inc(postPos)
+  # 2A: handle the builtin literal versions
+  if L.buf[postPos] in {'f', 'F', 'd', 'D', 'i', 'I', 'u', 'U'}:
     case L.buf[postPos]
     of 'f', 'F':
-      inc(postPos)
-      if (L.buf[postPos] == '3') and (L.buf[postPos + 1] == '2'):
+      if (L.buf[postPos + 1] == '3') and (L.buf[postPos + 2] == '2'):
```
timotheecour: how about this in followup PR?

`if (L.buf.skip(postPos + 1).continuesWith("32"))` would be more readable

introducing a `func continuesWith(a: openArray[char], b: string)`

and `template skip(a: openArray[char], start: int): untyped = toOpenArray(a, start, a.high)`

(and could be reused in cases below)


you could also introduce a template to avoid much code repetition, eg:

```
template onSuffix(suffix): bool =
  block foo:
    for j in 0..<suffix.len:
      if (L.buf.skip(postPos + j) != suffix[j]: break foo
    inc(postPos, suffix.len)
    internalSuffix = true
```
and then use it in all similar cases below:
```nim
elif onSuffix("32"): result.tokType = tkFloat32Lit
...
```

compiler/lexer.nim
```diff
@@ -386,82 +402,102 @@ proc getNumber(L: var Lexer, result: var Token) =
...
         inc(postPos, 3)
-      else:   # "f" alone defaults to float32
+        internalSuffix = true
+      elif (L.buf[postPos + 1] == '1') and
+           (L.buf[postPos + 2] == '2') and
+           (L.buf[postPos + 3] == '8'):
+        result.tokType = tkFloat128Lit
+        inc(postPos, 4)
+        internalSuffix = true
+      elif not (L.buf[postPos + 1] in SymChars): # standalone 'f'
```
timotheecour:
```suggestion
      elif L.buf[postPos + 1] notin SymChars: # standalone 'f'
```

ditto elsewhere

compiler/lexer.nim
```diff
@@ -386,82 +402,102 @@ proc getNumber(L: var Lexer, result: var Token) =
...
-      elif (L.buf[postPos] == '8'):
+        inc(postPos, 3)
+        internalSuffix = true
+      elif (L.buf[postPos + 1] == '8'):
         result.tokType = tkUInt8Lit
-        inc(postPos)
-      else:
+        inc(postPos, 2)
+        internalSuffix = true
+      elif not (L.buf[postPos + 1] in SymChars): # standalone 'u'
```
timotheecour:
```suggestion
      elif L.buf[postPos + 1] notin SymChars): # standalone 'u'
```

compiler/lexer.nim
```diff
@@ -386,82 +402,102 @@ proc getNumber(L: var Lexer, result: var Token) =
...
+        internalSuffix = true
     else:
-      lexMessageLitNum(L, "invalid number: '$1'", startpos)
-
-  # Is there still a literalish char awaiting? Then it's an error!
-  if  L.buf[postPos] in literalishChars or
-     (L.buf[postPos] == '.' and L.buf[postPos + 1] in {'0'..'9'}):
-    lexMessageLitNum(L, "invalid number: '$1'", startpos)
+      discard
+  # 2B: else look for user-definited types that are adjacent (no spaces)
```
timotheecour:
```suggestion
  # 2B: else look for user-defined types that are adjacent (no spaces)
```

compiler/lexer.nim
```diff
@@ -733,11 +780,11 @@ proc getEscapedChar(L: var Lexer, tok: var Token) =
 
 proc handleCRLF(L: var Lexer, pos: int): int =
   template registerLine =
-    let col = L.getColNumber(pos)
-
     when not defined(nimpretty):
+      let col = L.getColNumber(pos)
       if col > MaxLineLength:
         lexMessagePos(L, hintLineTooLong, pos)
+    discard
```
timotheecour:
```suggestion
```

(`when` is sufficient)

compiler/lexer.nim
```diff
@@ -1270,14 +1325,26 @@ proc rawGetTok*(L: var Lexer, tok: var Token) =
         # tkTripleStrLit -> tkGTripleStrLit
         inc(tok.tokType, 2)
     of '\'':
-      tok.tokType = tkCharLit
-      getCharacter(L, tok)
-      tok.tokType = tkCharLit
+      if (L.previousTokType == tkStrNumLit) and (tok.strongSpaceA == 0):
+        # if the previous token (with no prior whitespace) is a tkStrNumLit, then this is a numeric suffix
+        getSymbol(L, tok)
```
timotheecour:
```suggestion
        getSymbol(L, tok) # example: -12'big
```

compiler/lexer.nim
```diff
@@ -1270,14 +1325,26 @@ proc rawGetTok*(L: var Lexer, tok: var Token) =
...
+      elif L.previousTokType == tkAccent:
+        getSymbol(L, tok)
```
timotheecour:
```suggestion
        getSymbol(L, tok) # example: `'big`
```

doc/manual.rst
```diff
@@ -566,6 +566,19 @@ The type suffixes are:
...
 =================    =========================
 
+If the suffix is not predefined, then the suffix is assumed to be a call
+to a proc, template, macro or other callable identifier that is passed the
+string containing the literal. The identifier needs to contain the apostrophe
+and so will need to be escaped with backticks.
+
+.. code-block:: nim
+  import strutils
+  type nibble = distinct uint8
```
timotheecour: you're not using this type; did you intend `'nibble` to return a nibble?

would u4 be a better name?

doc/manual.rst
```diff
@@ -566,6 +566,19 @@ The type suffixes are:
...
+  proc `'nibble`(n: string): uint8 =
+    result = (parseInt(n) and 0x0F).uint8
+  
+  var x = 5'nibble
```
timotheecour: also add an example with extra args as you mentioned in PR:
```nim
proc `'foo`(a: string, b: int) = $(a, b)
assert -12'foo(2) == ("-12", 2)
```

tests/lexer/tminushandling.nim
```diff
@@ -0,0 +1,82 @@
+# Test numeric literals and handling of minus symbol
+
+import std/macros
+import strutils
```
timotheecour:
```suggestion
import std/[macros, strutils]
```
preferred style

tests/lexer/tminushandling.nim
```diff
@@ -0,0 +1,82 @@
...
+  "failed to include minus sign lexing float literal preceded by left paren"
+
+doAssert lispReprStr( -1.000'f64) == """(Float64Lit -1.0)""":
+  "failed to include minus sign lexing float literal preceded by a space"
+
+doAssert [-1].contains(minusOne):
+  "unable to handle negatives after square bracket"
+
+doAssert lispReprStr([-1]) == """(Bracket (IntLit -1))""":
+  "failed to include minus sign lexing int literal preceded by left bracket"
```
timotheecour: IMO these string messages in the assert are just noise; it's clear enough from context; maybe keep the ones you feel are most important for test comprehension and drop the other ones. Other tests also don't typically do that

tests/lexer/tstrnumlits.nim
```diff
@@ -0,0 +1,151 @@
+# Test tkStrNumLit
+
+import std/macros
+import strutils
```
timotheecour:
```suggestion
import std/[macros, strutils]
```

tests/lexer/tstrnumlits.nim
```diff
@@ -0,0 +1,151 @@
+# Test tkStrNumLit
```
timotheecour, line 5:
```suggestion
discard """
  targets: "c cpp js"
"""

# Test tkStrNumLit
```

tests/lexer/tstrnumlits.nim
```diff
@@ -0,0 +1,151 @@
...
+import std/macros
+import strutils
+import mlexerutils
+
+type
+  S = distinct string
+
+proc `'wrap`(number: string): S =
+  result = ("[[" & number & "]]").S
+
```
timotheecour:
```suggestion
proc `'wrap`(number: string): string =
  "{" & number & "}"
```
then the tests below simplify. Simplicity matters so we can easily add more tests, makes them more readable etc

(but maybe add just 1 single test with the previous version, say a wrap2, as sanity check)

tests/lexer/tstrnumlits.nim
```diff
@@ -0,0 +1,151 @@
...
+              Infix
+                Ident "&"
+                Infix
+                  Ident "&"
+                  StrLit "[["
+                  Ident "number"
+                StrLit "]]"
+            Ident "S"""":
+  proc `'wrap`(number: string): S =
+    result = ("[[" & number & "]]").S
```
timotheecour:
```nim
  proc `'wrap`(number: string): string = number
```
(simplifies the AST; the complexity of previous version is irrelevant to what we're testing in this file)

tests/lexer/tstrnumlits.nim
```diff
@@ -0,0 +1,151 @@
...
+doAssert string(1'wrap+ -1'wrap) == "[[1]]plus[[-1]]":
+  "unable to resolve a negative suffixed numeric literal following an operator"
+
+doAssert string(1'twrap) == "[[1]]":
+  "unable to resolve a template using a suffixed numeric literal"
+
+doAssert string(extraContext()) == "[[22.40]]":
+  "unable to return a suffixed numeric literal by an implicit return"
+
+doAssert string(0x5a3a'wrap) == "[[0x5a3a]]":
```
timotheecour: also add an example showing case is preserved eg with these examples:
```nim
0xdeadBEEF'wrap
0.1E12'wrap
0.0e12'wrap
0.0e+12'wrap
0.0e-12'wrap
0e-12'wrap
```

tests/lexer/tminushandling.nim
```diff
@@ -0,0 +1,82 @@
+# Test numeric literals and handling of minus symbol
+
+import std/macros
+import strutils
+import mlexerutils
+
```
timotheecour, line 9: you forgot one of the important tests:
```nim
assert -128'i8 == -128
```
which didn't work prior this PR

tests/lexer/tstrnumlits.nim
```diff
@@ -0,0 +1,151 @@
...
+
+proc `'d9`(number: string): S =   # proc starts with 'd' just like the d suffix
+  result = ("[[" & number & "]]").S
+
+proc `'i9`(number: string): S =   # proc starts with 'i' just like 'i64'
+  result = ("[[" & number & "]]").S
+
+proc `'u9`(number: string): S =   # proc starts with 'u' just like 'u8'
+  result = ("[[" & number & "]]").S
+
```
timotheecour: add this test:
```nim
template `'foo`(a: string, b: int) = $(a, b)
assert -12'foo(2) == ("-12", 2)
```

tests/lexer/tstrnumlits.nim
```diff
@@ -0,0 +1,151 @@
...
+  "lexer not handling upper-case exponent"
+
+doAssert 1.0E2 == 100.0:
+  "lexer not handling upper-case exponent with float"
+
+doAssert 1e2 == 100:
+  "lexer not handling lower-case exponent"
+
+doAssert $1E2'wrap == "[[1E2]]":
+  "lexer not passing exponent (e) marker with original case"
```
timotheecour: wrap this whole file inside a `template main()` so you can test at CT and RT without duplication, as described in https://github.com/timotheecour/Nim/issues/534#issuecomment-769565033

JohnAD:
For the most recent bug list from @timotheecour : 
1. Not really change-able unless we added the `suffixExpr` rather than use `dotExpr`. Something to keep in mind if we decide to later add that.
2. That is a serious bug; will start work on it today/tomorrow and add a test. [TODO]
3. Yes. I mod'd my personal version of the sublime highlighter code for my own purposes, but I'll consider doing a PR for the public NimLime also.
4. That makes sense. We will have to keep user-supplied suffixes out of the compiler core for a while.
5. I think I'm going to simplify the suffix scanning section in general; while keeping lexer performance up. Perhaps in a later PR, we can treat the built-ins as the same things as user-supplied suffixes. On a related note, I'm also going to make sure that suffixes are mostly resolvable at compile-time with suffixes running as macros. [TODO]
6. I will work on this today and add test. [TODO]
7. I will work on this today and add test. [TODO]
8. I will work on this today/tomorrow [TODO]
9. I will work on this today/tomorrow [TODO]

timotheecour:
> 8. I will work on this today/tomorrow [TODO]

I've sent out https://github.com/nim-lang/Nim/pull/17426 which fixes bug 8

timotheecour:
> 1. Not really change-able unless we added the suffixExpr rather than use dotExpr. Something to keep in mind if we decide to later add that.

the following works and fixes `bug 1`, you can use this (and improve it as needed)

```nim
diff --git a/compiler/renderer.nim b/compiler/renderer.nim
index 13ff6941f..1e5a3c2e8 100644
--- a/compiler/renderer.nim
+++ b/compiler/renderer.nim
@@ -1188,9 +1188,14 @@ proc gsub(g: var TSrcGen, n: PNode, c: TContext) =
     gcomma(g, n, c)
     put(g, tkBracketRi, "]")
   of nkDotExpr:
-    gsub(g, n, 0)
-    put(g, tkDot, ".")
-    gsub(g, n, 1)
+    if n[1].kind == nkIdent and n[1].ident.s[0] == '\'': # TODO: add a re-usable helper function isUserLitteral
+      assert n[0].kind == nkStrLit
+      put(g, tkStrNumLit, n[0].strVal)
+      gsub(g, n, 1) # TODO: this renders well but check whether this is the right node to add
+    else:
+      gsub(g, n, 0)
+      put(g, tkDot, ".")
+      gsub(g, n, 1)
   of nkBind:
     putWithSpace(g, tkBind, "bind")
     gsub(g, n, 0)
```

tests/lexer/tstrnumlits.nim
JohnAD: Done.

tests/lexer/tstrnumlits.nim
JohnAD: Done.

tests/lexer/tstrnumlits.nim
JohnAD: I've removed the `S` unique type refs and simplified things.

doc/manual.rst
JohnAD: Done.

doc/manual.rst
JohnAD: changed to `u4` and used distinct type

compiler/lexer.nim
JohnAD: I've simplified it using a `cmpSuffix` proc. While not inclusive as your template, it has the benefit of not asking the future programmer to remember what the template does. There are always pros/cons to everything of course.

compiler/lexer.nim
timotheecour: typo is still there

compiler/lexer.nim
timotheecour: comment still holds

tests/lexer/tstrnumlits.nim
JohnAD: Done.

compiler/lexer.nim
JohnAD: Sorry; I'm hitting "resolve conversation" as I'm editing the code prior to commit. :) You should see it now.

timotheecour: oh ya if figured. IIRC there's a way to batch replies, by clickign on "start a review" instead of "add a single comment" and then upon commit, click send review so they all appear at same time after commit. But either way is fine.

tests/lexer/tstrnumlits.nim
```diff
@@ -102,50 +44,84 @@ assertAST dedent """
...
+    doAssert 1234.56'wrap == 1234.56'f32a
+    doAssert 1234.56'wrap == 1234.56'd9
+    doAssert 1234.56'wrap == 1234.56'i9
+    doAssert 1234.56'wrap == 1234.56'u9
+    doAssert lispReprStr(1234.56'u9) == """(DotExpr (StrLit "1234.56") (Ident "\'u9"))""":
+      "failed to properly build AST for suffix that starts with u"
+    doAssert -128'i8 == (-128).int8
+
+  block: # extra parameter and case checks
+    proc `'wrap`(number: string): string =
```
timotheecour: this is defined many times, just define once at beginning of module and reuse it; you're already testing that such procs can be defined inside a template with the other ones (f9 etc)

tests/lexer/tstrnumlits.nim
```diff
@@ -102,50 +44,84 @@ assertAST dedent """
...
+      result = left & "plus" & right
+
+    doAssert 1'wrap == "[[1]]"
+    doAssert -1'wrap == "[[-1]]":
+      "unable to resolve a negative integer-suffix pattern"
+    doAssert lispReprStr(-1'wrap) == """(DotExpr (StrLit "-1") (Ident "\'wrap"))"""
+    doAssert 12345.67890'wrap == "[[12345.67890]]"
+    doAssert 1'wrap*1'wrap == "[[1]]times[[1]]":
+      "unable to resolve an operator between two suffixed numeric literals"
+    doAssert 1'wrap+ -1'wrap == "[[1]]plus[[-1]]":
```
timotheecour:
```suggestion
    doAssert 1'wrap + -1'wrap == "[[1]]plus[[-1]]":
```

(doesn't nim warn about this inconsistent spacing?) EDIT: indeed it does

tests/lexer/tstrnumlits.nim
```diff
@@ -102,50 +44,84 @@ assertAST dedent """
...
+    doAssert 0o5732'wrap == "[[0o5732]]"
+    doAssert 0b0101111010101'wrap == "[[0b0101111010101]]"
+    doAssert -38383839292839283928392839283928392839283.928493849385935898243e-50000'wrap == "[[-38383839292839283928392839283928392839283.928493849385935898243e-50000]]"
+    doAssert 1234.56'wrap == "[[1234.56]]":
+      "unable to properly account for context with suffixed numeric literals"
+
+  block: # verify that the i64, f32, etc builtin suffixes still parse correctly
+    const expectedF32: float32 = 123.125
+    proc `'wrap`(number: string): string =
+      result = "[[" & number & "]]"
```
timotheecour:
```suggestion
      "[[" & number & "]]"
```

(nit: simpler = better)

tests/lexer/tminushandling.nim
```diff
@@ -1,68 +1,16 @@
...
-
-doAssert 4 - 1 == 3:
-  "unable to handle subtraction with surrounding spaces with a numeric literal"
-
-doAssert 4 - one == 3:
-  "unable to handle subtraction with surrounding spaces with an identifier"
-
-# border cases that *should* generate compiler errors
-
+# border cases that *should* generate compiler errors:
```
timotheecour, line 13: also check for this, eg:
```nim
template bad() =
  x = 4 -1
doAssert not compiles(bad())
```

compiler/lexer.nim
```diff
@@ -410,85 +425,79 @@ proc getNumber(L: var Lexer, result: var Token) =
 
   # Second stage, find out if there's a datatype suffix and handle it
   var postPos = endpos
-  var hasSuffix = false
-  var internalSuffix = false
-  if L.buf[postPos] == '\'':
-    hasSuffix = true
+  let suffixMarker = (L.buf[postPos] == '\'')
```
timotheecour:
```suggestion
  let suffixMarker = L.buf[postPos] == '\''
```
more idiomatic and not ambiguous, even for humans

compiler/lexer.nim
```diff
@@ -410,85 +425,79 @@ proc getNumber(L: var Lexer, result: var Token) =
...
+  if suffixMarker:
     inc(postPos)
+    if (L.buf[postPos] notin SymStartChars):
```
timotheecour: ditto https://github.com/nim-lang/Nim/pull/17020/files/1f3b3315a903e3c0d62f7d795c67d7245cc8b86e..8133a25b7094ea51b9ae0450df9ac690eb775df7#r598331871

compiler/lexer.nim
```diff
@@ -410,85 +425,79 @@ proc getNumber(L: var Lexer, result: var Token) =
...
-        result.tokType = tkUInt8Lit
-        inc(postPos, 2)
-        internalSuffix = true
-      elif not (L.buf[postPos + 1] in SymChars): # standalone 'u'
-        result.tokType = tkUIntLit
-        inc(postPos)
-        internalSuffix = true
+  var internalSuffix = false
+  var endOfSuffix = postPos
+  while L.buf[endOfSuffix] in SymStartChars + {'0'..'9'}:
```
timotheecour:
```suggestion
  while L.buf[endOfSuffix] in SymChars:
```

tests/lexer/tstrnumlits.nim
JohnAD: Yes it does. I went ahead and added a version that is consistent to be complete. Is there a way to test for a warning?

JohnAD:
I've got a new version ready for commit, but I'll pause for a bit just in case you are in the middle of adding a conversation item.

compiler/lexer.nim
```diff
@@ -340,6 +340,21 @@ proc getNumber(L: var Lexer, result: var Token) =
...
+
+  proc cmpSuffix(L: Lexer, prev: int, endOfSuffix: int, chs: array[2, char]): bool =
+    if (endOfSuffix - prev) != 3:
+      return false
+    return L.buf[prev + 1] == chs[0] and L.buf[prev + 2] == chs[1]
+
+  proc cmpSuffix(L: Lexer, prev: int, endOfSuffix: int, chs: array[3, char]): bool =
+    if (endOfSuffix - prev) != 4:
+      return false
+    return L.buf[prev + 1] == chs[0] and L.buf[prev + 2] == chs[1] and L.buf[prev + 3] == chs[2]
```
timotheecour:
```suggestion
  proc cmpSuffix(L: Lexer, prev: int, endOfSuffix: int, chs: openArray[char]): bool =
    if (endOfSuffix - prev - 1) != chs.len:
      return false
    for i in 0..<chs.len:
      if L.buf[prev + 1 + i] != chs[i]: return false
    return true
```

tests/lexer/tstrnumlits.nim
timotheecour: there will be once i make a PR for https://github.com/nim-lang/Nim/issues/8803 and convince araq that this is indeed a good feature (as simple as `let ret = compilesWithContext(expr)`), until then testing for warnings can be done but is not as flexible, via a testament discard section with nimout etc); in your case, it's ok to not check for the warning (it's mostly irrelevant for this test file) and use both with correct spacing and without for completeness.

compiler/lexer.nim
JohnAD: Yours is cleaner but I suspect mine is faster as it avoids a loop for speculative cpu execution.

But I suspect clarity is more important, so I'll change it to yours.

timotheecour: I'd be surprised if a C compiler with optimization turned on doesn't unroll the loop. And in any case, this isn't a hotspot.

(please also add {.inline.} although in this case it probably makes no difference given that inlining is possible within same module)

tests/lexer/tstrnumlits.nim
```diff
@@ -123,5 +116,34 @@ template main =
...
+    doAssert deb(-12'bar) == "-12'bar"
+    # macro metawrap(): untyped =
+    #   func wrap1(a: string): string = "{" & a & "}"
+    #   func `'wrap2`(a: string): string = "{" & a & "}"
+    #   result = quote do:
+    #     let a1 = wrap1"-128"
+    #     let a2 = -128'wrap2
+    # metawrap()
+    # doAssert a1 == "{-128}"
+    # doAssert a2 == "{-128}"
```
timotheecour, line 146:
```suggestion
    macro `'bar`(a: static string): untyped = newLit(a.repr)
    doAssert -12'bar == "\"-12\""
    macro deb(a): untyped = newLit(a.repr)
    doAssert deb(-12'bar) == "-12'bar"
    when false: # xxx doesn't work yet
      macro metawrap(): untyped =
        func wrap1(a: string): string = "{" & a & "}"
        func `'wrap2`(a: string): string = "{" & a & "}"
        result = quote do:
          let a1 = wrap1"-128"
          let a2 = -128'wrap2
      metawrap()
      doAssert a1 == "{-128}"
      doAssert a2 == "{-128}"
```

* when false is better than commenting out when it parses ok (as here)
* the `nnkInfix` didn't add value to this test IMO (unrelated to litterals)

=> https://github.com/nim-lang/Nim/pull/17500

timotheecour:
for bug 6, relevant code to fix is in lib/packages/docutils/highlite.nim eg:
```nim
    of '\'':
      inc(pos)
      g.kind = gtCharLit
```

timotheecour:
* see bug 10 i added

timotheecour:
fix for bug 10:
```nim
diff --git a/compiler/lexer.nim b/compiler/lexer.nim
index c0ba217b5..efd78f983 100644
--- a/compiler/lexer.nim
+++ b/compiler/lexer.nim
@@ -1339,7 +1339,7 @@ proc rawGetTok*(L: var Lexer, tok: var Token) =
       if L.buf[L.bufpos] in SymChars+{'_'}:
         lexMessage(L, errGenerated, "invalid token: no whitespace between number and identifier")
     of '-':
-      if ((tok.strongSpaceA > 0) or (L.previousTokType in negationPrefixes)) and (L.buf[L.bufpos + 1] in '0'..'9'):
+      if ((tok.strongSpaceA > 0) or (L.previousTokType in negationPrefixes) or L.buf[L.bufpos - 1] == ' ') and (L.buf[L.bufpos + 1] in '0'..'9'):
         getNumber(L, tok)
         if L.buf[L.bufpos] in SymChars+{'_'}:
           lexMessage(L, errGenerated, "invalid token: no whitespace between number and identifier")
```


test for bug 10 (adapt as needed in your test file):
```nim
when true:
  func `'wrap`(a: string): string = "{" & a & "}"
  template toSuf1(): untyped =
    -12'wrap
  assert toSuf1() == "{-12}"

  template toSuf2(): untyped =
    var a = "adsf"
    -12'wrap
  assert toSuf2() == "{-12}"
```

=> fixed in devel, test in https://github.com/nim-lang/Nim/pull/17500

timotheecour:
- [x] i think we can close bug 2 because this works:
```nim
when true:
  func `'wrap`(a: string): string = "{" & a & "}"
  template toSuf(`'suf`): untyped =
    let x = -12'suf
    x
  assert toSuf(`'wrap`) == "{-12}"
```
(needs a test case; again, integrate as needed to reuse your `wrap`)


=> https://github.com/nim-lang/Nim/pull/17500

timotheecour:
* bug 8: is completely fixed by `genAst` (https://github.com/nim-lang/Nim/pull/17426) (test can be added after #17426 is merged)

* bug 9: `genAst` provides a reasonable workaround, but this can be improved in future work:
```nim
when defined case40:
  import std/genasts # pending #17426
  import std/macros
  func `'wrap2`(a: string): string = "{" & a & "}"
  macro bar(): untyped =
    func wrap1(a: string): string = "{" & a & "}"
    func `'wrap3`(a: string): string = "{" & a & "}"
    result = genAst():
      let a1 = wrap1"-128" # ok
      let a2 = -128'wrap2
      # let a3 = -128'wrap3 # xxx still doesn't work
      let a3 = `'wrap3`("-128") # ok
  bar()
```

doc/manual.rst
```diff
@@ -566,6 +566,19 @@ The type suffixes are:
...
 
+If the suffix is not predefined, then the suffix is assumed to be a call
+to a proc, template, macro or other callable identifier that is passed the
+string containing the literal. The identifier needs to contain the apostrophe
+and so will need to be escaped with backticks.
+
+.. code-block:: nim
+  import strutils
+  type u4 = distinct uint8 # a 4-bit unsigned integer aka "nibble"
+  proc `'u4`(n: string): u4 =
```
timotheecour, line 577: we still need an example with more than 1 param

timotheecour (approved):
LGTM, remaining comments and bug fixes can be done in followup PR
/cc @araq 

(EDIT: see bug 11 i added but it's more of a TODO for followup PR)

compiler/parser.nim
```diff
@@ -764,14 +777,17 @@ proc commandExpr(p: var Parser; r: PNode; mode: PrimaryMode): PNode =
...
   #| primarySuffix = '(' (exprColonEqExpr comma?)* ')'
   #|       | '.' optInd symbol generalizedLit?
   #|       | '[' optInd exprColonEqExprList optPar ']'
   #|       | '{' optInd exprColonEqExprList optPar '}'
   #|       | &( '`'|IDENT|literal|'cast'|'addr'|'type') expr # command syntax
   result = r
-
+  if baseTokType == tkStrNumLit:
+    result = dotExpr(p, result)
+    result = parseGStrLit(p, result)
```
Araq, line 790: Seems hacky and has no corresponding grammar update. I'll see if it can be done differently.

compiler/renderer.nim
```diff
@@ -1188,9 +1188,14 @@ proc gsub(g: var TSrcGen, n: PNode, c: TContext) =
     gcomma(g, n, c)
     put(g, tkBracketRi, "]")
   of nkDotExpr:
-    gsub(g, n, 0)
-    put(g, tkDot, ".")
-    gsub(g, n, 1)
+    if n[1].kind == nkIdent and n[1].ident.s[0] == '\'': # TODO: add a re-usable helper function isUserLitteral
+      assert n[0].kind == nkStrLit
```
Araq, line 1192: This is wrong btw,   ``a.`'foo`  `` was allowed.

timotheecour: ok we can just remove the assert and add a test

Araq:
I took over, see https://github.com/nim-lang/Nim/pull/17489

doc/manual.rst
timotheecour: fixed in devel

PR #17052, opened 2021-02-16
JsString and CString aliases for cstring

### LINKED ISSUE #342: Abstractions over cstring per backend
Possibly incomplete proposal

**Problem**: `cstring` has varying behavior across backends and VM, in C-like backends it's a char pointer which you have to allocate and increment the pointer to iterate through, in JS it's the native JS string type which indexes into strings instead of char codes, the char codes it does give are in UTF-16, and the type can be an efficient replacement for `string` in cases where the differences can be dealt with, but has to have a `null` value by default for compatibility with C cstrings. In VM though it is pretty much the same as `string` and can't be `nil`, leading to discrepancies with runtime behavior.

**Proposal**: Embrace the disharmonious nature of cstring across backends. Instead, have new `CString` and `JsString` types that are conditionally defined the same as `cstring` if the backend matches the type, and can be implemented in pure Nim for other backends if desired. This would help users and libraries safely and easily differentiate cstring behavior across backends.

`CString` and `JsString` (on top of `cstring`) could be implemented like this:

```nim
type cstring* {.magic: Cstring.}
proc len*(cstr: cstring) {.magic: "LengthStr".}
# document that cstring and procs on it behave differently across backends

when defined(js): # maybe the else: block should be used for docs
  type JsString* = cstring
  # maybe distinct is better?
else:
  type JsString* = object
    data: seq[uint16]
  proc len*(str: JsString) = str.data.len
  # there is probably a more efficient and correct representation than the above
  # this else block could also just be empty but public API getting deleted across backends could be bad

when defined(c) or defined(cpp) or defined(objc):
  # ^ or put in else block of JS, this is just a bit clearer
  type CString* = cstring
else:
  type CString* = ref seq[char]
  proc len*(str: CString) = str[].len
  # weird i know, but ditto comments as above
```

All existing `cstring` routines can stay with their varying results for each backend, but there could be clearer and more specific library code using JsString and CString. For example, `add(var cstring, cstring)` is an overload that gets defined only under JS. It could be rewritten as `add(var JsString, JsString)` so it's more clear what it does, which would use `magic: "ConStrStr"` or any magic it wants, and could even output directly as `a += b` if null cstrings are removed from JS. 

As for the VM, I think `cstring` should behave the same for `JsString` and `CString` individually as their runtime versions, but most people currently don't use `cstring` much in their compile time code. I don't know if it would be practical to have VM operations (and type representation) behave differently per backend, would like to hear opinions about this.

**Backwards compatibility**: Irrelevant to this proposal, only matters to following potential individual changes on `cstring` and magics on it, e.g. no more null for JS cstring. JS cstrings were "" by default for a while in 1.3, but this broke karax for example.

Related issues:

* https://github.com/nim-lang/Nim/issues/16686 (add for cstring wasn't documented because defined(js) was not true for nim doc)
* https://github.com/nim-lang/Nim/issues/15730 (compile time vs runtime cstring)
* https://github.com/nim-lang/Nim/issues/14350 (karax regression from default cstring "")
* https://github.com/nim-lang/Nim/issues/14229, https://github.com/nim-lang/Nim/issues/10911 (utf 16)

Probably many more

metagn:
closes https://github.com/nim-lang/RFCs/issues/342

Barebones implementation of the RFC. Changes to individual standard library modules (like just verbatim replacing `cstring` with `JsString` in lib/js) and implementing/stubbing `JsString` and `CString` for other backends can be done later. This amount of change is sufficient for user libraries.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -155,6 +155,15 @@ provided by the operating system.
   `ValueError` when the real command line is not available. `parseopt` was
   previously excluded from `prelude` for JS, as it could not be imported.
 
+- There are now backend-specific `JsString` and `CString` aliases for
+  `cstring`. The purpose is to make library definitions clearer, as `cstring`
+  can behave differently across backends.
+  
+  Currently the only way the types are used differently is that `JsString` has
+  an `add(var JsString, JsString)` overload. This overload was originally added
+  in version 0.8.14, but was never documented due to the `when` check
+  defining it not accounting for documentation.
+
 ## Language changes
 
 - `nimscript` now handles `except Exception as e`.
--- a/doc/manual.rst
+++ b/doc/manual.rst
@@ -1274,16 +1274,17 @@ cstring type
 ------------
 
 The ``cstring`` type meaning `compatible string` is the native representation
-of a string for the compilation backend. For the C backend the ``cstring`` type
-represents a pointer to a zero-terminated char array
-compatible with the type ``char*`` in Ansi C. Its primary purpose lies in easy
-interfacing with C. The index operation ``s[i]`` means the i-th *char* of
-``s``; however no bounds checking for ``cstring`` is performed making the
-index operation unsafe.
+of a string for the compilation backend. For the C, C++ and Objective-C
+backends the ``cstring`` type represents a pointer to a zero-terminated char
+array compatible with the type ``char*`` in Ansi C. For the JavaScript backend
+the ``cstring`` type represents the native JavaScript primitive string type.
+Its primary purpose lies in easy interfacing with backends. The index operation
+``s[i]`` means the i-th *char* of ``s``; however no bounds checking for
+``cstring`` is performed making the index operation unsafe.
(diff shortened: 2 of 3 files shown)


Araq:
IMHO we should alias ``cstring`` in favor of ``JsString`` or ``jsstring`` or ``jstring``. And we don't need an alias for `cstring` for the C backend.

metagn:
Well I thought CString could get exclusive behavior too, like a converter to `pointer` (#14097). I guess since it's the "mainstream" use there's not much need for a distinction, but it would help defining a possible emulation of it on JS like said in https://github.com/nim-lang/Nim/issues/14097#issuecomment-619256822.

> JsString or jsstring or jstring

Any preference? I didn't really consider which name people preferred.

Araq:
I prefer `jstring`.

metagn:
I think the RFC should stay open a little longer then

PR #17055, opened 2021-02-16
Fixes #17039 - checked field addr analysis

### LINKED ISSUE #17039: VM crash with object variant, adding seq to seq field
Encountered in nimscript. The issue seems to be that fields in object variant branches break when converting to `var T`.

### Example
```nim
type
  Obj1 = object
    case kind: bool
    of false:
      field: seq[int]
    else: discard

static:
  var obj1 = Obj1()
  obj1.field.add(@[])
```

### Current Output

```
....compiler\nim.nim(119) nim
compiler\nim.nim(84) handleCmdLine
compiler\main.nim(242) mainCommand
compiler\main.nim(213) compileToBackend
compiler\main.nim(90) commandCompileToC
compiler\modules.nim(177) compileProject
compiler\modules.nim(97) compileModule
compiler\passes.nim(180) processModule
compiler\passes.nim(73) processTopLevelStmt
compiler\sem.nim(607) myProcess
compiler\sem.nim(575) semStmtAndGenerateGenerics
compiler\semstmts.nim(2306) semStmt
compiler\semexprs.nim(1029) semExprNoType
compiler\semexprs.nim(2865) semExpr
compiler\semstmts.nim(2248) semStmtList
compiler\semexprs.nim(2917) semExpr
compiler\semstmts.nim(2204) semStaticStmt
compiler\vm.nim(2213) evalStaticStmt
compiler\vm.nim(2202) evalConstExprAux
compiler\vm.nim(716) rawExecute
lib\system\fatal.nim(53) sysFatal
```

### Expected Output
compiles

### Possible Solution

As an immediate solution, if you change this line in `opcLdObj`:

https://github.com/nim-lang/Nim/blob/6e267d28b3459aa447cae07ba61209438977cd5c/compiler/vm.nim#L714

to this line below it in `opcLdObjAddr`:

https://github.com/nim-lang/Nim/blob/6e267d28b3459aa447cae07ba61209438977cd5c/compiler/vm.nim#L730

it works. Maybe the wrong VM opcode is getting generated, but it's better to be safe than crashing the compiler.

### Additional Information

```
$ nim -v
Compiled at 2021-02-14

git hash: fff5819ee728aa62fd857982f799ecc0fac0fd5e
```

saem:
Checked field expressions, such as an object variant field access were
not being unwrapped during analysis of address taken.

This unwraps them to ensure the analysis is done for the field access,
but doesn't discard them.

PS: did some initial analysis, at least going by tests, opcLdObjAddr
should not require the conditional check based on kind for the src.
Will revisit later.

### DIFF
--- a/compiler/semexprs.nim
+++ b/compiler/semexprs.nim
@@ -707,6 +707,10 @@ proc analyseIfAddressTaken(c: PContext, n: PNode): PNode =
     if skipTypes(n[0].typ, abstractInst-{tyTypeDesc}).kind notin {tyVar, tyLent}:
       if n[0].kind == nkSym: incl(n[0].sym.flags, sfAddrTaken)
       result = newHiddenAddrTaken(c, n)
+  of nkCheckedFieldExpr:
+    checkMinSonsLen(n, 1, c.config)
+    n[0] = analyseIfAddressTaken(c, n[0])
+    result = n
   else:
     result = newHiddenAddrTaken(c, n)
 
--- a/tests/vm/t17039.nim
+++ b/tests/vm/t17039.nim
@@ -1,0 +1,10 @@
+type
+  Obj1 = object
+    case kind: bool
+    of false:
+      field: seq[int]
+    else: discard
+
+static:
+  var obj1 = Obj1()
+  obj1.field.add(@[])


saem:
🤔 hmm the premise of this seems off... will reconsider how to do this later -- I definitely misunderstood what was happening.

All of this really makes me wonder about using separate nodes/productions for `nkHiddenAddrTaken`, `nkCheckedFieldExpr`, seems better as properties and queries.

Araq:
Feel free to re-open once you have a better solution.

PR #17069, opened 2021-02-17
[minor] clean extccomp

ringabout:


compiler/extccomp.nim
```diff
@@ -12,9 +12,9 @@
 # from a lineinfos file, to provide generalized procedures to compile
 # nim files.
 
-import
-  ropes, os, strutils, osproc, platform, condsyms, options, msgs,
-  lineinfos, std / sha1, streams, pathutils, sequtils, times, strtabs
+import ropes, platform, condsyms, options, msgs, lineinfos, pathutils
+
+import std/[os, strutils, osproc, sha1, streams, sequtils, times, strtabs, json]
```
konsumlamm, line 17:
```suggestion
import platform, condsyms, msgs, lineinfos, pathutils

import std/[ropes, os, strutils, osproc, options, sha1, streams, sequtils, times, strtabs, json]
```
`ropes` and `options` are also standard library modules.

ringabout: not really, `ropes` and `options` are compiler modules too(they do in this module).

konsumlamm: Ah, my bad.

ringabout: never mind

PR #17084, opened 2021-02-18
Revert "follow #15818 and close #7109"

### LINKED ISSUE #7109: Better error message when attempting to use closure iterators with js backend
The current error looks like there is a bug. It should probably just say "Closure iterators are not supported by the js backend"

```nim
iterator iter: int {.closure.} =
    yield 1
    yield 2

var x = iter
```

```
jsiter.nim(5, 9) Error: internal error: symbol has no generated name: iter
```

Araq:
Reverts nim-lang/Nim#15823


Araq:
This was a breaking change, see https://forum.nim-lang.org/t/7523

We need to revert this. The proper fix seems to be to make closure iterators an error when they are **used** in the JS mode. Of course, we can also simply implement closure iterators for the JS backend.

xflywind:
> The proper fix seems to be to make closure iterators an error when they are used in the JS mode.
What do you mean?

PR #10604, opened 2019-02-08
move cookies and cgi to graveyard

narimiran:
This supersedes PR #9360, and closes #7810.

PR #17116, opened 2021-02-20
[stdlib] make cookies module modern

ringabout:
Ref https://github.com/nim-lang/Nim/pull/10604

`std/cookies` is not good enough, people are keeping creating a new one.

https://github.com/dom96/jester/blob/master/jester/private/utils.nim
https://github.com/planety/cookiejar
https://github.com/itsumura-h/nim-basolato/blob/master/src/basolato/core/utils.nim
https://github.com/achesak/nim-biscuits#about

BTW Python has deprecated `cgi` module at Python 3.8 and will remove `cgi` module at Python 3.10.

93.25% of browsers support sameSite attribute.
> This feature is backwards compatible. Browsers not supporting this feature will simply use the cookie as a regular cookie. There is no need to deliver different cookies to clients.

https://caniuse.com/?search=sameSite

### DIFF
--- a/lib/pure/cookies.nim
+++ b/lib/pure/cookies.nim
@@ -14,6 +14,8 @@ import std/[strtabs, times, options]
 
 type
   SameSite* {.pure.} = enum ## The SameSite cookie attribute.
+                            ## `Default` means that `setCookie`
+                            ## proc will not set `SameSite` attribute.
     Default, None, Lax, Strict
 
 proc parseCookies*(s: string): StringTableRef =
--- a/lib/pure/cookies.nim
+++ b/lib/pure/cookies.nim
@@ -9,18 +9,23 @@
 
 ## This module implements helper procs for parsing Cookies.
 
-import strtabs, times
+import std/[strtabs, times, options]
+
+
+type
+  SameSite* {.pure.} = enum ## The SameSite cookie attribute.
+    None, Lax, Strict
 
 proc parseCookies*(s: string): StringTableRef =
-  ## parses cookies into a string table.
+  ## Parses cookies into a string table.
   ##
   ## The proc is meant to parse the Cookie header set by a client, not the
   ## "Set-Cookie" header set by servers.
-  ##
-  ## Example:
-  ##
-  ## .. code-block::Nim
-  ##     doAssert parseCookies("a=1; foo=bar") == {"a": 1, "foo": "bar"}.newStringTable
+  runnableExamples:
+    import std/strtabs
+    let cookieJar = parseCookies("a=1; foo=bar") 
+    assert cookieJar["a"] == "1"
+    assert cookieJar["foo"] == "bar"
 
   result = newStringTable(modeCaseInsensitive)
   var i = 0
@@ -39,9 +44,10 @@ proc parseCookies*(s: string): StringTableRef =
 
 proc setCookie*(key, value: string, domain = "", path = "",
                 expires = "", noName = false,
-                secure = false, httpOnly = false): string =
+                secure = false, httpOnly = false, 
+                maxAge = none(int), sameSite = SameSite.Lax): string =
   ## Creates a command in the format of
-  ## ``Set-Cookie: key=value; Domain=...; ...``
+  ## `Set-Cookie: key=value; Domain=...; ...`
   result = ""
   if not noName: result.add("Set-Cookie: ")
   result.add key & "=" & value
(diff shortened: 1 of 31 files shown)


lib/pure/cookies.nim
```diff
@@ -39,9 +44,10 @@ proc parseCookies*(s: string): StringTableRef =
 
 proc setCookie*(key, value: string, domain = "", path = "",
                 expires = "", noName = false,
-                secure = false, httpOnly = false): string =
+                secure = false, httpOnly = false, 
+                maxAge = none(int), sameSite = SameSite.Lax): string =
```
ringabout: I can make `sameSite` default to `SameSite.None` and add overload with since. But it looks ugly.

lib/pure/cookies.nim
```diff
@@ -9,18 +9,23 @@
...
 
-import strtabs, times
+import std/[strtabs, times, options]
+
+
+type
+  SameSite* {.pure.} = enum ## The SameSite cookie attribute.
+    None, Lax, Strict
 
 proc parseCookies*(s: string): StringTableRef =
```
timotheecour, line 21: is `StringTableRef` even meaningful?

`nim r --eval:'import std/[cookies,strtabs]; echo parseCookies("a=1;Path=/; a=2;Path=/example;")'` prints `{a: 1, Path: /}` which seems wrong according to https://stackoverflow.com/questions/4056306/how-to-handle-multiple-cookies-with-the-same-name

ringabout:
see also https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite
> Standards related to the Cookie SameSite attribute recently changed such that:
The cookie-sending behavior if SameSite is not specified is SameSite=Lax. Previously the default was that cookies were sent for all requests.
Cookies with SameSite=None must now also specify the Secure attribute (they require a secure context/HTTPS).


https://tools.ietf.org/html/draft-west-cookie-incrementalism-00

lib/pure/cookies.nim
timotheecour: => moved this discussion to https://github.com/timotheecour/Nim/issues/602#issuecomment-782600451

lib/pure/cookies.nim
```diff
@@ -50,12 +58,19 @@ proc setCookie*(key, value: string, domain = "", path = "",
   if expires != "": result.add("; Expires=" & expires)
   if secure: result.add("; Secure")
   if httpOnly: result.add("; HttpOnly")
+  if maxAge.isSome: result.add("; Max-Age=" & $maxAge.unsafeGet)
+
+  if sameSite != SameSite.Default:
+    if sameSite == SameSite.None:
+      doAssert httpOnly, "Cookies with SameSite=None must specify the Secure attribute!"
```
metagn:
```suggestion
      doAssert secure, "Cookies with SameSite=None must specify the Secure attribute!"
```

I'm guessing it's supposed to be this?

ringabout: thanks

lib/pure/cookies.nim
```diff
@@ -9,18 +9,25 @@
...
 
-import strtabs, times
+import std/[strtabs, times, options]
+
+
+type
+  SameSite* {.pure.} = enum ## The SameSite cookie attribute.
+                            ## `Default` means that `setCookie`
+                            ## proc will not set `SameSite` attribute.
+    Default, None, Lax, Strict
```
timotheecour, line 19: std/compilesettings uses lower-case enum values, this is inconsistent; we should specify in nep1 which one is the "best practice"
followup: https://github.com/timotheecour/Nim/issues/622#issuecomment-786126727

lib/pure/cookies.nim
```diff
@@ -50,12 +58,19 @@ proc setCookie*(key, value: string, domain = "", path = "",
...
 
 proc setCookie*(key, value: string, expires: DateTime|Time,
                 domain = "", path = "", noName = false,
-                secure = false, httpOnly = false): string =
+                secure = false, httpOnly = false,
+                maxAge = none(int), sameSite = SameSite.Default): string =
   ## Creates a command in the format of
-  ## ``Set-Cookie: key=value; Domain=...; ...``
-  return setCookie(key, value, domain, path,
+  ## `Set-Cookie: key=value; Domain=...; ...`
```
timotheecour, line 73: pre-existing but this overload (with differently placed params) seems like a bad design, instead we should expose a proc (eg `toGmtString(t: DateTime|Time)` and let users call:
```nim
setCookie(..., t.toGmtString, ...)
```

lib/pure/cookies.nim
ringabout: It is already specified in nep1: Non-pure enum values should use camelCase whereas pure enum values should use PascalCase.

image

PR #17120, opened 2021-02-20
Add example for pattern matching on options

konsumlamm:
Follow-up to #17036. This PR adds an example for pattern matching on options. It uses `code-block`, since `fusion` isn't bundled anymore (see #16925). If there is a way to turn this into a `runnableExamples`, please let me know.

### DIFF
--- a/lib/pure/options.nim
+++ b/lib/pure/options.nim
@@ -64,7 +64,7 @@ supports pattern matching on `Option`s, with the `Some(<pattern>)` and
   of None():
     assert false
 
-  assertMatch(Some(Some(None())), some(some(none(int))))
+  assertMatch(some(some(none(int))), Some(Some(None())))
 ]##
 # xxx pending https://github.com/timotheecour/Nim/issues/376 use `runnableExamples` and `whichModule`
 
--- a/lib/pure/options.nim
+++ b/lib/pure/options.nim
@@ -7,22 +7,24 @@
 #    distribution, for details about the copyright.
 #
 
-## This module implements types which encapsulate an optional value.
-##
-## A value of type `Option[T]` either contains a value `x` (represented as
-## `some(x)`) or is empty (`none(T)`).
-##
-## This can be useful when you have a value that can be present or not. The
-## absence of a value is often represented by `nil`, but that is not always
-## available, nor is it always a good solution.
-##
-##
-## Basic usage
-## ===========
-##
-## Let's start with an example: a procedure that finds the index of a character
-## in a string.
-##
+##[
+This module implements types which encapsulate an optional value.
+
+A value of type `Option[T]` either contains a value `x` (represented as
+`some(x)`) or is empty (`none(T)`).
+
+This can be useful when you have a value that can be present or not. The
+absence of a value is often represented by `nil`, but that is not always
+available, nor is it always a good solution.
+
+
+Basic usage
+===========
+
+Let's start with an example: a procedure that finds the index of a character
+in a string.
+]##
+
 runnableExamples:
   proc find(haystack: string, needle: char): Option[int] =
     for i, c in haystack:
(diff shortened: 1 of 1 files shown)


lib/pure/options.nim
```diff
@@ -39,6 +39,29 @@ runnableExamples:
...
+## `None()` patterns.
+##
+## .. code-block:: nim
+##   {.experimental: "caseStmtMacros".}
+##
+##   import fusion/matching
+##
+##   case some(42)
+##   of Some(@a):
+##     assert a == 42
```
juancarlospaco:
```suggestion
##     doAssert a == 42
```
Same the other.

timotheecour:
> If there is a way to turn this into a runnableExamples, please let me know

* the following works, but not inside a `runnableExamples` because of this (IMO easily fixable) bug that removes `do` from  `runnableExamples`: https://github.com/nim-lang/Nim/issues/13491
```nim
when true:
  when (compiles do: import fusion/matching):
    case some(42)
    of Some(@a):
      assert a == 42
    of None():
      assert false
    Some(Some(None())) := some(some(none(int)))
```

* however, even after we fix https://github.com/nim-lang/Nim/issues/13491, IMO this pattern `when (compiles do: import fusion/matching):` is bad, as it violates scoping rules: `when foo: bar` shouldn't inject declarations in bar, refs https://github.com/nim-lang/Nim/issues/12030 and https://github.com/nim-lang/Nim/issues/16287; the correct fix is `whichModule` refs https://github.com/timotheecour/Nim/issues/376; so I suggest leaving is as you did and waiting for `whichModule`, maybe add: `### pending https://github.com/timotheecour/Nim/issues/376 use `whichModule`

lib/pure/options.nim
```diff
@@ -39,6 +39,29 @@ runnableExamples:
 ## inherits from `system.Defect` and should therefore never be caught.
 ## Instead, rely on checking if the option contains a value with the
 ## `isSome <#isSome,Option[T]>`_ and `isNone <#isNone,Option[T]>`_ procs.
+##
+##
+## Pattern matching
+## ================
+##
+## **Note:** This requires the [fusion](https://github.com/nim-lang/fusion) package.
```
juancarlospaco:
```suggestion
## .. Note:: This requires the [fusion](https://github.com/nim-lang/fusion) package.
```

For the new "Admonitions" thingy ?.

lib/pure/options.nim
timotheecour: but see https://github.com/nim-lang/Nim/pull/17077 ; we should probably update that section in contributing.rst to also cover code-block sections as it should act as `runnableExamples`

lib/pure/options.nim
```diff
@@ -39,6 +39,29 @@ runnableExamples:
...
+##
+##   import fusion/matching
+##
+##   case some(42)
+##   of Some(@a):
+##     assert a == 42
+##   of None():
+##     assert false
+##
+##   Some(Some(None())) := some(some(none(int)))
```
timotheecour: although some/Some/none/None is clear from the context, `:=` is not; can you add a short explanation?

konsumlamm: I could use `assertMatch` instead, which `:=` is an alias for.

lib/pure/options.nim
konsumlamm: Hmm, I haven't seen this yet, is it recommended to use it? If so, there should probably also be a PR to change this for all the other files.

juancarlospaco: https://nim-lang.github.io/Nim/rst.html#features

lib/pure/options.nim
timotheecour: => PR https://github.com/nim-lang/Nim/pull/17122

lib/pure/options.nim
timotheecour: it looks good IMO:

image

lib/pure/options.nim
```diff
@@ -39,6 +39,30 @@ runnableExamples:
...
+## Pattern matching
+## ================
+##
+## .. note:: This requires the [fusion](https://github.com/nim-lang/fusion) package.
+##
+## [fusion/matching](https://nim-lang.github.io/fusion/src/fusion/matching.html)
+## supports pattern matching on `Option`s, with the `Some(<pattern>)` and
+## `None()` patterns.
+##
+## .. code-block:: nim
```
timotheecour: can you at least use `##[` block comments so that the code is more easily copy-pastable?
(and make sure `nim doc` renders fine)
something like this:
```nim
##[
...

.. code-block:: nim
...
]##
```

lib/pure/options.nim
```diff
@@ -39,6 +39,30 @@ runnableExamples:
...
+##   import fusion/matching
+##
+##   case some(42)
+##   of Some(@a):
+##     assert a == 42
+##   of None():
+##     assert false
+##
+##   assertMatch(Some(Some(None())), some(some(none(int))))
+# xxx pending https://github.com/timotheecour/Nim/issues/376 use `whichModule`
```
timotheecour:
```suggestion
# xxx pending https://github.com/timotheecour/Nim/issues/376 use `runnableExamples` and `whichModule`
```
(otherwise it's not clear)

timotheecour (approved):
LGTM after final nits

lib/pure/options.nim
```diff
@@ -34,35 +36,37 @@ runnableExamples:
...
+
+  import fusion/matching
+
+  case some(42)
+  of Some(@a):
+    assert a == 42
+  of None():
+    assert false
+
+  assertMatch(Some(Some(None())), some(some(none(int))))
```
timotheecour: - [x] `assertMatch(Some(Some(None())), some(some(none(int))))` doesn't work:

Error: unhandled exception: Expected one of {nnkExprColonExpr} but node has kind nnkCall (assertion on (filename: "matching.nim", line: 857, column: 12)) [AssertionDefect]

`Some(Some(None())) := some(some(none(int)))` does work but would need a comment explaning what `:=` means

- [ ] the error message above is bad, should probably use `toLocation` from `lib/std/private/miscdollars.nim` instead (even if it's private; IMO fusion should be able to use things under `lib/std/private/` as it's intended to work in tandem with nim)

(EDIT: followup https://github.com/timotheecour/Nim/issues/603)

konsumlamm: Oh yeah, I swapped the arguments, it should be `assertMatch(some(some(none(int))), Some(Some(None())))`. Do you have any preference on `:=` vs `assertMatch`?

timotheecour: I prefer `assertMatch` (i generally dislike non-standard operators)

timotheecour (approved):

PR #17123, opened 2021-02-20
Fixes #17039 - checked field addr analysis

### LINKED ISSUE #17039: VM crash with object variant, adding seq to seq field

saem:
Checked field expressions, such as an object variant field access results in
occasionally broken address analysis crashing the VM. This guard added here
mimics guarded access in ldObjAddr. This is to prevent a crash, while a fix is
devised.

### DIFF
--- a/compiler/vm.nim
+++ b/compiler/vm.nim
@@ -711,7 +711,7 @@ proc rawExecute(c: PCtx, start: int, tos: PStackFrame): TFullReg =
     of opcLdObj:
       # a = b.c
       decodeBC(rkNode)
-      let src = regs[rb].node
+      let src = if regs[rb].kind == rkNode: regs[rb].node else: regs[rb].nodeAddr[]
       case src.kind
       of nkEmpty..nkNilLit:
         # for nkPtrLit, this could be supported in the future, use something like:
--- a/tests/vm/t17039.nim
+++ b/tests/vm/t17039.nim
@@ -1,0 +1,10 @@
+type
+  Obj1 = object
+    case kind: bool
+    of false:
+      field: seq[int]
+    else: discard
+
+static:
+  var obj1 = Obj1()
+  obj1.field.add(@[])


saem:
All tests passed locally, looking at the way the `nkCheckedFieldExpr` are generated this should be fine.

Let's see what happens in CI more broadly across platforms.

saem:
Nope, going to go with the hackier fix for now to prevent the compiler crash, will update the PR appropriately.

compiler/vm.nim
```diff
@@ -711,7 +711,7 @@ proc rawExecute(c: PCtx, start: int, tos: PStackFrame): TFullReg =
     of opcLdObj:
       # a = b.c
       decodeBC(rkNode)
-      let src = regs[rb].node
+      let src = if regs[rb].kind == rkNode: regs[rb].node else: regs[rb].nodeAddr[]
```
ringabout, line 714: Maybe something like this is better
```nim
      let kind = regs[rb].kind
      let src = 
        if kind == rkNode: regs[rb].node
        elif kind == rkNodeAddr: regs[rb].nodeAddr[]
        else: doAssert false, $kind
```

saem: The error shows up and can be localized without this and if the compile is release or danger it won't work, so it really doesn't add anything. So no, that's not better.

